PR #336, opened 2013-10-02
Generalize options

jchambers:
This is an alternative to https://github.com/antlr/antlr4/pull/326. As discussed there, another way to go about things would be to have `-D` set arbitrary properties that can be used by templates. This pull request does that by creating a combined pool of options that draws from (in order of increasing precedence) defaults, grammar-level options, and command-line options.

The `-D` flag can now be used to set arbitrary properties. If the given property is a recognized grammar option (e.g. `language`), it will override whatever is in the defaults/grammar definition. If it's not a recognized grammar option, it will still be set and can be read by the template as desired.

Do we think this is the right way to go? If so, I'll add unit tests. If not, I'm open to other suggestions.


jchambers:
Sorry to leave this for so long. It sounds like you folks have a plan for this, so I'll leave it in your hands. Cheers!

PR #448, opened 2014-01-24
Lookahead tweaks

sharwell:
* Fix code generation for `PlusBlock` to adhere to [the documentation](http://www.antlr.org/api/Java/org/antlr/v4/runtime/atn/ATNState.html). The only reason it worked previously is we never produced a `PlusBlockStartState` in the code generation process containing more than one alternative.
* Remove unnecessary call to `LA(1)` prior to entering an LL(1) positive closure with only one alternative
* Fix LL(1) analysis for non-greedy loops


parrt:
Fixed in https://github.com/antlr/antlr4/pull/532.

sharwell:
Only commit 0257a3a was addressed by #532, and only partially.

PR #532, opened 2014-04-02
Codegen fixes

sharwell:


parrt:
Sam says: The fix assumes that PlusBlockStartState never has more than one alternative. You can see from the documentation that it is not guaranteed to be the case forever, even though it is the case for now.

PR #697, opened 2014-09-03
Initial implementation of the baseContext rule option

### LINKED ISSUE #521: Context classes for syntactic restrictions
Occasionally a parser rule exists which is a syntactic restriction of another rule in the parser. For example, the unambiguous syntactic resolution to the nested `if`/`if`/`else` problem is used in the [JLS §19](http://docs.oracle.com/javase/specs/jls/se8/html/jls-19.html), with two of the rules as follows.

``` antlr
statement
    :   statementWithoutTrailingSubstatement
    |   labeledStatement
    |   ifThenStatement
    |   ifThenElseStatement
    |   whileStatement
    |   forStatement
    ;

statementNoShortIf
    :   statementWithoutTrailingSubstatement
    |   labeledStatementNoShortIf
    |   ifThenElseStatementNoShortIf
    |   whileStatementNoShortIf
    |   forStatementNoShortIf
    ;
```

In this case, `statement` produces a `StatementContext` class and `statementNoShortIf` produces a `StatementNoShortIfContext` class. However, `statementNoShortIf` is a syntactic restriction of the `statement` rule, and it would be very beneficial to post-processing (listeners and visitors) if both rules produced a `StatementContext`.

The JLS contains 7 restriction rules for the `NoShortIf`, and another 8 with the prefix `unann` where the use of annotations is restricted. This feature would greatly simplify the parse tree for these 15 rules.

sharwell:
## Summary

Fixes #521

This pull request implements a new `baseContext` option for rules. The option allows multiple rules to share a single generated context class, called the _base context_. Here is an example where it might be used:

``` antlr
statement
  : ifStatement
  | switchStatement
  | emptyStatement
  ;

statementNotEmpty
options { baseContext = statement; }
  : ifStatement
  | switchStatement
  ;
```

In this example, the parse tree for `statementNotEmpty` is identical to `statement` (including the listener and visitor behavior). The only observable difference is the `statementNotEmpty` rule is not able to match the `emptyStatement`. This includes the fact that the `getRuleIndex()` method for the tree produced by `statementNotEmpty` will be `RULE_statement`, _not_ `RULE_statementNotEmpty`. The parser does still include a constant `RULE_statementNotEmpty`, since it can be used by the parser interpreter, etc.
## Status

The implementation of the feature appears to be complete for most use cases, but unit tests are not yet completed (I did side-by-side comparisons of the output to ensure it was behaving properly in a number of scenarios).
- [x] Initial implementation
- [x] Unit tests
  - [x] Test basic use case produces code that emits the proper base context class for the parse tree (currently tested by inspection)
  - [x] Test `baseContext` option suppresses the emission of a rule context for the retargeted rule, including in listeners and visitors (currently tested by inspection)
  - [x] Test error messages
  - [x] Test labeled alternatives (currently tested by inspection)
  - [x] Test context accessor analysis (currently tested by inspection)
  - [x] ~~Test label analysis (currently tested by inspection, **fails**)~~ (see limitations below)
  - [x] ~~Test arguments, return values, and locals (currently produces the following error)~~ (see limitations below)
    
    > java.lang.UnsupportedOperationException: customized fields are not yet supported for customized context objects
- [x] Error message if some alternatives which produce a particular parse tree have alt labels, and some do not
  
  > error(165): rule _rule_: must label all alternatives in rules with the same base context, or none
- [x] Error message if `baseContext` option is assigned a value that is not another rule
  
  > error(166): rule _rule_: baseContext option value must reference a rule
- [x] Error message if `baseContext` is assigned a rule that itself has the `baseContext` specified (no transitive assignments allowed)
  
  > error(167): rule _rule_: base context must reference a rule that does not specify a base context
- [x] Error message if `baseContext` is specified for a lexer rule. This is currently impossible since lexer rules cannot have an `options{}` block, but should that syntax be allowed in the future this error analysis will already be in place.
  
  > error(168): rule _rule_: lexer rules cannot specify a base context
- [x] ~~Update code using `RuleFunction.addContextDecl` to properly add declarations to the correct context.~~ (Part of the limitation on labels/args/returns/locals)
- [x] ~~Update error 123 (`ALT_LABEL_REDEF`) to consider base context before reporting an error.~~ (Part of limitation on labeled alt names)

## Limitations

In the current implementation, a rule which uses the `baseContext` option has the following limitations:

* The rule cannot have arguments (this includes left recursive rules which implicitly have an argument)
* They cannot use `returns` or `locals`
* They cannot have labeled elements (labeled alternatives are fine)
* The names of labeled alternatives cannot alias the labeled alternatives from the base context

### DIFF
--- a/runtime-testsuite/test/org/antlr/v4/test/runtime/java/api/BaseContext.g4
+++ b/runtime-testsuite/test/org/antlr/v4/test/runtime/java/api/BaseContext.g4
@@ -1,0 +1,46 @@
+grammar BaseContext;
+
+mainRule
+	:	A* B
+	;
+
+mainRuleNoA
+options { baseContext = mainRule; }
+	:	B
+	;
+
+listLabelPrimary
+	:	B C?
+	;
+
+listLabelAlternative
+options { baseContext = listLabelPrimary; }
+	:	B+
+	;
+
+labeledAlts1
+	:	B	# contextName1
+	;
+
+labeledAlts2
+options { baseContext = labeledAlts1; }
+	:	B	# contextName2
+	;
+
+expr
+	:	B
+	|	expr A expr
+	;
+
+exprPrimaryOnly
+options { baseContext = expr; }
+	:	B
+	;
+
+indirectReference
+	:	a=mainRule b=mainRuleNoA
+	;
+
+A : 'A';
+B : 'B';
+C : 'C';
--- a/runtime-testsuite/test/org/antlr/v4/test/runtime/java/api/TestBaseContext.java
+++ b/runtime-testsuite/test/org/antlr/v4/test/runtime/java/api/TestBaseContext.java
@@ -1,0 +1,196 @@
+/*
+ * Copyright (c) 2012-2016 The ANTLR Project. All rights reserved.
+ * Use of this file is governed by the BSD 3-clause license that
+ * can be found in the LICENSE.txt file in the project root.
+ */
+package org.antlr.v4.test.runtime.java.api;
+
+import java.util.List;
+import org.antlr.v4.runtime.ANTLRInputStream;
+import org.antlr.v4.runtime.BailErrorStrategy;
+import org.antlr.v4.runtime.CommonTokenStream;
+import org.antlr.v4.runtime.misc.ParseCancellationException;
+import org.antlr.v4.runtime.tree.TerminalNode;
+import org.hamcrest.CoreMatchers;
+import org.junit.Assert;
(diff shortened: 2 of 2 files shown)


sharwell:
@parrt If you decide to include this in the future, let me know. I updated it substantially in tunnelvisionlabs/antlr4#47 to ensure it works with a mix of element labels, alt labels, and/or left recursive rules.

parrt:
@sharwell sounds good. thanks!

PR #748, opened 2014-10-26
Fixes for bild.py

bdkearns:


parrt:
howdy. going to try to cherry pick some stuff from here and put into 4.5...damn, just made 4.5 rc1. let me push that out and i can get this stuff in for 4.5 final. some of stuff is obvious. other parts are not. I'll try to bring in what i can and then you can try another pass to update for your needs. sorry for the mess.

bild.py
```diff
@@ -42,6 +43,8 @@
 PYTHON3_TARGET	= "../antlr4-python3"
 CSHARP_TARGET	= "../antlr4-csharp"
 
+JAVA_VERSION = os.environ.get('JAVA_VERSION', None)
```
parrt: ok, javac is piece of shit:

```
terence:master:~/github/bild $ javac -source 1.6 -target 1.6
javac: no source files
Usage: javac <options> <source files>
use -help for a list of possible options
terence:master:~/github/bild $ javac -source 1.7 -target 1.6
javac: source release 1.7 requires target release 1.7
```

why have -target if the damn thing won't gen earlier versions.  We have 1.7 code in some places, right?

ericvergnaud:
I don’t really have needs, except it working which is already the case :-)
The OrderedDict is a nice way of controlling the build sequence, although not required.
Using a global variable for JAVA_VERSION is definitely good and should actually be 1.6
I believe the properties stuff is maybe just adding unnecessary complexity 
Now that each target test suite is actually able to locate the runtime, this becomes overkilling

I think the rest is a

bild.py
ericvergnaud: We have no need for java 1.7 right now,
although we should start advertising that next major release will move to 1.7

parrt: this is fixed now. e.g., output in bild.log is:

[01/10/15 21:34:59 javac ./bild.py:127 bilder.py:452] javac -version
[01/10/15 21:35:00 javac ./bild.py:127 bilder.py:452] javac 1.7.0_25
[01/10/15 21:35:00 javac ./bild.py:127 bilder.py:454] javac -d /Users/parrt/antlr/code/antlr4/out/runtime -cp /Users/parrt/antlr/code/antlr4/out/runtime:runtime/Java/lib/org.abego.treelayout.core.jar -source 1.6 -target 1.6 -nowarn -Xlint -Xlint:-serial -g -sourcepath gen4:runtime/JavaAnnotations/src:runtime/Java/src /Users/parrt/antlr/code/antlr4/gen4/org/antlr/v4/runtime/tree/xpath/XPathLexer.java

bild.py
```diff
@@ -42,20 +40,27 @@
 PYTHON3_TARGET	= "../antlr4-python3"
 CSHARP_TARGET	= "../antlr4-csharp"
 
+JAVA_VERSION = os.environ.get('ANTLR_JAVA_VERSION', None)
```
ericvergnaud: this could be:
JAVA_VERSION = os.environ.get('ANTLR_JAVA_VERSION', '1.6')

bild.py
```diff
@@ -140,21 +145,30 @@ def mkjar():
 	mkjar_complete() # make it again with up to date XPath lexer
 	mkjar_runtime()	 # now build the runtime jar
 
+def mkdist():
```
ericvergnaud: just remove this since JAVA_VERSION is already initialized

parrt:
I think i have updated a few tiny things Eric didn't fix so I'll close this now. Thanks much, @bdkearns!

PR #761, opened 2014-11-22
Fix spacing in MANIFEST file

jrozner:
The manifest files in jars are apparently pretty picky with respect to
spacing and don't like having the spaces in front of the keys. If they
do the file is ignored and jar just generates a blank one, conveniently
leaving out the Main-Class directive making the jar useless. Given the
literal quotes in python the rest of the lines need to be moved over to
avoid adding extra whitespace.

### DIFF
--- a/bild.py
+++ b/bild.py
@@ -83,12 +83,12 @@ def mkjar_complete():
     copytree(src="tool/resources", trg="out")  # messages, Java code gen, etc...
     manifest = \
         """Main-Class: org.antlr.v4.Tool
-        Implementation-Vendor: ANTLR
-        Implementation-Title: ANTLR 4 Tool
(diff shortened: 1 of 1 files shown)


parrt:
looks like Eric fixed this as follows:
    manifest = \
        "Main-Class: org.antlr.v4.Tool\n" +\
        "Implementation-Title: ANTLR 4 Tool\n" +\
        "Implementation-Version: %s\n" +\
        "Implementation-Vendor: ANTLR\n" +\
        "Implementation-Vendor-Id: org.antlr\n" +\
        "Built-By: %s\n" +\
        "Build-Jdk: 1.6\n" +\
        "Created-By: http://www.bildtool.org\n" +\
        "\n"

 so I am closing.

PR #792, opened 2015-01-17
Restore groupId and version in POM

jcttrll:
There are quite a few issues with building via Maven right now; among them, the `tool/pom.xml` is missing `<groupId>` and `<version>` (previously defined by the parent POM that has been removed).

### DIFF
--- a/tool/pom.xml
+++ b/tool/pom.xml
@@ -1,7 +1,9 @@
 <project xmlns="http://maven.apache.org/POM/4.0.0" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/maven-v4_0_0.xsd">
     <modelVersion>4.0.0</modelVersion>
 
+    <groupId>org.antlr</groupId>
     <artifactId>antlr4</artifactId>
+    <version>4.5-SNAPSHOT</version>
 
     <name>ANTLR 4 Tool</name>
     <description>The ANTLR 4 grammar compiler.</description>


ericvergnaud:
Hi, 
maven build is no longer supported.
The pom's in Tools and runtime are meant for publishing to maven central only.
The only supported build method is via bold.py

Eric

Envoyé de mon iPhone

jcttrll:
OK, understood, and thanks for the info. However, I believe Maven will have an error even processing the POM to do the deployment to the Central Repository without a version and groupId.

ericvergnaud:
Those are passed by cmd line parameters, and seems to work with maven central snapshot repo.

Envoyé de mon iPhone

parrt:
yep, we definitely pass those on the command line to the deploy. Take a look at [how we deploy maven artifacts](https://github.com/antlr/antlr4/wiki/Deploying-ANTLR-mvn-artifacts).   let's hope that the release deploy works this way. If not we can add them back and so I will leave this open. thanks.

jcttrll:
That makes sense, and I'm confident that if you can deploy to the snapshot repo, you'll be able to do a release with the same mechanism. I have to say the `bild.py` and your bilder tool are a bit of a surprise and I expected to be able to use Maven on the POMs to build everything. :-)

I'd suggest closing this pull request; in the very unlikely case the release deploy doesn't honor your command line properties for some obscure reason, it will be obvious what the problem is from the Maven output (it will explicitly say it's missing these two properties).

Thank you both.

parrt:
Hi. Yep, bild.py is a response to complexity of modern build tools: ant, maven, gradle, etc...  I miss `make` with some improvements.

Issue #795, opened 2015-01-19
Excess token consumption during recovery


michaelpj:
The method `recoverInline` in `ANTLRErrorStrategy` returns a `Token`, however, in my generated parsers, I see it used like this:
```java
		try {
			enterOuterAlt(_localctx, 1);
			{
			setState(479);
			_la = _input.LA(1);
			if ( !(_la==TKminus || _la==TKplus) ) {
			_errHandler.recoverInline(this);
			}
			consume();
			}
		}
```
That is, the `Token` produced from `recoverInline` is ignored. 

This can lead to excess token consumption, as the `DefaultErrorStrategy` will call `consume` if it deletes a token, returning the matched token. (It calls it first in `singleTokenDeletion` to delete the token, but then again in `recoverInline`). But then the parser will `consume` once again, so we also consume the token *after* the one we wanted.

Needless to say, this can really mess up error recovery.

It's not clear to me from the documentation of `recoverInline` whether a successful return indicates that the token has been consumed or not. I suspect it does, since `singleTokenInsertion` does not in fact insert a token into the stream that could later be consumed. Hence, the fix is probably just to put the generated `consume` in an `else` block.

michaelpj:
This is with 4.4 - I haven't tried with 4.5 RC1 yet.

parrt:
The goal of in-line recovery is to modify the token stream such that there is a chance the parser can continue from the current error. Can you give me the grammar fragment? It still looks okay because if the current input symbol is not a plus or minus, the recovery will consume tokens until it sees one in look ahead (not consumed), which will then be consumed by the mechanism afterwards. We have not seen any bugs related to this. Is this merely an efficiency issue?

sharwell:
The result of `recoverInline` is assigned to the label variable if your token in the grammar is labeled.

There are three different behaviors the default implementation of `recoverInline` gives. Suppose you are trying to match a token `Y` followed by `Z`.

1. If the input sequence is LA(1)=`X`, LA(2)=`Y`, (i.e. skipping `X` will allow the parser to match `Y`), then we call `consume` to skip `X` and then return `LT(1)`. This code path should work fine. When the call returns, LA(1)=`Y` which means we *should* call `consume` (which it does).
2. If the input sequence is LA(1)=`Z`, then assume the `Y` is missing, and we construct a new token `Y`. When the call returns, LA(1)=`Z` which means we *should not* call `consume`. **This code appears to have a bug.**
[...]

PR #797, opened 2015-01-20
Fix excess token consumption after `recoverInline`

michaelpj:
Fixes: https://github.com/antlr/antlr4/issues/795

As discussed, this changes the code generation for `Parser`s so that they only `consume` when they do *not* need to `recoverInline`. I've added tests which fail before the change (in both cases, the `set` rule's final consumed token was `d`), and updated the documentation to warn that `recoverInline` will handle it's own token consumption.

(As a side note, I see that the build and test infrastructure has changed - it would be nice if there was a little bit of documentation about that. It took me a while to figure out the test generator, and I could never get `bild.py` to do what I wanted.) [Edit: parrt adding link to [building ANTLR itself](https://github.com/antlr/antlr4/wiki/How-to-build-ANTLR-itself)]

### DIFF
--- a/runtime/Java/src/org/antlr/v4/runtime/ANTLRErrorStrategy.java
+++ b/runtime/Java/src/org/antlr/v4/runtime/ANTLRErrorStrategy.java
@@ -63,6 +63,9 @@ public interface ANTLRErrorStrategy {
 	 * returns the {@link Token} instance which should be treated as the
 	 * successful result of the match.
 	 *
+   * <p>This method handles the consumption of any tokens - the caller should
+	 * <b>not</b> call {@link Parser#consume} after a successful recovery.</p>
+	 *
 	 * <p>Note that the calling code will not report an error if this method
 	 * returns successfully. The error strategy implementation is responsible
 	 * for calling {@link Parser#notifyErrorListeners} as appropriate.</p>
--- a/tool/resources/org/antlr/v4/tool/templates/codegen/Java/Java.stg
+++ b/tool/resources/org/antlr/v4/tool/templates/codegen/Java/Java.stg
@@ -647,8 +647,9 @@ setState(<m.stateNumber>);
 <capture>
 if ( <if(invert)><m.varName> \<= 0 || <else>!<endif>(<expr>) ) {
 	<if(m.labels)><m.labels:{l | <labelref(l)> = (Token)}><endif>_errHandler.recoverInline(this);
+} else {
+	consume();
 }
-consume();
 >>
 
 Wildcard(w) ::= <<
--- a/tool/test/org/antlr/v4/test/rt/gen/Generator.java
+++ b/tool/test/org/antlr/v4/test/rt/gen/Generator.java
@@ -585,6 +585,10 @@ public class Generator {
 				"aab",
 				"",
 				"line 1:1 extraneous input 'a' expecting {'b', 'c'}\n");
+		file.addParserTest(input, "SingleTokenDeletionConsumption", "T", "a",
+				"aabd",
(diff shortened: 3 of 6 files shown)


runtime/Java/src/org/antlr/v4/runtime/ANTLRErrorStrategy.java
```diff
@@ -63,6 +63,9 @@
 	 * returns the {@link Token} instance which should be treated as the
 	 * successful result of the match.
 	 *
+	 * This method handles the consumption of any tokens - the caller should
```
sharwell: This paragraph should be wrapped in `<p>` and `</p>` tags so it displays properly when the documentation is generated.

tool/resources/org/antlr/v4/tool/templates/codegen/Java/Java.stg
```diff
@@ -647,8 +647,9 @@ setState(<m.stateNumber>);
 <capture>
 if ( <if(invert)><m.varName> \<= 0 || <else>!<endif>(<expr>) ) {
 	<if(m.labels)><m.labels:{l | <labelref(l)> = (Token)}><endif>_errHandler.recoverInline(this);
+} else {
+  consume();
```
sharwell: Can you change this to a hard tab? You would not believe the file size difference we found by using hard tabs exclusively in the generated code.

tool/test/org/antlr/v4/test/rt/gen/Generator.java
```diff
@@ -597,6 +601,10 @@ private JUnitTestFile buildParserErrors() throws Exception {
 				"ad",
 				"",
 				"line 1:1 missing {'b', 'c'} at 'd'\n");
+		file.addParserTest(input, "SingleSetInsertionConsumption", "T", "a",
```
sharwell: What was the output of this test using the old code generation? It seems similar to the `SingleTokenInsertion` test, so I wanted to make sure it's actually capable of detecting the error.

My instinct was the best way to test this is looking at both the error message output and the resulting parse tree, but if this can detect the bug in the old code then I'm ok with it.

tool/test/org/antlr/v4/test/rt/gen/Generator.java
```diff
@@ -585,6 +585,10 @@ private JUnitTestFile buildParserErrors() throws Exception {
 				"aab",
 				"",
 				"line 1:1 extraneous input 'a' expecting {'b', 'c'}\n");
+		file.addParserTest(input, "SingleTokenDeletionConsumption", "T", "a",
```
sharwell: See the comment I made for `SingleSetInsertionConsumption`; this one relates to the `SingleTokenDeletion` test in the same way.

sharwell:
I made a few notes. One thing you might find interesting in the future - if you add the tests first and then the fix, you can always check out the commit containing the tests to see they fail, and then check out the commit with the fix to verify that the test now passes.

michaelpj:
Hm, I'd actually already rebased the commits into that order for precisely the reason you mention. I think the display was a github bug after force pushes - I've seen that before. Anyway, after pushing to fix the formatting issues it seems to have sorted itself out. You should be able to check out the first commit and see the tests fail.

As I mentioned in the PR description, the failure is that `$set.stop` reports the `d` token in both cases, showing that the `set` rule has consumed one token more than it should do. You also get a corresponding complaint about a missing `d` token in the stderr - since it's already been consumed, the final `'d'` in the grammar is unable to match.

This was the simplest way I could think of to exhibit the problem, so they're really regression tests. This area of the code could probably do with some more tests anyway, but it sounds like you have something in mind.

PR #805, opened 2015-01-26
Fix potential misuse of the DFA start state when initializing a decision

### LINKED ISSUE #804: Random ArrayIndexOutOfBounds exceptions coming from ParserATNSimulator
It looks like a transient problem, even in the environment where we saw the issue originally it hasn't been reproducible. Only happens in our automation environment (rehat 5.7 running on open stack running on I dunno what) and only occasionally.

It happens in here:

```java
synchronized (from) {
    if ( from.edges==null ) { 
        from.edges = new DFAState[atn.maxTokenType+1+1];
    }   

    from.edges[t+1] = to; // connect
}   
```

so in some case `atn.maxTokenType+1 > t` must be true.   

I checked and it looks like the field "edges" is always protected by its parent `DFAState` object except for in one suspicious place where it appears to be guarded by an instance of `DFA` in `DFA.java`'s `setPrecedenceDfa` method.  (all in 4.3)

It would seem on the face of it that you need to synchronize inside `setPrecedenceDfa` on the `DFAState` rather than the `DFA` (or perhaps in addition too?).  

Maybe we should just add an assert there.   

```java
assert t +1 < from.edges.length : "array index out of bounds!, t = " + t + ", edges.length=" + from.edges.length;
```

And perhaps add something about which threads are active and their stacks if it fails?

sharwell:
This should fix #804

**Note:** This code has not been compiled or tested as the build and test configuration has been removed from the `master` branch.

### DIFF
--- a/runtime/Java/src/org/antlr/v4/runtime/atn/ParserATNSimulator.java
+++ b/runtime/Java/src/org/antlr/v4/runtime/atn/ParserATNSimulator.java
@@ -369,36 +369,22 @@ public class ParserATNSimulator extends ATNSimulator {
 		// But, do we still need an initial state?
 		try {
 			DFAState s0;
-			/* Threading note: dfa.s0 is the final value set on the DFA instance
-			 * which affects branch choice in this method. If dfa.s0 is null
-			 * (i.e. it has not been set), then we take the "slow" path through
-			 * the initialization code to ensure a correct start state is used.
-			 */
-			if (dfa.s0 == null) {
+			if (dfa.isPrecedenceDfa()) {
+				// the start state for a precedence DFA depends on the current
+				// parser precedence, and is provided by a DFA method.
+				s0 = dfa.getPrecedenceStartState(parser.getPrecedence());
+			}
+			else {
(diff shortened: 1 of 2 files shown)


parrt:
Cool.  Did you get a chance to check whether or not it affects performance?

sharwell:
I am literally unable to compile the code anymore, so I cannot perform any form of functionality or performance testing.

A review of the code indicates that the fast path (taken after the DFA is initialized) contains exactly one extra read of a `volatile` field, but is otherwise unchanged.

sharwell:
The latest update changes the behavior to initialize `DFA.precedenceDfa` when the `DFA` instance is constructed. The `setPrecedenceDfa` method was marked deprecated (rather than removed) to avoid introducing a source- and binary-breaking change.

cberner:
Any idea when 4.5.1 will be released? We see this in our tests quite frequently, and it makes them fail non-deterministically

parrt:
I won't be able to release probably til May sometime. crazy busy.

parrt:
heading towards a 4.5.1 release soon

PR #812, opened 2015-02-02
Makes TestRig's dialog to save and restore its state

pavlo:
It changes the TestRig's JDialog to remember its state between launches. When it gets shutdown, it stores the following information using [standard Java Preferences API] (http://docs.oracle.com/javase/8/docs/technotes/guides/preferences/overview.html) and restores them during consequent launch:

* Dialog's width and height
* Dialog's position on screen
* The TreeView component's scale factor
* The placement of the splitpane divider


parrt:
Hi. Can you add yourself to the contributors.txt file?

pavlo:
Added. Thanks Terence!

PR #842, opened 2015-03-20
Added error checks.

peturingi:
Added error checks.
Removed a confusing variable 'current' from addAll(Collection<Integer> list).

In 'int get(int index)', it might be more robust to invert fromIndex and toIndex.

### DIFF
--- a/runtime/Java/src/org/antlr/v4/runtime/misc/IntegerList.java
+++ b/runtime/Java/src/org/antlr/v4/runtime/misc/IntegerList.java
@@ -58,8 +58,6 @@ public class IntegerList {
 		if (capacity < 0) {
 			throw new IllegalArgumentException();
 		}
-        if (capacity > MAX_ARRAY_SIZE)
-          throw new IllegalArgumentException();
 
 		if (capacity == 0) {
 			_data = EMPTY_DATA;
@@ -103,9 +101,12 @@ public class IntegerList {
 
 	public final void addAll(Collection<Integer> list) {
 		ensureCapacity(_size + list.size());
-		for (int x : list) {
-          _data[_size++] = x;
-        }
+		int current = 0;
+    		for (int x : list) {
+      			_data[_size + current] = x;
+      			current++;
+    		}
+    		_size += list.size();
 	}
 
 	public final int get(int index) {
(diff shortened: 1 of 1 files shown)


runtime/Java/src/org/antlr/v4/runtime/misc/IntegerList.java
```diff
@@ -57,6 +58,8 @@ public IntegerList(int capacity) {
 		if (capacity < 0) {
 			throw new IllegalArgumentException();
 		}
+        if (capacity > MAX_ARRAY_SIZE)
+          throw new IllegalArgumentException();
```
sharwell: :bulb: Considering `ArrayList` omits this check, I think it's safe for us to as well.

runtime/Java/src/org/antlr/v4/runtime/misc/IntegerList.java
```diff
@@ -100,13 +103,9 @@ public final void addAll(IntegerList list) {
...
 	public final void addAll(Collection<Integer> list) {
 		ensureCapacity(_size + list.size());
-		int current = 0;
 		for (int x : list) {
-			_data[_size + current] = x;
-			current++;
-		}
-
-		_size += list.size();
+          _data[_size++] = x;
```
sharwell: :exclamation: Even though it's not documented (and the class is not safe for concurrent operations on multiple threads), the original implementation is saf*er* for single-writer, multiple-reader scenarios. The new implementation contains a sequence point where `list.get(list.size() - 1)` returns an uninitialized value, even if memory fences are employed.

runtime/Java/src/org/antlr/v4/runtime/misc/IntegerList.java
```diff
@@ -268,6 +267,8 @@ public final int binarySearch(int fromIndex, int toIndex, int key) {
 		if (fromIndex < 0 || toIndex < 0 || fromIndex > _size || toIndex > _size) {
 			throw new IndexOutOfBoundsException();
 		}
+        if (fromIndex > toIndex)
```
sharwell: I think this condition makes sense. Can you update the code to match the surrounding formatting (hard tabs, and use braces)?

peturingi: Yes.

parrt:
Hi. we'll need you to add yourself to https://github.com/antlr/antlr4/blob/master/contributors.txt

peturingi:
Yes, done.

PR #870, opened 2015-05-12
Add goto to reserved words for java.

Pursuit92:
Goto isn't in the reserved words, so it doesn't get caught when compiling a grammar that uses it as a rule name.


parrt:
Hi. i know it's a simple change, but can you add yourself to contributors.txt?

Pursuit92:
Done, just amended the commit

parrt:
thanks and welcome aboard :)

PR #887, opened 2015-06-05
ignore subdirectory "gen"

hethi:
Hi,
I just followed the build instructions and found that git wouldn't ignore the generated subdirectory 'gen', so it might be added to .gitignore.
Have fun,
Herbert.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -34,6 +34,7 @@ nbactions*.xml
 # Generated files
 out/
 doc/
+gen/
 gen3/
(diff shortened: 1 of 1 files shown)


sharwell:
:exclamation: This pull request would exclude the following source folder:

https://github.com/antlr/antlr4/tree/master/tool/test/org/antlr/v4/test/rt/gen

parrt:
Thanks. I'll try /gen/

PR #888, opened 2015-06-05
Report an end position when reporting syntax errors

michaelpj:
Clients often want to report a span position for errors, rather than just a point position. Currently, an end position can be reconstructed from the token length for parser errors, but lexer errors do not have tokens attached. Nonetheless, we do know the ending index of the error, and it is helpful to report it.

For comparison, what I'm currently doing is using a regex to match the `token error at '...'` error message, pull the offending text out, and use the length of that. That's both slow and fragile, so something like this seems like a better approach.

This changes the interface of `ANTLRErrorListener`, so potentially breaks people's subclasses. Ideally, it would be nice to do this without changing the interface, but I don't see an obvious way to do that.

### DIFF
--- a/runtime/Java/src/org/antlr/v4/runtime/ANTLRErrorListener.java
+++ b/runtime/Java/src/org/antlr/v4/runtime/ANTLRErrorListener.java
@@ -30,13 +30,13 @@
 
 package org.antlr.v4.runtime;
 
+import java.util.BitSet;
+
 import org.antlr.v4.runtime.atn.ATNConfigSet;
 import org.antlr.v4.runtime.atn.DecisionInfo;
 import org.antlr.v4.runtime.atn.ParserATNSimulator;
 import org.antlr.v4.runtime.atn.PredictionMode;
 import org.antlr.v4.runtime.dfa.DFA;
-
-import java.util.BitSet;
 
 /** How to emit recognition errors. */
 public interface ANTLRErrorListener {
@@ -63,8 +63,10 @@ public interface ANTLRErrorListener {
 	 * 		  started production for the decision.
 	 * @param line
 	 * 		  The line number in the input where the error occurred.
-	 * @param charPositionInLine
+	 * @param startPositionInLine
 	 * 		  The character position within that line where the error occurred.
+	 * @param endPositionInLine
+	 * 		  The character position within that line where the error ended.
 	 * @param msg
 	 * 		  The message to emit.
 	 * @param e
@@ -76,7 +78,8 @@ public interface ANTLRErrorListener {
 	public void syntaxError(Recognizer<?, ?> recognizer,
 							Object offendingSymbol,
 							int line,
-							int charPositionInLine,
+							int startPositionInLine,
+							int endPositionInLine,
 							String msg,
 							RecognitionException e);
 
--- a/runtime/Java/src/org/antlr/v4/runtime/Lexer.java
+++ b/runtime/Java/src/org/antlr/v4/runtime/Lexer.java
@@ -29,14 +29,14 @@
  */
 package org.antlr.v4.runtime;
 
+import java.util.ArrayList;
+import java.util.EmptyStackException;
+import java.util.List;
+
 import org.antlr.v4.runtime.atn.LexerATNSimulator;
 import org.antlr.v4.runtime.misc.IntegerStack;
 import org.antlr.v4.runtime.misc.Interval;
 import org.antlr.v4.runtime.misc.Pair;
-
-import java.util.ArrayList;
-import java.util.EmptyStackException;
-import java.util.List;
 
 /** A lexer is recognizer that draws input symbols from a character stream.
  *  lexer grammars result in a subclass of this object. A Lexer object
@@ -383,7 +383,7 @@ public abstract class Lexer extends Recognizer<Integer, LexerATNSimulator>
 		String msg = "token recognition error at: '"+ getErrorDisplay(text) + "'";
 
 		ANTLRErrorListener listener = getErrorListenerDispatch();
-		listener.syntaxError(this, null, _tokenStartLine, _tokenStartCharPositionInLine, msg, e);
+		listener.syntaxError(this, null, _tokenStartLine, _tokenStartCharPositionInLine, _input.index(), msg, e);
 	}
 
 	public String getErrorDisplay(String s) {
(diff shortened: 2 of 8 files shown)


sharwell:
This is a high-impact breaking change, in the sense that it's a commonly-implemented interface for clients. I have some ideas on how to address the main concern (will add when I'm not typing on my phone), but I'm very much against changing this interface.

sharwell:
Hi @michaelpj,

There is a simple rule to follow to ensure this is never a problem: write your lexer in a way that ensures all input is tokenized without errors. See [this lexer grammar](https://github.com/tunnelvisionlabs/antlr4-grammar-postgresql/blob/4c4d8e6a8f3057b319365ec947e6e76cc5a22e14/src/com/tunnelvisionlabs/postgresql/PostgreSqlLexer.g4#L578-L607) for one example I'm particularly happy with.

If the specialized rules like you see in that link don't cover every case, you still include an [`ErrorCharacter`](https://github.com/tunnelvisionlabs/antlr4-grammar-postgresql/blob/4c4d8e6a8f3057b319365ec947e6e76cc5a22e14/src/com/tunnelvisionlabs/postgresql/PostgreSqlLexer.g4#L807-L816) rule to ensure your lexer will not report an error for any input, ever.

michaelpj:
As you correctly guessed, the input that we tend to get lexer errors for is quoted strings. Those being the most complex things that most people do in their lexers! ;)

However, I'm not sure that I want to never get lexer errors. The lexer seems like a perfectly good place to report that you have an unterminated string literal. Kicking it back to the parser to distinguish between the various kinds of valid or invalid string literal is fine, but I think it ought to be viable to report the errors at the point at which you first recognize them: in the lexer.

Regardless of whether it's better to avoid lexer errors, if we've *got* such a mechanism, it would be nice if it could provide us with the error information that we need! It's not like the information isn't there - it's particularly frustrating because the error message includes precisely the text that we want to cover with our error span!

sharwell:
> The lexer seems like a perfectly good place to report that you have an unterminated string literal. Kicking it back to the parser to distinguish between the various kinds of valid or invalid string literal is fine, but I think it ought to be viable to report the errors at the point at which you first recognize them: in the lexer.

This might seem like a good idea initially, but not when you view it in the context of the lexer's error recovery mechanism. If the lexer sees an unterminated quote and you don't have handling in for it like I created, the recovery behavior is to pretend the opening `"` character doesn't exist and proceed to tokenize the contents of the intended string as though they weren't in the string at all.

The primary purpose of the error reporting mechanism in the lexer is to let grammar developers know when they failed to account for some input within the lexer.

parrt:
Hi Michael. I think you're right that the end index should be provided in the listener but, as Sam says, it's a tough change to introduce at this point. could you ask the lexer for the end index (maybe not) or simply as it for token text and then end = start + text.length?

runtime/Java/src/org/antlr/v4/runtime/Lexer.java
```diff
@@ -383,7 +383,7 @@ public void notifyListeners(LexerNoViableAltException e) {
 		String msg = "token recognition error at: '"+ getErrorDisplay(text) + "'";
 
 		ANTLRErrorListener listener = getErrorListenerDispatch();
-		listener.syntaxError(this, null, _tokenStartLine, _tokenStartCharPositionInLine, msg, e);
+		listener.syntaxError(this, null, _tokenStartLine, _tokenStartCharPositionInLine, _input.index(), msg, e);
```
sharwell: Based on this line, I'm assuming the listener already has access to this information by the following:

```java
public void syntaxError(Recognizer<?, ?> recognizer, ...) {
  int endIndex = recognizer.getInputStream().index();
  ...
}
```

:question: Is this not correct?

runtime/Java/src/org/antlr/v4/runtime/ANTLRErrorListener.java
```diff
@@ -76,7 +78,8 @@
 	public void syntaxError(Recognizer<?, ?> recognizer,
 							Object offendingSymbol,
 							int line,
-							int charPositionInLine,
+							int startPositionInLine,
+							int endPositionInLine,
```
sharwell: :exclamation: This change assumes the token (or error) ends on the same line where recognition started. This is not always the case.

runtime/Java/src/org/antlr/v4/runtime/Parser.java
```diff
@@ -29,6 +29,13 @@
  */
 package org.antlr.v4.runtime;
 
+import java.util.ArrayList;
```
sharwell: :bulb: Your code formatting settings are moving things around. You can prevent these types of changes from appearing in a pull request by reviewing each file of a commit prior to committing it, and undoing any changes which aren't clearly part of the change you are working on.

runtime/Java/src/org/antlr/v4/runtime/Parser.java
```diff
@@ -558,7 +558,7 @@ public void notifyErrorListeners(Token offendingToken, String msg,
 		charPositionInLine = offendingToken.getCharPositionInLine();
 
 		ANTLRErrorListener listener = getErrorListenerDispatch();
-		listener.syntaxError(this, offendingToken, line, charPositionInLine, msg, e);
+		listener.syntaxError(this, offendingToken, line, charPositionInLine, charPositionInLine + offendingToken.getText().length(), msg, e);
```
sharwell: :exclamation: The [`getStopIndex()`](http://www.antlr.org/api/Java/org/antlr/v4/runtime/Token.html#getStopIndex()) and `getStartIndex()` methods likely provide better behavior here, considering users may explicitly set the text. The following expression would be a more accurate representation of this change:

```
charPositionInLine + offendingToken.getStopIndex() - offendingToken.getStartIndex() + 1
```

:exclamation: This line assumes `offendingToken` does not span multiple lines, but in practice many tokens do.

michaelpj:
Okay, I'm convinced that I can work around this more reasonably by interrogating the lexer for where it's got to - the fact that you've got access to the `Recognizer` in the error listener actually gives you significantly more options than I'd realised.

With that in mind, I'm going to close this, as an interface change probably isn't justified here. Indeed, as Sam rightly points out, to do this properly we really ought to be reporting `startLine`, `startIndex`, `stopLine`, and `stopIndex`. In fact, I've realised that my handling of error spans for parser errors is also wrong in the presence of tokens that span multiple lines (I don't have any of those, but still).

Given that this is (at least for me!) something that it's easy to get wrong, it *would* be nice if the user interface provided all the span information directly. Perhaps something to think about next time there's a major version and you feel like breaking interfaces ;)

parrt:
Indeed. Excellent analysis Michael.

PR #898, opened 2015-06-11
Check eof in tree

### LINKED ISSUE #896: EOF leaf not included in parent tree source interval
[Test case shows issue](https://github.com/parrt/antlr4/blob/check-eof-in-tree/tool/test/org/antlr/v4/test/tool/TestParserInterpreter.java#L72).  Tree shows `<EOF>` but source interval doesn't include it.  Whoops. Culprit is `Parser`:

```java
    public void exitRule() {
		_ctx.stop = _input.LT(-1); <-- ooops; doesn't count EOF
```

Issue is that consume() of EOF doesn't move p in token stream so `LT(-1)` doesn't include it. Not sure of best solution. Must detect match of EOF.  Can't test `LT(1)==EOF` as that can't distinguish "about to match EOF" from "matched EOF".

### LINKED ISSUE #897: Empty rule matched first gives invalid source interval
```
s : x A ;
x : ;
```

Gives invalid interval for x (INVALID -1..-2) not 0..-1.

parrt:
Fixes #896 and Fixes #897.

### DIFF
--- a/runtime/Java/src/org/antlr/v4/runtime/Parser.java
+++ b/runtime/Java/src/org/antlr/v4/runtime/Parser.java
@@ -177,6 +177,9 @@ public abstract class Parser extends Recognizer<Token, ParserATNSimulator> {
 	 */
 	protected int _syntaxErrors;
 
+	/** Indicates parser has match()ed EOF token */
+	protected boolean matchedEOF;
+
 	public Parser(TokenStream input) {
 		setInputStream(input);
 	}
@@ -217,6 +220,9 @@ public abstract class Parser extends Recognizer<Token, ParserATNSimulator> {
 	public Token match(int ttype) throws RecognitionException {
 		Token t = getCurrentToken();
 		if ( t.getType()==ttype ) {
+			if ( t.getType()==Token.EOF ) {
+				matchedEOF = true;
+			}
 			_errHandler.reportMatch(this);
 			consume();
 		}
@@ -630,7 +636,9 @@ public abstract class Parser extends Recognizer<Token, ParserATNSimulator> {
 	}
 
     public void exitRule() {
-		_ctx.stop = _input.LT(-1);
+		if ( matchedEOF ) _ctx.stop = _input.LT(1);
+		else _ctx.stop = _input.LT(-1);
+//		_ctx.stop = _input.LT(-1);
         // trigger event on _ctx, before it reverts to parent
         if ( _parseListeners != null) triggerExitRuleEvent();
 		setState(_ctx.invokingState);
@@ -908,6 +916,10 @@ public abstract class Parser extends Recognizer<Token, ParserATNSimulator> {
 
         return false;
     }
+
+	public boolean isMatchedEOF() {
+		return matchedEOF;
+	}
 
 	/**
 	 * Computes the set of input symbols which could follow the current parser
--- a/runtime/Java/src/org/antlr/v4/runtime/Parser.java
+++ b/runtime/Java/src/org/antlr/v4/runtime/Parser.java
@@ -177,7 +177,7 @@ public abstract class Parser extends Recognizer<Token, ParserATNSimulator> {
 	 */
 	protected int _syntaxErrors;
 
-	/** Indicates parser has match()ed EOF token */
+	/** Indicates parser has match()ed EOF token. See {@link #exitRule()}. */
 	protected boolean matchedEOF;
 
 	public Parser(TokenStream input) {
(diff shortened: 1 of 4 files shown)


runtime/Java/src/org/antlr/v4/runtime/Parser.java
```diff
@@ -217,6 +220,9 @@ public void reset() {
 	public Token match(int ttype) throws RecognitionException {
 		Token t = getCurrentToken();
 		if ( t.getType()==ttype ) {
+			if ( t.getType()==Token.EOF ) {
```
sharwell: :bulb: `t.getType()` could be replaced by `ttype` for efficiency.

runtime/Java/src/org/antlr/v4/runtime/ParserRuleContext.java
```diff
@@ -272,7 +272,14 @@ public TerminalNode getToken(int ttype, int i) {
 
 	@Override
 	public Interval getSourceInterval() {
-		if ( start==null || stop==null ) return Interval.INVALID;
+		if ( stop==null ) {
+			if ( start!=null ) {
+				return Interval.of(start.getTokenIndex(), start.getTokenIndex()-1); // empty
+			}
+			else {
+				return Interval.of(0, -1);
```
sharwell: This should be `Interval.INVALID` since we really have no idea what the source interval is.

parrt: ok

runtime/Java/src/org/antlr/v4/runtime/ParserRuleContext.java
```diff
@@ -272,7 +272,14 @@ public TerminalNode getToken(int ttype, int i) {
...
+			}
+		}
 		return Interval.of(start.getTokenIndex(), stop.getTokenIndex());
```
sharwell: :bulb: One would expect the total number of tokens from the input sequence contained in the source interval (including off-channel tokens) to be the following:

```java
Interval sourceInterval = getSourceInterval();
int tokenCount = sourceInterval.b - sourceInterval.a + 1;
```

However, in the specific case where off-channel tokens appear before an empty parse tree, `tokenCount` could be negative. This could be addressed by changing the `stop==null` condition above to `stop==null || stop.getTokenIndex()<start.getTokenIndex()`.

runtime/Java/src/org/antlr/v4/runtime/ParserRuleContext.java
```diff
@@ -272,7 +272,14 @@ public TerminalNode getToken(int ttype, int i) {
 
 	@Override
 	public Interval getSourceInterval() {
-		if ( start==null || stop==null ) return Interval.INVALID;
+		if ( stop==null ) {
```
sharwell: :exclamation: This method only works when `start` is not null. The first thing in the method should be:

```java
if (start == null) {
  return Interval.INVALID;
}
```

runtime/Java/src/org/antlr/v4/runtime/ParserRuleContext.java
parrt: ah. good catch

sharwell:
:exclamation: In addition to the items in the diff, `reset()` needs to be updated to set `matchedEOF` to false.

parrt:
i could swear i put that in. oh only in my plugin ;)
Ter

runtime/Java/src/org/antlr/v4/runtime/Parser.java
parrt: Might not be using CommonToken, right?

runtime/Java/src/org/antlr/v4/runtime/tree/SyntaxTree.java
```diff
@@ -42,10 +42,14 @@
 	 * Return an {@link Interval} indicating the index in the
 	 * {@link TokenStream} of the first and last token associated with this
 	 * subtree. If this node is a leaf, then the interval represents a single
-	 * token.
+	 * token and has interval i..i for token index i.
+	 *
+	 * <p>An interval of i..j for j &lt; i indicates an empty interval.</p>
```
sharwell: The `getSourceInterval()` method was updated in 81e2a654d772b630933d98ee6a8c1fe08a7d402a to ensure that an empty interval is always represented as i..i-1. I recommend changing this paragraph to the following:

```
<p>An interval of i..i-1 indicates an empty interval at position i in the input stream,
where 0 &lt;= i &lt;= the size of the input token stream.</p>
```

runtime/Java/src/org/antlr/v4/runtime/tree/SyntaxTree.java
```diff
@@ -42,10 +42,14 @@
...
 	 *
 	 * <p>If source interval is unknown, this returns {@link Interval#INVALID}.</p>
+	 *
+	 * <p>As a weird special case, the source interval for rules matched after
```
sharwell: :bulb: Since this is unlikely to appear in actual code, it might make more sense to omit this paragraph and instead file a bug for the known issue.

sharwell:
Two suggestions from reading the updated docs, but :+1:

PR #902, opened 2015-06-13
Add error nodes

parrt:
(crap.  I messed up and included the previous PR in this one; no big deal as I'm merging the other ahead of this.)

### DIFF
--- a/runtime/Java/src/org/antlr/v4/runtime/ParserInterpreter.java
+++ b/runtime/Java/src/org/antlr/v4/runtime/ParserInterpreter.java
@@ -218,7 +218,7 @@ public class ParserInterpreter extends Parser {
 					setState(atn.ruleToStopState[p.ruleIndex].stateNumber);
 					getContext().exception = e;
 					getErrorHandler().reportError(this, e);
-					getErrorHandler().recover(this, e);
+					recover(e);
 				}
 
 				break;
@@ -271,7 +271,7 @@ public class ParserInterpreter extends Parser {
 			case Transition.SET:
 			case Transition.NOT_SET:
 				if (!transition.matches(_input.LA(1), Token.MIN_USER_TOKEN_TYPE, 65535)) {
-					_errHandler.recoverInline(this);
+					recoverInline();
 				}
 				matchWildcard();
 				break;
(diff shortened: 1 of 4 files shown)


runtime/Java/src/org/antlr/v4/runtime/ParserInterpreter.java
```diff
@@ -404,4 +404,46 @@ public void addDecisionOverride(int decision, int tokenIndex, int forcedAlt) {
...
+				int expectedTokenType = ime.getExpectedTokens().getMinElement(); // get any element
+				String tokenText;
+				if ( expectedTokenType== Token.EOF ) tokenText = "<missing EOF>";
+				else tokenText = "<mismatched "+tok.getText()+">";
+
+				Token errToken =
+					getTokenFactory().create(new Pair<TokenSource, CharStream>(tok.getTokenSource(), tok.getTokenSource().getInputStream()),
+				                             expectedTokenType, tok.getText(),
+				                             Token.DEFAULT_CHANNEL,
+				                            -1, -1, // invalid start/stop
```
sharwell: :bulb: If you pass the start/stop from `tok` instead of invalid, you can avoid the expensive call to `tok.getText()` without losing the ability to show the text for an error.

runtime/Java/src/org/antlr/v4/runtime/ParserInterpreter.java
```diff
@@ -404,4 +404,46 @@ public void addDecisionOverride(int decision, int tokenIndex, int forcedAlt) {
...
+				                             tok.getLine(), tok.getCharPositionInLine());
+				_ctx.addErrorNode(errToken);
+			}
+			else { // NoViableAlt
+				Token tok = e.getOffendingToken();
+				Token errToken =
+					getTokenFactory().create(new Pair<TokenSource, CharStream>(tok.getTokenSource(), tok.getTokenSource().getInputStream()),
+				                             Token.INVALID_TYPE, tok.getText(),
+				                             Token.DEFAULT_CHANNEL,
+				                            -1, -1, // invalid start/stop
```
sharwell: :bulb: Same as the previous comment for `tok.getText()`.

runtime/Java/src/org/antlr/v4/runtime/ParserInterpreter.java
parrt: I think it's probably good to leave within valid location as we are conjuring it up out of thin air.

PR #913, opened 2015-06-17
Move TestPerformance.testExpressionGrammar to the runtime test suite

### LINKED ISSUE #909: Move TestPerformance.testExpressionGrammar to the runtime test suite
The [`testExpressionGrammar`](https://github.com/antlr/antlr4/blob/7d36891c827ef214e6c6e59c4f86772f6db89a2b/tool/test/org/antlr/v4/test/tool/TestPerformance.java#L1946-L1997) test is not target specific and tests a grammar and input which is known to expose performance problems associated with multiple algorithms in the ATN simulator. It should be added to the runtime test suite so other targets are validated in addition to Java.

sharwell:
Fixes #909


parrt:
Cool. i like it.

PR #914, opened 2015-06-18
add new interpreter that knows how to track alternatives in rule nodes

parrt:
some cleanup and improvement to other interpreter.

### DIFF
--- a/runtime/Java/src/org/antlr/v4/runtime/ParserInterpreter.java
+++ b/runtime/Java/src/org/antlr/v4/runtime/ParserInterpreter.java
@@ -321,6 +321,10 @@ public class ParserInterpreter extends Parser {
 		setState(transition.target.stateNumber);
 	}
 
+	/** Method visitDecisionState() is called when the interpreter reaches
+	 *  a decision state (instance of DecisionState). It gives an opportunity
+	 *  for subclasses to track interesting things.
+	 */
 	protected int visitDecisionState(DecisionState p) {
 		int predictedAlt = 1;
 		if ( p.getNumberOfTransitions()>1 ) {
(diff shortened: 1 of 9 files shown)


runtime/Java/src/org/antlr/v4/runtime/ParserInterpreter.java
```diff
@@ -100,6 +99,16 @@
 	protected int overrideDecision = -1;
 	protected int overrideDecisionInputIndex = -1;
 	protected int overrideDecisionAlt = -1;
+	protected boolean overrideDecisionReached = false; // latch and only override once; error might trigger infinite loop
```
sharwell: :bulb: This can be initialized to `true` if there is no override decision set, which would save some comparison operations during `visitDecision`.

tool/src/org/antlr/v4/tool/GrammarParserInterpreter.java
```diff
@@ -0,0 +1,318 @@
...
+	 */
+	protected final Grammar g;
+
+	protected BitSet decisionStatesThatSetOuterAltNumInContext;
+
+	/** Cache {@link LeftRecursiveRule#getPrimaryAlts()} and
+	 *  {@link LeftRecursiveRule#getRecursiveOpAlts()} for states in
+	 *  {@link #decisionStatesThatSetOuterAltNumInContext}.
+	 */
+	protected final Map<DecisionState, int[]> stateToAltsMap = new HashMap<DecisionState, int[]>();
```
sharwell: :bulb: If performance accessing this map matters, this would be much faster as an `ArrayList<int[]>`, indexed by decision number.

runtime/Java/src/org/antlr/v4/runtime/ParserInterpreter.java
parrt: mv'd to end of expr

tool/src/org/antlr/v4/tool/GrammarParserInterpreter.java
parrt: yeah, i like that better. gracias

PR #915, opened 2015-06-18
Small refactoring for flexibility of subclasses (token stream, tree view)

parrt:


runtime/Java/src/org/antlr/v4/runtime/BufferedTokenStream.java
```diff
@@ -209,7 +209,7 @@ public Token get(int i) {
 		if ( start<0 || stop<0 ) return null;
 		lazyInit();
 		List<Token> subset = new ArrayList<Token>();
-		if ( stop>=tokens.size() ) stop = tokens.size()-1;
+		if ( stop>=size() ) stop = size()-1;
```
sharwell: :exclamation: This is part of a bounds check for directly indexing `tokens`. The `tokens.get(i)` expression that follows needs to be updated to `this.get(i)` so the `size()` and `get(int)` methods are paired.

runtime/Java/src/org/antlr/v4/runtime/BufferedTokenStream.java
```diff
@@ -235,9 +235,9 @@ public Token LT(int k) {
 
 		int i = p + k - 1;
 		sync(i);
-        if ( i >= tokens.size() ) { // return EOF token
+        if ( i >= size() ) { // return EOF token
             // EOF must be last token
-            return tokens.get(tokens.size()-1);
+            return tokens.get(size()-1);
```
sharwell: :exclamation: This needs to be the following:

```java
return get(size() - 1);
```

runtime/Java/src/org/antlr/v4/runtime/BufferedTokenStream.java
```diff
@@ -290,11 +290,11 @@ public void setTokenSource(TokenSource tokenSource) {
      */
     public List<Token> getTokens(int start, int stop, Set<Integer> types) {
         lazyInit();
-		if ( start<0 || stop>=tokens.size() ||
-			 stop<0  || start>=tokens.size() )
+		if ( start<0 || stop>=size() ||
+			 stop<0  || start>=size() )
```
sharwell: :exclamation: These are currently bounds checks for direct access to `tokens`. (See previous comments for details.)

runtime/Java/src/org/antlr/v4/runtime/BufferedTokenStream.java
```diff
@@ -468,7 +468,7 @@ public String getText(Interval interval) {
 		int stop = interval.b;
         if ( start<0 || stop<0 ) return "";
         lazyInit();
-        if ( stop>=tokens.size() ) stop = tokens.size()-1;
+        if ( stop>=size() ) stop = size()-1;
```
sharwell: :exclamation: This is currently a bound checks for direct access to `tokens`. (See previous comments for details.)

runtime/Java/src/org/antlr/v4/runtime/BufferedTokenStream.java
parrt: makes sense

PR #917, opened 2015-06-18
added -root parameter to generate all targets

ericvergnaud:


runtime-testsuite/src/org/antlr/v4/testgen/TargetConfiguration.java
```diff
@@ -0,0 +1,36 @@
+package org.antlr.v4.testgen;
```
sharwell: :exclamation: The runtime test suite should not have knowledge of specific targets.

sharwell:
:bulb: Rather than embed target knowledge inside the test suite, consider updating **bild.py** to support generating the test suites for all (and/or some) targets as a separate build target.

sharwell:
:bulb: Actually it seems like the following command already regenerates the tests, with the exception of a few special cases.

```
./bild.py regen_tests
```

All that needs updating is the list of targets to generate (in particular, so it can generate the browser tests).

runtime-testsuite/src/org/antlr/v4/testgen/TargetConfiguration.java
ericvergnaud: This approach was agreed with Terence

ericvergnaud:
bild.py is for releasing antlr, not for generating tests that cannot be run during a release

runtime-testsuite/src/org/antlr/v4/testgen/TargetConfiguration.java
sharwell: The problem with the approach isn't that it won't work. My specific concerns are the following:

1. The runtime test suite is currently *fully* target-independent. Keeping it that way increases its maintainability and generality.
2. The functionality added by this change already exists. If it's not working as expected (or as needed), we should address the reasons why it's not working in practice rather than add a second piece of code to do the same thing.

sharwell:
> bild.py is for releasing antlr, not for generating tests that cannot be run during a release

It is my understanding that it already works for everything except browser-specific tests, and that browser-specific tests are the tests that aren't automatically run during a release due to limitations in the build infrastructure. The primary addition of this PR over what bild.py already does is the ability to generate those specific tests. Is that correct?

sharwell:
I would be comfortable with either of the following, and I think either would make sense. I *lean* towards the former.

1. Add the browser test generation to bild.py, but disable execution of these tests during the release build because our build infrastructure doesn't [currently] support dispatching the tests to other operating systems for a fully-automated complete set of tests. This leaves the option open for someone with more time than us to figure out how to include execution of browser tests in the release build, which would be awesome.

2. Add a script in antlr/antlr4-javascript which generates the browser tests, since they are specific to that repository. bild.py remains fully operational for regenerating other tests at-will.

Each of these approaches allows for the independent, on-demand regeneration of test code without adding complexity to the code in the test suite itself. If I've overlooked other issues that need to be considered which aren't handled by one or both of the above approaches, please let me know. :smile:

parrt:
after evaluating the issues, I am merging but will put a comment in that those directories must be kept in sync with the bild.py.

PR #918, opened 2015-06-19
add tests, refactor get-all-parse-tree stuff. add Trees support routines

parrt:


tool/src/org/antlr/v4/tool/GrammarParserInterpreter.java
```diff
@@ -64,7 +67,7 @@ public GrammarParserInterpreter(Grammar g, ATN atn, TokenStream input) {
 			  input);
 		this.g = g;
 		decisionStatesThatSetOuterAltNumInContext = findOuterMostDecisionStates();
-		stateToAltsMap = new int[g.atn.getNumberOfDecisions()][];
+		stateToAltsMap = new int[g.atn.states.size()][];
```
sharwell: :thought_balloon: You could use `DecisionState.decision` as the key instead of `ATNState.stateNumber` if you want to keep this array the smaller size.

runtime/Java/src/org/antlr/v4/runtime/tree/Trees.java
```diff
@@ -275,6 +289,47 @@ public static ParserRuleContext getRootOfSubtreeEnclosingRegion(ParseTree t,
...
+				}
+			}
+		}
+	}
+
+	/** Return first node satisfying the pred
+	 *
+ 	 *  @since 4.5.1
+	 */
+	public static Tree findNodeSuchThat(Tree t, Predicate<Tree> pred) {
```
sharwell: :bulb: `Predicate<? super Tree>`

parrt: Wouldn't we want `? extends` here? and technically isn't it okay to use as is because polymorphism gives me the guarantee I need.

sharwell: A `Predicate<T>` takes an argument `T` and returns a `boolean`. A `Predicate<Object>` clearly knows how to handle a `Tree` because a `Tree` is an `Object`. However, with the current definition you couldn't use the following:

```
static final Predicate<Object> TRUE = new Predicate<Object>() {
  @Override
  public boolean test(Object o) {
    return true;
  }
};
```

The type which means "A predicate which understands a `Tree` argument" is `Predicate<? super Tree>`.

tool/src/org/antlr/v4/tool/GrammarParserInterpreter.java
parrt: looks like I can't use the decision number because (x)* for left recursive rules with only a single alternative do not get a decision number.

runtime/Java/src/org/antlr/v4/runtime/tree/Trees.java
parrt: I wouldn't expect TRUE to work. In fact, I would want the opposite. I would want Trees and below only right?

PR #920, opened 2015-06-19
Add Maven build

jvanzyl:
Account for the following:
- deploying snapshots to OSSRH
- releasing to Maven Central
- Shaded JAR including the treelayout dependency
- OSGi manifest
- remove Eclipse IDE metadata for clean import
- remove IDEA IDE metadata for clean import


jvanzyl:
Happy to change, adjust anything. We'll just let it bake until everyone is happy. I'll just squash and force push on my fork.

sharwell:
@jvanzyl Try not to squash until the end - it makes it much easier to review changes that occurred after my initial review. :smile:

antlr4-maven-plugin/pom.xml
```diff
@@ -1,348 +1,151 @@
...
+      <artifactId>plexus-build-api</artifactId>
+      <version>0.0.7</version>
+    </dependency>
+    <!-- The version of ANTLR tool that this version of the plugin controls. 
+      We have decided that this should be in lockstep with ANTLR itself, other 
+      than -1 -2 -3 etc patch releases. -->
+    <dependency>
+      <groupId>org.antlr</groupId>
+      <artifactId>antlr4</artifactId>
+      <version>4.6-SNAPSHOT</version>
```
sharwell: :exclamation: `4.6-SNAPSHOT` doesn't seem to exist?

:question: Why not `${project.version}`

antlr4-maven-plugin/pom.xml
```diff
@@ -1,348 +1,151 @@
...
+      <groupId>org.apache.maven.plugin-tools</groupId>
+      <artifactId>maven-plugin-annotations</artifactId>
+      <version>3.2</version>
+      <scope>provided</scope>
+    </dependency>
+  </dependencies>
+
+  <build>
+    <plugins>
+      <plugin>
```
sharwell: :question: Do we not need the following? I'm concerned the **lifecycle-mapping-metadata.xml** file might not be included in the output with the current configuration. This file provides information for Eclipse m2e.

```xml
<resources>
  <resource>
    <directory>resources</directory>
  </resource>
</resources>
```

antlr4-maven-plugin/pom.xml
```diff
@@ -1,348 +1,151 @@
...
+            </goals>
+          </execution>
+          <execution>
+            <id>help-goal</id>
+            <goals>
+              <goal>helpmojo</goal>
+            </goals>
+          </execution>
+        </executions>
+      </plugin>
```
sharwell: :question: Do we need to include the **maven-site-plugin** plugin?
:question: Are we generating a site?

dain:
This opens correctly for me in IntelliJ now.  

The only remaining issue is that we need another `.gitignore` for `*.iml`.

antlr4-maven-plugin/pom.xml
jvanzyl: Good catch. Adjusted.

antlr4-testgen-maven-plugin/pom.xml
```diff
@@ -36,9 +36,11 @@
 <project xmlns="http://maven.apache.org/POM/4.0.0" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/maven-v4_0_0.xsd">
```
sharwell: :exclamation: The copyright header above this line is actually incorrect. It should read Terence and myself for this plugin (obviously my fault).

antlr4-testgen-maven-plugin/pom.xml
```diff
@@ -36,9 +36,11 @@
...
 
     <modelVersion>4.0.0</modelVersion>
-
-    <groupId>org.antlr</groupId>
-    <version>4.5.1-SNAPSHOT</version>
+    <parent>
```
sharwell: :bulb: This change makes the `<project.build.sourceEncoding>` property below unnecessary.

antlr4-maven-plugin/pom.xml
jvanzyl: This is in the parent pom.

antlr4-maven-plugin/pom.xml
jvanzyl: I started with the last maven build I could find in the tree. Happy to remove the site plugin. I'm not a fan myself.

parrt:
I'm going to delete the entire .idea  once we get this working :)

parrt:
mvn compile success!

pom.xml
```diff
@@ -0,0 +1,121 @@
...
+    <module>antlr4-maven-plugin</module>
+    <module>antlr4-testgen-maven-plugin</module>
+    <module>runtime-testsuite</module>
+  </modules>
+
+  <properties>
+    <project.build.sourceEncoding>UTF-8</project.build.sourceEncoding>
+    <project.reporting.outputEncoding>UTF-8</project.reporting.outputEncoding>
+    <antlr.testinprocess>true</antlr.testinprocess>
+    <java.version>1.6</java.version>
```
sharwell: :exclamation: The **maven-compiler-plugin** already has special properties defined for this. Rather than create a new `java.version` property, use the following instead:

```xml
<maven.compiler.source>1.6</maven.compiler.source>
<maven.compiler.target>1.6</maven.compiler.target>
```

:bulb: The build configuration currently omits warnings that should be reported (so we can correct the code if they appear. I'd like to see the following properties added here:

```xml
<maven.compiler.showWarnings>true</maven.compiler.showWarnings>
<maven.compiler.showDeprecation>true</maven.compiler.showDeprecation>
```

jvanzyl:
@parrt and @sharwell also note that I didn't re-export the tree layout bits in the OSGi bundle. I assume that's an implementation detail you don't want exposed, but if you do we can re-export. Forgot to mention that in the initial description.

pom.xml
```diff
@@ -0,0 +1,121 @@
...
+      <testResource>
+        <directory>test</directory>
+      </testResource>
+    </testResources>    
+    <pluginManagement>
+      <plugins>
+        <plugin>
+          <groupId>org.apache.maven.plugins</groupId>
+          <artifactId>maven-compiler-plugin</artifactId>
+          <configuration>
```
sharwell: :exclamation: The `<source>` and `<target>` properties can be removed here, in accordance with the properties change.

:bulb: I'd like to see the following added:

```xml
<compilerArgs>
  <arg>-Xlint</arg>
  <arg>-Xlint:-serial</arg>
</compilerArgs>
```

runtime/Java/pom.xml
```diff
@@ -1,103 +1,91 @@
...
-		</dependency>
-
-	</dependencies>
+  xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/maven-v4_0_0.xsd">
+  <modelVersion>4.0.0</modelVersion>
+  <parent>
+    <groupId>org.antlr</groupId>
+    <artifactId>antlr4-master</artifactId>
+    <version>4.5.1-SNAPSHOT</version>
+    <relativePath>../../pom.xml</relativePath>
```
sharwell: :question: Why not just `../..`?

runtime/Java/pom.xml
```diff
@@ -1,103 +1,91 @@
...
+        <groupId>org.apache.felix</groupId>
+        <artifactId>maven-bundle-plugin</artifactId>
+        <version>2.5.4</version>
+        <executions>
+          <execution>
+            <id>bundle-manifest</id>
+            <phase>process-classes</phase>
+            <configuration>            
+              <instructions>
+                <Bundle-SymbolicName>org.antlr.antlr4-runtime-osgi</Bundle-SymbolicName>
```
sharwell: :thought_balloon: These properties don't match the ones being produced by bild.py. The impact of the difference is not clear to me.

runtime-testsuite/pom.xml
```diff
@@ -1,57 +1,37 @@
...
+    <dependency>
+      <groupId>junit</groupId>
+      <artifactId>junit</artifactId>
+      <version>4.11</version>
+      <scope>test</scope>
+    </dependency>
+  </dependencies>
+
+  <build>
+    <sourceDirectory>src</sourceDirectory>
```
sharwell: :exclamation: This seems to be missing the test resources:

```xml
<resources>
  <resource>
    <directory>resources</directory>
  </resource>
</resources>
```

tool/pom.xml
```diff
@@ -1,133 +1,78 @@
...
+
+  <dependencies>
+    <dependency>
+      <groupId>junit</groupId>
+      <artifactId>junit</artifactId>
+      <version>4.11</version>
+      <scope>test</scope>
+    </dependency>
+    <dependency>
+      <groupId>org.seleniumhq.selenium</groupId>
```
sharwell: :exclamation: This dependency is not used by the build. It should be removed in order to avoid a large unnecessary download.

tool/pom.xml
```diff
@@ -1,133 +1,78 @@
...
+      <artifactId>selenium-java</artifactId>
+      <version>2.44.0</version>
+      <scope>test</scope>
+    </dependency>
+    <dependency>
+      <groupId>org.eclipse.jetty</groupId>
```
sharwell: :exclamation: This dependency is not used by the build. It should be removed in order to avoid a large unnecessary download.

tool/pom.xml
```diff
@@ -1,133 +1,78 @@
...
+        </configuration>
+        <executions>
+          <execution>
+            <goals>
+              <goal>antlr</goal>
+            </goals>
+          </execution>
+        </executions>
+      </plugin>
+    </plugins>
```
sharwell: :exclamation: This section is missing the **antlr4-testgen-maven-plugin**, which is used to generate the tests.

```xml
<plugin>
  <groupId>org.antlr</groupId>
  <artifactId>antlr4-testgen-maven-plugin</artifactId>
  <version>${project.version}</version>
  <executions>
    <execution>
      <goals>
        <goal>antlr4.testgen</goal>
      </goals>
      <configuration>
        <runtimeTemplates>test/org/antlr/v4/test/runtime/java/Java.test.stg</runtimeTemplates>
      </configuration>
    </execution>
  </executions>
</plugin>
```

antlr4-maven-plugin/pom.xml
sharwell: :thought_balloon: If we decide to remove the site plugin (it's not included in this pom currently BTW), we should consider removing the additional site content as well. It's been notoriously incorrect over time.

jvanzyl:
We can just make additional changes with more PRs.

tool/pom.xml
jvanzyl: @sharwell i say just plug that back in if it's a harmful omission.

sharwell:
:memo: The following two items from the initial description are not correct, and would require substantial changes. Rather than go through and point all of them out, I reviewed the change as though they were not intended features (i.e. bild.py remains the only way to create a release which is publicly distributed by the ANTLR organization).

> * deploying snapshots to OSSRH
> * releasing to Maven Central

runtime-testsuite/pom.xml
sharwell: This was answered above - the resources are specified in the parent pom.

sharwell:
This was merged already.

antlr4-maven-plugin/pom.xml
sharwell: This was corrected in #921

tool/pom.xml
ericvergnaud: This dependency is required in order to compile and run tests when creating a project from the pom.
It should not be removed.

tool/pom.xml
ericvergnaud: This dependency is required in order to compile and run tests when creating a project from the pom.
It should not be removed.

tool/pom.xml
sharwell: @ericvergnaud I looked through the artifact extensively to make sure the dependency really isn't required. It sounds like you're running into an issue with another artifact which happens to be resolved by including this dependency. This would indicate that the missing dependency errors are actually a bug in the other module, not this one.

To date, I haven't been able to reproduce an ANTLR environment where omitting this dependency causes failures. If you have an example, I would certainly be willing to help narrow down this issue with the hope of solving it while also not requiring ANTLR/Java developers to download this large dependency.

ericvergnaud: Have you tried running the javascript browser tests?

sharwell: I have not. Is there documentation for the process that you can point me to?

ericvergnaud: here is what I do:
 - generate the tests using the -tool option
 - in eclipse, refresh the javascript linked test folder
 - right-click -> run unit tests
when running in batch, sometimes they fail, due to browsers caching strategy
(hence why they are excluded from the build phase, not reliable enough)
but individually they all pass

sharwell: Thanks. I won't be able to get this tested until tomorrow evening. :smile:

PR #922, opened 2015-06-19
Add IDEA *.iml files to the ignores

jvanzyl:


.gitignore
```diff
@@ -17,6 +17,7 @@ nbactions*.xml
 
 # IntelliJ projects
 *.iws
+*.iml
 .idea/*
```
sharwell: :bulb: This line should just be:

```
.idea/
```

parrt:
i'll do manually.

Issue #907, opened 2015-06-16
Add test for rule named `type`


sharwell:
Not all targets support a parser rule named `type`. However, the Java target does support such a rule. The previous tests which covered this needed to be removed for cross-target support in #906. A new test should be added to specifically cover this case, and the test can be intentionally excluded from any targets which do not support such a rule.

PR #927, opened 2015-06-20
fix annoying warnings in non critical code

ericvergnaud:


sharwell:
I don't think we need to change the `BaseTest` in the `rt` package since it's no longer used by the new test generator. Changes to this code now increases the work involved in unifying the test suites.

parrt:
Hi Eric,Are these changes so that you can better compare the old and new mechanism?

sharwell:
The primary purpose of the test using  `DeclareListLocal` is ensuring that list works. Targets which do not support lists in the same way as Java should disable the test as not relevant to that language, but the test should retain its current form for at least Java  (and C# would use it add well).

ericvergnaud:
committed the wrong file actually
we can delete all legacy once this PR is merged

ericvergnaud:
In its current form, the test does not check that list works
and in C# it does not compile
if the purpose of the test is simply to check that the method does not crash, then the previous version which declared an object did the job
and did not cause any problem to other targets

runtime-testsuite/resources/org/antlr/v4/test/runtime/templates/CompositeParsers/DelegatorRuleOverridesLookaheadInDelegate.stg
```diff
@@ -23,7 +23,7 @@ masterGrammar(grammarName, slaveGrammarName) ::= <<
 grammar M;
 import S;
 prog : decl ;
-type : 'int' | 'float' ;
```
sharwell: :memo: #907 has us covered on the original rule name

runtime-testsuite/resources/org/antlr/v4/test/runtime/templates/CompositeParsers/DelegatorInvokesDelegateRuleWithArgs.stg
```diff
@@ -28,5 +28,5 @@ WS : (' '|'\n') -> skip ;
 
 slaveGrammar(grammarName) ::= <<
 parser grammar S;
-a[int x] returns [int y] : B {<write("\"S.a\"")> $y=1000;} ;
```
sharwell: :thought_balloon: I liked the readability of the extra space, but it doesn't work for python. The `more` template is serious attention to detail. I'm curious if it's required for one or more target languages.

runtime-testsuite/resources/org/antlr/v4/test/runtime/templates/LexerExec/CharSet.stg
```diff
@@ -1,21 +1,17 @@
...
 
-Options ::= [
-	"Ignore": "Currently unable to generate the required CRLF input sequence."
-]
-
 Grammar ::= [
 	"L": {<grammar("L")>}
 ]
 
-Input() ::= "34<\u000D><\n> 34"
```
sharwell: :exclamation: I'm planning to fix the handling of carriage returns to ensure the test suite includes coverage of these characters, which do not behave the same way as new line characters in ANTLR. Until it's working under the original design, I would prefer to not make any changes to this.

runtime-testsuite/resources/org/antlr/v4/test/runtime/templates/ParserErrors/TokenMismatch2.stg
```diff
@@ -10,7 +10,7 @@ Grammar ::= [
 
 Input() ::= "( ~FORCE_ERROR~ "
 
-Rule() ::= "a"
```
sharwell: :thought_balloon: I'm really curious how this passed the initial testing. It seems obviously wrong.

runtime-testsuite/resources/org/antlr/v4/test/runtime/templates/LexerExec/CharSet.stg
ericvergnaud: one basic rule in unit testing is that a test should only serve one purpose
I would suggest you write a new test for just that: handling of carriage returns
and have those tests running since they cover other stuff

runtime-testsuite/resources/org/antlr/v4/test/runtime/templates/CompositeParsers/DelegatorInvokesDelegateRuleWithArgs.stg
ericvergnaud: this is not for readability
the python more() is actually ‘; ‘

runtime-testsuite/resources/org/antlr/v4/test/runtime/templates/LexerExec/CharSet.stg
sharwell: Even if true, if carriage returns are working then there is no need to change this test. I prefer to address this change (and the other files with the same exclusion) separately.

runtime-testsuite/resources/org/antlr/v4/test/runtime/templates/CompositeParsers/DelegatorInvokesDelegateRuleWithArgs.stg
sharwell: Interesting. I think we could actually use the following as a more "ANTLR-y" way to do this. How do you feel about this as a general pattern for executing a sequence of multiple actions at a specific location in a test?

```
{<write("\"S.a\"")>} {$y=1000;}
```

parrt: I would ok with that and removes a template.

runtime-testsuite/resources/org/antlr/v4/test/runtime/templates/LexerExec/CharSet.stg
ericvergnaud: right now those tests are marked as ignore precisely because cr is not working
I changed them so they can be ran as part of the build in every target

I had already done that in the original test generator

my suggestion is that we take the change, and once you have fixed the cr issue, you provide a dedicated test for cr

runtime-testsuite/resources/org/antlr/v4/test/runtime/templates/CompositeParsers/DelegatorInvokesDelegateRuleWithArgs.stg
ericvergnaud: that would not insert a ‘;’ between the 2 productions, hence would not work in python

sharwell: @ericvergnaud Notice that my proposed form is two consecutive, independent actions. I would expect the Python target to emit them each on their own line, and thus not need the separating `;`.

ericvergnaud: if it works out of the bow, why not, but I will not modify the Python templates to support this
managing indents in Python is notoriously hazardous

tool/test/org/antlr/v4/test/tool/BaseTest.java
```diff
@@ -112,6 +113,7 @@
 import static org.junit.Assert.assertNotNull;
 import static org.junit.Assert.assertTrue;
 
+@SuppressWarnings("resource")
```
sharwell: :thought_balloon: According to [this StackOverflow answer](http://stackoverflow.com/a/1206001/138304), this message only applies to resources of type `Closeable`. We don't have any resources of type `Closeable` because it was first introduced in Java 7; Eclipse should be picking up this configuration but appears to not be. I'll look into this a bit more and try to automate Eclipse's selection of `-source 1.6 -target 1.6`. I'm expecting this to work even for users working with JDK 7 or JDK 8, but I haven't figured everything out yet.

I think we should only include this use of `@SuppressWarnings` if we can't get things to work using an updated build configuration that accurately reflects the limitations of our current compatibility goals.

runtime-testsuite/resources/org/antlr/v4/test/runtime/templates/LexerExec/CharSet.stg
sharwell: @ericvergnaud I am not opposed to separating the tests for `\r` from the tests for other items. However, the tests were not originally created with this separation; this proposed change removes the `\r` tests which we did have and does not replace them with an alternative.

runtime-testsuite/resources/org/antlr/v4/test/runtime/templates/CompositeParsers/DelegatorInvokesDelegateRuleWithArgs.stg
sharwell: :memo: I'm looking into this a bit more. We have tests that contain duplicate actions (e.g. **MultipleActions.stg**), but they use empty actions. This was an oversight in the current tests since I always expected users would have the ability to include multiple consecutive actions in either a lexer or parser. I'll have more information tonight when I can next run the Python tests.

runtime-testsuite/resources/org/antlr/v4/test/runtime/templates/LexerExec/CharSet.stg
ericvergnaud: It’s been like that for the past 6 months, after discussing with Ter.
Right now we don’t have \r tests since these tests are ignored
So better loose 50% coverage than 100%

tool/test/org/antlr/v4/test/tool/BaseTest.java
ericvergnaud: The thing is we should be closing those resources, but this requires a significant refactoring, which for test code is probably not worth it.
Those warnings show up even when source level is manually set to 1.6

sharwell: Looks like we forgot to close the pipes on some code paths in `BaseTest.execClass`. We should fix that rather than use `@SuppressWarnings`. Either way it would best be done as it's own separate pull request. I have a branch ready that I created for testing, but since you originally found the issue I'll let you submit a pull request if you want to. Just let me know. :smile:

runtime-testsuite/resources/org/antlr/v4/test/runtime/templates/CompositeParsers/DelegatorInvokesDelegateRuleWithArgs.stg
parrt: Looks like multi-actions puts 'em on diff lines:

```
a : {foo} {bar} A ;
```

gives

```python
            self.enterOuterAlt(localctx, 1)
            foo
            bar
            self.state = 4
            self.match(T.A)
```

parrt:
Guys, I'm going to dig through the conversation and commits and make the best compromise/decisions I can so we can get this out. Thanks for the insights.

runtime-testsuite/resources/org/antlr/v4/test/runtime/templates/CompositeParsers/DelegatorInvokesDelegateRuleWithArgs.stg
sharwell: :+1: That's awesome news. We can use this immediately to eliminate the need for the `more()` template, and I filed issue #931 to create new tests that ensure this basic functionality works for future targets as well.

parrt:
I have now merged in what I want. Thanks guys.

PR #928, opened 2015-06-20
delete legacy test generator and generated tests

ericvergnaud:


sharwell:
This looks pretty straightforward. :+1: from me *however* I do want to finish the new test review with Eric prior to merging this.

parrt:
Ok, we've looked over new testing stuff that Sam built and it appears ok to wack the old version.

PR #924, opened 2015-06-19
Maven tweaks

sharwell:
This pull request addresses a number of items identified while reviewing #920.

/cc @jvanzyl

PR #929, opened 2015-06-20
add generated tests

ericvergnaud:


sharwell:
The test generator was designed so the tests did not need to be included in source control. If the omission of this code from source control led to problems of some sort, I'd be very interested to get some additional information about that so I can make sure those concerns are fully addressed.

Regarding this change itself - including the tests in source control introduces the following problems:

1. It results in the test suite being stored twice in source control. The two copies can easily become out of sync with each other, and it's not easy to determine whether the generated code is a correct representation of the current runtime test suite. A number of minor issues are a direct result of this:
  * It's less clear how contributors should create and submit new runtime tests
  * It's more difficult to review proposed new tests for consistency and accuracy
  * Changes which affect the test suite increase in size, reducing the efficiency of the review process

2. The output of the test generator varies according to the version of Java used to generate tests. Even without making any changes, simply regenerating tests could result in someone making a large commit with no substantive changes.

3. The tests are fully regenerated by both bild.py and the Maven build. I'm not aware of any configuration where updates to the test suite are not immediately reflected in the generated code, even if the generated code does not appear in source control.

Due to the major procedural downsides associated with including this code in source control, I believe we should work to find another way to address any outstanding concerns related to the generated code without resorting to putting the generated code itself in source control.

ericvergnaud:
Comments in line.

> Le 21 juin 2015 à 04:49, Sam Harwell <notifications@github.com> a écrit :
> 
> The test generator was designed so the tests did not need to be included in source control.
> 
Maybe you are not aware that I designed the tests generator initially?
This was absolutely not the goal, it’s a side effect of any code generator. 
The only goal was to be able to generate the same set of tests for all runtime targets. 
If the above is now a goal, it was not discussed with the original author.
I’m quite happy with the improvements in this version, but this aspect was never in the discussion.
> If the omission of this code from source control led to problems of some sort, I'd be very interested to get some additional information about that so I can make sure those concerns are fully addressed.
> 
> Regarding this change itself - including the tests in source control introduces the following problems:
> 
> It results in the test suite being stored twice in source control. The two copies can easily become out of sync with each other, and it's not easy to determine whether the generated code is a correct representation of the current runtime test suite. A number of minor issues are a direct result of this
> 
where do you take it that the test suite is being stored twice?
> It's less clear how contributors should create and submit new runtime tests
I believe the comments in the generated code make it very clear.
> It's more difficult to review proposed new tests for consistency and accuracy
actually, it’s the exact opposite, because you can review the generated test first, which is much easier than reviewing the template first
> Changes which affect the test suite increase in size, reducing the efficiency of the review process
see above. If generated tests are not in git, then you need to review the templates, generate the tests and review them anyway
> 
> The output of the test generator varies according to the version of Java used to generate tests. Even without making any changes, simply regenerating tests could result in someone making a large commit with no substantive changes
> 
If that’s the case it’s a bug and it should be fixed. The generator produces text. It’s the first time I hear that the same java code would produce different text depending on the version of Java. Please file a bug with Oracle.
> .
> 
> The tests are fully regenerated by both bild.py and the Maven build. I'm not aware of any configuration where updates to the test suite are not immediately reflected in the generated code, even if the generated code does not appear in source control
> 
Are you not aware that only a subset of the standard runtime tests are generated by both bild.py and the Maven build? The javascript browser tests are not generated and should not be. The non standard tests and the tool tests are stored. 
Having the ability to pull binaries and regenerate code in the context of a release or CI is a requirement but that should not evolve into having the obligation to do so for day to day work. That’s a job for a CI, not for a developer.
> l.
> 
> Due to the major procedural downsides associated with including this code in source control, I believe we should work to find another way to address any outstanding concerns related to the generated code without resorting to putting the generated code itself in source control
> 
There is absolutely no additional downside associated with this change, since the existing generated tests are already in source control.

Maybe you foresaw a potential procedural upside in removing the generated tests. 

If so you did not foresee the major downsides of NOT including the generated tests, which are as follows:
 - makes it difficult and error prone to compare successive versions of the generated tests, so we can fix problems introduced by new templates. As demonstrated with the new generator, this is an absolute requirement. I was able to fix all the issues in a few hours only because I could compare with the previous generation.
 - makes it impossible to look at the tests without generating them. Means you can’t simply provide a link to github in a support discussion. 
 - can only create confusion for occasional contributors. They will see certain tests and not others? Why? How? Not very helpful I’m afraid.
 - steepens the learning curve for occasional contributors to produce new tests (they have to learn templates first). As such is anti-TDD.
 - increases the difficulty to understand a change in a test. A change in a template requires full knowledge of templates to be understood, which very few people have.
 - since by design the natural location for the generated tests is where the target templates reside, not storing them in source control requires complex rules in gitignore which are likely to be broken each time a new test file is generated.

I’m all in favor of simplicity, and smooth learning curves.

sharwell:
It's late so I don't have time to go into everything. I did want to point out that I filed issue antlr/stringtemplate4#118 regarding part of the differences that appear when generating code with different versions of Java.

ericvergnaud:
Don’t even take that time, it’s a complete waste.
Any time spent trying to convince people that there is only one correct way of doing things is a waste.
This is actually demonstrated by your current investment, where you want to build using maven rather than using bild.py. Correct?
Is anyone stopping you from doing that? No.
Did you consider the risk of having 2 diverging build tools? No.
Did you consider the additional cost of reviewing 2 build methods? No.
Did you factor in the fact that some of those tests cannot be automated due to platform limitations? No.
Did you ask the opinion of the original author of this generator? No.
You know what? Not a problem. You need it, do it.
Is anyone trying to stop you? No.
Is that because you are right to do this? No.
Terence needed bild.py, although I’m not fully supportive, I helped him, rather than tried to stop him. That was mere convenience.
You need a maven build? It is mere convenience.
I need generated tests to be in git? It is mere convenience.
So please stop immediately being an absolute pain in the ass, especially on subjects which relate to convenience, not correctness.
Stop immediately using your arrogant and assertive tone each time you comment on something.
If I recall correctly, you were completely wrong in 2 recent occasions, hence your credit for being a source of truth and correctness is depleted.
The only foreseeable outcome of your behavior is less people contributing to antlr.
The more happy people around, the better.

sharwell:
> - since by design the natural location for the generated tests is where the target templates reside, not storing them in source control requires complex rules in gitignore which are likely to be broken each time a new test file is generated.

Both Maven and Bild have standard locations for generated code which is already configured in **.gitignore** and does not place the generated output in the same location as the templates. Due to a bug in the test generator (which is fixed in #924, specifically commit 5b777fe88b84be838f31e620e2a31e4ab95db848), Eclipse was not automatically picking up this directory and presenting it in the Package Explorer. With this issue corrected, all a user needs to do is open the pom.xml and Eclipse will generate the tests and present them in Package Explorer as highlighted here:

image

> - makes it impossible to look at the tests without generating them. Means you can’t simply provide a link to github in a support discussion.

The **.stg** files in the test suite *are* the tests. The Java source code which is created from those during the build is an intermediate artifact used for executing the tests, but those intermediate files are no more the tests than the .class files which javac produces from them.

That said, the lack of ability to view the generated files online is an interesting limitation that I think would be nice to remove. However, neither ANTLR 3 nor ANTLR 4 store the code generated from a grammar in source control, and this has not proven to be a limiting factor for code reviews on dozens of projects which rely heavily on code generation during the build. I'm thinking a combination of a CI configuration along with using **maven-source-plugin**'s `test-source` goal may be able to provide this specific convenience without the overhead of placing generated files in source control.

> - steepens the learning curve for occasional contributors to produce new tests (they have to learn templates first). As such is anti-TDD.
> - increases the difficulty to understand a change in a test. A change in a template requires full knowledge of templates to be understood, which very few people have.

There is no avoiding the templates when creating tests. This is more complex than plain ol' Java files with JUnit tests, but generating tests from templates provides staggering advantages over individual tests when it comes to cross-target testing. The updated structure of the test suite is intended to be straightforward for new users to comprehend. In particular, only two files need to be changed to add a test. I don't have a single commit showing a test, but #909 shows a slightly more complex example and the set of commits [441587e8598ba25961e3cb20c88d4f425027ab8a, 4bee04f1140cfc138cf22a3a36759214637bfa98, 8fb7f32405e17cc290abf92d946d8764019bf6f9] show another.

As mentioned previously, the **.stg** files in the test suite *are* the tests. There is no need to review Java source code (or Python, JavaScript, or C# code for that matter) when reviewing a new test. The review for tests begins and ends with the templates, because the code generator handles the rest. If we encounter a case where a test does not produce the expected test result, evaluating the test on your local system would be an integral part of resolving the issue.

ericvergnaud:
Sam,

I thought I had been clear when I required that you stop immediately.

Eric

parrt:
ok, guys, this is a difficult choice to make as I see arguments on both sides. I've decided that I'm going to include the generated tests in the repo, at least at the beginning. Moreover, I'm going put all of the unit tests in one spot rather than across repositories. I'm going to put them in the runtime-testsuite subdirectory. If this causes problems for the mvn mojo/plugin thingie, we can probably work around it. In fact, I think if I simply put it under runtime-testsuite/test, mvn's pom.xml should ignore it completely.

parrt:
closing to do my own

ericvergnaud:
Hi,
I like the idea, but it requires an additional tweak to compile.
runtime-testsuite now has a dependency on tool, so need to add to the pom
also need to add selenium to be able to run browser tests I think
Eric

parrt:
Ok, Jason will update pom

On Mon, Jun 22, 2015 at 4:32 PM, ericvergnaud <notifications@github.com>
wrote:

> Hi,
> I like the idea, but it requires an additional tweak to compile.
> runtime-testsuite now has a dependency on tool, so need to add to the pom
> also need to add selenium to be able to run browser tests I think
> Eric
>
> > Le 23 juin 2015 à 04:21, Terence Parr <notifications@github.com> a
> écrit :
> >
> > Closed #929 <https://github.com/antlr/antlr4/pull/929>.
> >
> > —
> > Reply to this email directly or view it on GitHub <
> https://github.com/antlr/antlr4/pull/929#event-337122964>.
> >
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/antlr/antlr4/pull/929#issuecomment-114302205>.
>



-- 
Dictation in use. Please excuse homophones, malapropisms, and nonsense.

PR #942, opened 2015-06-30
Fold target repos into master (w/o histories I'm afraid)

parrt:
C#, javascript, python2, python3 repos have been folded into the main repository and Jason has gotten the maven stuff up to snuff. Tests for the runtime are generated when you ask for it otherwise the tests are run from the files held in the repository.


sharwell:
:question: Why not use `git subtree` to retain the history? Many people contributed to the various runtime targets over time and it would be unfortunate to lose that information.

parrt:
@sharwell I'll try to add history for runtime files themselves as I didn't have to mod those. should be "easy" enough.

PR #955, opened 2015-07-14
move swing related stuff out of runtime package into org.antlr.v4.gui

### LINKED ISSUE #576: create runtime lib w/o dependency on swing for micro devices
Requested from Groovy guys: https://twitter.com/jponge/statuses/469403167957204992

parrt:
Fixes #576

### DIFF
--- a/runtime/Java/pom.xml
+++ b/runtime/Java/pom.xml
@@ -11,14 +11,6 @@
   <name>ANTLR 4 Runtime</name>
   <description>The ANTLR 4 Runtime</description>
 
-  <dependencies>
-    <dependency>
-      <groupId>org.abego.treelayout</groupId>
-      <artifactId>org.abego.treelayout.core</artifactId>
-      <version>1.0.1</version>
-    </dependency>
-  </dependencies>
-
   <build>
     <sourceDirectory>src</sourceDirectory>
     <plugins>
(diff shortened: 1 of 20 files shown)


parrt:
only issue might be that org.antlr.v4.runtime.misc.TestRig is how you run the command-line grammar tester but now it's org.antlr.v4.gui.TestRig. maybe the answer is to alter the PR so that the package names are absolutely the same but would just be included with a different jar

sharwell:
> only issue might be that org.antlr.v4.runtime.misc.TestRig is how you run the command-line grammar tester but now it's org.antlr.v4.gui.TestRig. maybe the answer is to alter the PR so that the package names are absolutely the same but would just be included with a different jar

I think we have enough confusion surrounding TestRig for beginners already. I think keeping the same exact name is a very good idea.

parrt:
closing and rebuilding.

sharwell:
The other option here is doing what you have, but leaving a TestRig in the runtime in the same place it is now. The new TestRig would:

1. Attempt to locate the new TestRig in the classpath, and if that fails print a message that the tool jar must now be available if you want to run test rig.
2. If the new TestRig is found, print a message that TestRig has moved and the new one is being called automatically, then call its `main()` with the same arguments.

parrt:
interesting.

parrt:
Fixes #576

sharwell:
:+1:

PR #968, opened 2015-08-03
added antl4test-maven-plugin

teverett:
This is the source code previously at grammars-v4/support/anlt4test-maven-plugin

I am adding it to antlr4 so that with the next release of antlr4 it can be pushed to the maven  repo servers.  Additionally, it really doesn't belong on the grammars-v4 tree.


parrt:
Hi. i'll have to think about this one for next release. I'm kinda loth to add new stuff to tool proper at this point.

teverett:
I'm fine to just leave it here so you can merge it when the time is right.

parrt:
Hi. I'd rather keep this code separate to keep main tool simple etc...

PR #971, opened 2015-08-04
"int" is a reserved keyword.

hce:
closure-compiler complains about this with an error...

### DIFF
--- a/runtime/JavaScript/src/antlr4/atn/ATNDeserializer.js
+++ b/runtime/JavaScript/src/antlr4/atn/ATNDeserializer.js
@@ -594,10 +594,10 @@ var byteToHex = createByteToHex();
 ATNDeserializer.prototype.readUUID = function() {
 	var bb = [];
 	for(var i=7;i>=0;i--) {
-		var int = this.readInt();
+		var integer = this.readInt();
 		/* jshint bitwise: false */
-		bb[(2*i)+1] = int & 0xFF;
-		bb[2*i] = (int >> 8) & 0xFF;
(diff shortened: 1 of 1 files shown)


ericvergnaud:
Sorry, but I don't agree that int is a reserved keyword.
See: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Lexical_grammar

sharwell:
According to ECMA-262 §7.6.1, it seems @ericvergnaud is correct. Nevertheless, if this helps the closure-compiler avoid a warning is there any reason not to include it?

ericvergnaud:
Sounds like a closure-compiler bug, not the first one from what I can guess googling around.
There are hundreds of javascript rewriters out there, so will only accept this as a last resort.

ericvergnaud:
According to some people, the solution is to use --language_in ECMASCRIPT5_STRICT as an option to the closure compiler.

sharwell:
@ericvergnaud I don't see any disadvantages to the new code. The old name `int` wasn't serving any particular purpose that `integer` won't work for as well. My vote is :+1: on this. If you don't see any disadvantage we should go ahead and accept this change.

ericvergnaud:
@hce 
I'd like to get some feedback on the --language_in ECMASCRIPT5_STRICT approach. We you able to try it out?

hce:
When calling closure-compiler with --language_in ECMASCRIPT5_STRICT, the problem does indeed not occur.

Sorry for not responding sooner!

ericvergnaud:
Thanks.
Closing this since it is not required.

PR #973, opened 2015-08-05
Fix NameError in LexerATNSimulator.reset()

cooperra:
Without this line, one gets "NameError: name 'Lexer' is not defined" when calling reset(). This is the same way \__init__() assigns self.mode.


parrt:
Hi. can you sign antlr4/contributors.txt ?

parrt:
merged manually.

PR #979, opened 2015-08-19
Include files while scanning

HSorensen:
Yet another attempt to enable antlr to include additional files while scanning a source file.
Look at the provided JUnit for an example.

Comments and feedback is definitely needed.
Suggestions to better names etc.

Still todo: token location tracking is relative to the current file being read. This needs to be reflected when printing the token information.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -53,3 +53,4 @@ bilder.pyc
 bild.log
 
 bild_output.txt
+/bin/
--- a/.gitignore
+++ b/.gitignore
@@ -53,4 +53,3 @@ bilder.pyc
 bild.log
 
 bild_output.txt
-/bin/
--- a/contributors.txt
+++ b/contributors.txt
@@ -75,3 +75,5 @@ YYYY/MM/DD, github id, Full name, email
 2015/05/20, peturingi, Pétur Ingi Egilsson, petur@petur.eu
 2015/05/27, jcbrinfo, Jean-Christophe Beaupré, jcbrinfo@users.noreply.github.com
 2015/06/29, jvanzyl, Jason van Zyl, jason@takari.io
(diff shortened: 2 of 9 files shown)


parrt:
Hi.  unfortunately time does not allow me to evaluate something so big. I would be happy to handle something that is just allowing eclipse to use ANTLR more effectively however.

HSorensen:
@parrt I understand. 

This one pull request for all the needed for include. Is there a better way to make this more reviewable?

The total pull request is ~300 lines
The core functionality of this patch is ~175 lines, the testcase is another ~105 lines.

But even if you cannot review this can we discuss some of the implementation topics, like proper name for the grammar action.

Also note the JavaLR renaming stuff is ~20 lines, that you could merge and thus reduce this request.

parrt:
My biggest issue is not so much the code but the ramifications in terms of syntax or semantics on a tool that has slowly stabilized over 25 years :)   First, I have to understand the problem being solved etc. then I might find a better solution myself.   Regardless, it looks like it's mostly library changes which means that you can have your own super version of the runtime without me having to update the master, which could affect everybody from startups to twitter.

HSorensen:
Not sure if it counts for much in your book, but multiple input buffers has been supported by flex for a very long time so in that sense it is not a new concept. I would argue it was an oversight not to include this from the beginning:)
The pull request does touch the core Lexer.nextToken( ) loop, which of course should not be done lightly so I completely understand your reservations. Of course I cannot claim I understand all the semantics of your tool but I have taken care to minimize the impact of my suggested changes and if the feature is not used there should be zero impact on the existing semantics.
I can of course maintain my own fork for this and I am already building my own version of the antlr tool.
In the end I am sure twitter and many startups will be happy to use this new feature as well:)

parrt:
thanks :)  Yeah, I mostly build features that I have needed.  I guess I'll close this now but people can find later if needed.

PR #987, opened 2015-08-25
Improve hashing performance

ericvergnaud:
- adds a working test case (Python3 only) where performance can actually be measured
- changes hashing strategy
- fixes various typos


runtime/Python3/test/parser/clexer.py
```diff
@@ -0,0 +1,799 @@
+# Generated from C.bnf by ANTLR 4.5.1
```
sharwell: :question: Does this generated file need to be in source control?

sharwell:
:exclamation: This change includes large, generated files which I believe should be excluded from source control.

:question: Is there any way to add this as a test to the runtime test framework so all targets can benefit from it?

sharwell:
Based on their work in antlr/antlr4-python2#17, I believe @bdkearns is a good candidate for a reviewer of this pull request.

In addition, if @bdkearns wants to resubmit that pull request to this repository, I believe we should use it instead of this one (and only use the performance test from this pull request).

ericvergnaud:
The ability to test against a stable parser is critical to performance improvement.
Since C.bnf is not part of the antlr repo, stability cannot be achieved without the generated files.
This test runs from the Python IDE only, where there are no means to generate the files.
Adding this to the framework is a lot of work and adds no value since the test itself is not asserting anything.
Performance improvements are measured manually by switching branches.

ericvergnaud:
I don't need bdkearns to review this.
His behavior is inappropriate, and he is not a test driven developer.

sharwell:
> I don't need bdkearns to review this.
> His behavior is inappropriate, and he is not a test driven developer.

He found, analyzed, documented, and proposed a fix to this issue 11 months ago. I found his development strategy practical and his proposed changes clean and well-organized, and therefore highly value his opinion on the subject.

parrt:
gracias amigo

Issue #948, opened 2015-07-02
DefaultErrorStrategy.recover has some bad cases


michaelpj:
It looks like the current error recovery strategy has some pathological cases.

Consider the example in the documentation for `DefaultErrorStrategy.recover`:

```
a : '[' b ']'
  | '(' b ')'
  ;
b : c '^' INT ;
c : ID
  | INT
  ;
```

Given input `[]`, the current implementation constructs the combined contextual follow sets for the call chain `a -> b -> c`, and then consumes until it sees one of them. This results in syncing up to `]`, whence we can continue with `a`.

However, consider the following input: `[]^9]`. To the human eye this is a single malformed instance of `a`, with an extraneous `]`. However, currently we will synchronise in the same way, consuming only up until the `]`, and then falling over on the subsequent `^`. On the other hand, if we'd synced up to the `^` (which is in the follow set for `c`), we would have been fine.

---

To give a concrete example of how this is biting me, I have a grammar roughly like this:

```
main : (query | predicate)* ;
query : 'from' vardecl 'where' ... 'select' ... ;
vardecl: type lowerid ;
predicate: type lowerid '{' ... '}' ;
```

A query looks like this:

```
from int i
where ...
select ...
```

However, if we add an error:

```
from 'int i
where ...
select ...
```

things go wrong. I would like us to consume the `'`, `int` and `i` tokens, and resume the `query` rule from `where`. However, before we reach the `where` token, we see the `int` token, which could be the start of a `predicate`, and so is in our combined follow set. So we stop looking, and when we resume we start trying to parse a `predicate`, which doesn't go very well.

(This is a recurring problem for this grammar, as many constructs have `vardecl`s, and so the parser quite frequently ends up recovering into a `predicate`, which is usually not very helpful.)

The situation is analagous to the one I described before: we have an deep rule in the context stack which has a following token after some number of consumptions. However, before we reach that, we see a following token for a rule higher up in stack, and so we stop there.

---

My intuitive thought is that we should prioritise deeper rules when looking for following characters. But
1. This would be more expensive, since we'd need to look ahead for each rule in the stack in turn, until we found a following token.
2. I suspect this may have pathological cases of it's own. If we are willing to look far ahead (maybe even to EOF) for a following token, we may just discard too many tokens.

Hence I don't really have a good suggestion for what I'd like to happen here, except that I'd like it to avoid these particular cases!

sharwell:
[...]

PR #991, opened 2015-09-07
Reinstitute `sync` calls before `adaptivePredict`

michaelpj:
Inspired by https://github.com/antlr/antlr4/issues/948

There are four commented-out calls to `_errHandler.sync` in `Java.stg`. Two of these precede calls to `adaptivePredict`, and as discussed in the issue, it seems worthwhile to have them in, as they improve the chances of the predictor doing well. I've added a test case demonstrating the improvement: instead of getting `no viable alternative` error, we get an `extraneous input` error.

The other two calls appear in simple (LL1) alternative and optional blocks. Here we already make use of `recoverInline`, which does a good job in the examples I created. I am somewhat tempted to leave the calls in "just in case", but lacking any concrete example where there's an improvement, that seems somewhat arbitrary. Hence I've deleted those two calls, and added a test showing the current behaviour, which is identical with and without the `sync` calls.

(Sorry for the churn - regenerating the tests seems to have caused some unrelated reorderings in the generated sources.)


runtime-testsuite/resources/org/antlr/v4/test/runtime/templates/ParserErrors/SingleTokenDeletionBeforeAlt.stg
```diff
@@ -0,0 +1,28 @@
...
+
+Input() ::= "ac"
+
+Rule() ::= "a"
+
+Output() ::= <<
+>>
+
+Errors() ::= <<
+line 1:0 extraneous input 'a' expecting {'b', 'c'}
```
sharwell: :exclamation: You need an explicit `<\n>` at the end of the line.

runtime-testsuite/resources/org/antlr/v4/test/runtime/templates/ParserErrors/SingleTokenDeletionBeforePredict.stg
```diff
@@ -0,0 +1,28 @@
...
+
+Input() ::= "caaab"
+
+Rule() ::= "a"
+
+Output() ::= <<
+>>
+
+Errors() ::= <<
+line 1:0 extraneous input 'c' expecting 'a'
```
sharwell: :exclamation: You need an explicit `<\n>` at the end of the line.

sharwell:
:memo: This is related to issue #448.

:exclamation: This change appears to leave ANTLR in a state where the targets are no longer in sync. Now that the individual target languages are all part of the same repository, failing to update all languages at once will result in issues like antlr/antlr3#175.

:exclamation: The code generator is sensitive to the version of Java used for generating code. The tests need to be regenerated using the same version of Java that was originally used to generate them. This will reduce churn for items like the slave grammars, but not necessarily for the test methods themselves.

runtime-testsuite/resources/org/antlr/v4/test/runtime/templates/ParserErrors/SingleTokenDeletionBeforeAlt.stg
michaelpj: Bah - that will teach me to copy in the expected output verbatim.

michaelpj:
> This change appears to leave ANTLR in a state where the targets are no longer in sync.

Good point, I'll update the templates for the other targets in the analogous fashion. It's not entirely clear to me how I generate and run the tests for the other language targets - is there an easy way to do that? They don't seem to be checked in, so I don't need to check in the updated versions, but it would be nice to be able to verify that I'm not taking things out of sync.

> The code generator is sensitive to the version of Java used for generating code.

That's a little unfortunate. Do you know which version this is? I guess I can just try a few and see which one causes the least churn...

michaelpj:
> That's a little unfortunate. Do you know which version this is? I guess I can just try a few and see which one causes the least churn...

I found it - you want Java 8, but not 7 or 6. A little bit odd, and probably worth mentioning somewhere.

parrt:
Hi.  I'm still kind of out on sabbatical. I don't think we have gone to a requirement for Java 8 yet have we?

Regarding running the tests, mvn install should do the trick for all targets. travis-ci seems to be happy with your changes.

@sharwell  are you in favor of incorporating this change?

sharwell:
I'm apprehensive about removing the calls before the LL(1) constructs. The general rule we intended to follow is placing a call to `sync` before evaluating the LL(1) symbol used by during decision making (in both the LL(1) and non-LL(1) cases). I think we should leave these calls in place unless we can prove they are redundant.

If those calls are added back, then yes I'd be in favor of incorporating this. The performance overhead of these calls is all but eliminated [when using the `BailErrorStrategy`](https://github.com/antlr/antlr4/blob/master/runtime/Java/src/org/antlr/v4/runtime/BailErrorStrategy.java#L95) with the two-stage parsing technique.

michaelpj:
Sorry, I think I've been unclear - this PR doesn't *remove* any existing calls, I merely did not reinstate some of the commented out calls and just removed them entirely. I only did that because I couldn't construct a test case where it made a difference, but I may well be missing something!

@sharwell am I right in thinking that by your argument we should reinstate the sync calls before *all* the LL1 constructs? In the diff that would be the ones in `LL1AltBlock`, `LL1OptionalBlock` and `LL1OptionalBlockSingleAlt`?

sharwell:
@michaelpj Yes, that is what I meant.

parrt:
Hmm...I hesitate to change the error handling without a thorough investigation; maybe we should leave those commented out or not included as Michael has done.

I wonder if we could do a clean pull request on this... it looks like it was meant to insert those calls but now we are saying we don't want to insert them so I'm not sure of the goal anymore. ;)

michaelpj:
The original goal of this PR was just to insert the calls in `AltBlock` and `OptionalBlock`, which is what it does now. Those are also tested. I also deleted commented out code, which, naturally, does nothing :wink:

It sounds like @sharwell thinks it's desirable to also reinstitute the commented-out calls in the other blocks as well. That seems likely to be safe to me, as `sync` should be a safe, fast operation, but it was sufficiently marginal that I didn't manage to squeeze out any obvious test cases that exercise it.

Perhaps we could merge this as is, and open an issue for further investigation of our `sync`-ing policy?

parrt:
Adding to my todo list to examine branch with all commits done and compare to master.

parrt:
Hmm...I'm getting C# warnings about the sync():

```
TParser.cs(177,18): error CS1061: Type `Antlr4.Runtime.IAntlrErrorStrategy' does not contain a definition for `sync' and no extension method `sync' of type `Antlr4.Runtime.IAntlrErrorStrategy' could be found. Are you missing an assembly reference?
```

from `mvn install`.

sharwell:
I believe it's `Sync` in the C# target.

michaelpj:
Sorry, that notification slipped thorough my net! I've updated the patch to use the right `Sync` command.

I'm not sure why I didn't see the errors from the C# code myself - do you have to generate it separately before running `mvn install`? I do have `mono` installed.

parrt:
I think mvn was not running all target tests earlier just java ;)

parrt:
Looks like there are lots of failures. not sure why. hmm... 
https://travis-ci.org/antlr/antlr4/jobs/91798624

michaelpj:
It would really help if I could run the non-Java tests... Great that Travis does, though.

The problem is clear enough - I assumed that the commented out parts were still valid, which they're clearly not - the JS and Python ones need `this` and `self` qualifiers.

PR #995, opened 2015-09-16
Fix maven test scope

ericvergnaud:
It was recently noticed that travis-ci only tests the java runtime. This somewhat breaks the purpose of the ci. This PR clarifies test generation paths and enables testing of all targets except browsers

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -7,9 +7,12 @@ jdk:
   - oraclejdk7
   - oraclejdk8
 before_install: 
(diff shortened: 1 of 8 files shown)


sharwell:
:question: If we exclude just the C# tests, do the other 3 pass? It's not everything but that alone would be a huge step forward.

ericvergnaud:
no, right now we just got java and python2

sharwell:
For python3, could it be that travis has it named just **python3** (see antlr/antlr4-python3#41 for a related Ubuntu issue)?

If you can get Java and Python 2 working together, that's better than Java by itself so I say we get that merged ASAP and then move forward with the others as we can.

PR #997, opened 2015-09-17
Added proper visit function in ParseTreeVisitor

krzkaczor:
It provides proper implementation for visit function which is essential for visitors. It fires appropriate visit function in visitors. It behaves in the same way as java implementation.

I have also fixed bug that prevented running runtime on nodejs. There is no importScripts symbol in scope so JS throws exception - we need to use typeof operator to avoid it.

### DIFF
--- a/contributors.txt
+++ b/contributors.txt
@@ -75,3 +75,4 @@ YYYY/MM/DD, github id, Full name, email
 2015/05/20, peturingi, Pétur Ingi Egilsson, petur@petur.eu
 2015/05/27, jcbrinfo, Jean-Christophe Beaupré, jcbrinfo@users.noreply.github.com
 2015/06/29, jvanzyl, Jason van Zyl, jason@takari.io
+2015/08/18, krzkaczor, Krzysztof Kaczor, krzysztof@kaczor.io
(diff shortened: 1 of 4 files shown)


ericvergnaud:
Hi,
thanks for this.
If not already done, you need to sign the contributors agreement (see contributors.txt)
Eric

ericvergnaud:
The NodeJS bug is already fixed by another PR, so please exclude from this one.

krzkaczor:
@ericvergnaud done x2

ericvergnaud:
Ok, now we just need Terence to bless it.

sharwell:
@krzkaczor Thanks for the contribution and getting it updated so quickly. Can you do one of the following as well:

1. Add one or more unit tests which cover this functionality
2. Create an issue to add unit tests covering this functionality at a later time

If you decide to go with the first option and run into problems, @ericvergnaud should be able to answer any questions you have.

:memo: This is a non-blocking request, meaning the pull request can be merged before you do one of these items. I'm just commenting so we don't lose track of this functionality. :+1:

krzkaczor:
@sharwell Thanks :) I wanted to do so in the first place but I don't know where can I find tests? I looked at package.json test command and it's not implemented. 

Now I found runtime-testsuite directory but tests are written in java and I am not sure how proper test should look like.

I will  create issue for that and maybe I will time in future to implement it.

PR #998, opened 2015-09-17
ANTLRFileStream,java: field fileName

HSorensen:
Simplify ANTLRFileStream by using the field "name" for super class ANTLRInputStream
* Change field fileName to use the field name from ANTLRInputStream.
* Change load( ) method to private
* Remove getSourceName()

### DIFF
--- a/runtime/Java/src/org/antlr/v4/runtime/ANTLRFileStream.java
+++ b/runtime/Java/src/org/antlr/v4/runtime/ANTLRFileStream.java
@@ -38,26 +38,22 @@ import java.io.IOException;
  * when you construct the object.
  */
 public class ANTLRFileStream extends ANTLRInputStream {
-	protected String fileName;
-
+	private String encoding;
+	
 	public ANTLRFileStream(String fileName) throws IOException {
 		this(fileName, null);
 	}
 
 	public ANTLRFileStream(String fileName, String encoding) throws IOException {
-		this.fileName = fileName;
-		load(fileName, encoding);
+		this.name = fileName;
+		this.encoding=encoding;
+		load();
 	}
 
-	public void load(String fileName, String encoding)
+	private void load()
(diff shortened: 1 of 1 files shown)


sharwell:
While this would have almost certainly been an improvement in the early days of ANTLR 4, I'm concerned because this introduces breaking changes in an API that people are likely depending on. At this point I'm not comfortable with the impact of the change on existing users.

HSorensen:
If someone really use the existing interface to first use the interface to
create a ANTLRFileStream with one name and then use load with a different
name I think it is already broken.
The contract for the filename in the interface is the getSourceName().

Would it be better to use @deprecated for the fileName field and the
load(filename,encoding) method?

All the tests passes with the changes.
IMO, if someone is depending on the current interface there should be a
test-case for their needs.

sharwell:
> If someone really use the existing interface to first use the interface to create a ANTLRFileStream with one name and then use load with a different name I think it is already broken.

Yes, you are correct. I think it would make sense to update the `load` method to write to the `fileName` field. The more likely problems which I refer to as breaking changes would be the following:

1. A class can currently extend `ANTLRFileStream` and override the `load` method. All such code will break if this pull request is merged.
2. A class can currently extend `ANTLRFileStream` and access the `fileName` field.

> IMO, if someone is depending on the current interface there should be a test-case for their needs.

I agree completely, but the reality is this project doesn't include tests for all the cases people use. Instead, this project historically tends to rely on the code review process to identify potentially breaking changes before they make it into the code. With tens of millions of users spanning nearly every country on earth, every breaking change is not a matter of *if*, but *how many*.

HSorensen:
You are right. 
If you must preserve existing behavior you are right, but that also means you cannot update fileName from load( ) in-case some is using the existing behaviour:)
. 
I created a second version using @Deprecated.

PR #1018, opened 2015-10-09
Update doc comment in C# base listener and visitor classes

brwml:
When using labeled rules, the doc comments in the emitted C# base listener and visitor classes contain a reference that does not exist. This produces a compiler warning. However, the corresponding interfaces are correct. It seems appropriate to use those comments for the base classes.

### DIFF
--- a/contributors.txt
+++ b/contributors.txt
@@ -77,4 +77,4 @@ YYYY/MM/DD, github id, Full name, email
 2015/06/29, jvanzyl, Jason van Zyl, jason@takari.io
 2015/08/18, krzkaczor, Krzysztof Kaczor, krzysztof@kaczor.io
 2015/09/18, worsht, Rajiv Subrahmanyam, rajiv.public@gmail.com
-2015/10/06, brwml, Bryan Wilhelm, bryan.wilhelm@live.com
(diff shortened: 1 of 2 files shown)


brwml:
Two commits were added in the same area to handle rule method resolution when left recursion was used in the grammar. The compiler issues a warning and assumes the parameterless method. This change explicitly uses the parameterless method for both labeled and unlabeled rules.

sharwell:
:thought_balloon: The latest updates might break the comments for rules with arguments. I'm on my phone right now; need to look again from a computer later.

brwml:
Yes. I was able to confirm that. We seem to be in a place where we can't completely eliminate the warning without adding more context. Would you prefer I revert the latter two commits for the time being?

sharwell:
@brwml If you revert the last two commits, the compiler will actually produce the behavior we want in the output. One option is to add the following alongside the other disabled warnings in the generated code (see the `fileHeader` template):

```csharp
// Ambiguous reference in cref attribute
#pragma warning disable 419
```

parrt:
@sharwell Is this now ready for me to merge?

sharwell:
@parrt :+1:

parrt:
ok, merging manually

PR #1020, opened 2015-10-13
Visitor function is now searched by context's constructor name

krzkaczor:
I changed the way that visit function searches next function to call. Previous solution didn't support labels. 

Is there any way to find label similar to `ctx.ruleindex`? I couldn't found anything so I came up with hacky solution that transforms constructor name. However it works perfectly fine for me.

### DIFF
--- a/runtime/JavaScript/src/antlr4/tree/Tree.js
+++ b/runtime/JavaScript/src/antlr4/tree/Tree.js
@@ -100,8 +100,9 @@ var visitAtom = function(visitor, ctx) {
 		return;
 	}
 
-	var name = ctx.parser.ruleNames[ctx.ruleIndex];
-	var funcName = "visit" + Utils.titleCase(name);
+	var exactCtxName = ctx.constructor.name.replace(/Context$/, '');
+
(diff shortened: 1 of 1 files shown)


parrt:
@ericvergnaud Does this PR have your blessing?

ericvergnaud:
Need to look into this

ericvergnaud:
Hi,
the correct method for calling the appropriate method of the visitor is to call the 'accept' method of the visited rule context, passing the visitor as a parameter.
Eric

krzkaczor:
Thanks for your comment. I will try to present better fix in few days.

parrt:
Hi guys. is this going anywhere?

krzkaczor:
Uhh hard to tell. I will try to take a look at this thing again today and fix PR. I have used that fix in my fork for some time now... ;)

parrt:
@krzkaczor @ericvergnaud any update? Should we try to fix this for merge or just drop it?

ericvergnaud:
drop it, the proposed strategy is wrong

PR #1025, opened 2015-10-21
Adding getTokenSource in BufferedTokenStream

hkff:
detected after calling parser.compileParseTreePattern ()

### DIFF
--- a/contributors.txt
+++ b/contributors.txt
@@ -78,3 +78,4 @@ YYYY/MM/DD, github id, Full name, email
 2015/08/18, krzkaczor, Krzysztof Kaczor, krzysztof@kaczor.io
 2015/09/18, worsht, Rajiv Subrahmanyam, rajiv.public@gmail.com
 2015/10/08, fedotovalex, Alex Fedotov, me@alexfedotov.com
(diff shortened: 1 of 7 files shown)


ericvergnaud:
Hi,
thanks for this, but to be honest I am not supportive of this PR.
The preferred patterns for accessing members vary across languages.
In Java, it is getters, in C# it is properties.
In Python and JavaScript, the preferred pattern is direct access, since the behavior can be controlled through the __getattr__ method in Python and the Object.createProperty feature in JavaScript.
I appreciate that the current Python and JavaScript code bases still largely mimic the Java implementation, due to their origin, but the long term plan is to gradually adopt natural language patterns.
Eric

hkff:
Hi,

I totally agree with that. So the other option is to change all the calls to getTokenSource by a direct access.

Walid

ericvergnaud:
Cool!
Please proceed :-)
Eric

hkff:
Done :)

parrt:
Hi. I will also need you to sign the contributors.txt file at the root of the repository. thanks!

hkff:
Done.

PR #1026, opened 2015-10-21
fix #1023: getExpectedTokens() returns out of context tokens on conse…

### LINKED ISSUE #1023: JavaScript: getExpectedTokens() behaves incorrectly on consecutive parser runs
I'm using an antlr4 generated JavaScript parser for code completion purposes with a custom language and noticed a strange behaviour resulting in out-of-context suggestions of the code completer.
For debugging purposes and to being able to reproduce this issue, I simplified my grammar to the following:

*TestLexer.g4*
```
lexer grammar TestLexer;

KEY_ENT : 'entity';
KEY_MEMB : 'members';
KEY_FOO : 'foo';
KEY_BAR : 'bar';
KEY_A : 'a';
KEY_B : 'b';
KEY_C : 'c';
CBO : '{';
CBC : '}';
ID : [A-Z][a-z]*;
COLON : ':';
SEMICOLON : ';';
STR : '"' (~'"')* '"';
WS  :  [ \t\r\n]+ -> skip;
ERROR_TOKEN : .;
```

*TestParser.g4*
```
parser grammar TestParser;
options {tokenVocab=TestLexer;}

root : entity+;
entity : KEY_ENT ID CBO entityContent CBC;
entityContent : members;
members : KEY_MEMB CBO member* CBC;
member : (KEY_A | KEY_B) COLON STR SEMICOLON;
```

The generated antlr4 JavaScript parser is run similarly to the basic JavaScript usage example. Consider the following code to reproduce my problem:
```javascript
var antlr4 = require('antlr4/index.js');
var TestLexer = require('./TestLexer.js');
var TestParser = require('./TestParser.js');

var ErrorListener = function(errors) {
  antlr4.error.ErrorListener.call(this);
  this.errors = errors;
  return this;
};

ErrorListener.prototype = Object.create(antlr4.error.ErrorListener.prototype);
ErrorListener.prototype.constructor = ErrorListener;
ErrorListener.prototype.syntaxError = function(rec, sym, line, col, msg, e) {
  this.errors.push(msg);
};

var input = 'entity Foo { members { a : "foo bar"; b : "foo"; X } }';

var errors = [];

// first parser run 
var chars = new antlr4.InputStream(input);
var lexer = new TestLexer.TestLexer(chars);
var tokens  = new antlr4.CommonTokenStream(lexer);

var parser = new TestParser.TestParser(tokens);
parser.buildParseTrees = true;
var listener = new ErrorListener(errors);

parser.removeErrorListeners();
parser.addErrorListener(listener);
parser.root();
console.log(errors); // --> extraneous input 'X' expecting {'a', 'b', '}'}'

// to this point everthing is like expected
// now we simulate a second run with newly instantiated objects on the very same input

errors = [];
chars = new antlr4.InputStream(input);
lexer = new TestLexer.TestLexer(chars);
tokens  = new antlr4.CommonTokenStream(lexer);

parser = new TestParser.TestParser(tokens);
parser.buildParseTrees = true;
listener = new ErrorListener(errors);

parser.removeErrorListeners();
parser.addErrorListener(listener);
parser.root();
console.log(errors); // --> extraneous input 'X' expecting {'entity', 'a', 'b', '}'}' 

// now 'entity' pops up within the expected tokens, which is not a valid expectation
// in the current context of the erroneous token 'X'
```
Looks like a bug to me. Even if all the objects are newly instantiated like above, some state seems to remain and leads to the faulty behaviour, for the output is always correct on the first run.

Thanks in advance for helping me with this issue. In the context of my actual, quite complex grammar, I get lots of faulty tokens from the root context, which is quite annoying because my code completion shows incorrect suggestions.
Martin

martin-probst:
recognizer.getExpectedTokens() returns the cached object from the ATN state. Therefore this object must not be modified to prevent unforeseeable behaviour. IntervalSet.addSet added the errorRecoverySet to the original cached expectedTokens, what lead to the behaviour observed in #1023 .

### DIFF
--- a/contributors.txt
+++ b/contributors.txt
@@ -78,3 +78,4 @@ YYYY/MM/DD, github id, Full name, email
 2015/08/18, krzkaczor, Krzysztof Kaczor, krzysztof@kaczor.io
 2015/09/18, worsht, Rajiv Subrahmanyam, rajiv.public@gmail.com
 2015/10/08, fedotovalex, Alex Fedotov, me@alexfedotov.com
+2015/10/21, martin-probst, Martin Probst, martin-probst@web.de
--- a/runtime/JavaScript/src/antlr4/error/ErrorStrategy.js
+++ b/runtime/JavaScript/src/antlr4/error/ErrorStrategy.js
@@ -269,7 +269,8 @@ DefaultErrorStrategy.prototype.sync = function(recognizer) {
     case ATNState.PLUS_LOOP_BACK:
     case ATNState.STAR_LOOP_BACK:
         this.reportUnwantedToken(recognizer);
-        var expecting = recognizer.getExpectedTokens();
+        var expecting = new IntervalSet();
+        expecting.addSet(recognizer.getExpectedTokens());
         var whatFollowsLoopIterationOrRule = expecting.addSet(this.getErrorRecoverySet(recognizer));
         this.consumeUntil(recognizer, whatFollowsLoopIterationOrRule);
         break;


ericvergnaud:
Much better!
Approved

ericvergnaud:
Many thanks.
Please sign the license agreement as part of this PR

sharwell:
@martin-probst Thanks for taking the time to not only report this, but to track it down and send a fix! Once you update the [contributors.txt](https://github.com/antlr/antlr4/blob/master/contributors.txt) file it should be ready.

PR #1031, opened 2015-10-28
Update TestRig.java help text with correct package name

HSorensen:
The help and error message from the TestRig program refers to the old `runtime.misc` instead of the new `gui` package:
```
java org.antlr.v4.runtime.misc.TestRig GrammarName startRuleName
  [-tokens] [-tree] [-gui] [-ps file.ps] [-encoding encodingname]
  [-trace] [-diagnostics] [-SLL]
  [input-filename(s)]
Use startRuleName='tokens' if GrammarName is a lexer grammar.
Omitting input-filename makes rig read from stdin.
```


parrt:
thanks!

PR #1046, opened 2015-11-19
Extra 'null' check in TestRig#process

abego:
After loading and casting the parserClass in TestRig#process using `cl.loadClass(parserName).asSubclass(Parser.class)` there is a check if the class returned by `#asSubclass` is not null:

```
parserClass = cl.loadClass(parserName).asSubclass(Parser.class);
if ( parserClass==null ) {
	System.err.println("Can't load "+parserName);
}
```

The method `#asSubclass` will never return `null`, but throw an `ClassCastException` when the cast is not valid, therefore the check can be removed.

See also: [TestRig#process Documentation](http://docs.oracle.com/javase/6/docs/api/java/lang/Class.html#asSubclass(java.lang.Class))


parrt:
Hi Udo! Heh, can you "sign" the antlr/antlr4/contributors.txt file so that I can keep the license clean? you can just add it as a commit or create a new pull request and I will merge it before I merge anything else.

abego:
Done. You will find a new pull request.

PR #1060, opened 2015-12-09
fix #1058

### LINKED ISSUE #1058: Javascript visitor modify prototype methods
Hello Terence and the team. I'm so impressed with Antlr that I wanted to donate but PayPal is not available in my country (at least as far as I know it is very complicated to move some money to a PayPal account). I'm working with the javascript runtime and found the following error in tree\Tree.js:

ParseTreeListener.prototype.visitTerminal = function(node) {
};

ParseTreeListener.prototype.visitErrorNode = function(node) {
};

These 2 methods must belong to the prototype of ParseTreeVisitor not ParseTreeListener (I modified it in my project and it worked). Everything else works great! You are the best programmers!

ericvergnaud:
missing visitor methods in javascript runtime

### DIFF
--- a/runtime/JavaScript/src/antlr4/tree/Tree.js
+++ b/runtime/JavaScript/src/antlr4/tree/Tree.js
@@ -95,6 +95,13 @@ ParseTreeVisitor.prototype.visit = function(ctx) {
 	}
 };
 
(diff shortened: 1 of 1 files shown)


sharwell:
Overall :+1: 

:question: Is there a way to include a test that covers this case?

PR #1063, opened 2015-12-16
Fix mistake in tokenIndex bounds check

pboyer:
I noticed a minor mistake in the JavaScript runtime. From what I can tell the `tokenIndex` property is never assigned and thus is `undefined`

Note that in JavaScript,

```js
undefined < anyInteger
```

returns `false`

### DIFF
--- a/contributors.txt
+++ b/contributors.txt
@@ -84,3 +84,4 @@ YYYY/MM/DD, github id, Full name, email
 2015/10/21, martin-probst, Martin Probst, martin-probst@web.de
 2015/10/21, hkff, Walid Benghabrit, walid.benghabrit@mines-nantes.fr
 2015/11/25, abego, Udo Borkowski, ub@abego.org
+2015/12/23, pboyer, Peter Boyer, peter.b.boyer@gmail.com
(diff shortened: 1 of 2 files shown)


ericvergnaud:
Hi,
thanks for this.
Have you signed the contributors agreement?
Eric

sharwell:
@pboyer Before we can accept this, you'll need to update the pull request to add a commit to sign [the contributor agreement](https://github.com/antlr/antlr4/blob/master/contributors.txt).

I would prefer to see a test covering this case, but it does appear to resolve a bug.

:memo: I am not a JavaScript developer, so @ericvergnaud will need to review this as well.

pboyer:
Thanks, guys! I'll take a look after work. Can you point out where I might add a test for this?

sharwell:
@pboyer I'm not sure a test for this case is easy to write with the current infrastructure, but @ericvergnaud should be able to point you in the right direction if there's a good place for it. :smile:

pboyer:
Hi @ericvergnaud and @sharwell!

I added my name to the contributors agreement. I have some more improvements for the js runtime but I'll send them in separate PR's.

PR #1064, opened 2015-12-16
Constructors should only call non-overridable methods

kirill-vlasov:
This pull request is focused on resolving occurrences of sonar issue Fixing squid:S1699 - Constructors should only call non-overridable methods
You can find more information about the issue here: https://dev.eclipse.org/sonar/coding_rules#q=squid:S1699

Please let me know if you have any questions.

Kirill Vlasov


sharwell:
:-1: This is a source and binary breaking change that doesn't appear to resolve any known bugs. Due to the number of users who rely on ANTLR, I believe API stability should have priority in cases like this so I don't think we should merge it.

parrt:
Thanks, but as Sam said, probably can't accept this update.

PR #1065, opened 2015-12-16
%n should be used in place of \n to produce the platform-specific line separator

kirill-vlasov:
This pull request is focused on resolving occurrences of sonar issue squid:S2275 - %n should be used in place of \n to produce the platform-specific line separator
You can find more information about the issue here: https://dev.eclipse.org/sonar/coding_rules#q=squid:S2275
Please let me know if you have any questions.
Kirill Vlasov

### DIFF
--- a/runtime/Java/src/org/antlr/v4/runtime/atn/ATNState.java
+++ b/runtime/Java/src/org/antlr/v4/runtime/atn/ATNState.java
@@ -187,7 +187,7 @@ public abstract class ATNState {
 			epsilonOnlyTransitions = e.isEpsilon();
 		}
 		else if (epsilonOnlyTransitions != e.isEpsilon()) {
-			System.err.format(Locale.getDefault(), "ATN state %d has both epsilon and non-epsilon transitions.\n", stateNumber);
+			System.err.format(Locale.getDefault(), "ATN state %d has both epsilon and non-epsilon transitions.%n", stateNumber);
 			epsilonOnlyTransitions = false;
 		}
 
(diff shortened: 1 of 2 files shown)


sharwell:
This seems like a neutral change.

runtime/Java/src/org/antlr/v4/runtime/atn/ATNState.java
```diff
@@ -187,7 +187,7 @@ public void addTransition(int index, Transition e) {
 			epsilonOnlyTransitions = e.isEpsilon();
 		}
 		else if (epsilonOnlyTransitions != e.isEpsilon()) {
-			System.err.format(Locale.getDefault(), "ATN state %d has both epsilon and non-epsilon transitions.\n", stateNumber);
+			System.err.format(Locale.getDefault(), "ATN state %d has both epsilon and non-epsilon transitions.%n", stateNumber);
```
sharwell: This should be impossible to hit. It's mostly here to make sure that if a code change is made (e.g. a tree-based optimization algorithm for grammars) which breaks this condition, it gets reported.

tool/src/org/antlr/v4/gui/PostScriptDocument.java
```diff
@@ -84,7 +84,7 @@ public String getPS() {
 	public void boundingBox(int w, int h) {
 		boundingBoxWidth = w;
 		boundingBoxHeight = h;
-		boundingBox = String.format(Locale.US, "%%%%BoundingBox: %d %d %d %d\n", 0,0,
+		boundingBox = String.format(Locale.US, "%%%%BoundingBox: %d %d %d %d %n", 0,0,
```
sharwell: The PostScript files are unlikely to be read as text. According to the language specification, LF and CRLF are treated the same way, so the changes in this file don't seem to help or hurt.

parrt:
Hiya! Thanks but as Sam says, I think \n is ok and that is really just generated stuff anyway.

PR #1079, opened 2015-12-23
Fix some issues with the Javascript language binding

dtymon:
This pull request fixes some issues I found in the Javascript language binding.

Thanks,
David

### DIFF
--- a/contributors.txt
+++ b/contributors.txt
@@ -84,3 +84,4 @@ YYYY/MM/DD, github id, Full name, email
 2015/10/21, martin-probst, Martin Probst, martin-probst@web.de
 2015/10/21, hkff, Walid Benghabrit, walid.benghabrit@mines-nantes.fr
 2015/11/25, abego, Udo Borkowski, ub@abego.org
+2015/12/17, sebadur, Sebastian Badur, sebadur@users.noreply.github.com
--- a/doc/listeners.md
+++ b/doc/listeners.md
@@ -8,7 +8,7 @@ By default, ANTLR-generated parsers build a data structure called a parse tree o
 
 The interior nodes of the parse tree are phrase names that group and identify their children. The root node is the most abstract phrase name, in this case `stat` (short for statement). The leaves of a parse tree are always the input tokens. Parse trees sit between a language recognizer and an interpreter or translator implementation. They are extremely effective data structures because they contain all of the input and complete knowledge of how the parser grouped the symbols into phrases. Better yet, they are easy to understand and the parser generates them automatically (unless you turn them off with `parser.setBuildParseTree(false)`).
 
-Because we specify phrase structure with a set of rules, parse tree subtree roots correspond to grammar rule names. ANTLR has a ParseTreeWalker that knows how to walk these parse trees and trigger events in listener implementation objects that you can create. ANTLR the tool generates listener interfaces for you also unless you, unless you turn that off with a commandline option. You can also have it generate visitors. For example from a Java.g4 grammar, ANTLR generates:
+Because we specify phrase structure with a set of rules, parse tree subtree roots correspond to grammar rule names. ANTLR has a ParseTreeWalker that knows how to walk these parse trees and trigger events in listener implementation objects that you can create. The ANTLR tool generates listener interfaces for you also, unless you turn that off with a commandline option. You can also have it generate visitors. For example from a Java.g4 grammar, ANTLR generates:
 
 ```java
 public interface JavaListener extends ParseTreeListener<Token> {
@@ -27,11 +27,10 @@ Assuming you've created a listener object called `MyListener`, here is how to ca
 JavaLexer lexer = new JavaLexer(input);
 CommonTokenStream tokens = new CommonTokenStream(lexer);
 JavaParser parser = new JavaParser(tokens);
-ParserRuleContext<Token> tree = parser.compilationUnit(); // parse
- 
-ParseTreeWalker walker = new ParseTreeWalker(); // create standard walker
+JavaParser.CompilationUnitContext tree = parser.compilationUnit(); // parse a compilationUnit
+
 MyListener extractor = new MyListener(parser);
-walker.walk(extractor, tree); // initiate walk of tree with listener
+ParseTreeWalker.DEFAULT.walk(extractor, tree); // initiate walk of tree with listener in use of default walker
 ```
 
 Listeners and visitors are great because they keep application-specific code out of grammars, making grammars easier to read and preventing them from getting entangled with a particular application.


ericvergnaud:
Hi,

thanks for this.

Sorry if the below sounds a bit negative but there are to many changes in there for a single commit.
Could you split them on a per issue basis, so we can track changes/comments/tests?

Some of the proposed changes are no brainers (typos/missing requires)

Others look like bug fixes but it’s unclear which bug is being fixed.
Do you have reproducible examples of why they would be useful?

Some are purely stylistic, and I will not accept them.

Also you will need to sign the contributors agreement.

Eric

dtymon:
Hi Eric,

Can you identify which of the changes you consider to be purely stylistic?

I use Emacs which I have configured to trim trailing whitespace which I concede has added more changes to the diffs than is necessary. Apart from these, all changes have been made for a reason.

The only other changes I can see that you might think are stylistic are the 'numChildren' type changes. However, these have been made in sections of code that are high volume calls.

I have developed an HTML template rendering engine using ANTLR and these loops can get executed millions of times per page render. Unlike other languages like C++, Javascript has no means to optimise the evaluation of the loop termination condition and, as such, evaluates '.length' on each iteration. If length was a simple scalar this would not be an issue. However, in this instance it is defined as a JS getter with a non-trivial implementation, thus adding considerable unnecessary overhead to the loop.

This behaviour is confirmed by placing a breakpoint in the getter implementation.

By simply evaluating this once outside the for() loop, given the loop does not alter the length of the list, a measurable performance improvement can be gained.

I have over 30 commits, bug fixes and improvements to the ANTLR Javascript binding, that have reduce render times by over 40%. Some of the modifications bring the implementation closer to Terence's Java implementation, which I have used as a reference implementation where necessary.

I would like to think that the majority of these changes would be accepted into the main ANTLR branch as I see them as improving the overall quality and performance of the code base. If not, then the changes will be committed to my fork of the project for others to use as they see fit.

Thanks,
David

ericvergnaud:
Hi David,

thanks for your patience.

I have no doubt that the proposed changes were made for a reason, but since that reason is not explicit there is no way for me to understand it.

Although the JavaScript runtime is young, it currently works very well for many users. Any change can be seen as both an improvement and a risk, and only an explanation or a reproducible test can help arbitrate the decision to include or not.

As an example, it's great that you have identified performance improvements, but since it was not stated in your commit proposal, it does look like purely stylistic. 

This is the perfect example why proposing fully documented and small dedicated commits is important.  

I believe ++n vs n += 1 is also purely stylistic, but maybe you can prove otherwise? 

Eric

ericvergnaud:
Hi again,

it would also be great if you could provide your performance framework, and how you measure this 40% improvement.

Eric

dtymon:
Hi Eric,

Thanks for your feedback. I'll try to give you a better idea of what we are using ANTLR JS for which hopefully might explain the changes we have made. Also, I haven't had a chance to thank you yet for providing the JS binding as it has proven invaluable for us.

We have developed an asynchronous HTML rendering engine for Node.js using ANTLR to define the grammar and generate the parse tree. It is quite a significant grammar that allows for conditional branching, looping and logical expressions. I have attached our lexer and parser files and a small sample HTML file:

[BmdTemplateLexer.g4.txt](https://github.com/antlr/antlr4/files/71264/BmdTemplateLexer.g4.txt)
[BmdTemplateParser.g4.txt](https://github.com/antlr/antlr4/files/71263/BmdTemplateParser.g4.txt)
[sample.html.txt](https://github.com/antlr/antlr4/files/71274/sample.html.txt)

As you point out and I can confirm, the master version of ANTLR JS successfully parses files conforming to the grammar. However, it was considerably slower than the Java binding, more than we thought could be attributed to the strengths of Java over JS. So we started to investigate if we could figure out why it was that much slower.

This investigation took around 2 months. The approach we took was to look at the decision making process of the Java and JS bindings and where they deviated. I don't have a full grasp of Terence's work so used a fairly rudimentary approach of putting verbose debug in both bindings and analysing them for differences and tracing it back to the code to try to understand why there were differences. When necessary, we used the Java implementation as a reference to see if it had been correctly translated into JS.

Some of the things we found were not be immediately obvious just looking at the code. For example, dumping out the DFAState hash string showed that it can be extremely large. Here is a sample with a key length of 1047 chars:

```
[(225,1,[$]), (230,2,[$]), (235,3,[$]), (238,4,[$]), (243,5,[$]), (248,6,[$]), (254,7,[$]), (258,8,[$]), (261,9,[$]), (268,10,[$]), (275,11,[$]), (280,12,[$]), (288,13,[$]), (298,14,[$]), (303,15,[$]), (306,16,[$]), (314,17,[$]), (319,18,[$]), (327,19,[$]), (335,20,[$]), (346,21,[$]), (354,22,[$]), (363,23,[$]), (367,24,[$]), (372,25,[$]), (377,26,[$]), (384,27,[$]), (386,28,[$]), (388,29,[$]), (390,30,[$]), (392,31,[$]), (394,32,[$]), (396,33,[$]), (398,34,[$]), (400,35,[$]), (402,36,[$]), (404,37,[$]), (406,38,[$]), (408,39,[$]), (410,40,[$]), (413,41,[$]), (416,42,[$]), (418,43,[$]), (421,44,[$]), (424,45,[$]), (426,46,[$]), (428,47,[$]), (430,48,[$]), (432,49,[$]), (434,50,[$]), (437,51,[$]), (440,52,[$]), (443,53,[$]), (446,54,[$]), (448,55,[$]), (451,56,[$]), (453,57,[$]), (455,58,[$]), (467,59,[$]), (472,60,[$]), (476,60,[$]), (458,61,[485 $]), (462,62,[490 $]), (494,63,[$]), (499,63,[$]), (505,63,[$]), (511,63,[$]), (544,64,[537 $]), (546,64,[537 $]), (538,64,[$]), (583,65,[580 $]), (585,66,[$]), (592,67,[$]), (596,67,[$])]
```

I am not sure if it is our grammar that can result in such long strings. Compounding the length of the hash is, when they differ, they often differ near the end. This means that it is very expensive to do an equality check or use them as keys in JS Objects. We found it was much more efficient to use something like the Murmur hashes that are used in the Java binding.

During this process, it was difficult to raise bugs against ANTLR JS since we were on a journey and we had no idea where it was going to take us. We have reached the end of that journey and believe we have a variant of ANTLR JS that has considerable benefits for anyone with significant grammars they would like to implement in JS. We would like to contribute these back to the community to say thanks to the ANTLR team for helping us out.

As for the 40% performance improvement, here is some output of a test utility we have. We pass it in HTML files containing our rendering language and it attempts to generate a parse tree for it, effectively looking for syntax errors in the HTML templates. We currently have around 7300 files to validate. Running it with our version takes in the region of 13.5 seconds to complete:
```shell
$ rm -f node_modules/antlr4; ln -sf node_modules/antlr4-dtymon antlr4
$ time find content -name \*.html | node deploy/antlr-verify
node   12.30s user 0.91s system 95% cpu 13.770 total
$ time find content -name \*.html | node deploy/antlr-verify
node   12.25s user 0.42s system 98% cpu 12.902 total
$ time find content -name \*.html | node deploy/antlr-verify
node   12.05s user 0.41s system 99% cpu 12.527 total
$ time find content -name \*.html | node deploy/antlr-verify
node   12.28s user 1.00s system 95% cpu 13.865 total
$ time find content -name \*.html | node deploy/antlr-verify
node   12.24s user 0.69s system 97% cpu 13.283 total
```
If we replace our `antlr4` node module with the current 4.5.1 version, it takes around 22 seconds to complete:
```shell
$ rm -f node_modules/antlr4; ln -sf node_modules/antlr4-master antlr4
$ time find content -name \*.html | node deploy/antlr-verify
node   20.58s user 0.76s system 97% cpu 21.866 total
$ time find content -name \*.html | node deploy/antlr-verify
node   22.07s user 0.81s system 98% cpu 23.152 total
$ time find content -name \*.html | node deploy/antlr-verify
node   20.69s user 0.56s system 99% cpu 21.449 total
$ time find content -name \*.html | node deploy/antlr-verify
node   21.15s user 0.76s system 98% cpu 22.249 total
$ time find content -name \*.html | node deploy/antlr-verify
node   20.82s user 0.47s system 98% cpu 21.539 total
```
Our version is approximately 38.5% faster than the current 4.5.1 version.

As for changing `i += 1` to `++i`, you are correct, it is completely stylistic. However, `i += 1` is very Python-esque given its lack of pre and post increment.

I hope this gives you a better understanding of how we got to where we are now.

Thanks,
David

ericvergnaud:
Hi David,

great! except you've just worsten the issue of piling up multiple changes in a single PR.
Can you PLEASE submit a PR per independent change, so we can walk through it calmly and have the ability to revert it independently?

Eric

dtymon:
Hi Eric,

Yeah sorry, I did not intend for all of these commits to be handled in one PR. I've tried to isolate the changes in each commit to be one logical fix/improvement.

I will create a separate PR for each and explain what each one tries to address.

Most of the early commits are independent of each other so you should be able to pick and choose. Same of the later commits though are dependent on earlier commits. All of our changes are now committed to our fork.

Would you like me to create all of the PRs now, or just start off slowly rather than inundate you with PRs?

Thanks,
David

PR #1083, opened 2015-12-24
JS: LL1Analyzer passing PredictionContext incorrectly to ATNConfig ctor

dtymon:
The PredictionContext should be passed to the ATNConfig constructor in the first argument, the `params` object. Instead, it is being passed as the second argument which is intended to be the config.


ericvergnaud:
blessed

PR #1084, opened 2015-12-24
JS: LexerActionExecutor cached hash string name clashes with function

dtymon:
LexerActionExecutor caches its hash string in a member called
'hashString'. However, the class also has a method with the
same name which leads to unexpected results.

The member has been renamed to '_hashString' to avoid the name
clash.


ericvergnaud:
blessed

Issue #1002, opened 2015-09-21
Ignore case new syntax


KvanTTT:
What do you think about new lexer command for case ignoring?

``` ANTLR
Procedure: 'procedure' -> ignoreCase;
```

Or to use regex-like syntax (which gives more flexibility, for example ignore part of word):

``` ANTLR
Procedure: '\iprocedure\-i';
Procedure: '\i' 'procedure' '\-i';
```

Now I have to use the following not pretty syntax:

``` ANTLR
Procedure: P R O C E D U R E;
fragment A: [aA];
fragment B: [bB];
fragment C: [cC];
fragment D: [dD];
fragment E: [eE];
fragment F: [fF];
fragment G: [gG];
fragment H: [hH];
fragment I: [iI];
fragment J: [jJ];
fragment K: [kK];
fragment L: [lL];
fragment M: [mM];
fragment N: [nN];
fragment O: [oO];
fragment P: [pP];
fragment Q: [qQ];
fragment R: [rR];
fragment S: [sS];
fragment T: [tT];
fragment U: [uU];
fragment V: [vV];
fragment W: [wW];
fragment X: [xX];
fragment Y: [yY];
fragment Z: [zZ];
```

This feature especially important for case-insensivity languages, such as Pascal, PHP, etc.

jimidle:
Don't create all to fragment rule just override LA in the input stream and always return toLower. 
Jim






On Mon, Sep 21, 2015 at 4:04 AM -0700, "Ivan Kochurkin" <notifications@github.com> wrote:












What do you think about new lexer command for case ignoring?

Procedure: 'procedure' -> ignoreCase;



Or to use regex-like syntax (which gives more flexibility, for example ignore part of word):

Procedure: '\iprocedure\-i';
Procedure: '\i' 'procedure' '\-i';



Now I have to use the following syntax:

Procedure: P R O C E D U R E;
fragment A: [aA];
fragment B: [bB];
fragment C: [cC];
fragment D: [dD];
fragment E: [eE];
fragment F: [fF];
fragment G: [gG];
fragment H: [hH];
fragment I: [iI];
fragment J: [jJ];
fragment K: [kK];
fragment L: [lL];
fragment M: [mM];
fragment N: [nN];
fragment O: [oO];
fragment P: [pP];
fragment Q: [qQ];
fragment R: [rR];
fragment S: [sS];
fragment T: [tT];
fragment U: [uU];
fragment V: [vV];
fragment W: [wW];
[...]

PR #1092, opened 2016-01-08
[RFC] Case Insensitivity Proof of Concept

KvanTTT:
At present time there are quite a lot of case insensitive languages: [Pascal](https://github.com/antlr/grammars-v4/tree/master/pascal), [PHP](https://github.com/antlr/grammars-v4/tree/master/php), [SQLite](https://github.com/antlr/grammars-v4/tree/master/sqlite), []() [TSQL](https://github.com/antlr/grammars-v4/tree/master/tsql) and other. In the usual case this feature implemented via fragment rules (code generation approach, see [here](https://github.com/antlr/grammars-v4/blob/master/pascal/pascal.g4#L453) for example) or via overriding LA method in the input stream (runtime approach as mention @jimidle [here](https://github.com/antlr/antlr4/issues/1002)).

I implemented new option and syntax in ANTLR to support case insensitivity. In my approach it is possible to declare caseInsensitive and caseSensitive modes (it can be used in grammar for parsing both PHP code (caseInsensitive keywords) and JavaScript code (caseSensitive keywords):
```ANTLR
mode JavaScript, caseSensitive;
Break: 'break';
Do: 'do';
Instanceof: 'instanceof';
...

mode PHP, caseInsensitive;
Abstract: 'abstract';
Array: 'array';
As: 'as';
```

Also it is possible to declare caseInsensitive option for entire grammar (combined grammar also supported):
```ANTLR
grammar pascal;
program
   : programHeading 'interface'? block '.'
   ;
programHeading
   : 'program' identifier ('(' identifierList ')')? ';'
   | 'unit' identifier ';'
   ;
// Another parser rules
...

INTERFACE: 'interface';
PROGRAM: 'program';
UNIT: 'unit';
// Another tokens with string literals (without fragment rules).
```

Moreover it is possible to declare caseSensitive and caseInsensitive tokens in the same mode like this:
```ANTLR
lexer grammar L;
Token_1: 'a';

mode DEFAULT_MODE, caseSensitive;
Token_3: 'd'
```

See unit tests for detail and other cases.

Сase insensitivity implemented via code generation approach, but it can be replaced with runtime approach too if necessary. This feature improve grammar readability and make grammar creation more easy.

If this pull request overall is good, I'll finish remaining issues with separated lexer and parser and other. Otherwise I'll try to separate fixed symbol issues and make another pull request. Suggestions are welcome.


KvanTTT:
Sorry, I did not run runtime tests. I'll fix it.

KvanTTT:
I have not idea why "Method L:_serializeATN () is too complex" exception occured for C# runtime. :unamused:

ericvergnaud:
Probably because the recursive string concatenation ("abc" + "bcd" + … ) is now exceeding the compiler stack limit

KvanTTT:
@ericvergnaud, no, this issue related to lexer. Maybe method has too big size. Local tests are passed (Windows 7). I tried to disable "inline" optimization as described [here](https://bugzilla.xamarin.com/show_bug.cgi?id=5092), but test still failed.
I don't understangd why it occured after my pull request.

ericvergnaud:
lexer also has its own serializedATN, and the failing method is TestHugeLexer
I haven’t look at your PR, but given the nature of the proposal, I guess it almost doubles the size of the ATN.

KvanTTT:
@ericvergnaud, I understand you, but in this test serializedATN size should not be changed because of caseInsensitive option is not enabled. But I'll look at it in more detail.

parrt:
Any change to the meta-language requires deep thought on my part. Not sure when I can devote the time.

parrt:
Hi. I'm going to close not because it's not an excellent job but because it's a fairly significant change and I'm nervous about unintended consequences.

PR #1094, opened 2016-01-14
Only run the maven plugin when needed.

Pursuit92:
For grammar-heavy projects, running the generator every time results in a lot of extra compilation that slows down builds. 

This makes it so that the antlr tool is only run when the tokens file either doesn't exist, or is older than the last modification to its grammar.

### DIFF
--- a/antlr4-maven-plugin/src/main/java/org/antlr/mojo/antlr4/Antlr4Mojo.java
+++ b/antlr4-maven-plugin/src/main/java/org/antlr/mojo/antlr4/Antlr4Mojo.java
@@ -372,6 +372,22 @@ public class Antlr4Mojo extends AbstractMojo {
         scan.addSourceMapping(mapping);
         Set<File> grammarFiles = scan.getIncludedSources(sourceDirectory, null);
 
+        // We don't want the plugin to run for every grammar, regardless of whether
+        // it's changed since the last compilation. Check the mtime of the tokens vs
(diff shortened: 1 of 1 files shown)


jimidle:
Keying the generation of code based on the tokens file is not an option as
it is not the sole indicator of change.

IFF we wish this functionality, then the timestamp of the generated
code/outputs must be checked against the timestamp of the source grammar
file.

This is not as simple a task as it first seems to be - I do not think we
can accept this pull request - it would break many a development build.

Jim

Pursuit92:
Is it ever the case that a tokens file is *not* generated when the antlr tool is run?

KvanTTT:
I like this idea (actually I implemented this in my private project).
For reliability the SHA-1 hash (or another) of respective grammar can be also added to the file header in comment.

parrt:
The tokens file is always generated but it might not change the contents.

Pursuit92:
That's fine, this will work as long as the mtime on the tokens file changes, which it will if the the generator gets run.

parrt:
Hm...I'm pretty sure this will work. Some people make their own .tokens files but ANTLR always generates one based upon:

```java
	/** What is the name of the vocab file generated for this grammar?
	 *  Returns null if no .tokens file should be generated.
	 */
	public String getVocabFileName() {
		return g.name+VOCAB_FILE_EXTENSION;
	}
```

I'm approving it.

PR #1100, opened 2016-02-01
fixes bug in visit()

spurcell:
In the common case where a rule has labeled productions, visit() would look for a function named visitX where X was the rule name, not the *label* name - and that function isn't even generated by the generator, since it generates functions for each labeled rule.

E.g. in the rule below, the visitor would be generated with functions visitWrap(), visitAdd(), and visitMult() - but the visit function logic would attempt to call visitExpr(), which is *not* generated.

```
expr
    : '(' expr ')'	# wrap
    | expr '+' expr	# add
    | expr '*' expr	# mult
    ;
```

Also: the package on NPM is seriously out of date and there's critical stuff missing (e.g. a working visitor) - an update would be appreciated!

### DIFF
--- a/contributors.txt
+++ b/contributors.txt
@@ -88,3 +88,5 @@ YYYY/MM/DD, github id, Full name, email
 2015/12/17, sebadur, Sebastian Badur, sebadur@users.noreply.github.com
 2015/12/23, pboyer, Peter Boyer, peter.b.boyer@gmail.com
 2015/12/24, dtymon, David Tymon, david.tymon@gmail.com
+2016/01/31, sp, Seth Purcell, sp@users.noreply.github.com
+
--- a/runtime/JavaScript/src/antlr4/tree/Tree.js
+++ b/runtime/JavaScript/src/antlr4/tree/Tree.js
@@ -107,8 +107,8 @@ var visitAtom = function(visitor, ctx) {
 		return;
 	}
 
-	var name = ctx.parser.ruleNames[ctx.ruleIndex];
-	var funcName = "visit" + Utils.titleCase(name);
+	// trim off 'Context' and prepend 'visit'
+	var funcName = 'visit' + ctx.constructor.name.substr(0, ctx.constructor.name.length - 7);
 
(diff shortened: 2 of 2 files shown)


ericvergnaud:
Hi,

there is indeed a bug, but the correct method for calling the appropriate method of the visitor is to call the 'accept' method of the visited rule context, passing the visitor as a parameter.

Eric

parrt:
@ericvergnaud should I accept or should be changed to accept from visit?

ericvergnaud:
@parrt should be changed

Envoyé de mon iPhone

spurcell:
I was trying to get the implementation to conform to the examples in section 4.2 of the ANTLR bible e.g.:

``` Java
@Override
public Integer visitMulDiv(LabeledExprParser.MulDivContext ctx) {
    int left = visit(ctx.expr(0)); // get value of left subexpression 
    int right = visit(ctx.expr(1)); // get value of right subexpression 
    if ( ctx.op.getType() == LabeledExprParser.MUL ) 
    return left * right; return left / right; // must be DIV
}
```

Note the use of `visit(ctx.expr(0))` rather than `ctx.expr(0).accept()`. Usage of `accept()` is mentioned once on page 243 of my copy but never appears in an example; I was going by the only docs I had. (Also, the workaround for languages without double dispatch is obviously not required in JavaScript.)

Fleshing out the JS docs to include a visitor example in the supported style (and actions) would make this package much more user-friendly.

ericvergnaud:
Hi,
you are welcome to provide such an example.
Eric

jamespedid:
What version is this fixed in?

I downloaded antlr4 on npm and I am running into this issue with the visitor pattern, doing the example in the antlr4 book.

BurtHarris:
Good catch @jamespedid: its not fixed, and there is no active bug tracking this issue.   If you could log one it's probably the best way of moving it forward.

The status of this Pull Request reflects the fact it was never merged.   @spurcell didn't seem to understand @ericvergnaud's feedback, and simply closed the pull request.   Its probably just a communications mixup.

Issue #1048, opened 2015-11-20
"Unary prefix" cases preferred ahead of "primary" cases in left-recursive rules


michaelpj:
```
grammar repro;

expr : literal                      
     | op expr                    
     | left=expr op right=expr
     ;

literal : '-'? Integer;

op : '+' | '-' ;

Integer : [0-9]+ ;

WS  :  [ \t\r\n\u000C]+ -> channel(1) ;
```

```
echo "-1" | java -cp antlr-4.5.1-complete.jar:. org.antlr.v4.gui.TestRig repro expr -tree
(expr (op -) (expr (literal 1)))
```

That is, `-1` parses as a unary minus applied to a literal 1, rather than a literal -1, despite the fact that the latter rule is an earlier alternative.

Now, I know that ANTLR rewrites left-recursive rules to parse them sensibly. Examining the log, I can see the following rewrite of the rule:
```
2015-11-20 11:29:01:089 left-recursion LogManager.java:49 expr
    :   ( {} op<tokenIndex=13> expr<tokenIndex=15,p=2> 
        | literal<tokenIndex=9> 
        )
        (
          {precpred(_ctx, 1)}?<p=1> op<tokenIndex=23> right=expr<tokenIndex=27,p=2>
        )*
    ;
```
It looks like we put the recursive-but-not-left-recursive alternatives first, then the non-recursive (primary) alternatives, and then the left-recursive alternatives. This surprises me - I would have thought that we'd try the non-recursive alternatives first, but that doesn't appear to be the case.

Is this the intended behaviour?

michaelpj:
To clarify my surprise a bit more: since this is an example of an ambiguity, I would have expected the order of alternatives to affect which version is chosen.

parrt:
Hmm...yes, this seems very odd. Looking at your PR now too.

PR #1121, opened 2016-02-22
Retain ordering of non-left-recursive alternatives during rule transformation

michaelpj:
During the left-recursive rule transformation, we walk the grammar collecting both "prefix" and "other" alternatives. These rules differ primarily in that we only want to attach a precedence to the "prefix" rules, as these ones are recursive. After that, they are all put together into a single block at the start of the generated rule.

However we track the `prefixAlts` and `otherAlts` separately, and when we write out the generated rule we put all the `prefixAlts` before all the `otherAlts`. That means we lose the information about any alternative ordering between these rules, which can lead to surprising behaviour (see https://github.com/antlr/antlr4/issues/1048).

I can't see an _a priori_ reason why the "prefix" alternatives should have to come before the "other" alternatives at this point, so this PR retains the ordering by simply tracking them in the same list, ensuring that the rewritten rules appear in the same order that we encountered their sources.

The tests in the first commit fail on master (by parsing `-1` as `(expr (op -) (literal 1))` rather than `(expr (literal - 1))`), but pass with this PR.


parrt:
I'll merge manually so I can tweak name of `prefixAlts`.

michaelpj:
Thanks!

parrt:
thanks you!

PR #1130, opened 2016-03-07
Fix Javadoc errors

gagern:
This fixes all errors (but not the numerous warnings) reported by Oracle JDK 1.8.0_72 in a simple run of `mvn javadoc:javadoc`. Mostly unescaped `>`, `<`, `&`, but also self-closed `<p/>` and similar. Plus I deleted two commented-out functions, since the function body got included in the doc comment, and since git will be able to recover them so removing them completely helps keep the sources cleaner.


parrt:
can you add yourself to contributors.txt?

gagern:
@parrt: Done that

PR #1131, opened 2016-03-07
Set stop token in postamble

### LINKED ISSUE #1129: $stop not available in @after
The [parser rule attributes](https://github.com/antlr/antlr4/blob/master/doc/actions.md#parser-rule-attributes) documentation claims:

> **`stop` (Token):** The last nonhidden channel token to be matched by the rule. When referring to the current rule, this attribute is available only to the after and finally actions.

But apparently the token is not available even in the `@after` action.

Example:

```
grammar test ;

top
    : e { System.out.println("e.start = " + $e.start + ", e.stop = " + $e.stop); }
    ;

e
    @after { System.out.println("start = " + $start + ", stop = " + $stop); }
    : '(' e ')'
    | OTHER+
    ;

OTHER: .;
```

Running `(1+1)` through the `TestRig` I see

```
start = [@1,1:1='1',<3>,1:1], stop = null
start = [@0,0:0='(',<1>,1:0], stop = null
e.start = [@0,0:0='(',<1>,1:0], e.stop = [@4,4:4=')',<2>,1:4]
```

I observed this in ANTLR 4.5.2 and could reproduce it in a 4.5.3 snapshot based on d0fb48c2b89128158950f4b32a61d50d52d981ea.

gagern:
This ensures that the stop token has been set before the `@after` action gets executed.

This is one way to fix #1129. Another would be adding the corresponding line directly to the method template. But the postamble-based approach should be more flexible in the long run.

### DIFF
--- a/tool/src/org/antlr/v4/codegen/DefaultOutputModelFactory.java
+++ b/tool/src/org/antlr/v4/codegen/DefaultOutputModelFactory.java
@@ -30,6 +30,7 @@
 
 package org.antlr.v4.codegen;
 
+import org.antlr.v4.codegen.model.Action;
 import org.antlr.v4.codegen.model.CodeBlockForOuterMostAlt;
 import org.antlr.v4.codegen.model.OutputModelObject;
 import org.antlr.v4.codegen.model.RuleFunction;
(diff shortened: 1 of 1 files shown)


gagern:
NB: In #1130 I've added myself to `contributors.txt`.

parrt:
Small tweak to only generate that if after or finally action exist.

PR #1137, opened 2016-03-15
JS: Do deep Actions array comparison in LexerActionExecutor equals()

dtymon:
According to the corresponding Java implementation, the equals() method for LexerActionExecutor should be doing an equals() test on each of the actions rather than testing the Array reference is equal.


parrt:
@ericvergnaud blessed?

ericvergnaud:
blessed

PR #1138, opened 2016-03-15
JS: Fix copy-and-paste error in OR SemanticContext toString() method

dtymon:
The OR toString() method was incorrectly added to the AND prototype


parrt:
@ericvergnaud blessed?

ericvergnaud:
blessed

PR #1139, opened 2016-03-15
JS: SemanticPredicate should be SemanticContext

dtymon:
AND.evalPrecedence() method is attempting to invoke a method on a non-existent class "SemanticPredicate". It should be using "SemanticContext" instead.


parrt:
@ericvergnaud blessed?

ericvergnaud:
blessed

PR #1140, opened 2016-03-15
JS: ATNConfig can incorrectly change "alt" of 0 to null

dtymon:
When an ATNConfig instance is created with an alt of 0, it is possible for it to be incorrectly changed to null due to its "falsey" nature.


parrt:
@dtymon are you in contributors.txt?

dtymon:
Yes Terence, I have added myself to contributors.txt in the past. Thanks.

ericvergnaud:
blessed

PR #1146, opened 2016-03-25
Add Vocabulary.getMaxTokenType()

### LINKED ISSUE #904: Vocabulary interface missing an iterable or equivalent method
Depreciation of Recognizer.getTokenNames() presently leaves no equivalent or convenient method to iterate over the valid set of symbolic token names.  Since the underlying String[] is now private, cannot iterate based on size. 

Adding a Vocabulary.getSymbolicNames() method, returning a String[], would be helpful.

msteiger:
Fixes #904

### DIFF
--- a/runtime/Java/src/org/antlr/v4/runtime/Vocabulary.java
+++ b/runtime/Java/src/org/antlr/v4/runtime/Vocabulary.java
@@ -37,6 +37,14 @@ package org.antlr.v4.runtime;
  * @author Sam Harwell
  */
 public interface Vocabulary {
+
+	/**
+	 * Returns the highest token type value. It can be used to iterate from
+	 * zero to that number, thus querying all stored entries.
+	 * @return the highest token type value
+	 */
+	int getMaxTokenType();
+
 	/**
 	 * Gets the string literal associated with a token type. The string returned
 	 * by this method, when not {@code null}, can be used unaltered in a parser
@@ -85,7 +93,7 @@ public interface Vocabulary {
 	 *
 	 * <ul>
 	 *  <li>Tokens created by lexer rules.</li>
-	 *  <li>Tokens defined in a {@code tokens{}} block in a lexer or parser
+	 *  <li>Tokens defined in a <code>tokens{}</code> block in a lexer or parser
 	 *  grammar.</li>
 	 *  <li>The implicitly defined {@code EOF} token, which has the token type
 	 *  {@link Token#EOF}.</li>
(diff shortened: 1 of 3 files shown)


runtime/Java/src/org/antlr/v4/runtime/Vocabulary.java
```diff
@@ -85,7 +93,7 @@
 	 *
 	 * <ul>
 	 *  <li>Tokens created by lexer rules.</li>
-	 *  <li>Tokens defined in a {@code tokens{}} block in a lexer or parser
+	 *  <li>Tokens defined in a <code>tokens{}</code> block in a lexer or parser
```
msteiger: Fixes invalid curly brackets in JavaDoc.

runtime/Java/src/org/antlr/v4/runtime/VocabularyImpl.java
```diff
@@ -94,6 +96,8 @@ public VocabularyImpl(String[] literalNames, String[] symbolicNames, String[] di
 		this.literalNames = literalNames != null ? literalNames : EMPTY_NAMES;
 		this.symbolicNames = symbolicNames != null ? symbolicNames : EMPTY_NAMES;
 		this.displayNames = displayNames != null ? displayNames : EMPTY_NAMES;
+		this.maxTokenType = Math.max(this.displayNames.length,
```
msteiger: This is very conservative as it even include display names. Pinging @sharwell for feedback on this one.

parrt:
hi. have you added yourself to contributors.txt?

msteiger:
I've added myself to the file. Thanks!

parrt:
ok, i hope to do a 4.5.3 soon.

parrt:
This isn't quite right I think. off-by-one:

```
		this.maxTokenType =
			Math.max(this.displayNames.length,
					 Math.max(this.literalNames.length, this.symbolicNames.length));
```

That is the length but max token type is less by 1. I.e., A grammar with two tokens:

```
a : A B ;
```

gets:

```
        private static final String[] _LITERAL_NAMES = {
        };
        private static final String[] _SYMBOLIC_NAMES = {
                null, "A", "B"
        };
```

Max length is 3 so max token type must be 1 less: 2.

msteiger:
Sorry, my fault. You're absolutely right.

mike-lischke:
If you only have empty names that - 1 will result in a -1 for maxTokenType. Is that intended or should this value be clipped to 0?

parrt:
hmm...good question. Actually -1 since max 0 would indicate 0 was a valid token index like tokens[0].

mike-lischke:
Ok, thanks.

parrt:
A related note: https://github.com/antlr/antlr4/issues/1309 maxTokenType for the ATN is set improperly due to bug in Grammar.getMaxTokenType(). :(

Issue #1148, opened 2016-03-26
Rules named 'return' generates invalid python2 code


Wingpad:
I was building a grammar with a rule called "return" and when I compiled it using the Python2 target it did not catch it as a reserved word and generated the code `def return(self):` which is invalid python syntax. This is an issue since the reserved word `tuple` is caught and the ANTLR target prevents auto-generated code that would otherwise fail.

parrt:
Fixed by https://github.com/antlr/antlr4/pull/1149

PR #1149, opened 2016-03-26
Fixing Issue #1148

Wingpad:
Just added return to the list of python2 keywords. It appears to be missing in the Python3 Target as well.


ericvergnaud:
blessed

PR #1154, opened 2016-03-30
Add ability to set parse tree internal node super class with option c…

### LINKED ISSUE #1152: Alt num is sometimes super useful in the parse tree internal nodes
Easiest solution is to add option: `-DtreeSuperClass=MyContext`, which could override `Parser`'s `enterOuterAlt()` method. This would mirror current `superClass` option. Then add a `ParserRuleContext` method called `getAltNumber()` that defaults to nothing: 0 (alts are 1..n). should be binary backward compatible with 4.5.2 and provides significant new power.

I guess I’ll leave the factory concept til later as it’s orthogonal.

parrt:
…ontextSuperClass. Provide impl in Java target that has altNum backing field.  Add test across targets to set/get alt num. Fixes #1152.


runtime/Java/src/org/antlr/v4/runtime/RuleContextWithAltNum.java
```diff
@@ -0,0 +1,20 @@
...
+ *
+ *  options {contextSuperClass=org.antlr.v4.runtime.RuleContextWithAltNum;}
+ *
+ *  that provides a backing field / impl for the outer alternative number
+ *  matched for an internal parse tree node.
+ *
+ *  I'm only putting into Java runtime as I'm certain I'm the only one that
+ *  will really every use this.
+ */
+public class RuleContextWithAltNum extends ParserRuleContext {
```
sharwell: :thought_balloon: I think this needs to override `copyFrom`, or labeled outer alternatives won't work with it.

PR #1157, opened 2016-04-01
Execute commands for test suite using the PATH

### LINKED ISSUE #1156: locateTool in runtime test suite too restrictive
On my MacPorts system, the python test suite fails to run out of the box because my `python3.5` binary is installed in `/opt/local/bin`, not in `/usr/bin/` or  `/usr/local/bin/` as expected by `BasePythonTest.locateTool`. Can't you use the `PATH` environment variable? Or even better yet, just name the binary and leave it to `ProcessBuilder` to launch the command by name only, using whatever `PATH` convention is to be applied on the current platform?

The `node` locator has a similar issue. But it doesn't throw a `RuntimeException` as the Python version does, but instead throws a `NullPointerException` since `locateTool` there just returns `null` which then is used to construct a `File`. The `csharp` thingy is a bit better on the one hand, since it calls `mono` by name without a path. But on the other hand it's worse since it attempts to locate `xbuild` using those hardwired paths without allowing for a system property override. It looks as though it attempty to run all the tests, but has `null` as the output for each of them. At this point I gave up trying to run the complete suite on my system.

On the whole it would be nice if you could just selectively skip tests for some of these languages, without failing the test phase altogether. Not sure whether Maven/Surefire/Junit is up to that task, though.

gagern:
This switches from some hardwired search paths to using command names only, relying on the system's `PATH` to locate the appropriate binary.
If that fails, there are system properties to override each setting, now even for C#.

This fixes #1156 for me on MacPorts Mac, but it should be tested on Ubuntu and Windows, too.

### DIFF
--- a/doc/building-antlr.md
+++ b/doc/building-antlr.md
@@ -112,6 +112,18 @@ antlr4-tool-testsuite/4.5.2-SNAPSHOT/antlr4-tool-testsuite-4.5.2-SNAPSHOT.jar
 
 Note that ANTLR is written in itself, which is why maven downloads antlr4-4.5.jar for boostrapping 4.5.2-SNAPSHOT purposes.
 
+If the test suite fails to find some required tool, it is possible to specify a name or full path
+using a system property, i.e. using a `-Dkey=value` argument to the `mvn` invocation.
+The following keys are understood:
+
+| language   | key                        | default value (documentation might be outdated)          |
+|------------|----------------------------|----------------------------------------------------------|
+| Python     | `antlr-python2-python`     | `python2.7`                                              |
+| Python     | `antlr-python3-python`     | `python3.5`                                              |
+| JavaScript | `antlr-javascript-nodejs`  | `nodejs` if found, `node` otherwise                      |
+| C#         | `antlr-javascript-msbuild` | default MSBuild 12.0 path on Windows, `xbuild` otherwise |
+| C#         | `antlr-javascript-mono`    | omitted on Windows, `mono` otherwise                     |
+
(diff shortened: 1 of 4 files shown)


parrt:
Hmm..this is risky given our recent change to runtime tests. does this PR still look reasonable?

parrt:
@gagern should I try to integrate with new runtime tests?

parrt:
I kinda of improved things in https://github.com/antlr/antlr4/pull/1376  but couldn't take all changes as we've improved runtime test rig since the original pull request. I'll close this one.

gagern:
What would be risky here?

I still think the general idea of my PR, i.e. using `PATH` instead of hard-coded paths, is the sane thing to do. I won't be able to port this to whatever changes to master there were in the meantime, particularly not if that new PR runs the same risk of lying around unnoticed until it becomes obsolete. If you really want this, and can promise a timely review, I may reconsider.

parrt:
@gagern Is it just a `PATH` check update? That sounds better than our current looking in likely spots approach.

gagern:
Well, my approach here had three components.

1. One system property per tool so you can configure everything, as I wrote in the new documentation paragraph.
2. Auto-detected versions use name instead of full path, so they will use `PATH` to actually find the program in question. This is in my opinion the most important aspect.
3. To distinguish `nodejs` from `node`, try executing the former and fall back to the latter. This same approach could in theory be used for systems which don't have `python2` but do have `python`.

parrt:
Cool. ok, I added a separate specific issue for the PATH. https://github.com/antlr/antlr4/issues/1407

PR #1161, opened 2016-04-07
JS: ParserRuleContext can change invokingState from 0 to null

dtymon:
If an instance of a class deriving from ParserRuleContext is created
with an invokingState of 0 then the constructor will incorrectly
change this to null due to the falsey nature of 0.

A similar problem exists in the base class RuleContext, where it
will change an invokingState of 0 to -1. Using the Java source as
a reference, this appears to be a bug. The invokingState should
be set to -1 when the parent is null.

It looks like the code is trying to watch for an invokingState of
null or undefined. However, I think this is unnecessary since the
rule context classes generated from the grammar already handle
these cases in their constructor, for example:

function ContentContext(parser, parent, invokingState) {
        if(parent===undefined) {
            parent = null;
        }
        if(invokingState===undefined || invokingState===null) {
                invokingState = -1;
        }

### DIFF
--- a/runtime/JavaScript/src/antlr4/ParserRuleContext.js
+++ b/runtime/JavaScript/src/antlr4/ParserRuleContext.js
@@ -59,8 +59,6 @@ var ErrorNodeImpl = Tree.ErrorNodeImpl;
 var Interval = require("./IntervalSet").Interval;
 
 function ParserRuleContext(parent, invokingStateNumber) {
-	parent = parent || null;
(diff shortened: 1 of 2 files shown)


ericvergnaud:
Hi,
can you describe the problem this would solve?
Eric

parrt:
@dtymon can you describe the problem this solves?

parrt:
closing due to lack of response from original author.

PR #1164, opened 2016-04-08
avoid null entries for ambigous tree list

WolfgangFahl:
might help in fixing https://github.com/antlr/intellij-plugin-v4/issues/258


parrt:
Hi. i'll need you to sign antlr4/contributors.txt.

parrt:
Turns out, i need to fix; see fixed version here https://github.com/antlr/intellij-plugin-v4/blob/master/src/java/org/antlr/intellij/plugin/preview/ShowAmbigTreesDialog.java#L246

PR #1171, opened 2016-04-13
Remove unnecessary semicolons from Python targets.

renatahodovan:
The Python templates contained superfluous semicolons after
the sync call of errHandler. This patch removes them.

PR #1172, opened 2016-04-13
Add missing default pass statements to Python targets.

renatahodovan:
The default pass statements were missing from the LL1AltBlock's
template in case of Python targets, which generated conditional statements
without body in some cases and ended in syntax error. The patch fixes this.

### DIFF
--- a/tool/resources/org/antlr/v4/tool/templates/codegen/Python2/Python2.stg
+++ b/tool/resources/org/antlr/v4/tool/templates/codegen/Python2/Python2.stg
@@ -319,7 +319,7 @@ self.state = <choice.stateNumber>
 token = self._input.LA(1)
 <choice.altLook,alts:{look,alt| <cases(ttypes=look)>
     <alt>
-	}; separator="\nel">
+    pass}; separator="\nel">
 else:
     <error>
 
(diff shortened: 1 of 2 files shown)


ericvergnaud:
Can you provide a minimal grammar to reproduce the faulty behavior so we can include it in our test suite?

renatahodovan:
Sure. A minimal repro case would be:
```
grammar Test;

start:
    ('a' |)
    ;
```

and the related code in the generated source is:

```py
            if token in [TestParser.T__0]:
                self.state = 2
                self.match(TestParser.T__0)

            elif token in [TestParser.EOF]:

            else:
                raise NoViableAltException(self)
```

Should I attach this to the patch (if yes what is the right place to do)?

ericvergnaud:
Hi,
no I guess the place is correct.
I'll try and write a test for this.
Have you signed the contributors file?
Eric

renatahodovan:
I've added myself in another (still open) PR: https://github.com/antlr/antlr4/pull/1171/

renatahodovan:
Without this patch, the Python{2,3} targets are broken. Any chance of landing it any time soon?

renatahodovan:
@ericvergnaud could you please take a look at this one too? (This pending PR is the main reason I need to provide an alternative jar file to my ANTLR-based [tool](https://github.com/renatahodovan/picireny#install) .)

ericvergnaud:
blessed

parrt:
This Fixes #1252

PR #1174, opened 2016-04-15
change runtime.misc to gui (new TestRig package)

umaranis:


parrt:

PR #1186, opened 2016-04-30
Fix typo in the parser of the Python3 runtime.

renatahodovan:
The last parameter name of Parser::enterRule in the Python3 runtime
was mistakenly combined with its type. The patch fixes this.


ericvergnaud:
Thanks!
blessed.
Eric

PR #1187, opened 2016-05-06
Fix initialization of ATNConfig when computing SLL prediction termination condition.

renatahodovan:
The PredictionMode::hasSLLConflictTerminatingPrediction method aims to
create ATNConfig objects from another ATNConfig and SemanticContext
objects. In case of the Python targets, the initialization
happened without keyword arguments. Since the called **init**
method had default values set for all the parameters, the parameter
substitution worked by indices. As a consequence, the first ATNConfig
parameter was wrongly interpreted as an ATNState and the SemanticContext as
an alternative. The patch fixes this by adding the missing keywords.

### DIFF
--- a/runtime/Python2/src/antlr4/atn/PredictionMode.py
+++ b/runtime/Python2/src/antlr4/atn/PredictionMode.py
@@ -213,7 +213,7 @@ class PredictionMode(object):
                 # dup configs, tossing out semantic predicates
                 dup = ATNConfigSet()
                 for c in configs:
-                    c = ATNConfig(c,SemanticContext.NONE)
+                    c = ATNConfig(config=c, semantic=SemanticContext.NONE)
                     dup.add(c)
                 configs = dup
             # now we have combined contexts for configs with dissimilar preds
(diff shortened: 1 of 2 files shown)


renatahodovan:
One possible reproduction is using the <a href='https://github.com/antlr/grammars-v4/blob/master/ecmascript/ECMAScript.PythonTarget.g4'>ECMAScript grammar</a> for the input ```a = 1 + 2 ``` and forcing the parser to use SLL prediction. As a result, the state field of ATNConfig will be an ATNConfig object and we will fail in ATNConfig::\_\_hash\_\_ with the following error message:

```
  File "/usr/local/lib/python3.5/site-packages/antlr4/atn/ATNConfig.py", line 111, in __hash__
    return hash((self.state.stateNumber, self.alt, self.context, self.semanticContext))
AttributeError: 'ATNConfig' object has no attribute 'stateNumber'
```

renatahodovan:
Without this patch, the Python{2,3} targets are broken. Any chance of landing it any time soon?

parrt:
@ericvergnaud these look correct to me. merging but let me know if it's an issue. I examined ATNConfig.py to be sure.

PR #1189, opened 2016-05-08
Update python-target.md

davek2:
Update documentation so that Python example will work.  Example "MyGrammar" has a lexer rule "StartRule" as the root.  This should be changed to parser rule "startRule".


parrt:

PR #1198, opened 2016-05-20
Update ANTLRParser.g

flip111:
Grammar seemed okay, fixed some spelling (pun intended)

by the way there were a bunch of non-comment things with `V3` in them, which seems odd since this file is in the V4 section. But i didn't dare to touch that.


parrt:
Hi. can you sign the contributors.txt file?

flip111:
I don't want to be liable for anything https://github.com/antlr/antlr4/blob/master/contributors.txt#L32 you can take the change as is or discard them whatever you want.

parrt:
I full understand.

PR #1210, opened 2016-06-07
Add C++ target to ANTLR4

mike-lischke:


parrt:
Actually it was merged from parrt/antlr4 via https://github.com/antlr/antlr4/pull/1346  Not sure why this says merged.

PR #1215, opened 2016-06-21
Added badges

adamretter:
- build status
- java version
- license
- latest download

You can see the visual impact here - https://github.com/adamretter/antlr4/tree/patch-1#readme


sharwell:
Interesting. Some of these I haven't seen before. My notes, which may or may not have anything to do with what @parrt will think:

* I like having the build status badge at the top.
* I don't like having a red badge (the download badge) so close to the build badge. To me it looks like one build is passing and one is failing. The bigger problem may be the fact that it has added *yet another* place where the version number needs to be incremented when a release is made.
* The other badges are useful information, but should they be badges? I don't know.

parrt:
yeah the red badge is kind of misleading. I'm okay with the other badges, but as Sam said, I'm not sure they should be badges. Perhaps I should simply say this out loud in the readme?

adamretter:
Heh I used the Red from the Antlr website, I thought you would like it. Not
a problem though.

I added the badges because I think visually it helps you find those simple
facts very easily. But well.. design is of course personal.

If you would like me to modify the PR, im happy to do so

parrt:
thanks! i've done a simpler version and pushed so i'm closing for 4.6.

PR #1230, opened 2016-07-10
Use relative links in README

twz123:
According to [GitHub Help: Relative links in READMEs](https://help.github.com/articles/relative-links-in-readmes/) it's cooler to use relative links. You'll also get the correct Markdown rendering.


parrt:
Hi. thanks! github recently changed and i need full paths, but those old ones were bad. i'll close as new readme seems ok.

PR #1231, opened 2016-07-10
Add an iterative version of the ParseTreeWalker

twz123:
I hit a `StackOverflowError` while traversing deeply nested trees using the standard `ParseTreeWalker`, so I decided to implement an iterative version of it using a linked data structure instead of recursion during traversal.


runtime/Java/src/org/antlr/v4/runtime/tree/IterativeParseTreeWalker.java
```diff
@@ -0,0 +1,151 @@
...
+ * An iterative (read: non-recursive) pre-order and post-order tree walker that
+ * doesn't use the thread stack but it's own heap-based stack using linked data
+ * structures. Makes it possible to process deeply nested parse trees.
+ */
+public class IterativeParseTreeWalker extends ParseTreeWalker {
+    public static final ParseTreeWalker DEFAULT = new IterativeParseTreeWalker();
+
+    @Override
+    public void walk(final ParseTreeListener listener, final ParseTree t) {
+        for (WalkerNode<?> currentNode = createNode(null, t, 0); currentNode != null;) {
```
sharwell: :bulb: This implementation requires at least one allocated object for every node in the tree. You could avoid all these allocations by instead using two stacks:

```java
Deque<SyntaxTree> treeStack = new ArrayDeque<SyntaxTree>();
IntegerStack currentIndex = new IntegerStack();
```

Where the node you are currently visiting is `treeStack.peek()` and its index in the parent is `currentIndex.peek()`.

runtime/Java/src/org/antlr/v4/runtime/tree/IterativeParseTreeWalker.java
```diff
@@ -0,0 +1,151 @@
...
+
+package org.antlr.v4.runtime.tree;
+
+/**
+ * An iterative (read: non-recursive) pre-order and post-order tree walker that
+ * doesn't use the thread stack but it's own heap-based stack using linked data
+ * structures. Makes it possible to process deeply nested parse trees.
+ */
+public class IterativeParseTreeWalker extends ParseTreeWalker {
+    public static final ParseTreeWalker DEFAULT = new IterativeParseTreeWalker();
```
sharwell: :question: Why are we not replacing the default `ParseTreeWalker` implementation with the iterative version? *At minimum* I don't see any reason to keep using the recursive implementation for the default walker `ParseTreeWalker.DEFAULT`.

runtime/Java/src/org/antlr/v4/runtime/tree/IterativeParseTreeWalker.java
twz123: You're right. I thought if I'd use a `LinkedList` for the stack, then I'd have object allocation anyways, just hidden inside the `LinkedList`. But I see that `ArrayDeque` does a better job here.

runtime/Java/src/org/antlr/v4/runtime/tree/IterativeParseTreeWalker.java
twz123: Well, I didn't want to be too intrusive with this PR and directly replace the recursive walker. If you think that the recursive one can be replaced, I'll happily do that :smile:

runtime/Java/src/org/antlr/v4/runtime/tree/IterativeParseTreeWalker.java
```diff
@@ -0,0 +1,151 @@
+/*
+ * [The "BSD license"]
+ *  Copyright (c) 2012 Terence Parr
```
mike-lischke: Isn't that copyright wrong? Neither exists this file since 2012 nor was it written by Sam + Ter.

sharwell: The year is particularly interesting. It turns out keeping the year constant and using the year when the code base started provides the longest and strongest copyright protections. If you want to see some of the details on this, take a look at the long and probably very boring comments in DotNetAnalyzers/StyleCopAnalyzers#1661.

The use of mine and Terence's names has to do with copyright assignment, which is (to the best of my knowledge) common practice in a number of circles. There are three ways we work to ensure the original author of patches remains recognized for the life of the project:

1. The **contributors.txt** file, which is an essential and core part of this project. It's not distributed, but the distribution links back to it.
2. The Git commit history.
3. The pull requests and issues filed by everyone.

:memo: I am not a lawyer

mike-lischke: Hmm, you certainly have your reasons, but to me it feels wrong. The header says the file exists since 2012, which is plain wrong. If you always use the same year then you wouldn't need individual copyrights in each file, but just can have a central COPYRIGHT file and save the work to maintain a per-file copyright. Anway, even though I don't agree with your reasoning, I certainly don't want to open a big discussion about this here. Thanks for explaining.

twz123: I'm fine with everything here. Just tell me what, in your opinion, should be the copyright header, or if I can remove that completely? I must admit that I just copy/pasted the copyright notice from another file, trying to adhere to the code standards of this project, and not paying much attention to the header's content.

parrt:
@twz123 we need a sig in the contributors.txt file :)

parrt:
@ericvergnaud What is best way to "broadcast" to target authors that a new bit of functionality has been added to one target?

ericvergnaud:
Have them subscribe to all changes (which is what I do)

mike-lischke:
@parrt Isn't the ANTLR mailinglist supposed to serve as a central communication means between ANTLR developers? Even though there are many user questions the main purpose should be the discussions/exchange about the development of ANTLR, IMO. So, you could send out a notification mail about the addition and have target authors reply with their time estimates for implementation (or, if there is no reply for particular target you know nobody is taking care) and also report the completion.

Alternatively, you could add a comment with all the known target authors in Github when you merge a new feature, so they will all get a mail about this. Better than to subscribe to all changes.

mike-lischke:
@parrt I believe it would be very helpful if each change/addition of runtime code must also name an existing testcase (or a new one, if there is none yet) in the PR, which tests the changed/added code. Target authors then can use the same test for their implementations.

parrt:
@ericvergnaud @mike-lischke how about I create a team with target folks and then ref the team in comments that they should notice?

parrt:
@twz123 we need a sig in the contributors.txt file :)

mike-lischke:
@parrt Yes, a Github team could work as well.

ericvergnaud:
let’s try that and see how it goes

twz123:
@parrt Added myself to contributors at the date of my first commit. If I should better do it using the current date, just tell me. Would you like me to rebase this thing?

runtime/Java/src/org/antlr/v4/runtime/tree/IterativeParseTreeWalker.java
twz123: Shall I wipe out the header from this file, then?

parrt:
Note to @antlr/antlr-targets that a new tree walker is enter the runtime, if you would like to implement it.

PR #1247, opened 2016-08-09
Fix the type hint of SemanticContext.filterPrecedencePredicates parameter.

renatahodovan:
SemanticContext.filterPrecedencePredicates() takes a set as parameter
in Python3 but the hint tells list instead. The patch fixes this.


parrt:
Manually fixed https://github.com/antlr/antlr4/commit/c0b5a40bcf0927c472dc42bd90360b8c2ec51085 due to conflicts. thanks @renatahodovan for all your fixes!

PR #1248, opened 2016-08-09
Fix comparisons in ATNConfigSet.__contains__ of Python targets.

renatahodovan:
The **contains** method of ATNConfigSet used different hashing
function (hash) for indexing the config dictionary than the
getOrAdd method (hashCodeForConfigSet) which filled it.
Furthermore, they also used different methods for comparing
ATNConfig objects. The patch makes them consistent.


parrt:

ericvergnaud:
blessed

PR #1251, opened 2016-08-09
Fix the initialization of conflictingAlts in execATN in Python targets.

renatahodovan:
The Java target initializes the conflictingAlt local variable
based on the conflictingAlts property of the target state.
However, the Python targets reset it to None. The patch makes the
initializations consistent.


parrt:
Looks correct to me. @ericvergnaud any issue?

PR #1259, opened 2016-08-11
Repalce string exceptions in Python targets.

renatahodovan:
Raising string exception in Python is not supported.
PredictionContext.getReturnState attempted this what is
fixed by the patch.


parrt:
Per http://stackoverflow.com/questions/2052390/manually-raising-throwing-an-exception-in-python perhaps we want more specific error. I'll use `raise IllegalStateException` and do manually. thanks!

parrt:
Fixed in https://github.com/antlr/antlr4/commit/144baa9adece96039b65b8069883f7382232925a

PR #1263, opened 2016-08-13
Replace JDialog with JFrame

vonloxley:
I was resizing the dialog about three dozen times the day before, so I thought it could use a maximize button.

### DIFF
--- a/tool/src/org/antlr/v4/gui/TreeViewer.java
+++ b/tool/src/org/antlr/v4/gui/TreeViewer.java
@@ -285,8 +285,8 @@ public class TreeViewer extends JComponent {
     private static final String DIALOG_DIVIDER_LOC_PREFS_KEY    = "dialog_divider_location";
     private static final String DIALOG_VIEWER_SCALE_PREFS_KEY   = "dialog_viewer_scale";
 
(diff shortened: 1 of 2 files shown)


danielsun1106:
+1

parrt:
Cool. can you sign contributors.txt?

vonloxley:
Hi Terence,

I understand your need to defend against possible litigation, the proposed contract however is formulated in a way that leaves some open questions:

1. Point two: I can only claim that third party rights (a formulation I feel is way too broad) aren’t violated at the time of contribution, not at the time of incorporation into the product or for the future. Furthermore I am able to certify that my contribution is free of third party rights in my country. I can`t possible do so for other countries where, for example, patent laws may be quite different.
2. Point three: I can only accept liability up to the amount of money I received for my work.

Therefore, I feel sorry for not being able to sign this contract.

Regards,

Niki

parrt:
@vonloxley I fully understand. thanks anyway. I'll do myself.

parrt:
Fixed with https://github.com/antlr/antlr4/commit/d580f2184f7ca7b8fd106a3fbc5a7bb12be97260

PR #1229, opened 2016-07-09
.NET Core Support

lecode-official:
I added a new solution for the .NET Runtime, which supports .NET Core. Also made some code changes, to fix issues with missing APIs.

PR #1271, opened 2016-08-21
Ported to .NET Standard 1.5.

lextm:
The steps I did for this port,

1. Create a PCL project.
1. Switch to .NET Standard profile 1.5.
1. Add files in vs2013 project to this new project.
1. Add a new defined constant NETCORE.


lextm:
I used a different approach to port it to .NET Core against .NET Standard 1.5, without almost no code change compared to #1229. The limitation right now is I don't know how to set up multi-targeting to .NET 3.5/4.0/4.5/4.6.

Once Microsoft finishes .NET Core on MSBuild (full csproj support), then there might be even better ways to migrate.

lextm:
Close this. .NET Core tooling is not yet ready.

Issue #1115, opened 2016-02-16
TokenStreamRewriter for Python


RedTailedHawk:
Has any work been done migrating TokenStreamRewriter to Python?  If not, I may have some bandwidth to port it over.  Is there a general estimate you could give as to the level of effort (easy, medium, hard)?

Thanks.
-Larry

ericvergnaud:
Hi,

no work has been done. Not sure about an estimate, but certainly less than a day worth of work.

RedTailedHawk:
OK thanks.  If I get a solid window of time, I'll give it a shot.

ericvergnaud:
:-)

lygav:
@ericvergnaud 
Hello, so i've implemented the TokenStreamRewriter to Python2 run-time.
Please look into PR #1273 
I would very appreciate feedback especially regarding how to integrate with auto generated test suite.
If all goes well i'll expand on this with implementing additional parts for Python2 and Python3 runtimes.

Thanks

ericvergnaud:
@lygav thanks for this, it would indeed be great if you could add some tests, which is now much simpler, see https://github.com/antlr/antlr4/blob/master/doc/adding-tests.md

Issue #550, opened 2014-04-16
TokenStreamRewriter doesn't preserve overrides order


mazurkin:
Output order differs when many overrides are inserted with the same index.

I use `TokenStreamRewriter` in Listener. When a `placeholder` rule is entered I put `<b>` tag into rewriter with `insertBefore()`. When the placeholder is exited I put `</b>` tag into rewriter with `insertAfter()`. Everything is fine with call sequence but `rewriter.getText()` return something like the following when placeholders are side-by-side and not separated with some other lexer token.

```
<b>[placeholder1]<b></b>[placeholder1]</b>
```

PR #1273, opened 2016-09-02
add TokenStreamRewriter to Python2 runtime

lygav:
Hello, i've added the TokenStreamRewriter to Python2 runtime.
If all goes well i hope to expand on this contribution.

Please advise about how to integrate into the auto-generated test suit if this is necessary.

P.S: Closes ISSUE #1115

### DIFF
--- a/runtime/Python2/src/antlr4/TokenStreamRewriter.py
+++ b/runtime/Python2/src/antlr4/TokenStreamRewriter.py
@@ -1,0 +1,231 @@
+from StringIO import StringIO
+from antlr4.Token import Token
+
+from antlr4.CommonTokenStream import CommonTokenStream
+
+
+class TokenStreamRewriter(object):
+    DEFAULT_PROGRAM_NAME = "default"
+    PROGRAM_INIT_SIZE = 100
+    MIN_TOKEN_INDEX = 0
+
+    def __init__(self, tokens):
+        """
+        :type  tokens: antlr4.BufferedTokenStream.BufferedTokenStream
+        :param tokens:
+        :return:
+        """
+        super(TokenStreamRewriter, self).__init__()
+        self.tokens = tokens
+        self.programs = {self.DEFAULT_PROGRAM_NAME: []}
+        self.lastRewriteTokenIndexes = {}
+
+    def getTokenStream(self):
+        return self.tokens
+
+    def rollback(self, instruction_index, program_name):
+        ins = self.programs.get(program_name, None)
+        if ins:
+            self.programs[program_name] = ins[self.MIN_TOKEN_INDEX: instruction_index]
+
+    def deleteProgram(self, program_name=DEFAULT_PROGRAM_NAME):
+        self.rollback(self.MIN_TOKEN_INDEX, program_name)
+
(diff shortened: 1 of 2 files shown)


parrt:
@ericvergnaud this would be a nice addition. can you bless it?

ericvergnaud:
blessed

ericvergnaud:
blessed

runtime/Python2/src/antlr4/TokenStreamRewriter.py
```diff
@@ -0,0 +1,231 @@
...
+        for i, iop in enumerate(rewrites):
+            if any((iop is None, not isinstance(iop, TokenStreamRewriter.InsertBeforeOp))):
+                continue
+            prevInserts = [op for op in rewrites[:i] if isinstance(iop, TokenStreamRewriter.InsertBeforeOp)]
+            for prevIop in prevInserts:
+                if prevIop.index == iop.index:
+                    iop.text += prevIop.text
+                    rewrites[i] = None
+            # look for replaces where iop.index is in range; error
+            prevReplaces = [op for op in rewrites[:i] if isinstance(rop, TokenStreamRewriter.ReplaceOp)]
```
ericvergnaud: should this not be is instance(op, ...

ericvergnaud:
@lygav please add some tests, following https://github.com/antlr/antlr4/blob/master/doc/adding-tests.md

lygav:
@ericvergnaud Thank you for taking time to review the PR.
I'll add the needed tests.

runtime/Python2/src/antlr4/TokenStreamRewriter.py
lygav: Seems you are right.
I'll fix that, and will add tests and will review the code again.

lygav:
@ericvergnaud Hello, although TokenStreamRewriter belongs to the 'runtime' it looks like  TokenStreamRewriter has tests only as part of the "tool" testsuite, and only for Java.
The only other implementation is in C# 'runtime' and i just cannot see where is the test for it?
How do you propose to create test for it then:
 - create python test in the Pyhon2 runtime dir (will not run with the build?)
 - try to create TokenStreamDescriptor (or similar name..)

ericvergnaud:
Hi,
create TokenStreamRewriterDescriptor is the way to go
Eric

lygav:
Thanks Eric, i am going to try to figure this out.

parrt:
@lygav I'll move my java tests to runtime.

parrt:
Actually, it looks like a major pain to make this general. So far we test ANTLR itself and not runtime lib stuff that is accessed only with code. I'm not sure it's worth the effort for me at this time to make a general mechanism. It would be easier to have runtime tests access just library stuff in the target language. Better than not having tests I guess. @ericvergnaud open to suggestions.

ericvergnaud:
Ok so let's leave as is and create independent tests in the respective runtimes.
We'll see on the long term whether there is a pattern that is worth sharing across runtimes.

lygav:
Hello @ericvergnaud and @parrt , please look into the pull request with all the tests
replicated after the Java runtime.
Plus further enhancements to make Python2 runtime to be of the same consistent feel.
Also fixed another issue from some time ago (#550)

Thank you

PR #1279, opened 2016-09-13
Minor link fixes

mrec:
"What if I need ASTs not parse trees for a compiler, for example?" wasn't hyperlinked and read oddly, and the two links in the following para were to raw .md files rather than HTML.


parrt:
Hi.  can you add yourself to the contributors list? it's overkill for just documentation, but since documentation could contain code, I'm doing a blanket request to have everyone add themselves.

mrec:
Never mind, can't resolve conflicts via web UI and not worth the faff at this point.

parrt:
i fully understand, @mrec , thanks anyway! I'll do it manually.

PR #1288, opened 2016-09-23
Javascript performance. 

ghosthope:
A new string.hash function has been added  (https://github.com/garycourt/murmurhash-js/blob/master/murmurhash3_gc.js). Performance of the function increased in two times


ericvergnaud:
Hi,
haven't check the numbers, but this looks great!
I would expect an even greater performance improvement by refactoring the Set class, which calls object.hashString before calling String.hashCode.
The idea would be to through all the hash key components using murmur3 rather than hash the corresponding string.
Eric

parrt:
@ericvergnaud so this is blessed?

parrt:
@ghosthope Can you sign contributors.txt too?

ericvergnaud:
I need to look at it again, will do that soon

ericvergnaud:
blessed, although I may move this to a reusable hash class in the future

ghosthope:
@parrt 
How can i do this?

ericvergnaud:
yes

parrt:
@ghosthope Can you sign contributors.txt too?

parrt:
@ghosthope add a commit to https://github.com/antlr/antlr4/blob/master/contributors.txt that has your "sig" on the bottom.

parrt:
Hi @ghosthope, I think your last commit was "blank"; i.e., not signature added.

PR #1290, opened 2016-09-24
TestParseTrees is ok

sridharxp:
Java Array to String works differently from Go Array to String (in fmt.Println).
So TestTokenAndRuleContextString is actually a success
Go creates [a b c] whereas java [a, b, c]


ericvergnaud:
Hi,

thanks for this, it’s great to welcome a new language target, and relay super that you have managed to leverage the test framework.
However it seems a lot of non-Go files have also been changed.
Can you clean up this PR to only include changes required for the Go target?

Eric

sridharxp:
Hi
I shall add this info to golang fork.

PR #1292, opened 2016-09-26
Add AppVeyor.yml for Windows contious integration

BurtHarris:
This adds a simple configuration file to the root directory which enables use of appveyor.com for Windows-based integration testing.   See a sample build from my fork at    https://ci.appveyor.com/project/BurtHarris/antlr4

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -6,4 +6,3 @@ test_script:
   - mvn install -q -Dantlr-python2-python="C:\Python27\python.exe" -Dantlr-python3-python="C:\Python35\python.exe" -Dantlr-javascript-nodejs="C:\Program Files (x86)\nodejs\node.exe" --batch-mode
 build:
   verbosity: minimal
-  
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -4,3 +4,6 @@ build_script:
   - mvn compile -q --batch-mode 
 test_script:
   - mvn install -q -Dantlr-python2-python="C:\Python27\python.exe" -Dantlr-python3-python="C:\Python35\python.exe" -Dantlr-javascript-nodejs="C:\Program Files (x86)\nodejs\node.exe" --batch-mode
+build:
+  verbosity: minimal
+  
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -1,4 +1,4 @@
-version: '4.5.4-SNAPSHOT+Appveyor.{build}'
+version: '4.5.4-SNAPSHOT+AppVeyor.{build}'
 os: Windows Server 2012
 build_script:
   - mvn compile --batch-mode 
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -1,0 +1,6 @@
+version: '4.5.4-SNAPSHOT+Appveyor.{build}'
+os: Windows Server 2012
+build_script:
+  - mvn compile --batch-mode 
+test_script:
+  - mvn install -Dantlr-python2-python="C:\Python27\python.exe" -Dantlr-python3-python="C:\Python35\python.exe" -Dantlr-javascript-nodejs="C:\Program Files (x86)\nodejs\node.exe" --batch-mode
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -1,6 +1,6 @@
 version: '4.5.4-SNAPSHOT+AppVeyor.{build}'
 os: Windows Server 2012
 build_script:
-  - mvn compile --batch-mode 
+  - mvn compile -q --batch-mode 
 test_script:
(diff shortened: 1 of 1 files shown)


BurtHarris:
@parrt, I think you asked for integration with AppVeyor.com.   

It was easier than I thought.   While I don't understand all the log spew, it seems like a step in the right direction.   

You'll need to configure an account on AppVeyor.com pointing to the official repository, but that also seems easy.

KvanTTT:
Can you try to use `verbosity: minimal` in build section option to decrease log size?
Add build status badge to README.

BurtHarris:
@KvanTTT, I'm not sure where you are suggesting I would add the `verbosity: minimal`, but it sounds like a very good idea.   It appears adding a `-q` to the mvn command line can reduce the spew.

I think we want the badge to reflect the builds on the official fork rather than mine, so I don't think I can propose an edit for this until that's set-up.

KvanTTT:
```YML
build:
  verbosity: quiet|minimal|normal|detailed
```
See also an official [appveyor.yml reference](https://www.appveyor.com/docs/appveyor-yml/).

BurtHarris:
OK, thanks, I'll try that.

appveyor.yml
```diff
@@ -0,0 +1,9 @@
+version: '4.5.4-SNAPSHOT+AppVeyor.{build}'
```
sharwell: :bulb: This will quickly become incorrect. You can improve on this by moving the version handling to the build script [like this](https://github.com/Microsoft/CodeContracts/blob/master/appveyor.yml#L5). To determine the specific build number for the call to `Update-AppveyorBuild`, you can extract it from the **pom.xml** file.

appveyor.yml
```diff
@@ -0,0 +1,9 @@
...
+os: Windows Server 2012
+build_script:
+  - mvn compile -q --batch-mode
```
sharwell: :question: Does AppVeyor come with Maven installed? If not, you can install it using Chocolatey [like this](https://github.com/sharwell/antlr4cs/blob/master/build/appveyor-install.ps1#L11) followed by setting `M2_HOME` [like this](https://github.com/sharwell/antlr4cs/blob/master/appveyor.yml#L7).

appveyor.yml
```diff
@@ -0,0 +1,9 @@
+version: '4.5.4-SNAPSHOT+AppVeyor.{build}'
+os: Windows Server 2012
+build_script:
+  - mvn compile -q --batch-mode 
+test_script:
+  - mvn install -q -Dantlr-python2-python="C:\Python27\python.exe" -Dantlr-python3-python="C:\Python35\python.exe" -Dantlr-javascript-nodejs="C:\Program Files (x86)\nodejs\node.exe" --batch-mode
+build:
+  verbosity: minimal
+
```
sharwell: :bulb: I would prefer to see the file end with a trailing newline character

appveyor.yml
```diff
@@ -0,0 +1,9 @@
+version: '4.5.4-SNAPSHOT+AppVeyor.{build}'
+os: Windows Server 2012
+build_script:
+  - mvn compile -q --batch-mode 
+test_script:
+  - mvn install -q -Dantlr-python2-python="C:\Python27\python.exe" -Dantlr-python3-python="C:\Python35\python.exe" -Dantlr-javascript-nodejs="C:\Program Files (x86)\nodejs\node.exe" --batch-mode
```
sharwell: 💭 I have no experience with the dependencies listed here (w.r.t. AppVeyor). Do you know if they come pre-installed on AppVeyor?

KvanTTT: All AppVeyor preinstalled tools listed here: [Build Worker installed software](https://www.appveyor.com/docs/installed-software/).

appveyor.yml
KvanTTT: Yes, Maven is [available in AppVeyor](https://www.appveyor.com/docs/installed-software/#java).

BurtHarris:
Thank you @sharwell  & @KvanTTT I've pushed a change to eliminate the trailing whitespace on the last line. I found Maven was already installed on the default appveyor worker image part way through working on this.  How convinent. 👍

parrt:
hiya. so does this correctly build antlr on a windows machine? that would be nice.

BurtHarris:
Yes, it seems to.   The tests say they are passing.

parrt:
@BurtHarris How would i know if this stuff works? Do i need to become member of AppVeyor etc...?  How can I test?

KvanTTT:
@parrt, yes, you should sign up to AppVeyor, link your account with GitHub account, select ANTLR repository and setup build config manually via Web UI or add `appveyor.yml` file to the repository. After it you will be able to build ANTLR project and run unit-tests as you can do it with Travis-CI.

parrt:
@KvanTTT ok, created account. Will investigate

KvanTTT:
I also suggest to add AppVeyor build status on README page.

parrt:
added badge.

PR #1294, opened 2016-09-27
Added @Nullable and @NotNull annotations in generated Java code.

johnw42:


ericvergnaud:
Hi,
thanks for this but we actually removed these annotations a few releases ago because we believed that they add more noise than value.
I'll let @parrt eventually change his mind.
Eric

sharwell:
:memo: My fork (under Maven group ID com.tunnelvisionlabs) of this project still emits these annotations if you are interested in using them.

parrt:
Yep, I could see an argument for annotations in generated code but I removed them from the code I have to look at. ;)

tool/resources/org/antlr/v4/tool/templates/codegen/Java/Java.stg
```diff
@@ -210,14 +215,25 @@ public class <file.grammarName>BaseVisitor\<T> extends AbstractParseTreeVisitor\
...
+@Override public T visit<lname; format="cap">(<notNull(file)><file.parserName>.<lname; format="cap">Context ctx) { return visitChildren(ctx); \}}; separator="\n">
 }
 >>
 
 fileHeader(grammarFileName, ANTLRVersion) ::= <<
 // Generated from <grammarFileName; format="java-escape"> by ANTLR <ANTLRVersion>
 >>
 
+importNullAnnotations(elem) ::= <<
+<if(elem.nullAnnotationsPackage)>
```
parrt: seems like we should just always generate a ref to the appropriate package rather than an option.

parrt:
Hmm...this has way too many changes just to get some notnull stuff in the output. I'm going to close. I could see a templates only change.

PR #1301, opened 2016-10-07
Update python-target.md

cagdasgerede:
Added missing import statement for sys


parrt:

Issue #1302, opened 2016-10-12
Allow retrieving the underlying original input.


nielsbasjes:
The ParseTree.getText() method returns the concatenation of all the tokens that are used in a specific ParseTree. In there the skipped characters have been removed in the lexer.

In some specific cases there is a need to get the underlying full (unmodified) source text including all the characters that were skipped (i.e. spaces, comments and such).

I propose a new method getSourceText to provide this.

nielsbasjes:
To give you some insight [my use case:](https://github.com/nielsbasjes/yauaa) I need this is a situation where I use Antlr4 to do pattern detection in text and when a pattern is found I need to extract the original that has been matched.

parrt:
@nielsbasjes Can't we solve this by just using channel not skip in lexer? Any runtime change could mean ripple effect across all targets.

nielsbasjes:
My usecase is a situation where I want to find patterns and afterwards extract the unmodified matched string.
The reason I did it this way is that I initially had the spaces explicitly in my lexer and parser yet this made it very hard to read; a lot of SPACE+ in the parser . Using this construct it became a lot better to read and maintain.
I figured others may have a similar situation so that is why I submitted this change proposal.

parrt:
If you send the whitespace to the parser on a hidden channel, you don't have to have matching of whitespace inside the parser but you still get the full input stream.

nielsbasjes:
Today I did some digging through the code and noticed that the javadoc of the getText method says:

	/** Return the combined text of all child nodes. This method only considers
	 *  tokens which have been added to the parse tree.
	 *  <p>
	 *  Since tokens on hidden channels (e.g. whitespace or comments) are not
	 *  added to the parse trees, they will not appear in the output of this
	 *  method.
	 */

So I apparently do not understand your last comment. 
How does putting the spaces on a hidden channel (instead of doing a skip) relate to getting the full input stream? 
Even with the skip I get everything from the input stream so from what I understand right now it makes no difference (hence I put up this change proposal).

parrt:
You ask the stream not the tree for the text.

zspitz:
@parrt I understand that the original source shouldn't be part of the parse tree, as the parse tree works off of the token stream, which already is filtered by channel. But could the tree hold and expose some reference to the parser instance? Then I could write the following:

    var parser = tree.Parser;
    var originalSource = parser.TokenStream.TokenSource.InputStream.ToString();

[...]

PR #1303, opened 2016-10-12
Create ParseTree.getSourceText to retrieve the original underlying input

nielsbasjes:
This implements #1302


parrt:
See comment here https://github.com/antlr/antlr4/issues/1302  Not sure I'm ready to alter all runtimes for this case. thanks anyway!

PR #1304, opened 2016-10-12
Add .editorconfig

nielsbasjes:
An EditorConfig file allows easy config of the IDE editor when loading a new project (like Antlr4).
See http://editorconfig.org/ for more info.

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -1,9 +1,7 @@
 root = true
 
-[*]
+[*.{java,stg}]
 charset = utf-8
-end_of_line = lf
 insert_final_newline = true
 trim_trailing_whitespace = true
 indent_style = tab
-
--- a/.editorconfig
+++ b/.editorconfig
@@ -1,0 +1,9 @@
+root = true
+
+[*]
+charset = utf-8
+end_of_line = lf
+insert_final_newline = true
+trim_trailing_whitespace = true
+indent_style = tab
+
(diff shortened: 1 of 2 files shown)


.editorconfig
```diff
@@ -0,0 +1,9 @@
+root = true
+
+[*]
```
sharwell: ❗️ This matches some files in source control, but not others. I would prefer to limit this to the ones we know, which would probably start with this:

```
[*.{java,stg}]
```

.editorconfig
```diff
@@ -0,0 +1,9 @@
...
+charset = utf-8
+end_of_line = lf
```
sharwell: ❗️ This (`end_of_line`) is not correct for all systems. Line endings for a project in Git should be controlled by **.gitattributes**, so this line needs to be omitted from **.editorconfig** altogether.

.editorconfig
```diff
@@ -0,0 +1,9 @@
...
+insert_final_newline = true
+trim_trailing_whitespace = true
```
sharwell: :bulb: This (`trim_trailing_whitespace`) is likely to cause churn in the code base over time. I would leave it out but it could go either way.

.editorconfig
nielsbasjes: Ok

.editorconfig
nielsbasjes: I'll remove it but I tend to disagree. By enforcing the editor to insert the correct end-of-line the need for making git do a conversion can be dropped.

parrt:
hi @nielsbasjes  Can you add yourself to the contributors.txt file?

nielsbasjes:
@parrt Done

PR #1316, opened 2016-11-03
TestGenerator improvements including Windows NPE fix

harriman:
This series of commits fixes a couple of problems for people running TestGenerator on Windows - including a Null Pointer Exception, a difference in the code generated, and line separator issues.

This series also includes some TestGenerator usability improvements which I hope will be helpful for people working on new targets for ANTLR.


harriman:
(I inadvertently committed contributors.txt with \r\n line endings.  Actually only one line is added.  Another commit later in the series fixes the line endings. Grumble.)

runtime-testsuite/resources/org/antlr/v4/test/runtime/templates/Index.stg
```diff
@@ -13,5 +13,5 @@ TestFolders ::= [
 	"SemPredEvalLexer": [],
 	"SemPredEvalParser": [],
 	"Sets": [],
-	"Visitors": []
+	"Visitors": "JavaScript/Node"
```
sharwell: ❗️ I prefer to not add target-specific filtering to the test suite templates themselves. You can already exclude specific tests in the target-specific templates using the `IgnoredTests` map. If you want to modify the behavior so it omits the test altogether (as opposed to just adding the `@Ignore` annotation), that would be preferable to a third exclusion mechanism - all of which remain undocumented AFAIK.

runtime-testsuite/src/org/antlr/v4/testgen/TestGenerator.java
```diff
@@ -30,202 +30,315 @@
...
-
-	/** Execute from antlr4 root dir:
-	 * *
-	 * $ java TestGenerator -o output-root-dir -templates -viz
-	 *
-	 * Example:
-	 *
-	 * $ java org.antlr.v4.testgen.TestGenerator -root /Users/parrt/antlr/code/antlr4
-	 */
+	public final static String[] allTargets = {
```
sharwell: :bulb: There is a mix of hard tabs and spaces though the changes in this file. It would help to normalize them to all hard tabs.

runtime-testsuite/src/org/antlr/v4/testgen/TestGenerator.java
```diff
@@ -30,202 +30,315 @@
...
+     * @param arg       the argument to examine
+     * @param moreArgs  The rest of the arguments.  If arg is recognized as an
+     *                  option that requires a value, the value is consumed from
+     *                  this iterator; TestGenException is thrown if not enough.
+     * @return true if option was recognized and consumed; false if unrecognized
+     * @throws TestGenException
+     */
+    public boolean acceptOption(String arg, Iterator<String> moreArgs) {
+        try {
+            if ("-root".equals(arg))
```
sharwell: :bulb: Several control flow blocks in new code are missing braces. Obviously this was also the case for the original code. It would help if you could add them to the new code, otherwise I can submit a pull request in the future to address it separately. (I wasn't going to comment, but this is a long control flow block so it really stood out to me.)

runtime-testsuite/src/org/antlr/v4/testgen/TestGenerator.java
```diff
@@ -30,202 +30,315 @@
...
+                rootDir = new File(moreArgs.next());
+            else if ("-outdir".equals(arg))
+                outDir = new File(moreArgs.next());
+            else if ("-templates".equals(arg))
+                testTemplatesRoot = new File(moreArgs.next());
+            else if ("-encoding".equals(arg))
+                encoding = moreArgs.next();
+            else if ("viz".equals(arg))
```
sharwell: ❗️ Should be `-viz`

runtime-testsuite/src/org/antlr/v4/testgen/TestGenerator.java
```diff
@@ -30,202 +30,315 @@
...
+        } catch (NoSuchElementException ex) {
+            throw new TestGenException("Option requires a value: %s", arg);
+        }
+    }
+
+    /** Consume one or more args, each arg but the last ending with a comma. */
+    protected static void acceptOptionMulti(Collection<String> out, String arg, Iterator<String> moreArgs) {
+        try {
+            String v = moreArgs.next();
+            while (v.endsWith(",")) {
```
sharwell: ❗️ This does not seem like the expected command line argument handling. As a user, I would expect one of the following to work:

* Passing the same command line argument with the flag multiple times, e.g. `-exclude A -exclude B`. This appears to work based on what I see in the code review.
* Passing the command line once, but having one comma-separated value, e.g. `-exclude A,B`. If a space were allowed, I would expect the required form to be `-exclude "A, B"` (quotes used to make one argument that contains a space). Neither of these forms currently works.

Assuming the first of these forms really does work, my recommendation is simply to remove the handling for `,` and update the comment for the method.

runtime-testsuite/src/org/antlr/v4/testgen/TestGenerator.java
```diff
@@ -30,202 +30,315 @@
...
-			}
-		}
+	protected void generateCodeForFoldersInIndex(STGroup targetGroup, File targetOutputDir) {
+        Collection<String> testFolders = new ArrayList<String>(20);
+        Collection<String> excludeFolders = new TreeSet<String>();
+
+		STGroup rootIndex = new STGroupFile(new File(testTemplatesRoot, "Index.stg").getPath());
+
+        // let output file line endings be the same kind found in top Index.stg
+        lineSeparator = detectLineSeparator(rootIndex.getFileName(), lineSeparator);
```
sharwell: 💭 Now that we have **.gitattributes**, this should not matter. Git will normalize line endings anyway when the source files are committed.

runtime-testsuite/pom.xml
```diff
@@ -106,11 +109,15 @@
...
                 <phase>generate-test-sources</phase>
                 <goals>
-                  <goal>java</goal>
+                  <goal>exec</goal>
                 </goals>
                 <configuration>
-                  <mainClass>org.antlr.v4.testgen.TestGenerator</mainClass>
+                  <executable>java</executable>
                   <arguments>
+                    <argument>-Dline.separator="&#x000a;"</argument>
```
sharwell: :question: Rather than use this, why not just update `StrlenStringMap` to normalize line endings before returning the length?

runtime-testsuite/src/org/antlr/v4/testgen/TestGenerator.java
```diff
@@ -266,11 +380,20 @@ private void importLanguageTemplates(STGroup testGroup, STGroup languageGroup) {
 		}
 	}
 
+    public String render(ST template) {
```
sharwell: 📝 This method seems unnecessary since the line endings for the affected files is being managed by Git itself.

sharwell:
@harriman 
> (I inadvertently committed contributors.txt with \r\n line endings. Actually only one line is added. Another commit later in the series fixes the line endings. Grumble.)

I'm wondering if you are using an old version of Git. On my system I can't get it to commit anything other than normalized line endings no matter how I save the file. And despite what the raw diff says, commit f5dd31f02720935276fd1c1a896e1fc2c86576ad was actually a NOP from the perspective of the Git client.

image

harriman:
On github's website, when I looked at
      the changes, appallingly
      every line showed different.  Sure enough, it had \r\n in my 
      working directory... the one file I didn't bother to diff before
      uploading.  So I committed a fix.  Happy if no harm resulted.
      
      I'm on git version 2.9.0.windows.1
      
      I have all my tools configured to leave line ends untouched if
      possible.  I never use \r\n line ends, always \n, so I never have
      to worry about whether a file is "text" or "binary".  At least in
      theory.  :)
      
      On 2016-11-03 06:00, Sam Harwell wrote:
    
    
      @harriman 
      
        (I inadvertently committed contributors.txt with \r\n line
          endings. Actually only one line is added. Another commit later
          in the series fixes the line endings. Grumble.)
      
      I'm wondering if you are using an old version of Git. On my
        system I can't get it to commit anything other than normalized
        line endings no matter how I save the file. And despite what the
        raw diff says, commit f5dd31f was actually a NOP
        from the perspective of the Git client.
      
      —
        You are receiving this because you were mentioned.
        Reply to this email directly, view
          it on GitHub, or mute
          the thread.
      
        
          
          
        
        
      
      {"api_version":"1.0","publisher":{"api_key":"05dde50f1d1a384dd78767c55493e4bb","name":"GitHub"},"entity":{"external_key":"github/antlr/antlr4","title":"antlr/antlr4","subtitle":"GitHub repository","main_image_url":"https://cloud.githubusercontent.com/assets/143418/17495839/a5054eac-5d88-11e6-95fc-7290892c7bb5.png","avatar_image_url":"https://cloud.githubusercontent.com/assets/143418/15842166/7c72db34-2c0b-11e6-9aed-b52498112777.png","action":{"name":"Open in GitHub","url":"https://github.com/antlr/antlr4"}},"updates":{"snippets":[{"icon":"PERSON","message":"@sharwell in #1316: @harriman \r\n\u003e (I inadvertently committed contributors.txt with \\r\\n line endings. Actually only one line is added. Another commit later in the series fixes the line endings. Grumble.)\r\n\r\nI'm wondering if you are using an old version of Git. On my system I can't get it to commit anything other than normalized line endings no matter how I save the file. And despite what the raw diff says, commit f5dd31f02720935276fd1c1a896e1fc2c86576ad was actually a NOP from the perspective of the Git client.\r\n\r\nimage\r\n"}],"action":{"name":"View Pull Request","url":"https://github.com/antlr/antlr4/pull/1316#issuecomment-258135495"}}}

harriman:
Hi sharwell,  Thanks very much for
      looking at my PR!
      I'm going to LLVM Dev Con today, so I'll read over
      your comments this evening.  Regards, ... Kurt Harriman
      
      On 2016-11-03 05:55, Sam Harwell wrote:
    
    
      @sharwell requested changes on this pull request.
      
      In runtime-testsuite/resources/org/antlr/v4/test/runtime/templates/Index.stg:
      > @@ -13,5 +13,5 @@ TestFolders ::= [
 	"SemPredEvalLexer": [],
 	"SemPredEvalParser": [],
 	"Sets": [],
-	"Visitors": []
+	"Visitors": "JavaScript/Node"

      ❗️ I prefer to not add
        target-specific filtering to the test suite templates
        themselves. You can already exclude specific tests in the
        target-specific templates using the IgnoredTests
        map. If you want to modify the behavior so it omits the test
        altogether (as opposed to just adding the @Ignore
        annotation), that would be preferable to a third exclusion
        mechanism - all of which remain undocumented AFAIK.
      
      In runtime-testsuite/src/org/antlr/v4/testgen/TestGenerator.java:
      >  
 public class TestGenerator {
 	
-	public final static String[] targets = {"CSharp", "Java", "Python2", "Python3", "JavaScript/Node", "JavaScript/Safari", "JavaScript/Firefox", "JavaScript/Explorer", "JavaScript/Chrome"};
-
-	/** Execute from antlr4 root dir:
-	 * *
-	 * $ java TestGenerator -o output-root-dir -templates -viz
-	 *
-	 * Example:
-	 *
-	 * $ java org.antlr.v4.testgen.TestGenerator -root /Users/parrt/antlr/code/antlr4
-	 */
+	public final static String[] allTargets = {

      💡 There is a mix of hard tabs and
        spaces though the changes in this file. It would help to
        normalize them to all hard tabs.
      
      In runtime-testsuite/src/org/antlr/v4/testgen/TestGenerator.java:
      > -	private String getTargetName() {
-		return targetName;
-	}
+    /**
+     * Consume a command line option.
+     * 
+     * @param arg       the argument to examine
+     * @param moreArgs  The rest of the arguments.  If arg is recognized as an
+     *                  option that requires a value, the value is consumed from
+     *                  this iterator; TestGenException is thrown if not enough.
+     * @return true if option was recognized and consumed; false if unrecognized
+     * @throws TestGenException
+     */
+    public boolean acceptOption(String arg, Iterator<String> moreArgs) {
+        try {
+            if ("-root".equals(arg))

      💡 Several control flow blocks in
        new code are missing braces. Obviously this was also the case
        for the original code. It would help if you could add them to
        the new code, otherwise I can submit a pull request in the
        future to address it separately. (I wasn't going to comment, but
        this is a long control flow block so it really stood out to me.)
      
      In runtime-testsuite/src/org/antlr/v4/testgen/TestGenerator.java:
      > +     *                  option that requires a value, the value is consumed from
+     *                  this iterator; TestGenException is thrown if not enough.
+     * @return true if option was recognized and consumed; false if unrecognized
+     * @throws TestGenException
+     */
+    public boolean acceptOption(String arg, Iterator<String> moreArgs) {
+        try {
+            if ("-root".equals(arg))
+                rootDir = new File(moreArgs.next());
+            else if ("-outdir".equals(arg))
+                outDir = new File(moreArgs.next());
+            else if ("-templates".equals(arg))
+                testTemplatesRoot = new File(moreArgs.next());
+            else if ("-encoding".equals(arg))
+                encoding = moreArgs.next();
+            else if ("viz".equals(arg))

      ❗️ Should be -viz
      
      In runtime-testsuite/src/org/antlr/v4/testgen/TestGenerator.java:
      > +                acceptOptionMulti(targets, arg, moreArgs);
+            else if ("-exclude".equals(arg))
+                acceptOptionMulti(excludes, arg, moreArgs);
+            else
+                return false;   // option not recognized
+            return true;        // ok, consumed
+        } catch (NoSuchElementException ex) {
+            throw new TestGenException("Option requires a value: %s", arg);
+        }
+    }
+
+    /** Consume one or more args, each arg but the last ending with a comma. */
+    protected static void acceptOptionMulti(Collection<String> out, String arg, Iterator<String> moreArgs) {
+        try {
+            String v = moreArgs.next();
+            while (v.endsWith(",")) {

      ❗️ This does not seem like the
        expected command line argument handling. As a user, I would
        expect one of the following to work:
      
        Passing the same command line argument with the flag
          multiple times, e.g. -exclude A -exclude B. This
          appears to work based on what I see in the code review.
        Passing the command line once, but having one
          comma-separated value, e.g. -exclude A,B. If a
          space were allowed, I would expect the required form to be -exclude
            "A, B" (quotes used to make one argument that
          contains a space). Neither of these forms currently works.
      
      Assuming the first of these forms really does work, my
        recommendation is simply to remove the handling for ,
        and update the comment for the method.
      
      In runtime-testsuite/src/org/antlr/v4/testgen/TestGenerator.java:
      > -				Map<String, Object> templateNames = testIndex.rawGetDictionary("TestTemplates");
-				if ( templateNames != null && !templateNames.isEmpty() ) {
-					final ArrayList<String> sortedTemplateNames = new ArrayList<String>(templateNames.keySet());
-					Collections.sort(sortedTemplateNames);
-					generateTestFile(testIndex, targetGroup, testDir, sortedTemplateNames, targetFolder);
-				}
-			}
-		}
+	protected void generateCodeForFoldersInIndex(STGroup targetGroup, File targetOutputDir) {
+        Collection<String> testFolders = new ArrayList<String>(20);
+        Collection<String> excludeFolders = new TreeSet<String>();
+
+		STGroup rootIndex = new STGroupFile(new File(testTemplatesRoot, "Index.stg").getPath());
+
+        // let output file line endings be the same kind found in top Index.stg
+        lineSeparator = detectLineSeparator(rootIndex.getFileName(), lineSeparator);

      💭 Now that we have .gitattributes,
        this should not matter. Git will normalize line endings anyway
        when the source files are committed.
      
      In runtime-testsuite/pom.xml:
      >                    <arguments>
+                    <argument>-Dline.separator="&#x000a;"</argument>

      ❓ Rather than use this, why not
        just update StrlenStringMap to normalize line
        endings before returning the length?
      
      In runtime-testsuite/src/org/antlr/v4/testgen/TestGenerator.java:
      > @@ -266,11 +380,20 @@ private void importLanguageTemplates(STGroup testGroup, STGroup languageGroup) {
 		}
 	}
 
+    public String render(ST template) {

      📝 This method seems unnecessary
        since the line endings for the affected files is being managed
        by Git itself.
      —
        You are receiving this because you authored the thread.
        Reply to this email directly, view
          it on GitHub, or mute
          the thread.
      
        
          
          
        
        
      
      {"api_version":"1.0","publisher":{"api_key":"05dde50f1d1a384dd78767c55493e4bb","name":"GitHub"},"entity":{"external_key":"github/antlr/antlr4","title":"antlr/antlr4","subtitle":"GitHub repository","main_image_url":"https://cloud.githubusercontent.com/assets/143418/17495839/a5054eac-5d88-11e6-95fc-7290892c7bb5.png","avatar_image_url":"https://cloud.githubusercontent.com/assets/143418/15842166/7c72db34-2c0b-11e6-9aed-b52498112777.png","action":{"name":"Open in GitHub","url":"https://github.com/antlr/antlr4"}},"updates":{"snippets":[{"icon":"PERSON","message":"@sharwell requested changes on #1316"}],"action":{"name":"View Pull Request","url":"https://github.com/antlr/antlr4/pull/1316#pullrequestreview-6993472"}}}

harriman:
tparr's improvements to TestGenerator in ccad5bd9bcb8bd3df866d0d89ea4040b23beca77 are comprehensive. So I'm closing this. I'll keep sharwell's comments in mind for future reference. Thanks!

PR #1348, opened 2016-11-08
Fixes for the C++ target tests on Travis

mike-lischke:


parrt:
Something's not right here.  it does not have a number of changes. For example, consider doc/go-target.md. It also seems to be a massive number of changes. Perhaps you need to do a rebase on master and then cut another PR? ouch.  you are definitely behind.  take a look at LexerFile.java

PR #1353, opened 2016-11-13
Grammar dependency management for Maven plugin

### LINKED ISSUE #1233: maven plugin dependency tracking ignores imported files
The implementation from #1094 doesn't consider the modification times of imported grammar files to determine if the grammar needs to be rebuilt.

I think properly tracking this would require that the antlr tool produce a dependency file that can be consumed by later runs. That's something that might take some time to get implemented, in the meantime it'd be good to have a configuration to the plugin to toggle the "smart" regeneration behavior on/off. It'd be my preference that it defaults to off since that will produce correct results in the default configuration but allow individuals to turn it on if they need faster builds and know it wouldn't cause them any problems.

marcohu:
Adds grammar dependency management to the ANTLR maven plugin to have grammars processed when a referenced grammar or token file changes

Fixes #1233


tool/src/org/antlr/v4/misc/Graph.java
```diff
@@ -44,13 +45,13 @@
...
-		T payload;
-		List<Node<T>> edges; // points at which nodes?
+		public T payload;
+		public List<Node<T>> edges = Collections.emptyList(); // points at which nodes?
 
 		public Node(T payload) { this.payload = payload; }
 
 		public void addEdge(Node<T> n) {
-			if ( edges==null ) edges = new ArrayList<Node<T>>();
+			if ( edges==Collections.EMPTY_LIST ) edges = new ArrayList<Node<T>>();
```
marcohu: Thanks for your remark. There is indeed no official contract. emptyList() of Oracle Java and OpenJDK  return EMPTY_LIST and I think other implementations do the same. I use this idiom in a few places and never had any fallout. But I can change it to play safe.

parrt:
@marcohu I'll need you to sign the contributors.txt file :)

marcohu:
@parrt Done!

antlr4-maven-plugin/src/main/java/org/antlr/mojo/antlr4/Antlr4Mojo.java
```diff
@@ -234,6 +235,8 @@ public void execute() throws MojoExecutionException, MojoFailureException {
 
         Log log = getLog();
 
+		outputEncoding = validateEncoding(outputEncoding);
```
marcohu: Fixes a possible bug when no project encoding is provided, introduced with https://github.com/antlr/antlr4/commit/34f0c66a89abfa6d843c34d2f5ed9478a0ed7ef5.

I did not bother to submit a new PR for this.

parrt:
@marcohu I'm going to have to just try you on this plugin thing as I don't really grok mvn. :) thanks!

marcohu:
I'm not a Maven fan either. When it works, it's usually pretty convenient. The ANTLR Maven plugin not so much without automatic change detection.

I will remove my output encoding fix and push for a hopefully last revision later today.

parrt:
I'm in a hurry so I'll just merge @marcohu :)

marcohu:
Ok. Hope everything works as expected. The extra check certainly doesn't hurt and things can be cleaned up later.

PR #1371, opened 2016-11-18
CPP runtime :  Added a specialized ctor for Any for nullptr.

FloorGoddijn:
* There was an inconvenience that an Any created with a nullptr returned false for isNull(), resolved by adding the ctor
* removed warning [forcing value to bool 'true' or 'false'] when calling is<>


parrt:
@mike-lischke can you evaluate?

mike-lischke:
Yes, patch looks good.

PR #1373, opened 2016-11-19
Fix go get command in the Go target instructions

ereyes01:
Great work, I was able to make the example work, and play around with it a little further. However, I had to modify your `go get` command to what I edited in this PR. Thanks for adding Go support!


parrt:

PR #1377, opened 2016-11-20
Swift target new test

hanjoes:


parrt:
pulling in manually from tweaks made in my fork instead.

Issue #1327, opened 2016-11-06
Swift target does not build on JDK 6


ewanmellor:
The Swift target at https://github.com/ewanmellor/antlr4/tree/swift-target does not currently build on JDK 6.  It is using java.lang.ProcessBuilder.Redirect, which was added in JDK 7.

I don't know if this is a problem for merging into mainline.  @parrt ?

parrt:
Hmm...If we can leave it as compatible with jdk 6 that would be nice. I don't think we have made an across-the-board requirement for jdk 7 yet.

ewanmellor:
@parrt Can I persuade you otherwise? ;-)

Java 6 hasn't had a publicly available update since April 2013 (according to [Wikipedia](https://en.wikipedia.org/wiki/Java_version_history#Java_SE_6)) but Oracle have made dozens of security fixes under their support program.  That means that the public version is known to be insecure.  I don't think that we (or anyone) should be supporting it unless you have a user with a compelling need.

The use of ProcessBuilder.Redirect is in the Swift BaseTest, so it's not critical and I'm sure that we could rewrite it, but just as a matter of policy I don't think it's a good idea to cling onto JDK 6.

(I'm sure Travis would appreciate us dropping a platform too.)

parrt:
Hmm...ok, anybody object to requiring Java 7 now? @sharwell @ericvergnaud ?

ericvergnaud:
I support dropping Java 6.

Envoyé de mon iPhone

parrt:
Added https://github.com/antlr/antlr4/issues/1352

marcohu:
Does this involve only the build environment requirement or will the runtime also depend on Java 7?

parrt:
just the build

marcohu:
Great! I would seriously consider going straight to Java 8 then.

Do you plan to use something like the Animal Sniffer plugin to ensure JDK 6 compatibility for the runtime?

sharwell:
@marcohu We already build the release using a newer JDK, and use the bootclasspath build option to ensure JDK 6 compatibility.

It seems like the use of ProcessBuilder.Redirect is likely not required. Can you clarify where it is used? We can replace it with the mechanism already used in the Java target tests to redirect/consume output from a separate process.

marcohu:
@sharwell I'm actually using your version, bootclasspath checking is not implemented universally. Don't know whether this is intentional. Compiling the maven plugin yields a warning for me:

[...]

PR #1383, opened 2016-11-20
Enforce JDK runtime requirements

marcohu:
Use 1.6 version for the compiler and employ animal sniffer to enforce Java 6 compatibility.

### DIFF
--- a/runtime/Java/pom.xml
+++ b/runtime/Java/pom.xml
@@ -10,6 +10,51 @@
   <name>ANTLR 4 Runtime</name>
   <description>The ANTLR 4 Runtime</description>
 
+  <profiles>
+    <profile>
+      <id>jdk6</id>
+      <activation>
+        <activeByDefault>true</activeByDefault>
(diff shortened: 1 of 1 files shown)


parrt:
We actually just decided on jdk 1.7 min for 4.6 ;)

marcohu:
I've asked about the runtime requirement some days ago and you mentioned that Java 7 is only required for the build system, but the runtime will stay on 6 a bit longer. See here: https://github.com/antlr/antlr4/issues/1327#issuecomment-260126256.

Has that view changed and Java 7 is the new baseline for the runtime as well?

parrt:
Hmm...yeah, I think i'll just make it all v7. Is there a reason you need 6?

parrt:
Hmm..I'd rather not add yet more to pom files. I've been holding line at jdk 6 for a long time. I think i'll go with 7 for 4.6 release.  You can always compile locally for your customers.

PR #1399, opened 2016-11-23
Implement left-recursive rule optimization

### LINKED ISSUE #994: Another poor performance about expression parsing
I'm using ANTLR-4.4 with a grammar file modified from [sqlite.g4](https://github.com/antlr/grammars-v4/blob/master/sqlite/SQLite.g4),and find  some complex expressions cost too much time to parse.And then I found  [issue#192](https://github.com/antlr/antlr4/issues/192) .

That grammar file works fast with the antlr version I'm using now.But when I add some rules I used in my grammar file,it get slow down.

To make the issue simple,the grammar I use is just like this:

```
grammar expr;

expr: ID
    | 'not' expr
    | expr 'and' expr
    | expr 'or' expr
    | expr 'between' expr 'and' expr
    ;

ID: [a-zA-Z_][a-zA-Z_0-9]*;
WS: [ \t\n\r\f]+ -> skip;
ERROR: .;
```

and the sample file is just same like

```
not X1 and not X2 and not X3 and not X4 and not X5 and not X6 and not X7 and not X8 and not X9 and not X10 and not X11 and not X12 or
    X1 and not X2 and not X3 and not X4 and not X5 and not X6 and not X7 and not X8 and not X9 and not X10 and not X11 and not X12 or
not X1 and     X2 and not X3 and not X4 and not X5 and not X6 and not X7 and not X8 and not X9 and not X10 and not X11 and not X12
```

I tried this with the antlr plugin in IDEA,just 3 lines of the expressions will make the parsing time unacceptable.

### LINKED ISSUE #1398: Implement dynamic disambiguation of left-recursive rules during closure
## Background

Since the early days of ANTLR 4, we've known that long lookahead inside of left-recursive (LR) rules poses "interesting challenges" for performance. The earliest approaches to this problem were attempts to inline the evaluation of special "precedence predicates" during prediction, rather than treat them as standard predicates that are only evaluated after an SLL conflict is reached. Unfortunately, after attempting to implement this I realized a major flaw in the design - LR rules are fundamentally *ambiguous*, and rely on ANTLR's "first alternative wins" decision making strategy to choose the expected alternative.

Eventually we realized that the solution to much of the performance problem associated with LR rules could be solved by *disambiguating* the ATN prior to making a decision. This is accomplish by eliminating known ambiguous configurations before running the `adaptivePredict` algorithm.

> :warning: One extremely valuable mechanism in ANTLR 4 is the ability to detect and report ambiguous parse trees. It's critically important that the disambiguation algorithm only eliminate alternatives which are ambiguous as a direct result of writing a LR rule in the grammar.

The initial implementation appeared in #401, with notable bugs identified in #509 and #679 (now fixed).

## Remaining problem

While the above work dramatically improved performance in real-world scenarios where LR rules were used, it failed to meet the performance expectations of users working with grammars and inputs that require "looking through" a LR rule to make a prediction. As a simple example, consider the following grammar fragment for a made-up language:

```antlr
statement
  : 'if' expression
  | 'if' expression 'else' expression
  ;

expression
  : ID
  | expression ('*' | '/') expression
  | expression ('+' | '-') expression
  ;
```

In this example, the expression following an `if` keyword in the language must be fully consumed by the `adaptivePredict` algorithm only to determine whether or not it is followed by an `else` keyword. I refer to this situation as "looking through *expression* to make a decision in *statement*". The performance impact of grammars with this characteristic is tremendous - the `adaptivePredict` algorithm will explore all possible paths through *expression* during the prediction! This is clearly not what the user intended when *expression* was written as an LR rule.

For reference, here is the ATN for `statement`:

And here is the ATN for `expression`:

Several issues have been filed which appear to be manifestations of this issue in some form:

* #994 
* #1232 
* #1320 
* #1323 

## Dynamic disambiguation

Early work on generalizing the `applyPrecedenceFilter` algorithm to "look through" situations focused on ways to apply the precedence filter to decisions traversed deep inside the `adaptivePredict` algorithm. The implementation plan looked something like this:

1. Augment the `PredictionContext` class to keep track of precedence levels when a precedence rule is invoked
1. During `closure` operations, identify precedence decisions using the same algorithm we already use to create precedence DFAs
1. When a precedence decision is reached, suspend the closure operation and perform a closure on just the specific configuration which reached a precedence decision
1. Using the information provided by the prediction context and limited closure from the precedence decision, run the `applyPrecedenceFilter` algorithm
1. Add the resulting filtered set of configurations back to the result of the primary `closure` operation being performed

However, the subtleties of this approach kept me from ever getting it implemented (even as a prototype).

The real breakthrough came today when I took a much closer look at the syntactic *ambiguity* produced by LR rules. Remember that aside from computing the DFA start state, `adaptivePredict` asks only one thing: which alternative provides some viable path to consume more input symbols than any other alternative? For a true ambiguity, then for some input(s) there is more than one such alternative whether we are in SLL or LL prediction modes. More importantly, `adaptivePredict` doesn't care at all *how* that longest sequence is reached. For example, consider the following input with the grammar listed above:

```
if a * b + c else d
```

Clearly the intended final parse tree for this expression would look like the following:

```
(statement 'if' (expression (expression (expression 'a') '*' (expression 'b')) '+' (expression 'c')) 'else' (expression 'd'))
```

However, when `adaptivePredict` is attempting to choose the first or second alternative of the *statement* rule, the precedence within the *expression* is totally irrelevant. A parse looking like the following would lead to exactly the same final result as the desired expression:

```
(statement 'if' (expression (expression 'a') '*' (expression (expression 'b') '+' (expression 'c'))) 'else' (expression 'd'))
```

This property can be generalized by the following:

> During a closure operation inside `adaptivePredict`, the elimination of any configuration which is ambiguous with another configuration that predicts the same alternative will not affect the final decision made by `adaptivePredict`.

Leveraging this property we can redefine the implementation plan to eliminate the opposite set of configurations from the configurations eliminated by `applyPrecedenceFilter`. It also turns out this is a wonderful idea because it is much simpler from an implementation standpoint. The `StarLoopEntryState` that forms a precedence decision has two outgoing edges:

1. The first edge stays inside the LR rule, and matches some terminal which follows a left-recursive invocation of the LR rule.
2. The second alternative leads to the stop state of the LR rule, and steps into the enclosing context.

If any return edge from the rule stop state leads to the right edge of a left-recursive invocation of the rule, then anything reachable by following the first edge is also reachable by following the second edge. Knowing this, we can implement dynamic disambiguation by intentionally failing to
include the target of the first edge in the closure operation. With the edge eliminated, one specific class of ambiguities, specifically those created by the use of LR rules in a grammar, are reduced to an unambiguous single alternative during `adaptivePredict`. The fact that the path taken through the grammar during `adaptivePredict` doesn't match the path taken through the grammar to produce the final parse tree does not change the overall outcome.

sharwell:
Fixes #1398
Fixes #994

:memo: For #1323 I'm observing a 10,000:1 performance improvement. Yeah, *that* much! :tada: /cc @sufail

### DIFF
--- a/runtime/Java/src/org/antlr/v4/runtime/atn/ParserATNSimulator.java
+++ b/runtime/Java/src/org/antlr/v4/runtime/atn/ParserATNSimulator.java
@@ -1552,6 +1552,60 @@ public class ParserATNSimulator extends ATNSimulator {
         }
 
 		for (int i=0; i<p.getNumberOfTransitions(); i++) {
+			// This block implements first-edge elimination of ambiguous LR
+			// alternatives as part of dynamic disambiguation during prediction.
+			// See antlr/antlr4#1398.
+			if (i == 0
+				&& p.getStateType() == ATNState.STAR_LOOP_ENTRY
+				&& ((StarLoopEntryState)p).isPrecedenceDecision
+				&& !config.context.hasEmptyPath()) {
+
+				// When suppress is true, it means the outgoing edge i==0 is
+				// ambiguous with the outgoing edge i==1, and thus the closure
+				// operation can dynamically disambiguate by suppressing this
+				// edge during the closure operation.
+				boolean suppress = true;
+
+				// Require all return states to return back to the same rule
+				// that p is in.
+				int limit = config.context.size();
+				for (int j = 0; j < limit; j++) {
+					ATNState returnState = atn.states.get(config.context.getReturnState(j));
+					suppress = suppress && returnState.ruleIndex == p.ruleIndex;
+				}
+
+				// Further check to make sure this isn't a 0-precedence entry.
+				// See antlr/antlr4#679.
+				if (suppress) {
+					RuleStopState ruleStopState = atn.ruleToStopState[p.ruleIndex];
+					for (int j = 0; j < limit; j++) {
+						for (Transition transition : ruleStopState.transitions) {
+							if (transition.getSerializationType() != Transition.EPSILON) {
+								continue;
+							}
+
+							if (((EpsilonTransition)transition).outermostPrecedenceReturn() != p.ruleIndex) {
+								continue;
+							}
+
+							int returnStateNumber = config.context.getReturnState(j);
+							suppress = returnStateNumber != transition.target.stateNumber;
+							if (!suppress) {
+								break;
+							}
+						}
+
+						if (!suppress) {
+							break;
+						}
+					}
+				}
+
+				if (suppress) {
+					continue;
+				}
+			}
+
 			Transition t = p.transition(i);
 			boolean continueCollecting =
 				!(t instanceof ActionTransition) && collectPredicates;
--- a/runtime/Java/src/org/antlr/v4/runtime/atn/ParserATNSimulator.java
+++ b/runtime/Java/src/org/antlr/v4/runtime/atn/ParserATNSimulator.java
@@ -1560,44 +1560,17 @@ public class ParserATNSimulator extends ATNSimulator {
 				&& ((StarLoopEntryState)p).isPrecedenceDecision
 				&& !config.context.hasEmptyPath()) {
 
+				StarLoopEntryState precedenceDecision = (StarLoopEntryState)p;
+
 				// When suppress is true, it means the outgoing edge i==0 is
 				// ambiguous with the outgoing edge i==1, and thus the closure
 				// operation can dynamically disambiguate by suppressing this
 				// edge during the closure operation.
 				boolean suppress = true;
-
-				// Require all return states to return back to the same rule
-				// that p is in.
-				int limit = config.context.size();
-				for (int j = 0; j < limit; j++) {
-					ATNState returnState = atn.states.get(config.context.getReturnState(j));
-					suppress = suppress && returnState.ruleIndex == p.ruleIndex;
-				}
-
-				// Further check to make sure this isn't a 0-precedence entry.
-				// See antlr/antlr4#679.
-				if (suppress) {
-					RuleStopState ruleStopState = atn.ruleToStopState[p.ruleIndex];
-					for (int j = 0; j < limit; j++) {
-						for (Transition transition : ruleStopState.transitions) {
-							if (transition.getSerializationType() != Transition.EPSILON) {
-								continue;
-							}
-
-							if (((EpsilonTransition)transition).outermostPrecedenceReturn() != p.ruleIndex) {
-								continue;
-							}
-
-							int returnStateNumber = config.context.getReturnState(j);
-							suppress = returnStateNumber != transition.target.stateNumber;
-							if (!suppress) {
-								break;
-							}
-						}
-
-						if (!suppress) {
-							break;
-						}
+				for (int j = 0; j < config.context.size(); j++) {
+					if (!precedenceDecision.precedenceLoopbackStates.get(config.context.getReturnState(j))) {
+						suppress = false;
+						break;
 					}
 				}
 
--- a/runtime/Java/src/org/antlr/v4/runtime/atn/ATNDeserializer.java
+++ b/runtime/Java/src/org/antlr/v4/runtime/atn/ATNDeserializer.java
@@ -36,8 +36,11 @@ import org.antlr.v4.runtime.misc.Pair;
 
 import java.io.InvalidClassException;
 import java.util.ArrayList;
+import java.util.BitSet;
+import java.util.HashMap;
 import java.util.List;
 import java.util.Locale;
+import java.util.Map;
 import java.util.UUID;
 
 /**
@@ -527,6 +530,9 @@ public class ATNDeserializer {
 	 * @param atn The ATN.
 	 */
 	protected void markPrecedenceDecisions(ATN atn) {
+		// Map rule index -> precedence decision for that rule
+		Map<Integer, StarLoopEntryState> rulePrecedenceDecisions = new HashMap<>();
+
 		for (ATNState state : atn.states) {
 			if (!(state instanceof StarLoopEntryState)) {
 				continue;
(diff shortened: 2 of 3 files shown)


runtime/Java/src/org/antlr/v4/runtime/atn/ParserATNSimulator.java
```diff
@@ -1552,6 +1552,55 @@ protected void closure_(ATNConfig config,
...
+			// See antlr/antlr4#1398.
+			if (i == 0 && p.getStateType() == ATNState.STAR_LOOP_ENTRY && ((StarLoopEntryState)p).isPrecedenceDecision) {
+				boolean suppress = !config.context.isEmpty();
+				int limit = config.context.size();
+				if (config.context.hasEmptyPath()) {
+					limit--;
+				}
+
+				// Require all return states to return back to the same rule
+				// that p is in.
```
sharwell: :memo: We are eliminating lookahead for one outgoing edge. This will affect *all* of the ATN configurations represented by the prediction context graph, so we need to verify that all of the configurations are affected by the ambiguity, hence why we verify the target of all return states.

runtime/Java/src/org/antlr/v4/runtime/atn/ParserATNSimulator.java
```diff
@@ -1552,6 +1552,55 @@ protected void closure_(ATNConfig config,
         }
 
 		for (int i=0; i<p.getNumberOfTransitions(); i++) {
+			// This block implements first-edge elimination of ambiguous LR
+			// alternatives as part of dynamic disambiguation during prediction.
+			// See antlr/antlr4#1398.
+			if (i == 0 && p.getStateType() == ATNState.STAR_LOOP_ENTRY && ((StarLoopEntryState)p).isPrecedenceDecision) {
+				boolean suppress = !config.context.isEmpty();
+				int limit = config.context.size();
+				if (config.context.hasEmptyPath()) {
```
sharwell: ❗️ This is likely a bug. If we have an empty path we should automatically *not* suppress the edge.

sharwell: :arrow_right: Fixed

runtime/Java/src/org/antlr/v4/runtime/atn/ParserATNSimulator.java
```diff
@@ -1552,6 +1552,60 @@ protected void closure_(ATNConfig config,
...
+						}
+
+						if (!suppress) {
+							break;
+						}
+					}
+				}
+
+				if (!suppress) {
+					break;
```
sharwell: ❗️ Should be the following (copy paste error):

```java
if (suppress) {
  continue;
}
```

runtime/Java/src/org/antlr/v4/runtime/atn/ParserATNSimulator.java
```diff
@@ -1552,6 +1552,60 @@ protected void closure_(ATNConfig config,
...
+						for (Transition transition : ruleStopState.transitions) {
+							if (transition.getSerializationType() != Transition.EPSILON) {
+								continue;
+							}
+
+							if (((EpsilonTransition)transition).outermostPrecedenceReturn() != p.ruleIndex) {
+								continue;
+							}
+
+							ATNState returnState = atn.states.get(config.context.getReturnState(j));
```
sharwell: :bulb: No need to get the `ATNState` if we just need the state number.

sharwell:
@qwerppoo I ran your unit test from #994 using this branch. I had to modify it to use `System.nanoTime()` and increase the number of iterations to 2000, but it seems overall you're looking at an average ratio of ~1.03. Iteration 2000 completed in 14ms.

runtime/Java/src/org/antlr/v4/runtime/atn/ParserATNSimulator.java
```diff
@@ -1552,6 +1552,60 @@ protected void closure_(ATNConfig config,
...
+						for (Transition transition : ruleStopState.transitions) {
+							if (transition.getSerializationType() != Transition.EPSILON) {
+								continue;
+							}
+
+							if (((EpsilonTransition)transition).outermostPrecedenceReturn() != p.ruleIndex) {
+								continue;
+							}
+
+							int returnStateNumber = config.context.getReturnState(j);
```
parrt: This can move out of inner loop.

runtime/Java/src/org/antlr/v4/runtime/atn/ParserATNSimulator.java
```diff
@@ -1552,6 +1552,60 @@ protected void closure_(ATNConfig config,
...
+				int limit = config.context.size();
+				for (int j = 0; j < limit; j++) {
+					ATNState returnState = atn.states.get(config.context.getReturnState(j));
+					suppress = suppress && returnState.ruleIndex == p.ruleIndex;
+				}
+
+				// Further check to make sure this isn't a 0-precedence entry.
+				// See antlr/antlr4#679.
+				if (suppress) {
+					RuleStopState ruleStopState = atn.ruleToStopState[p.ruleIndex];
```
parrt: Can't we test something simpler here? Perhaps we verify that all returnStateNumber values are the same after checking they are in the same rule?  Why bother with FOLLOW edges?

runtime/Java/src/org/antlr/v4/runtime/atn/ParserATNSimulator.java
```diff
@@ -1552,6 +1552,60 @@ protected void closure_(ATNConfig config,
...
+					for (int j = 0; j < limit; j++) {
+						for (Transition transition : ruleStopState.transitions) {
+							if (transition.getSerializationType() != Transition.EPSILON) {
+								continue;
+							}
+
+							if (((EpsilonTransition)transition).outermostPrecedenceReturn() != p.ruleIndex) {
```
parrt: Can you help me understand why precedence DFA stuff comes into play here?

sharwell: `outermostPrecedenceReturn` labels edges in the ATN which return from an LR rule invocation where the precedence was 0. All the return edges for non-zero precedence go to the block end state that we're looking for.

parrt:
I'm closing after manually pulling most of this into my own version (that is easy for my brain to handle ha!). https://github.com/antlr/antlr4/pull/1404

PR #1423, opened 2016-11-29
Fix for redeclared error is generated Go code. 

wxio:
@parrt @pboyer 
Clean pull request for redeclared var error.

Create block so _x and _m var are created in a local scope.

### DIFF
--- a/runtime/Go/antlr/lexer.go
+++ b/runtime/Go/antlr/lexer.go
@@ -282,6 +282,10 @@ func (b *BaseLexer) setInputStream(input CharStream) {
 	b.tokenFactorySourcePair = &TokenSourceCharStreamPair{b, b.input}
 }
 
+func (b *BaseLexer) GetTokenSourceCharStreamPair() *TokenSourceCharStreamPair {
+	return b.tokenFactorySourcePair
+}
+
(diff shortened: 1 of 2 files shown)


millergarym:
@parrt Question about pull request etiquette. Should these changes have been separate pull requests?

sridharxp:
Checking...Skipped some tests  (5+7) in ` TestPerformance` and `TestVisitors` Other tests passed ok.
After inclusion of `GetTokenSourceCharStreamPair` method there are no failures which surfaced after recent changes in Go.stg

parrt:
@antlr/antlr-targets just a note that I'm out of action til this coming weekend

pboyer:
Hey folks, 

Generally, yes, I would prefer to see this as separate PR's. But, their impact should each be fairly limited and there are only two. So, we'll let you get away with it this time. :)

@millergarym We need your changes in contributors.txt to merge this.

millergarym:
@pboyer my pull request for the contributor.txt change containing my details has already been accepted by Terrance.

pboyer:
@millergarym Sorry I didn't see that! Thank you so much!

PR #1426, opened 2016-11-29
Fix a dangerous c-cast in the cpp codegen

Naios:
Following code was generated by the cpp backend for the demo project:

```cpp
antlrcpp::Any TParser::StatContext::accept(tree::ParseTreeVisitor *visitor) {
  if (dynamic_cast<TParserVisitor*>(visitor) != nullptr)
    return ((TParserVisitor *)visitor)->visitStat(this);
//           ~~~~~~~~~~~~~~~~
  else
    return visitor->visitChildren(this);
}
```

There is a C-Cast used after the class was checked through a dynamic cast.
This could lead to undefined behaviour when casting classes with virtual bases.

This PR changes the cpp template in order to cache the result returned by the dynamic cast to remove the c-cast.

### DIFF
--- a/tool/resources/org/antlr/v4/tool/templates/codegen/Cpp/Cpp.stg
+++ b/tool/resources/org/antlr/v4/tool/templates/codegen/Cpp/Cpp.stg
@@ -1048,8 +1048,8 @@ virtual antlrcpp::Any accept(tree::ParseTreeVisitor *visitor) override;
 VisitorDispatchMethod(method) ::=  <<
 
 antlrcpp::Any <parser.name>::<struct.name>::accept(tree::ParseTreeVisitor *visitor) {
-  if (dynamic_cast\<<parser.grammarName>Visitor*>(visitor) != nullptr)
-    return ((<parser.grammarName>Visitor *)visitor)->visit<struct.derivedFromName; format="cap">(this);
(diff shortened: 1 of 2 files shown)


tool/resources/org/antlr/v4/tool/templates/codegen/Cpp/Cpp.stg
```diff
@@ -1048,8 +1048,8 @@ virtual antlrcpp::Any accept(tree::ParseTreeVisitor *visitor) override;
 VisitorDispatchMethod(method) ::=  <<
 
 antlrcpp::Any <parser.name>::<struct.name>::accept(tree::ParseTreeVisitor *visitor) {
-  if (dynamic_cast\<<parser.grammarName>Visitor*>(visitor) != nullptr)
-    return ((<parser.grammarName>Visitor *)visitor)->visit<struct.derivedFromName; format="cap">(this);
+  if (auto parserVisitor = dynamic_cast\<<parser.grammarName>Visitor*>(visitor))
```
sharwell: 💭 I typically prefer to see the assignment on the line before the `if` (just so an assignment isn't in the conditional). That said:

1. This is generated code.
2. I'm not familiar enough with this target to know what it's coding conventions are. The form seen here may very well be preferred for this project.

parrt:
@mike-lischke Do you bless this PR?

PR #1432, opened 2016-11-30
Add heartbeat thread to gen minimal output for travis, appveyor to avoid timeout.

parrt:


parrt:
@antlr/antlr-targets this should prevent 10 minute timeouts on travis etc...  Also note that I've re-ignored  new performance tests for node,python as they don't seem to be finishing. I'm off for 3 days and want travis etc... to build while I'm gone.

Issue #1397, opened 2016-11-22
Go target generated code for Java.g4 grammar doesn't compile


ereyes01:
Before submitting an issue to ANTLR, please check off these boxes:
 
- [*] I am not submitting a question on how to use ANTLR; instead, go to [antlr4-discussion google group](https://groups.google.com/forum/#!forum/antlr-discussion) or ask at [stackoverflow](http://stackoverflow.com/questions/tagged/antlr4)
- [*] I have done a search of the existing issues to make sure I'm not sending in a duplicate

### Expected behavior

`antlr4 -Dlanguage=Go Java.g4` should generate Go code that compiles. Java.g4 was taken from github.com/antlr/grammars-v4

Other info:

antlr4 command is: `alias antlr4='java -Xmx500M -cp "/home/ereyes/.m2/repository/org/antlr/antlr4/4.6-SNAPSHOT/antlr4-4.6-SNAPSHOT.jar:$CLASSPATH" org.antlr.v4.Tool'`

```
$ go version
go version go1.7.1 linux/amd64
```

### Actual behavior

The generated code does not compile. Here's what the Go compiler says:

```
$ go install ./
# /home/ereyes/code/antlr4-java/parser
./java_lexer.go:726: syntax error: unexpected _input, expecting comma or )
./java_lexer.go:739: syntax error: unexpected _input, expecting comma or )
```

### Steps to reproduce the behavior

```
$ cd ~/code
$ git clone https://github.com/antlr/grammars-v4
$ mkdir -p antlr4-java/src/parser
$ export GOPATH=$HOME/code/antlr4-java
$ cd antlr4-java/src/parser
$ cp ~/code/grammars-v4/java/Java.g4 ./
$ antlr4 -Dlanguage=Go Java.g4
$ go get github.com/antlr/antlr4/runtime/Go/antlr
$ go install ./
```

I looked at the error and attempted to fix by applying the patch below, but it still doesn't fix the problem:

```
diff --git a/java_lexer.go b/java_lexer.go
index b7b70c6..4b5114f 100644
--- a/java_lexer.go
+++ b/java_lexer.go
@@ -723,7 +723,7 @@ func (p *JavaLexer) JavaLetter_Sempred(localctx antlr.RuleContext, predIndex int
 			return Character.isJavaIdentifierStart(_input.LA(-1))
 
 	case 1:
[...]

PR #1436, opened 2016-12-01
Fix for #1397

pboyer:
### Problem 

The issue here was that I used a := short variable declaration in the template, when there could be multiple assignments of this variable. Hence, duplicate assignment compiler errors were encountered. 

### Solution 

Here, I follow the model set by `Java.stg` and use a single variable declaration. I observed no regressions in the test suite and @ereyes01 was kind enough to test it out on the `Java.g4` grammar. 

### Question

I need a verification that the `currentRule.hasLookaheadBlock` field will not provide "false positives". Specifically, we need to be sure that `_alt` is not declared and not used. Go will emit a compile error if a variable is declared and not used. If this is not guaranteed I can provide a simple "proxy" usage to silence the error.

PTAL @parrt.

This is a fix for #1397.

### DIFF
--- a/tool/resources/org/antlr/v4/tool/templates/codegen/Go/Go.stg
+++ b/tool/resources/org/antlr/v4/tool/templates/codegen/Go/Go.stg
@@ -407,7 +407,7 @@ func (p *<parser.name>) <currentRule.name; format="cap">(<currentRule.args:{a | 
 	}()
 	
 	<if(currentRule.hasLookaheadBlock)>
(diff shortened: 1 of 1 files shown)


pboyer:
Thanks @parrt! Did you see the question above?

parrt:
@pboyer I poked around and I'm confident it only gets generated when there is more than one alt and it's a plus or star loop. It should only define that when it needs it.

PR #1438, opened 2016-12-01
Create a new Xcode playground for Swift.

ewanmellor:
This takes the example from
https://github.com/janyou/Antlr-Swift-Runtime/tree/master/Test and packages
it into an Xcode playground.  This allows the user to interact with the
grammar directly inside Xcode, and serves as a live demo of the Swift support.

Update the docs to match.


sharwell:
💭 This seems like it should be a separate project (separate repository).

ewanmellor:
@sharwell Why?  It's an example of how to use the Swift runtime.  Like live documentation or a tutorial basically.  It belongs with the rest of the docs (and I'd expect people to update it like they should update the docs).

sharwell:
@ewanmellor Even in that case it seems to me like it would be a small, self-contained repository which the documentation links to. The final decision on it is up to @parrt though.

:memo: If it is decided to be kept in this repository, I would recommend *at least* removing the generated files from source control.

parrt:
Interesting to have some "live doc". Can we avoid adding generated files to the repo?  I'm unsure if "sample projects" like this belong in the repo. Imagine this multiplied by 3 IDE for each language. Perhaps the doc part goes in the repo.

ewanmellor:
If we remove the generated code, then the user would have to figure out how to make Antlr work before they could try the playground.  That's not the end of the world, but it harms the value of the tutorial in my opinion.

parrt:
It's small so i'm going to include. if there becomes lots of these for various IDEs, we can make a separate project.

PR #1447, opened 2016-12-03
Update VS2015 cpp runtime project

redxdev:
The VS2015 project for the cpp runtime references a nonexistant file (Tree.cpp) and doesn't reference a file required for compilation and linking (ParseTree.cpp). Additionally, an include directive doesn't compile as ParseTree.cpp is in a subfolder. This should fix the above issues with VS2015.


parrt:
@mike-lischke @davesisson do you do VS? Can you check this PR?

PR #1448, opened 2016-12-04
Optimize javascript runtime

ericvergnaud:
A signifiant improvement by switching from strings to murmur3 for set/map keys .
On the DropLoopEntryBranchInLRRule_4 performance test modified to check
performance with gradual complexity, the results show a 30% improvement


parrt:
I think you're still working on this one, right @ericvergnaud ?

ericvergnaud:
No that's good to go

Envoyé de mon iPhone

PR #1452, opened 2016-12-05
Augment vocabulary

parrt:


runtime/Java/src/org/antlr/v4/runtime/Vocabulary.java
```diff
@@ -148,4 +157,45 @@
...
+	 *  associated with token type i.  Token types without literals
+	 *  (such as {@code ID : [A-Z]+;}) have null entries
+	 *  in the returned array. The first possibly valid entry is
+	 *  {@link Token#MIN_USER_TOKEN_TYPE}.
+	 *
+	 * @return The non-null array of literal names assigned to tokens.
+	 *
+	 * @since 4.6
+	 */
+	String[] getLiteralNames();
```
sharwell: ❓ Why are these added? I'm concerned users will be confused because the returned arrays are not all the same length.

I would prefer to not expose mutable data through this interface, plus this can be a source and binary breaking change for user implementations of the `Vocabulary` interface.

runtime/Java/src/org/antlr/v4/runtime/VocabularyImpl.java
```diff
@@ -156,7 +156,7 @@ public int getMaxTokenType() {
 
 	@Override
 	public String getLiteralName(int tokenType) {
-		if (tokenType >= 0 && tokenType < literalNames.length) {
+		if (tokenType >= Token.MIN_USER_TOKEN_TYPE && tokenType < literalNames.length) {
```
sharwell: :memo: This change and the one in `getSymbolicName` has no real impact (the entries at position 0 are already `null`.

runtime/Java/src/org/antlr/v4/runtime/VocabularyImpl.java
```diff
@@ -178,7 +178,7 @@ public String getSymbolicName(int tokenType) {
 
 	@Override
 	public String getDisplayName(int tokenType) {
-		if (tokenType >= 0 && tokenType < displayNames.length) {
+		if (tokenType >= Token.MIN_USER_TOKEN_TYPE && tokenType < displayNames.length) {
```
sharwell: ❗️ This prevents a user from assigning a custom display name to `Token.INVALID_TYPE`. Adding this as a restriction doesn't seem to provide any benefit to the user.

tool-testsuite/test/org/antlr/v4/test/tool/TestVocabulary.java
```diff
@@ -32,10 +32,13 @@
 import org.antlr.v4.runtime.Token;
 import org.antlr.v4.runtime.Vocabulary;
 import org.antlr.v4.runtime.VocabularyImpl;
+import org.antlr.v4.tool.Grammar;
 import org.junit.Assert;
 import org.junit.Before;
 import org.junit.Test;
 
+import static junit.framework.TestCase.assertEquals;
```
sharwell: ❗️ `org.junit.Assert.assertEquals`

tool-testsuite/test/org/antlr/v4/test/tool/TestVocabulary.java
```diff
@@ -64,7 +67,7 @@ public void testVocabularyFromTokenNames() {
 		Vocabulary vocabulary = VocabularyImpl.fromTokenNames(tokenNames);
 		Assert.assertNotNull(vocabulary);
 		Assert.assertEquals("EOF", vocabulary.getSymbolicName(Token.EOF));
-		for (int i = 0; i < tokenNames.length; i++) {
+		for (int i = Token.MIN_USER_TOKEN_TYPE; i <= vocabulary.getMaxTokenType(); i++) {
```
sharwell: ❗️ The original test was correct. We want to test the full range of the original array. If you want, you could add an assertion that `vocabulary.getMaxTokenType()` matches `tokenNames.length - 1`.

runtime/Java/src/org/antlr/v4/runtime/VocabularyImpl.java
parrt: Token.MIN_USER_TOKEN_TYPE is more clear and is actually correct unlike previous 0

tool-testsuite/test/org/antlr/v4/test/tool/TestVocabulary.java
parrt: what's the diff?

tool-testsuite/test/org/antlr/v4/test/tool/TestVocabulary.java
parrt: thanks

runtime/Java/src/org/antlr/v4/runtime/Vocabulary.java
sharwell: :memo: I'm still 👎 on the change. `Vocabulary` works well in code for representing this information, where `String[]` leads to ambiguity. This change just makes it easier for users to move away from best practices.

tool-testsuite/test/org/antlr/v4/test/tool/TestVocabulary.java
sharwell: This import is from the old JUnit 3 framework. Everywhere else we use the one I mentioned.

runtime/Java/src/org/antlr/v4/runtime/VocabularyImpl.java
parrt: this is impl not the interface

tool-testsuite/test/org/antlr/v4/test/tool/TestVocabulary.java
parrt: wow. damn junit. well, I'm using it elsewhere too. oh well.

tool-testsuite/test/org/antlr/v4/test/tool/TestVocabulary.java
```diff
@@ -64,7 +67,8 @@ public void testVocabularyFromTokenNames() {
 		Vocabulary vocabulary = VocabularyImpl.fromTokenNames(tokenNames);
 		Assert.assertNotNull(vocabulary);
 		Assert.assertEquals("EOF", vocabulary.getSymbolicName(Token.EOF));
-		for (int i = 0; i < tokenNames.length; i++) {
+		Assert.assertEquals(vocabulary.getMaxTokenType(), tokenNames.length-1);
+		for (int i = Token.MIN_USER_TOKEN_TYPE; i <= vocabulary.getMaxTokenType(); i++) {
```
parrt: just adding what you wanted

runtime/Java/src/org/antlr/v4/runtime/VocabularyImpl.java
parrt: actaully i'll just note size is not guaranteed to be max tokentype + 1

tool-testsuite/test/org/antlr/v4/test/tool/TestVocabulary.java
sharwell: Yes, the new assertion before the for loop is a good check to add. There is still a problem with the for loop initializer, which now skips checking i==0.

parrt:
@antlr/antlr-targets I updated C++, Swift, C# targets for 3 new methods in Vocabulary. Seems ok.

mike-lischke:
Hmm, IMO, having both `getSymbolicName(i)` and `getSymbolicNames` is a waste of electrons. Both deliver the same information in just marginally different form. However, I prefer the new `getSymbolicNames` function because the returned array is iteratable and still can directly be indexed like the single-result form. I also agree with Sam that such redundancy can confuse people.

parrt:
@mike-lischke redundancy in getting an array and single elements?

parrt:
@sharwell @mike-lischke perhaps I move the methods out of interface and put in implementation class as convenience method.

mike-lischke:
Well, **I** would wonder why there is a getElement() if there is also a getElements() because getElement() is just a subset of the other one, but your mileage may vary :-)

parrt:
Thank you for your feedback. I'm closing this in favor or https://github.com/antlr/antlr4/pull/1463

PR #1458, opened 2016-12-07
Check labels for tokens with different types 

### LINKED ISSUE #1409: Add support to label rules of mixed types
Currently it's allowed to label rules of mixed types, however it compiles to an invalid code.

```antlr
example: value=integer | value=string;
```
The generated parser will contain a `value` property which type is `StringContext` (because it was declared later). IMHO, the most appropriate behaviour is to declare the property as `ParserRuleContext` (all rules) or `ParseTree` (rules and tokens).

KvanTTT:
fixes #1409

### DIFF
--- a/tool-testsuite/test/org/antlr/v4/test/tool/TestSymbolIssues.java
+++ b/tool-testsuite/test/org/antlr/v4/test/tool/TestSymbolIssues.java
@@ -309,4 +309,39 @@ public class TestSymbolIssues extends BaseJavaToolTest {
 
 		testErrors(test, false);
 	}
+
+	// https://github.com/antlr/antlr4/issues/1409
+	@Test public void testLabelsForTokensWithMixedTypes() {
+		String[] test = {
+				"grammar L;\n" +
+				"\n" +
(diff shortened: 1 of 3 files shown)


parrt:
excellent work, @KvanTTT !

sharwell:
@KvanTTT What about this case?

```antlr
rule
  : t=a
  | t+=a+
  ;
```

KvanTTT:
@sharwell, yes it's also covered by [checkForTypeMismatch](https://github.com/antlr/antlr4/blob/11117e02d4e01fd7a99fa54258d7d0abadfaf7db/tool/src/org/antlr/v4/semantics/SymbolChecks.java#L153) method and test from [TestSymbolIssues.java](https://github.com/antlr/antlr4/blob/11117e02d4e01fd7a99fa54258d7d0abadfaf7db/tool-testsuite/test/org/antlr/v4/test/tool/TestSymbolIssues.java#L53): `label x type mismatch with previous definition: TOKEN_LIST_LABEL!=TOKEN_LABEL`.

PR #1460, opened 2016-12-08
Fix AttributeError when trying to write an immutable variable in Python.

renatahodovan:
The `removeOne` function of IntervalSet tries to directly
rewrite the start field of an immutable range variable when
splitting an existing interval. This causes AttributeError which
is fixed by the patch.

### DIFF
--- a/runtime/Python2/src/antlr4/IntervalSet.py
+++ b/runtime/Python2/src/antlr4/IntervalSet.py
@@ -158,7 +158,7 @@ class IntervalSet(object):
                 # split existing range
                 elif v<i.stop-1:
                     x = Interval(i.start, v)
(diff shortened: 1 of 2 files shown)


ericvergnaud:
Hi,
thanks for this.
Could you provide a simple test case that triggers this piece of code, that is not covered by the current test suite?
Eric

renatahodovan:
The test case that triggered the failure for me was:

`__attribute__(())`

This was parsed with the official [C grammar](https://github.com/antlr/grammars-v4/blob/master/c/C.g4) using the `gccAttributeSpecifier` as start rule.

parrt:
@ericvergnaud do we have a similar error in python3 target?

Issue #1459, opened 2016-12-07
[golang target]  Specific grammar generates code that fails to compile


syn-ful:
Before submitting an issue to ANTLR, please check off these boxes:
 
- [x] I am not submitting a question on how to use ANTLR; instead, go to [antlr4-discussion google group](https://groups.google.com/forum/#!forum/antlr-discussion) or ask at [stackoverflow](http://stackoverflow.com/questions/tagged/antlr4)
- [x] I have done a search of the existing issues to make sure I'm not sending in a duplicate

Please include information about the expected behavior, actual behavior, and the smallest grammar or code that reproduces the behavior. If appropriate, please indicate the code generation targets such as Java, C#, ...  Pointers into offending code regions are also very welcome.

### Expected behavior
Go code that successfully compiles.

### Actual behavior
When trying to compile Golang code from a grammar that has a mixture of non-greedy subrules and two or more "?" wildcard characters, the code fails to compile.  I'm using this grammar successfully with a Python2 target.

```
$ go run test_parse.go test.txt 
# _/Users/ccrook/git-repos/config_parser/antlr_playground/go_parser
go_parser/test_parser.go:337: no new variables on left side of :=
```

### Steps to reproduce the behavior

Directory structure:
```
├── Test.g4
├── go_parser
│   ├── Test.tokens
│   ├── TestLexer.tokens
│   ├── test_base_listener.go
│   ├── test_lexer.go
│   ├── test_listener.go
│   └── test_parser.go
├── test.txt
├── test_parse.go
```

Grammar:
  ***Notice multiple "?" wildcard chars in the `foo` parser rule.***

```
grammar Test;

init
  : .*? (foo .*?)+ ;

foo
  : 'A' 'B'? 'C'?
  ;
```

ANLTR4 command: 
***Note: I'm pointing to a 4.6 version of the code (compiled from source 12/6).***
```
java -jar ../lib/antlr4-4.6-SNAPSHOT.jar -Dlanguage=Go -o go_parser Test.g4
```

Attempt to complile/run go code:
```
$ go run test_parse.go test.txt 
# _/Users/ccrook/git-repos/config_parser/antlr_playground/go_parser
go_parser/test_parser.go:337: no new variables on left side of :=
```

The compiled go code attempts to declare the "la_" variable multiple times:
```
	la_ := p.GetInterpreter().AdaptivePredict(p.GetTokenStream(), 3, p.GetParserRuleContext())

	if la_ == 1 {
		{
[...]

PR #1462, opened 2016-12-08
Fix for #1459

pboyer:
## Issue 

#1459 shows that `la_` may end up being redeclared in generated code.

## Fix

This variable doesn't need to be declared at all. We can just use inline it.

## Testing

All tests continue to pass. I used @syn-ful's example code to confirm this now compiles and from basic experimentation, it seems to work as expected.

PTAL @parrt CC @syn-ful

### DIFF
--- a/tool/resources/org/antlr/v4/tool/templates/codegen/Go/Go.stg
+++ b/tool/resources/org/antlr/v4/tool/templates/codegen/Go/Go.stg
@@ -644,10 +644,7 @@ p.GetErrorHandler().Sync(p)
 
 <endif>
 
-
-la_ := p.GetInterpreter().AdaptivePredict(p.GetTokenStream(), <choice.decision>, p.GetParserRuleContext())
-
-switch la_ {
+switch p.GetInterpreter().AdaptivePredict(p.GetTokenStream(), <choice.decision>, p.GetParserRuleContext()) {
 <if(alts)>
 <alts:{alt | case <i>:
 	<alt>}; separator="\n\n">
@@ -658,10 +655,10 @@ switch la_ {
 OptionalBlock(choice, alts, error) ::= <<
 p.SetState(<choice.stateNumber>)
 p.GetErrorHandler().Sync(p)
-la_ := p.GetInterpreter().AdaptivePredict(p.GetTokenStream(), <choice.decision>, p.GetParserRuleContext())
+
 <if(alts)>
 
-<alts:{alt | if la_ == <i><if(!choice.ast.greedy)>+1<endif> {
+<alts:{alt | if p.GetInterpreter().AdaptivePredict(p.GetTokenStream(), <choice.decision>, p.GetParserRuleContext()) == <i><if(!choice.ast.greedy)>+1<endif> {
 	<alt>
 }; separator="} else ">
 <endif>


tool/resources/org/antlr/v4/tool/templates/codegen/Go/Go.stg
```diff
@@ -658,10 +655,10 @@ switch la_ {
 OptionalBlock(choice, alts, error) ::= <<
 p.SetState(<choice.stateNumber>)
 p.GetErrorHandler().Sync(p)
-la_ := p.GetInterpreter().AdaptivePredict(p.GetTokenStream(), <choice.decision>, p.GetParserRuleContext())
+
 <if(alts)>
 
-<alts:{alt | if la_ == <i><if(!choice.ast.greedy)>+1<endif> {
+<alts:{alt | if p.GetInterpreter().AdaptivePredict(p.GetTokenStream(), <choice.decision>, p.GetParserRuleContext()) == <i><if(!choice.ast.greedy)>+1<endif> {
```
sharwell: ❗️ This will cause the full prediction for a decision to execute multiple times (massive negative performance impact)

tool/resources/org/antlr/v4/tool/templates/codegen/Go/Go.stg
```diff
@@ -658,10 +655,10 @@ switch la_ {
 OptionalBlock(choice, alts, error) ::= <<
 p.SetState(<choice.stateNumber>)
 p.GetErrorHandler().Sync(p)
-la_ := p.GetInterpreter().AdaptivePredict(p.GetTokenStream(), <choice.decision>, p.GetParserRuleContext())
```
sharwell: ❓ Have you tried changing this to use `=` instead of `:=`? It seems like you should able to forward declare this variable and then use it here without redeclaring it.

tool/resources/org/antlr/v4/tool/templates/codegen/Go/Go.stg
pboyer: @sharwell Not totally sure, but I think you're wrong about that. 

The case shown in #1459 have different parameter sets. In this case, the `decision` id is different in the two invocations.

The code that was previously causing a compiler error is now re-written as:

```
if p.GetInterpreter().AdaptivePredict(p.GetTokenStream(), 3, p.GetParserRuleContext()) == 1 {
		{
			p.SetState(22)
			p.Match(TestParserT__1)
		}

	}
	p.SetState(26)
	p.GetErrorHandler().Sync(p)

	if p.GetInterpreter().AdaptivePredict(p.GetTokenStream(), 4, p.GetParserRuleContext()) == 1 {
		{
			p.SetState(25)
			p.Match(TestParserT__2)
		}

	}
```

sharwell: If you use [this Java.g4 grammar](https://github.com/tunnelvisionlabs/antlr4ts/blob/master/benchmark/Java.g4), you should see an example of the pathological case inside the generated code for the `annotation` rule.

parrt:
if we can avoid it, we should avoid repeat adaptivepredict calls.

pboyer:
@sharwell I definitely appreciate the extra scrutiny here and certainly agree we should avoid the extra calls!

However, I'm feeling like this criticism may be undue. The Java templates, as far as I can tell, would suffer from the same issue.

https://github.com/antlr/antlr4/blob/master/tool/resources/org/antlr/v4/tool/templates/codegen/Java/Java.stg#L532

If I'm wrong, maybe you guys can point out why and I can take a similar approach. Otherwise, it may take me a few cycles to get a fix in.

parrt:
@pboyer I'm ok with the extra call. It's your target and usually won't be noticed. I think we avoid double calls to prediction for speed reasons. We might make an unnecessary single call but I'm thinking we never double up.

sharwell:
@parrt For an optional block with *N* alternatives, this implementation will call `adaptivePredict` up to *N* times (with the worst case always occurring when the optional item does not appear in the input).

sharwell:
@pboyer The Java code avoids the problem by using a `switch` statement instead of a sequence of `if` statements. The prediction is made once, and the code branches to the `case` associated with the result of the single call to `adaptivePredict`.

pboyer:
@sharwell Thanks! It did not occur to me that the `if` block might be generated recursively in the templates. It definitely makes more sense to have a switch. I'll try out the Java.g4 grammar you suggested and see if I can do something more efficient.

Issue #1464, opened 2016-12-08
[Tool][CppTarget]Build cpp demo generate warning: extra ‘;’


gaulouis:
Hello,

When i build the CppTarget demo on Linux, with gcc, i got some pollution warning.

Issue #1457, opened 2016-12-06
[Cpp] namespace pollution of the antlr4 namespace from generated lexer/parser


Naios:
50646a6da8

Generated C++ parsers and lexers are making the `antlr4` scope globally available through a
`using namespace antlr4` statement in the generated *Parser.h and *Lexer.h which leads to a namespace pollution, especially when not using a custom namespace (`-package` option).

See the sources from the cpp target demo project:

TParser.h
```c++
/* parser postinclude section */

using namespace antlr4;

/* parser context section */
```

The expected behaviour is that included headers never make namespaces globally available,
which means the `using namespace antlr4` statement needs to be removed but this could break existing
code including generated headers while not fully qualifying classes in the `antlr4` namespace.

See http://stackoverflow.com/questions/5849457/using-namespace-in-c-headers for details.

Naios:
I could open a PR for it, however I wanted to ask first whether it will be accepted because of the possible breakage of code not declaring classes as part of the `antlr4` namespace as explained above.

sharwell:
I don't think we've had a first release of this target, so now would be the time to do it if we're going to do it. @mike-lischke What do you think?

gaulouis:
I can do it tonight if someone can do the Java [generate one file per class](https://github.com/antlr/antlr4/issues/1456) else I continue my search...

mike-lischke:
Right, while I made sure not to use `using namespace` in any of the runtime files I did that for the generated files and this can easily be fixed. @Naios if you are willing to do that and open a PR for merging that then go ahead. Every help is appreciated.

Issue #1443, opened 2016-12-02
Conversion error when build cpp target


ivan-khudyashev:
Error when make test grammar.
> antlr4_runtime/cpp/demo/Linux/main.cpp:28:42: error: conversion from ‘antlrcpptest::TParser::MainContext*’ to non-scalar type ‘Ref<antlr4::tree::ParseTree> {aka std::shared_ptr<antlr4::tree::ParseTree>}’ requested
>    Ref<tree::ParseTree> tree = parser.main();
> demo/CMakeFiles/antlr4-demo.dir/build.make:62: ошибка выполнения рецепта для цели «demo/CMakeFiles/antlr4-demo.dir/Linux/main.cpp.o»
> make[2]: *** [demo/CMakeFiles/antlr4-demo.dir/Linux/main.cpp.o] Ошибка 1
> CMakeFiles/Makefile2:252: ошибка выполнения рецепта для цели «demo/CMakeFiles/antlr4-demo.dir/all»
> make[1]: *** [demo/CMakeFiles/antlr4-demo.dir/all] Ошибка 2
> Makefile:127: ошибка выполнения рецепта для цели «all»
> make: *** [all] Ошибка 2

Sequence of actions.
I. Download zip-archive project https://github.com/antlr/antlr4/archive/master.zip
II. Unpack zip-archive into directory: `sompath/antlr4_runtime/cpp`
III. Download into directory `sompath/antlr4_runtime/cpp` file antlr4-4.5.4-SNAPSHOT.jar from here:
http://www.soft-gems.net/index.php/tools/49-the-antlr4-c-target-is-here
link for download: http://www.soft-gems.net/supplement/download.php?ID=105
IV. Execute command from Readme.md
1. `cd sompath/antlr4_runtime/cpp`
2. `mkdir build && mkdir run && cd build`
3. `cmake .. -DANTLR_JAR_LOCATION=sompath/antlr4_runtime/cpp/antlr4-4.5.4-SNAPSHOT.jar -DWITH_DEMO=True`
4. `make`
**ERROR**

Conditions:
OS: Ubuntu 16.04.1 LTS
`java -version`:  
> openjdk version "1.8.0_111"
> OpenJDK Runtime Environment (build 1.8.0_111-8u111-b14-2ubuntu0.16.04.2-b14)
> OpenJDK 64-Bit Server VM (build 25.111-b14, mixed mode)

`g++ --version`: 

[...]

PR #1465, opened 2016-12-08
Fix build cpp demo

gaulouis:
Thanks you to consider my "contribution".

@see https://github.com/antlr/antlr4/issues/1457 (namespace pollution)
@see https://github.com/antlr/antlr4/issues/1464 (warnings polution)
@see https://github.com/antlr/antlr4/issues/1443 (build Linux error)

### DIFF
--- a/runtime/Cpp/runtime/src/IntStream.cpp
+++ b/runtime/Cpp/runtime/src/IntStream.cpp
@@ -1,32 +1,6 @@
-﻿/*
- * [The "BSD license"]
- *  Copyright (c) 2016 Mike Lischke
- *  Copyright (c) 2013 Terence Parr
- *  Copyright (c) 2013 Dan McLaughlin
- *  All rights reserved.
- *
- *  Redistribution and use in source and binary forms, with or without
- *  modification, are permitted provided that the following conditions
- *  are met:
- *
- *  1. Redistributions of source code must retain the above copyright
- *     notice, this list of conditions and the following disclaimer.
- *  2. Redistributions in binary form must reproduce the above copyright
- *     notice, this list of conditions and the following disclaimer in the
- *     documentation and/or other materials provided with the distribution.
- *  3. The name of the author may not be used to endorse or promote products
- *     derived from this software without specific prior written permission.
- *
- *  THIS SOFTWARE IS PROVIDED BY THE AUTHOR ``AS IS'' AND ANY EXPRESS OR
- *  IMPLIED WARRANTIES, INCLUDING, BUT NOT LIMITED TO, THE IMPLIED WARRANTIES
- *  OF MERCHANTABILITY AND FITNESS FOR A PARTICULAR PURPOSE ARE DISCLAIMED.
- *  IN NO EVENT SHALL THE AUTHOR BE LIABLE FOR ANY DIRECT, INDIRECT,
- *  INCIDENTAL, SPECIAL, EXEMPLARY, OR CONSEQUENTIAL DAMAGES (INCLUDING, BUT
- *  NOT LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS OR SERVICES; LOSS OF USE,
- *  DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER CAUSED AND ON ANY
- *  THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY, OR TORT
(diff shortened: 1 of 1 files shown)


runtime/Cpp/demo/TLexer.g4
```diff
@@ -66,7 +66,7 @@ OpenCurly: '{' -> pushMode(Mode1);
 CloseCurly: '}' -> popMode;
 QuestionMark: '?';
 Comma: ',' -> skip;
-Dollar: '$' -> more, mode(Mode1), type(DUMMY);
```
KvanTTT: I guess the tokens section
```ANTLR
tokens {
	DUMMY	
}
```
can be also removed for now. It's not used anywhere.

gaulouis: Thanks,

I will remove it...

parrt:
@mike-lischke do you bless this?

gaulouis:
"do you bless this?"

No !

1] options{ superClass} bugged
```
ParserHeader(parser, funcs, atn, sempredFuncs, superClass = {Parser}) ::= <<
<namedActions.context>

class <parser.name> : public antlr4::<superClass> {
```
need to be replaced by 
```
ParserHeader(parser, funcs, atn, sempredFuncs, superClass = {antlr4::Parser}) ::= <<
```

2] Lexer namespace pollution not clean

3] antlr4 namespace present in impelementation file
I need to add: ``LexerRuleContextHeader() ::= "antlr4::RuleContext"`` according to ``LexerRuleContext() ::= "RuleContext"``

parrt:
Ok, so I should close or let you keep working on this branch/PR?

gaulouis:
I would like to familiarize myself with git.

Please leave it open so I can have a full review proccess.

PS: These are just small modifications.

gaulouis:
I have break some thing. I'am sorry. How can i fix it ?

tool/resources/org/antlr/v4/tool/templates/codegen/Cpp/Cpp.stg
```diff
@@ -263,10 +263,10 @@ bool <if (r.factory.g.lexer)><lexer.name><else><parser.name><endif>::<r.name>Sem
 
 //--------------------------------------------------------------------------------------------------
 
-ParserHeader(parser, funcs, atn, sempredFuncs, superClass = {Parser}) ::= <<
+ParserHeader(parser, funcs, atn, sempredFuncs, superClass = {antlr4::Parser}) ::= <<
 <namedActions.context>
 
-class <parser.name> : public antlr4::<superClass> {
+class <parser.name> : public <superClass> {
```
mike-lischke: That will not work well. A superclass doesn't necessarily have to be from the antlr4 namespace, but could be an own class.

mike-lischke:
PLEASE don't mix many unrelated changes into a single PR. It's very difficult to review them and almost impossible to accept or reject if something is wrong. Make only related changes into a PR and rather issue several PRs then.

Some notes (though I'd like to make them on individual PRs):

1. Don't change the C++ template lightheaded. The problem with multiple semicolons is known and there is no solution currently. The culprit here is that sometimes templates are used in other templates in various nesting levels. Sometimes a semicolon is provided by a higher level, sometimes it is not.
2. The demo: the grammar is more than just a demonstration. It contains everything to make ANTLR generate all necessary constructs, so that we can see if the code generation works properly. Don't remove elements from the grammar! It's not primarily thought as a demonstration for the end user (the demo app is however). There is currently no other way to test completeness of code generation.
3. Bug #1443 is only a missing update of the linux demo app. I simply haven't come around to do that.
4. Some template changes are plain wrong (e.g. the antlr4:: prefix for <SuperClass>).

So, I recommend to close this PR and make the changes instead into individual PRs for individual reviews.

gaulouis:
Thank you for your valuable informations.

KvanTTT:
@mike-lischke I agree with you: the tokens section should be existitng, but the `type` command should be used in another case.

gaulouis:
I have do:
```
-Dollar: '$' -> more, mode(Mode1), type(DUMMY);

+Dollar: '$' -> more, mode(Mode1);
+Ampersand: '&' -> type(DUMMY);
```
However, I have no rules that use Dollar or Ampersand. Do i need to create a rule ?

mike-lischke:
@gaulouis No, it's just fine like that. It doesn't count that we don't use a rule, the code generation still kicks in for every rule we have in a grammar. Your change is fine, just check that the demo app still parses the example input, so that new users have an easy start.

Issue #1461, opened 2016-12-08
[Grammar] warning "incompatible commands more and type" treated as error


gaulouis:
Hello,

I have build Antlr4 from git and compiled the antlr4-cpp target succefully.
When i try to build the runtime/Cpp/demo, i got the error : ``warning(179): TLexer.g4:69:34: incompatible commands more and type``

The raison is than org.antlr.v4.tool.ErrorType have been modified. And therefore the grammar used in [Cpp/demo](https://github.com/antlr/antlr4/blob/master/runtime/Cpp/demo/TLexer.g4) is obsolete.

```
Dollar: '$' -> more, mode(Mode1), type(DUMMY);
```
Please. How to correct incompatibilities of commands in the grammar ?

Related issues:
- https://github.com/antlr/antlr4/issues/1388
- https://github.com/antlr/antlr4/issues/348

KvanTTT:
You can not use `more` and `type(DUMMY)` commands simultaniously, it's senselessly. See the table in my abovementioned issue. It seems to me that you should remove `type(DUMMY)` command in your case bacause of tokens with `DUMMY` type will never appear. They always will have `Dot` type from `Mode1`.

gaulouis:
Thank you very much for your valuable quick help.
I delete "type (DUMMY)" and everything works fine.

PR #1468, opened 2016-12-09
Fix compilation error for CppTarget demo on Linux

gaulouis:


gaulouis:
Now, if i want to fix the #1461 . What i need to do ?
- Create a new PR with only grammar modification
- Create a new PR with grammar modification and #1443 fix
- Append a patch file on this thread
- Wait for this PR will be merged

parrt:

PR #1469, opened 2016-12-09
Fix conflict lexer command in CppTarget demo

gaulouis:
Since Tool generates errors if token have several identical commands, the demo grammar in CppTarget throw an error at compile time.

This Pull Request fix this, by creating a new token.


parrt:

PR #1470, opened 2016-12-09
Issue 1457

### LINKED ISSUE #1457: [Cpp] namespace pollution of the antlr4 namespace from generated lexer/parser

gaulouis:
This Pull Request fix #1457 

- Ensured that there is no "using namespace" in the generated header files

- Ensured that there is any "using namespace" in runtime/src headers included by antlr-runtime.h:
``$]find runtime/Cpp/runtime/src/ -type f -name *.h -print | xargs grep "namespace" | grep using``

### DIFF
--- a/contributors.txt
+++ b/contributors.txt
@@ -126,3 +126,4 @@ YYYY/MM/DD, github id, Full name, email
 2016/11/29, Naios, Denis Blank, naios@users.noreply.github.com
 2016/12/01, samtatasurya, Samuel Tatasurya, xemradiant@gmail.com
 2016/12/03, redxdev, Samuel Bloomberg, sam@redxdev.com
+2016/12/11, Gaulouis <gaulouis.com@gmail.com>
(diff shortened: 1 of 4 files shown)


parrt:
Hi. please sign contributors.txt :)

parrt:

Issue #1076, opened 2015-12-17
JavaScript visitors not visiting children


RedTailedHawk:
My generated visitors are working fine in Python and Java, but not in JavaScript.  Could it be that the generated JavaScript is just a stub and does not visit the children?

```
// Visit a parse tree produced by rql2Parser#expression.
rql2Visitor.prototype.visitExpression = function(ctx) {
};
```

I'm using ANTLR 4.5.1 to generate the code.  Is there a later version I should be using?

Thanks for any help.
-Larry

juniperoserra:
The issue is that there is a bug in ParseTreeVisitor.visit.

Inside the "`map`" clause where it says, "`ctx.accept(this)`" it should say "`child.accept(this)`" 

It can be patched from the outside.
I'll try to set up a pull request.

`ParseTreeVisitor.prototype.visit = function(ctx) {
    if (Array.isArray(ctx)) {
        return ctx.map(function(child) {
            return child.accept(this);
        }, this);
    } else {
        return ctx.accept(this);
    }
};`

PR #1471, opened 2016-12-09
Call visitErrorNodes properly in Python targets.

renatahodovan:
The consume method of the Parser class calls visitTerminal on both
TerminalNodes and ErrorNodes even if the comment above states that
ErrorNodes should be visited by visitErrorNodes. This behaviour
is also inconsitent with the Java target. The patch fixes this in
both Python targets.

### DIFF
--- a/runtime/Python2/src/antlr4/Parser.py
+++ b/runtime/Python2/src/antlr4/Parser.py
@@ -328,7 +328,10 @@ class Parser (Recognizer):
                 node = self._ctx.addTokenNode(o)
             if hasListener:
                 for listener in self._parseListeners:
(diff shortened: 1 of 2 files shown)


parrt:
@ericvergnaud blessed?

parrt:
Does this fix or is it similar to https://github.com/antlr/antlr4/issues/1076

renatahodovan:
@parrt It seems that JavaScript target also lacks of the call of visitErrorNode. So yes, it has a similar problem. (The other visitors seem to be working already.) I can upload a fix for that too.

ericvergnaud:
blessed

PR #1472, opened 2016-12-09
Fix 1298

parrt:
`sync()` was could add error nodes that were ignored when we converted regular RuleContext nodes to those for alt labels.


parrt:
@antlr/antlr-targets Note that this bug fix is applicable to all targets. A small modification to the `ParserRuleContext.copyFrom()` method.

PR #1474, opened 2016-12-09
fix IntervalSet

### LINKED ISSUE #1265: A bug of the antlr4 python3 runtime 4.5.3
In the file IntervalSet.py, the statement at line 147 is,
     _i.start = v + 1_.
In which the variable _i_ is of type _range_, so its member _start_ is readonly.
Maybe it can be replaced by, 
    _self.intervals[self.intervals.index(i)] = range(v + 1, i.stop)._

parrt:

Fixes #1265


renatahodovan:
@parrt Could you please have a look at #1460 ? I've already proposed the same fix for both Python targets.

PR #1475, opened 2016-12-10
Fix travis runtime-test Cpp

gaulouis:
Since divide to conquer with travis. CppTarget test unit are break

This Pull Request fix the command line argument order.

Replace:
``g++ -std=c++11 -I/xxx -L/yyy -lantlr4-runtime -o a.out zzz.cpp``
By:
``g++ -std=c++11 zzz.cpp -o a.out -I/xxx -L/yyy -lantlr4-runtime``


gaulouis:
CppTarget build correctly both with clang++ and g++ but g++ throw error on test. Need more fix on template.

Ambigus travis jobs labels:
TARGET=cpp CXX=g++-5
instead of:
TARGET=cpp CXX=clang++

Issue #1298, opened 2016-09-30
Single-token deletion error node not in parse tree


lionelplessis:
Hi there,

When single-token deletion kicks in and skips a token in a rule with alternative labels (it entered the rule but didn't find the alternative yet) the error node that is created is not attached to the parse tree. If the extra token is present after a valid rule alternative is found then the error node representing the token is correctly present in the parse tree.

This is due to the ParseRuleContext.CopyFrom method not copying children (at least error nodes). 
See the comments on the generated code:

``` java
private VariableExprContext variableExpr(int _p) throws RecognitionException {
   ParserRuleContext _parentctx = _ctx;
   int _parentState = getState();
   VariableExprContext _localctx = new VariableExprContext(_ctx, _parentState); // Base class of alt rule with labels (correct alt not found yet).
   VariableExprContext _prevctx = _localctx;
   int _startState = 6;
   enterRecursionRule(_localctx, 6, RULE_variableExpr, _p);
   int _la;
   try {
      int _alt;
      enterOuterAlt(_localctx, 1);
      {
      setState(36);
      _errHandler.sync(this); // Single-token deletion skips extra token and adds an error node to the children of _localctx (VariableExprContext).
      switch ( getInterpreter().adaptivePredict(_input,3,_ctx) ) {  // A valid alternative is found.
      case 1:
         {
         _localctx = new LiteralExprContext(_localctx); // Simple fields(properties are copied here but not children (which include the sinlge-token deletion error node).
         _ctx = _localctx;
         _prevctx = _localctx;

         setState(33);
         literal();
         }
         break;
```

Some more context and a concrete example:
We are using Antlr 4.5.3 to build a proper grammar/parser for an existing template engine. Target languages are C# and java but we have been working mainly with C# so far. We have both a lexer and parser files to use Antlr modes.
We are writing and testing the grammar using the Antlr plugin for IntelliJ. Note that for this issue the **IntelliJ plugin is misleading** as it shows the missing error node in the tree which is not the case for the TestRig (-gui) (parse tree vs listener issue??).

In our template engine, dynamic expressions in text must be inside _${_ expr _}_ . Expressions mainly support basic operators, a few functions, identifiers and number literals. Our lexer file has a catch-all rule to match any unrecognized character and we skip (hidden channel) whitespace characters.

So expression _${abc}_ is valid. Expression _${ ? abc ? }_ with 2 single invalid '?' tokens before and after a valid labeled alternative produces a parse tree without the error node for the first token.
[...]

PR #1476, opened 2016-12-10
Implement the fix of #1298 for Python targets.

renatahodovan:
The patch adapts the fix in PR #1472 for Python targets.


parrt:
Looks correct. thanks!

PR #1479, opened 2016-12-10
[Cpp] Explain how to install the Cpp tools

alvarogarcia7:


KvanTTT:
This is not actual. The 4.6 version is upcoming.

parrt:
hi. yeah, we're coming out with 4.6 so this stuff will change.

PR #1493, opened 2016-12-12
Realign CSharp runtime with Java runtime, implements #1441

ericvergnaud:


parrt:
@ericvergnaud this mixes js with C# commits. Was that intentional?

ericvergnaud:
No but it doesn't hurt.

Envoyé de mon iPhone

parrt:
ok. thanks; working on C++ at moment. hang on...

PR #1498, opened 2016-12-13
Fixes #1299. matching a set didn't properly notify error strategy of a successful match

### LINKED ISSUE #1299: Matched tokens being marked as error
I found a case where the parser is marking matched tokens found inside error nodes as an error. For instance, given the malformed expression `1 * 2 +++`, this is the resulting tree:

antlr4_parse_tree

However, it's the expected tree:

image

Notice that the operators `+` are being marked as an error in the first tree, although they have being recognized as operators. 

It worth noting that the former tree was generated using the TestRig, while the latter is a screenshot of the tree generated in the IntelliJ plugin. This issue is most likely related to antlr/intellij-plugin-v4#180.

parrt:


parrt:
@antlr/antlr-targets An issue in other targets too. I changed and will see what build says.

Issue #1077, opened 2015-12-20
ANTLR4 problem with malformed(?) backslash-specified unicode literal 


SimonSntPeter:
Trivial bug. Seems to produce invalid java, which then bombs. Problem may be that unicode chars of form "\u..." are expected to be of 4 digits.
This should reproduce

```
fragment MONEYSIGN : [\u24\uA2] ;
```

gets me a java stack dump starting

```
Exception in thread "main" java.lang.NumberFormatException: For input string: "24\u"
```

Curiously this doesn't except

```
fragment MONEYSIGN : [\u24] ;
```

Using antlr 4.5.1

SimonSntPeter:
Above isn't very helpful. Better code to reproduce is

```
grammar MSSQL;

moneysymbol : MONEYSYMBOL ;

MONEYSYMBOL : [\u24\uA2] ;  

SKIPWS : [ \t\r\n]+ -> skip ;
```

and start the parse at moneysymbol. Book says unicode must be in form \uXXXX so this really is minor. However if you just have "MONEYSYMBOL : [\u24] ;" then it doesn't crash but fails with this on the command line (when given input $, which is character \u24):

```
line 1:0 token recognition error at: '$'
[@0,3:2='<EOF>',<-1>,2:0]
enter   moneysymbol, LT(1)=<EOF>
line 2:0 missing MONEYSYMBOL at '<EOF>'
exit    moneysymbol, LT(1)=<EOF>
```

which I take to mean that it's expecting something but didn't get it.

KvanTTT:
I'll fix it and another lexical issues like this:
```ANTLR
INVALID_RANGE:         'GH'..'LM';
INVALID_RANGE_2:       'F'..'A' | 'Z';
INVALID_CHAR_SET:      [f-az];
```

PR #1517, opened 2016-12-15
Character issues

KvanTTT:
fixed https://github.com/antlr/antlr4/issues/1077 and another character issues (see tests for detail).
Some changes are taken from https://github.com/antlr/antlr4/pull/1092

### DIFF
--- a/runtime-testsuite/test/org/antlr/v4/test/runtime/descriptors/LexerExecDescriptors.java
+++ b/runtime-testsuite/test/org/antlr/v4/test/runtime/descriptors/LexerExecDescriptors.java
@@ -302,30 +302,6 @@ public class LexerExecDescriptors {
 
 	}
 
-	public static class CharSetWithReversedRange extends BaseLexerTestDescriptor {
-		public String input = "9";
-		/**
-		A
-		[@0,0:0='9',<1>,1:0]
-		[@1,1:0='<EOF>',<-1>,1:1]
-		 */
-		@CommentHasStringValue
-		public String output;
-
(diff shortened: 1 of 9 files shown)


mike-lischke:
While you are that would it be possible for you to add full Unicode sets (everything up to 0x10FFFF) in the code generator? That looks like something related to your changes here, so you might have a quick idea how to implement that.

parrt:
There's a lot going on here. Not sure I can make all of these right before a release.  E.g., I don't agree that 'keyword' and ID should yield a warning. we'll see millions of them in a typical grammar.

KvanTTT:
@mike-lischke I think full Unicode sets should be implemented in separated pull request. This one it too big and fixes the bugs, not brings a new features.

KvanTTT:
> E.g., I don't agree that 'keyword' and ID should yield a warning

@parrt 'keyword' and ID does not yield a warning here. This PR about incorrect, duplicated and empty char sets.

parrt:
ah. ok, well, let me pull and sniff some more.

parrt:
@mike-lischke I don't see a quick way to do 32-bit unicode. certainly makes everything hard as char is 16 bits in Java.

parrt:
@KvanTTT  please note for future reference how i like my `else` on a separate line than `}` ;)

KvanTTT:
@parrt ok, I got it :)

parrt:
I pulled in your stuff into a separate branch, made some tweaks and will use a separate PR so I'm closing this.

PR #1523, opened 2016-12-15
New cpp option

parrt:
Updated PR from @mike-lischke to use grammar option not command-line option.


mike-lischke:
Sorry, I don't understand why you changed that to a grammar option. This new option goes hand hand with things like the package option. Why forcing different syntaxes on the user for very similar switches? Also, this setting is very C++ specific, so making this a grammar option is the wrong place IMO. It's something you specify only for the C++ target code generation. The original PR was as good as it could possibly be.

Side note: the .md files are like HTML. Their formatting is determined by the viewer, hence you should not manually break lines.

parrt:
@mike-lischke the editor in my ide didn't break and i was in a hurry. wanted to read.  re option. it mirrors option superClass. It is similar to -package but I leave that for backward compatibility; that should really be an option too. Adding a command-line tool option is a big deal.

mike-lischke:
Sorry to disagree, but superClass is totally different. It's an option that applies to all targets. With a grammar option we have now again something very target specific in a grammar. We should rather strive to get such things out of the grammars. But you are the supreme dictator :-)

parrt:
tool options should be few. also you would pollute the tool option space for one target?  making superClass a tool option makes way more sense than exportMacro.  I don't even like adding the new option you have. ;)  I would have told C++ users to add a -D to the C++ compile to get macro in place. I was opposed to option but had to get 4.6 out and you weren't online.

PR #1528, opened 2016-12-19
Fix for VS 2013 runtime builds (C++)

mike-lischke:
There is an additional revision in the PR (about the new setting) which only contains a reverted doc file. It's not the full revision (I only reverted the manual line breaks). So it should be save to merge as well.


parrt:
@mike-lischke Can we take out the new command-line option? ;)

mike-lischke:
@parrt Not sure what you mean. There is no code wrt. the command line option (it might seem so when you look through the individual commits, but the final change is only in text). See the tab "Files Change 4".

mike-lischke:
Ok, hope now it's fine.

parrt:
Yep, now it's gone. thanks!

PR #1548, opened 2016-12-24
adapt code to compile with msys2 mingw compiler

adrpo:
- add required extra flags to the C++ compiler in CMakeLists.txt
- guard some code with _MSC_VER instead of WIN32

[cut]
Thank you for proposing a contribution to the ANTLR project. In order to accept changes from the outside world, all contributors must "sign" the  [contributors.txt](https://github.com/antlr/antlr4/blob/master/contributors.txt) contributors certificate of origin. It's an unfortunate reality of today's fuzzy and bizarre world of open-source ownership.

Make sure you are already in the contributors.txt file or add a commit to this pull request with the appropriate change. Thanks!
[/cut]

### DIFF
--- a/contributors.txt
+++ b/contributors.txt
@@ -128,3 +128,4 @@ YYYY/MM/DD, github id, Full name, email
 2016/12/03, redxdev, Samuel Bloomberg, sam@redxdev.com
 2016/12/11, Gaulouis, Gaulouis, gaulouis.com@gmail.com
 2016/12/22, akosthekiss, Akos Kiss, akiss@inf.u-szeged.hu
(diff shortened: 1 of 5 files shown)


parrt:
@mike-lischke do you bless this?

mike-lischke:
Hmm, the change from _WIN32 to _MSC_VER is actually unneeded, but it doesn't harm. Ok, ship it.

Biswa96:
> the change from _WIN32 to _MSC_VER is actually unneeded

It is needed for non-MSVC based toolchain. The `#pragma warning` directive is MSVC specific. So, in other toolchain, like gcc or clang mingw, that directive will show warnings.

PR #1560, opened 2016-12-28
Add method to ANTLRMessage to return the offending token as a v4 token.

grosenberg:
Provides a convenience conversion method to ANTLRMessage to permit the existing `offendingToken` field to be retrieved as a v4 token rather than its native v3 type.


grosenberg:
Strange.  Cause of the build failure was that the C++ tests took too long to complete.

parrt:
Hi @grosenberg I can't see any changes except whitespace. Can you point out what was added?

grosenberg:
Line 57 is the added method.

Re whitespace, is there a source format spec for the repo? Better if Github could be told to ignore space/tab whitespace differences in its compare function.

sharwell:
@grosenberg Yes there is a **.editorconfig** file in the repository. You'll need to update your IDE and/or install a plug-in so it's recognized automatically.

grosenberg:
@sharwell Thanks. My formatter settings are fully consistent with those. I did a full file reformat -- my SOP -- and converted all existing 4 space indents to tabs. I will avoid that in the future or, if needed, ensure that it is isolated in a 'fixes whitespace only' commit.

grosenberg:
> What is the motivation for this? Not sure it would be widely used and in that case easier/better to just keep the function in your own code.

Throughout the entirety of the v4 API, this is the one place where there is an explicit exposure of a v3 artifact, namely the `offendingToken`.  No doubt that other embedded tool developers have implemented this conversion as I have since my first version of AntlrDT for v4.

The problem with an own-code fix is that it requires an exported exposure of the v3 runtime. In the context of a complex, multi-plugin system, it is somewhat extreme to work around a single v4 limitation. Not a blocker, but still not 'nice.'

The nice solution is to implement the fix at the source of the problem, as proposed.  The added code is a minimal single method converting between two fundamental objects whose design is unlikely to change, and introduces no side effects or external dependencies.

As such, there is no expected cost to having added the method -- the fixed class is part of the tool, so there is no burden on the runtime.

sharwell:
> Throughout the entirety of the v4 API, this is the one place where there is an explicit exposure of a v3 artifact, namely the `offendingToken`.

The entire ANTLR 4 Tool is written using ANTLR 3. For example, Grammar.ast is a V3 tree.

sharwell:
> Really? `org.antlr.v4.Tool` is not a v4 artifact?

Sorry, I assumed we were talking about runtime artifacts getting exposed by the tool.

I don't believe a basic understanding of using the ANTLR 3 runtime types is an unrealistic expectation for working directly with code that uses it heavily inside and out. In the IDE integrations that work closely with the ANTLR 4 tool, we use the ANTLR 3 trees produced by the tool in many, *many* more ways than just `ANTLRMessage`. I'm concerned that adding this one method will have limited use in practice, and actually making this practical will just cascade into a duplicate of the long-desired request to make the tool use ANTLR 4 internally for parsing (no issue exists that I've seen).

My main recommendation is to stick with ANTLR 3 when working with the ANTLR 4 tool. Aside from that, if you want an ANTLR 4 token instead of an ANTLR 3 token you could always make this a static method of a utility class in your project which could work with both `ANTLRMessage` and any other case where tokens appear in the tool APIs.

parrt:
Hi, yeah, I guess I'll close as it's not worth belaboring.

PR #1566, opened 2016-12-30
Write visitor tests with specific tests for expected runtime semantics

sharwell:
* Add tests for runtime visitor implementations with (hopefully) complete coverage of semantics
* Remove previous visitor tests because they didn't test anything that the current tests do not cover, and didn't reflect the expected behavior of generated visitors


parrt:
until i can decide on an api lib testing policy, i'm gonna hold off / close this one

PR #1575, opened 2017-01-04
Update LexerATNSimulator.cpp

jkra:
added unicode line breaks to LexerATNSimulator::consume() for correct line and position

[cut]
Thank you for proposing a contribution to the ANTLR project. In order to accept changes from the outside world, all contributors must "sign" the  [contributors.txt](https://github.com/antlr/antlr4/blob/master/contributors.txt) contributors certificate of origin. It's an unfortunate reality of today's fuzzy and bizarre world of open-source ownership.

Make sure you are already in the contributors.txt file or add a commit to this pull request with the appropriate change. Thanks!
[/cut]

### DIFF
--- a/runtime/Cpp/runtime/src/atn/LexerATNSimulator.cpp
+++ b/runtime/Cpp/runtime/src/atn/LexerATNSimulator.cpp
@@ -604,11 +604,15 @@ void LexerATNSimulator::setCharPositionInLine(size_t charPositionInLine) {
 
 void LexerATNSimulator::consume(CharStream *input) {
   size_t curChar = input->LA(1);
-  if (curChar == '\n') {
-    _line++;
+  if ((curChar == '\n')
+      || (curChar == 0x2028 /*'\u2028'*/)
+      || (curChar == 0x2029 /*'\u2029'*/)
+      || (curChar == 0x0085 /*'\u0085'*/)
+      || (curChar == 0x000C /*'\u000C'*/)) {
+    ++_line;
     _charPositionInLine = 0;
   } else {
-    _charPositionInLine++;
+    ++_charPositionInLine;
   }
   input->consume();
 }


sharwell:
:memo: @parrt This review is blocked on #1574.

sharwell:
I lean :-1: on this change for the following reasons:

* `consume` is a performance-critical method inside the lexer. Small changes here have a larger than expected impact on the overall performance of ANTLR 4 lexers.
* These line endings are rarely seen in user scenarios, but everyone would be paying the performance hit for checking for them.
* `Lexer.consume` is only one place where line endings can be tracked. Two more places are the following:
  * The grammar could be modified so line breaks (including these Unicode line breaks) are their own token. There are multiple ways to create an action so the lexer increments the current line number when a Unicode line break token is encountered.
  * A more memory-efficient way to track line endings is storing the indexes in the input where a line starts, and using a binary search to determine the line/column from a token's offset within an input stream. This reduces the size of every token by 2 integers, and can be accomplished using a custom `CharStream`, `Token`, and `TokenFactory` without making any changes to `Lexer`.

parrt:
@mike-lischke any thoughts?

mike-lischke:
While I basically agree with Sam's arguments, I also see the need for full Unicode support. However, it's probably better to start this in the reference runtime and propagate the changes (e.g. a different line count handling like Sam suggested) to the other runtimes.

@sharwell However, also your approach would require to scan at least once for line breaks. That is only more efficient when the current way is repeatedly reading the same chars (e.g. because of backtracking).

Given that the Unicode line breaks don't appear often (I personally have never seen of them, except some I created myself), we should reject this patch.

sharwell:
@mike-lischke As I mentioned, the alternative method of tracking line numbers is a *memory* optimization. In code where millions of token instances are created and the average number of tokens per line of the input is greater than 1, tracking line numbers separately from the token instances would reduce overall memory pressure during parsing.

parrt:
ok, i'm gonna close.

jkra:
oh... ok ...good, no problem... but:
it is propably more or less a gut-feeling-decicion ("Bauchgefühl" as we say in german ;)). I don't think that 4 more ifs are a big deal. Maybe you have other profiling tools, but with this naive approach (based on http://stackoverflow.com/questions/6805026/is-switch-faster-than-if):

```
#include <iostream>
#include <vector>

#define MAX_COUNT (1 << 29)
size_t line = 0;

size_t testIf_1(const std::vector<char>& charVector)
{
  clock_t start = clock();
  for (size_t i = 0; i < charVector.size(); ++i)
  {
    const size_t curChar = charVector.at(i);
    if ((curChar == 0x000C)) {
      ++ line;
    }
  }
  clock_t end = clock();
  size_t result = 1000 * (end - start) / CLOCKS_PER_SEC;

  return result;
}

size_t testIf_5(const std::vector<char>& charVector)
{
  clock_t start = clock();
  for (size_t i = 0; i < charVector.size(); ++i)
  {
    const size_t curChar = charVector.at(i);
    if ((curChar == '\n')
        || (curChar == 0x2028 /*'\u2028'*/)
        || (curChar == 0x2029 /*'\u2029'*/)
        || (curChar == 0x0085 /*'\u0085'*/)
        || (curChar == 0x000C /*'\u000C'*/)) {
      ++ line;
    }
  }
  clock_t end = clock();
  size_t result = 1000 * (end - start) / CLOCKS_PER_SEC;

  return result;
}

int main()
{
  std::cout<<"Starting with "<<MAX_COUNT<<" numbers ..."<<std::endl;

  std::vector<char> charVector(MAX_COUNT, 0x000C);

  line = 0;
  size_t testIf_1_ms = testIf_1(charVector);
  std::cout<<"If_1 statement: "<<testIf_1_ms<<" ms with lines: "<<line<<std::endl;

  line = 0;
  size_t testIf_5_ms = testIf_5(charVector);
  std::cout<<"If_5 statement: "<<testIf_5_ms<<" ms with lines: "<<line<<std::endl;

  return 0;
}
```

I don't see any differences. If we had a lot of '||' -> yes we would have a performance problem ... but here ... propably not...

and yes, I have these use cases with my text editors using unicode... the solutions of sam are too complicated for me, as an antlr user...

but anyway and as mentioned -> I'm fine with this decision and thanks for antlr4 -> and especially for the cpp-runtime... it is such a really great tool... :)

p.s.
maybe changing of '_line++' -> '++_line' (and similiar) could be done ... for performance reasons ;)

mike-lischke:
@jkra My recommendation is based mostly on the thought that such a change should be done in all runtimes, not just C++ (but thanks for actually proving that we don't need to be concerned about a potential performance hit).  I'm nagging @parrt already for a while about full Unicode support :-D

jkra:
@mike-lischke ok... maybe in one of the next releases...

PR #1576, opened 2017-01-04
Implement support for optional getters

sharwell:
This analysis is required for proper code generation in the TypeScript target when strict null checks are enabled. It also applies to targets intending to differentiate optional values from required values.

### DIFF
--- a/tool/src/org/antlr/v4/codegen/model/ElementFrequenciesVisitor.java
+++ b/tool/src/org/antlr/v4/codegen/model/ElementFrequenciesVisitor.java
@@ -21,12 +21,30 @@ import java.util.Deque;
 import java.util.Map;
 
 public class ElementFrequenciesVisitor extends GrammarTreeVisitor {
+	/**
+	 * This special value means "no set", and is used by {@link #minFrequencies}
+	 * to ensure that {@link #combineMin} doesn't merge an empty set (all zeros)
+	 * with the results of the first alternative.
+	 */
+	private static final FrequencySet<String> SENTINEL = new FrequencySet<String>();
+
 	final Deque<FrequencySet<String>> frequencies;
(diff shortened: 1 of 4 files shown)


parrt:
Looks interesting. do we have a few examples that exercise this stuff? i guess we have one issue you just resolved, right?

sharwell:
@parrt I used this to fix tunnelvisionlabs/antlr4ts#265. That issue includes an example which highlights the difference in the generated code produced by this target.

I also mentioned optional values. In a hypothetical modified Java target, one could use `Optional<TContext>` instead of `TContext` for the return value of optional accessors. In such a target, all generated accessors would have the property that the return value is never null. Perhaps it's not a great idea for performance reasons, but I think it serves as an interesting example.

sharwell:
@parrt The Swift target may also benefit from this change, but I don't know enough about its handling of nullable types to say for sure.

tool/src/org/antlr/v4/codegen/model/ElementFrequenciesVisitor.java
```diff
@@ -134,6 +183,11 @@ protected void exitSubrule(GrammarAST tree) {
 			for (Map.Entry<String, MutableInt> entry : frequencies.peek().entrySet()) {
 				entry.getValue().v = 2;
 			}
+
+			int multiplier = tree.getType() == POSITIVE_CLOSURE ? 1 : 0;
```
parrt: can we change this to a boolean rather than a multiplier? ;)  Ain't that fortran66 style hahah

sharwell: I simplified this even more :smile:

parrt:
@janyou @hanjoes might be interesting to Swift target.

PR #1581, opened 2017-01-06
Remove the "shared" visitor tests

sharwell:
These tests were modeled after the listener tests, but were not implemented against the standard visitor API semantics. The tests only execute in the Node.js target, and there is no straightforward way to implement these tests across multiple targets. New tests which cover both the traversal behavior and the runtime semantics are now available and should be implemented by each target.


parrt:
ok, then we can add back for javascript or was it python target that had them?

sharwell:
@parrt The tests were for the JavaScript target, but instead of adding the same tests the code should use tests that cover the API, starting with mirroring the ones I added for Java.

notsonotso:
I hope this gets fixed... wasted a lot of time on visitors, and now it turns out the impl is bogus.

Issue #1539, opened 2016-12-21
go target: copyright notices are interpreted as package documentation


willfaught:
See https://godoc.org/github.com/antlr/antlr4/runtime/Go/antlr.

The copyright notices appear right above the package declarations. For example:

```go
/* Copyright (c) 2012-2016 The ANTLR Project. All rights reserved.
 * Use of this file is governed by the BSD 3-clause license that
 * can be found in the LICENSE.txt file in the project root.
 */
package antlr
```

This results in all copyright notices in all the files being concatenated together for the package documentation.

A simple fix is to add a blank line between the comment and the declaration to remove the association between the two.

Note that that comment format isn't idiomatic Go. Multi-line comments are typically only used for long package documentation. Something like this would read better for Go:

```go
// Copyright (c) 2012-2016 The ANTLR Project. All rights reserved.
// Use of this file is governed by the BSD 3-clause license that
// can be found in the LICENSE.txt file in the project root.

package antlr
```

Or, since there are no conventional line length limits in Go, even:

```go
// Copyright (c) 2012-2016 The ANTLR Project. All rights reserved.
// Use of this file is governed by the BSD 3-clause license that can be found in the LICENSE.txt file in the project root.

package antlr
```

Also note that there's currently no link for Go runtime documentation on antlr.org. The link above is what you'd want for that.

Issue #1361, opened 2016-11-16
Split Go runtime into its own repository


millergarym:
Go has its own build system. A separate repository for the Go runtime would make for a better development experience. Currently doing a  `go get github.com/antlr/antlr4` downloads the entire Antlr4 repository. The only files needed are the Go runtime source. Additionally, it would be nice to have a copy of the jar and an example is this repository.

I have created an example repository at [https://github.com/millergarym/antlr4-go](https://github.com/millergarym/antlr4-go).

Cheers

parrt:
Hi. I agree in general but for antlr4 I'd like to keep all runtimes in the same repo so I'm forced to keep everything in sync. If I change the tool, all my tests run automatically.

millergarym:
Hi Terence,
There are a couple of ways that I can think of that would be achieve both goal.

The first is what I have done with the proof of concept repo, which is to manually update an antlr4-go repo.
The second would be for the build to update target specific repos.
The third is for the antlr4-go directory to be a subtree of so other repository.

I think subtrees and submodules are more problems than they are worth, so wouldn't really suggest this option. 

There is no short term issue with the first, but in the long term is will get out of sync and github.com/antlr/antlr4-go is better option for a repo.

Having the build system update target specific repo is probably the best option.
This could be done for each successful build or on releases.
It looks like the release process would need a bit of work to accommodate this.

Happy to help but probably not with the build process, I am no maven maven.

Separately, is there any documentation on the runtime test "architecture".

Cheers
Gary

parrt:
@millergarym yeah, still, having build update a target etc... is complicated.  Downloading extra cruft with `go get` is invisible and doesn't require additional machinery :)

millergarym:
@parrt except there isn't a built jar in `go get` download. 

Does the current build process (travis etc.) leave an accessible jar?
If so please point me at it.
Because of this I included antlr4-4.5.4-SNAPSHOT.jar in my Go gettable repo.

To get from the current repo to a jar requires a jdk, maven and a reasonable amount of knowledge. This is quite a bit of additional machinery for non-Java developers.

parrt:
@millergarym Doesn't `go get` return just the go stuff you need? Why do you need the jar?

millergarym:
To run the antlr tool and generate Go source from the .g4 files.

PR #1583, opened 2017-01-07
Use single-line comments for copyright

willfaught:
Multi-line comments are typically only used for long package documentation in Go. This changes the copyright file header multi-line comment to multiple single-line comments to be idiomatic. All the Go runtime files, and only them.

### DIFF
--- a/runtime/Go/antlr/atn.go
+++ b/runtime/Go/antlr/atn.go
@@ -1,7 +1,6 @@
-/* Copyright (c) 2012-2016 The ANTLR Project. All rights reserved.
- * Use of this file is governed by the BSD 3-clause license that
- * can be found in the LICENSE.txt file in the project root.
- */
+// Copyright (c) 2012-2016 The ANTLR Project. All rights reserved.
+// Use of this file is governed by the BSD 3-clause license that
+// can be found in the LICENSE.txt file in the project root.
 
 package antlr
 
--- a/runtime/Go/antlr/atn_config.go
+++ b/runtime/Go/antlr/atn_config.go
@@ -1,7 +1,6 @@
-/* Copyright (c) 2012-2016 The ANTLR Project. All rights reserved.
- * Use of this file is governed by the BSD 3-clause license that
- * can be found in the LICENSE.txt file in the project root.
- */
+// Copyright (c) 2012-2016 The ANTLR Project. All rights reserved.
+// Use of this file is governed by the BSD 3-clause license that
+// can be found in the LICENSE.txt file in the project root.
 
 package antlr
 
--- a/runtime/Go/antlr/atn_config_set.go
+++ b/runtime/Go/antlr/atn_config_set.go
@@ -1,7 +1,6 @@
-/* Copyright (c) 2012-2016 The ANTLR Project. All rights reserved.
- * Use of this file is governed by the BSD 3-clause license that
- * can be found in the LICENSE.txt file in the project root.
- */
+// Copyright (c) 2012-2016 The ANTLR Project. All rights reserved.
+// Use of this file is governed by the BSD 3-clause license that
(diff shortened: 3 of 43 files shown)


sharwell:
I'm :-1: on this for a few reasons:

1. The current headers are no longer causing a problem (#1539)
1. The script @parrt uses to verify headers across the repository probably doesn't handle these comments in its current form (it could likely be updated, but this is just one issue of several)
1. Every change to the header comments causes a massive, and at this point totally unnecessary, merge conflict for me across several external projects which I keep in sync with this repository

This repository is already not an "idiomatic Go project" (#1361) so I'm not seeing the benefit here.

:memo: Normally I would actually be in favor of a change like this, but having just recently spent several hours dealing with the aforementioned merge conflicts, every proposed change to the headers now is a reminder of that.

parrt:
@willfaught yeah, my copyright script will have trouble with this I think. Thanks for expending the effort on this though.  @sharwell  out of curiosity, why would Go target files cause a conflict in your fork? If you are keeping that target, couldn't you just pull in those changes as they are the top of the file? why would it conflict with anything?

sharwell:
> why would Go target files cause a conflict in your fork?

Currently there is no implementation of the Go target for my optimized release. Any change to the Go or another runtime causes a conflict. It's unavoidable (and totally not a problem) for general maintenance of targets, and in most cases required changes are quite scope-limited (affect few files).

willfaught:
@parrt Doesn't your script handle # Python comments? Can't it be changed to
handle // comments similarly?

@samharwell That issue is only about having a shorter, more convenient
import path and a smaller runtime download; it doesn't have anything to do
with Go idioms. All Go code should be idiomatic. I'll be making a lot of
minor changes to the Go runtime soon beyond just the copyright headers. For
example, the code hasn't even been fully formatted with gofmt yet.

In my opinion, external projects shouldn't influence this project's
development, at least for compatible changes. If no runtime changes can be
made, then how can even bugs be fixed? If Go development was frozen upon
being merged from pboyer/antlr4, then it shouldn't have been merged yet and
we should back it out. There is still much polish work to be done.

I sympathize with having to deal with merge conflicts manually, but won't
these conflicts be merged automatically?

On Sat, Jan 7, 2017 at 10:09 AM Sam Harwell <notifications@github.com>
wrote:



why would Go target files cause a conflict in your fork?




Currently there is no implementation of the Go target for my optimized
release. Any change to the Go or another runtime causes a conflict. It's
unavoidable (and totally not a problem) for general maintenance of targets,
and in most cases required changes are quite scope-limited (affect few
files).




—
You are receiving this because you were mentioned.
Reply to this email directly, view it on GitHub
<https://github.com/antlr/antlr4/pull/1583#issuecomment-271099804>, or mute
the thread

ericvergnaud:
I tend to agree with Will on this one.
In short, target authors should have the final word on their code, otherwise the loss of control is likely to lead to the loss of authors 
Eric
Envoyé de mon iPhone

sharwell:
@willfaught @ericvergnaud Note that my comments *only* hold for the copyright headers. They have no bearing whatsoever on anything that appears below that.

willfaught:
@sharwell Gotcha.

Wouldn't changes below the copyright stuff also cause the merge conflicts you mentioned?

sharwell:
@willfaught sure but there I'm with what Eric said.

I'm actually the one who first pushed for moving to shorter headers (I've moved to a 2 line header in most of my own projects). I also agree with changing things that are causing a problem. But here it seems like a minor style detail that doesn't really impact anything, and other developers reading the code are unlikely to be surprised to see the headers looking consistent through the repository.

parrt:
I'm approving this PR. Still not sure how it could affect Sam's stuff but whatever.

PR #1585, opened 2017-01-07
The final fix (hopefully) for alternative labels check in left recursive rules

### LINKED ISSUE #1543: type mismatch between left and right labels in left-recursive rule
We thought we fixed this in https://github.com/antlr/antlr4/issues/195#issuecomment-268737026 but here's another case reported:

@ferhaterata says: provided a Grammar that ANTLR generates recognizer in 4.5.1 and 4.5.3 but fails in 4.6.1. The labels 'left' and 'right' represent two different types. Sorry for preparing a verbose grammar.

```
grammar Test;

module: formula* ;

formula:
      left=expression 'in'  right=expression    #in
    | left=formula    'and' right=formula       #and
    ;

expression:
      left=expression '+' right=expression      #union
    | variable                                  #var
    ;

variable: IDENTIFIER;

IDENTIFIER : [a-zA-Z];
``` 
It throws

> error(75): Test.g4:2:13: label left=expression type mismatch with previous definition: left=formula
> error(75): Test.g4:5:57: label right=formula type mismatch with previous definition: right=expression

In fact, this one works fine,

```
grammar test1;

example:
    left=integer '*' right=integer  #integerExample
  | left=string  '+' right=string   #stringExample
  ;

integer: IDENTIFIER;
string: INTEGER;

STRING: [a-zA-Z];
INTEGER: [a-zA-Z];
```

I assume the problem occurs once repetitions are allowed in labels.

KvanTTT:
fix #1543

### DIFF
--- a/tool-testsuite/test/org/antlr/v4/test/tool/TestSymbolIssues.java
+++ b/tool-testsuite/test/org/antlr/v4/test/tool/TestSymbolIssues.java
@@ -341,6 +341,48 @@ public class TestSymbolIssues extends BaseJavaToolTest {
 		testErrors(test, false);
 	}
 
+	// https://github.com/antlr/antlr4/issues/1543
+	@Test public void testLabelsForTokensWithMixedTypesLRWithLabels() {
+		String[] test = {
+				"grammar L;\n" +
+				"\n" +
+				"expr\n" +
+				"    : left=A '+' right=A        #primary\n" +
+				"    | left=expr '-' right=expr  #sub\n" +
+				"    ;\n" +
+				"\n" +
+				"A: 'a';\n" +
+				"B: 'b';\n" +
+				"C: 'c';\n",
+
+				""
+		};
+
+		testErrors(test, false);
+	}
+
+	// https://github.com/antlr/antlr4/issues/1543
+	@Test
+	public void testLabelsForTokensWithMixedTypesLRWithoutLabels() {
+		String[] test = {
+				"grammar L;\n" +
+				"\n" +
+				"expr\n" +
+				"    : left=A '+' right=A\n" +
+				"    | left=expr '-' right=expr\n" +
+				"    ;\n" +
+				"\n" +
+				"A: 'a';\n" +
+				"B: 'b';\n" +
+				"C: 'c';\n",
+
+				"error(" + ErrorType.LABEL_TYPE_CONFLICT.code + "): L.g4:3:0: label left type mismatch with previous definition: TOKEN_LABEL!=RULE_LABEL\n" +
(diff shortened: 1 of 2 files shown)


KvanTTT:
@sharwell, I removed your tests and expanded my test. Please review.

tool-testsuite/test/org/antlr/v4/test/tool/TestSymbolIssues.java
```diff
@@ -315,51 +330,12 @@ public void testSetUp() throws Exception {
...
 				"C: 'c';\n",
 
-				"error(" + ErrorType.LABEL_TYPE_CONFLICT.code + "): L.g4:8:15: label t2=b type mismatch with previous definition: t2=a\n" +
-				"error(" + ErrorType.LABEL_TYPE_CONFLICT.code + "): L.g4:11:17: label t3+=c type mismatch with previous definition: t3+=a\n"
-		};
-
-		testErrors(test, false);
-	}
-
-	// https://github.com/antlr/antlr4/issues/1543
```
sharwell: :bulb: I would prefer to keep this and the following as distinct tests since they target a specific issue.

parrt:
@sharwell you ok with changes now for the label mismatch?

tool/src/org/antlr/v4/semantics/SymbolChecks.java
```diff
@@ -128,54 +131,102 @@ public void checkForTokenConflicts(List<GrammarAST> tokenIDRefs) {
...
+		else {
+			return findAltLabelName(label.parent);
+		}
+	}
+
+	private void checkForTypeMismatch(Rule r, LabelElementPair prevLabelPair, LabelElementPair labelPair) {
 		// label already defined; if same type, no problem
 		if (prevLabelPair.type != labelPair.type) {
-			String typeMismatchExpr = labelPair.type + "!=" + prevLabelPair.type;
+			// TODO: replace rule token for left-recursive rule with correct token.
```
sharwell: ❓ What will this fix?

sharwell: 💭 It appears this is referring to the fact that the warning is typically reported on the actual label redefinition, but for left-recursive rules the warning is reported on the enclosing rule. My recommendation is to:

1. Remove this TODO line
2. File an issue describing...
  * The current behavior
  * The desired behavior
  * A mention to this pull request (where the current behavior was introduced)
  * A mention to the `Token` selection strategy as the source of the problem.

sharwell:
@parrt The tests seemed fairly comprehensive, so this is likely working now. I did have the one remaining question/suggestion for the TODO line.

tool/src/org/antlr/v4/semantics/SymbolChecks.java
KvanTTT: I guess this is related to the left recursive rules transformation. Transformed rules have other token positions. Unfortunately I have no idea how to fix it easily.

PR #1586, opened 2017-01-07
Reset BufferedTokenStream.fetchedEOF when calling setTokenSource

### LINKED ISSUE #1584: BufferedTokenStream cannot be reused after EOF
Currently `BufferedTokenStream` cannot be reused after reaching EOF. The `fetchedEOF` flag is set to true when the first EOF is encountered, and `setTokenSource` does not reset it. This prevents the stream from ever attempting to fetch more tokens from the source, leaving it as empty.

Credit for finding this issue goes to @JanCervak, who reported it in tunnelvisionlabs/antlr4cs#136.

sharwell:
Fixes #1584

@antlr/antlr-targets This bug may affect other targets.


parrt:
@antlr/antlr-targets Please note a small change to fix a bug in your targets.

mike-lischke:
I already added that fix to the C++ runtime while working on it.

hanjoes:
fixed in PR #1606

ericvergnaud:
fixed in PR #1616

PR #1591, opened 2017-01-08
Channel names & constants in lexer, improved modes record format.

### LINKED ISSUE #1554: Ability to get a channel name from integer id in lexer
At present time we can get a token name via `Vocabulary` interface. Also we can get a mode name from `modeNames` array in Lexer. But it's impossible to get a channel name from lexer.

KvanTTT:
Added channelNames to lexer for all runtimes (C++, C#, Go, Java, JavaScript, Python2, Python3).
Added missed channel constants to lexer for Python 2, Python 3, JavaScript, Go runtimes.
Mode record format now is the same as tokens and channels mode (comma-separated constants).

Fixes #1554.


KvanTTT:
Sorry, I forgot to update some files. Will fix them.

KvanTTT:
Can we remove generated files for XPath and Hello grammars?

sharwell:
> Can we remove generated files for XPath and Hello grammars?

💭 As far as I can tell they were only included by mistake

runtime/Cpp/runtime/src/tree/xpath/XPathLexer.cpp
```diff
@@ -46,14 +47,14 @@ const atn::ATN& XPathLexer::getATN() const {
...
-    case 4: IDAction(dynamic_cast<RuleContext *>(context), actionIndex); break;
+    case 4: IDAction(dynamic_cast<antlr4::RuleContext *>(context), actionIndex); break;
 
   default:
     break;
   }
 }
 
-void XPathLexer::IDAction(RuleContext * /*context*/, size_t actionIndex) {
+void XPathLexer::IDAction(antlr4::RuleContext *context, size_t actionIndex) {
```
mike-lischke: Are these changes manual or generated? The outcommented context name is necessary to silence a compiler warning (unused parameter), so such things shouldn't be changed lightly.

KvanTTT: These files generated with the 4.6 version of ANTLR. On my opinion, ALL generated files should be excluded from the repository.

parrt:
@KvanTTT @mike-lischke I think I included XPath as there is a chicken-egg issue whereby we need to build antlr 4.6.1 with that file but that file should be 4.6.1 generated. Hmm...what are the issues if we remove it?

KvanTTT:
I think you can merge this request now and remove/update XPath autogenerated files later.

parrt:
@KvanTTT Can you rebase on master? Got some conflicts. thanks!

KvanTTT:
Certainly!

runtime/Java/src/org/antlr/v4/runtime/LexerInterpreter.java
```diff
@@ -31,11 +32,11 @@
...
 
 	@Deprecated
-	public LexerInterpreter(String grammarFileName, Collection<String> tokenNames, Collection<String> ruleNames, Collection<String> modeNames, ATN atn, CharStream input) {
-		this(grammarFileName, VocabularyImpl.fromTokenNames(tokenNames.toArray(new String[tokenNames.size()])), ruleNames, modeNames, atn, input);
+	public LexerInterpreter(String grammarFileName, Collection<String> tokenNames, Collection<String> ruleNames, Collection<String> channelNames, Collection<String> modeNames, ATN atn, CharStream input) {
+		this(grammarFileName, VocabularyImpl.fromTokenNames(tokenNames.toArray(new String[tokenNames.size()])), ruleNames, channelNames, modeNames, atn, input);
 	}
 
-	public LexerInterpreter(String grammarFileName, Vocabulary vocabulary, Collection<String> ruleNames, Collection<String> modeNames, ATN atn, CharStream input) {
+	public LexerInterpreter(String grammarFileName, Vocabulary vocabulary, Collection<String> ruleNames, Collection<String> channelNames, Collection<String> modeNames, ATN atn, CharStream input) {
```
parrt: won't this be a breaking change? Can we add this ctor and leave the old one to call this one with null for column names?

tool/resources/org/antlr/v4/tool/templates/codegen/Java/Java.stg
```diff
@@ -898,9 +898,16 @@ public class <lexer.name> extends <superClass; null="Lexer"> {
 		<lexer.tokens:{k | <k>=<lexer.tokens.(k)>}; separator=", ", wrap, anchor>;
 	<if(lexer.channels)>
 	public static final int
-		<lexer.channels:{k | <k>=<lexer.channels.(k)>}; separator=", ", wrap, anchor>;
+		<lexer.channels:{c | <c>=<lexer.channels.(c)>}; separator=", ", wrap, anchor>;
 	<endif>
-	<rest(lexer.modes):{m| public static final int <m> = <i>;}; separator="\n">
+	<if(rest(lexer.modes))>
+	public static final int
+		<rest(lexer.modes):{m | <m>=<i>}; separator=", ", wrap, anchor>;
```
parrt: is this the same as the one-liner?

runtime/Java/src/org/antlr/v4/runtime/LexerInterpreter.java
KvanTTT: I can add a new ctor, but it seems to me not very clear.

KvanTTT:
> I'm just worried about backward compatibility. Anyone using LexerInterpreter would get a compile error in 4.6.1, right?

Yes. But the fix is very easy.

parrt:
Yes, but stuff should be backward compatible as a general goal unless we have a good reason not to. :)

KvanTTT:
Ok, will fix it.

tool/resources/org/antlr/v4/tool/templates/codegen/Java/Java.stg
KvanTTT: I changed mode format from
```Java
public static final int MODE1 = 1;
public static final int MODE2 = 2;
```

to the tokens format:

```Java
public static final int
    MODE1 = 1, MODE2 = 2;
```

Channels also have the same format. This is more convenient. Are you agree with me?

parrt: agreed!

KvanTTT:
I added `Deprecated` attribute to Java ctor and `Obsolete` attribute to C# ctor. I do not change C++ and Swift runtimes because of they are relative new runtimes.

PR #1594, opened 2017-01-10
Remove lower case formatting on Go types & super import

wxio:
Since case formatting in Go denotes private verses public the format={lower} is getting in the way.
Currently the only workaround to place the generated code in the same package as the types being used, and these types can't be exported. This is inconvenient and unnecessary. I see know downside to leave the case on the type as set in the grammar.

The `import "./<superClass>"` also get in the way. I tried using superClass to add fields to a parser. This works nicely, but the import needed to be removed by hand. Relative imports are of very limited use in Go. Also in the grammar developer needs an import for the superClass it is trivial to place the import statement in the `@members` block.

### DIFF
--- a/runtime-testsuite/resources/org/antlr/v4/test/runtime/templates/Go.test.stg
+++ b/runtime-testsuite/resources/org/antlr/v4/test/runtime/templates/Go.test.stg
@@ -325,5 +325,5 @@ func pred(v bool) bool {
 
 Invoke_pred(v) ::= <<pred(<v>)>>
 ContextRuleFunction(ctx, rule) ::= "<ctx>.<rule>"
-StringType() ::= "String"
+StringType() ::= "string"
 ContextMember(ctx, subctx, member) ::= "<ctx>.<subctx>.<member; format={cap}>"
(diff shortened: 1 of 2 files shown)


willfaught:
I see "String undefined" test failures.

It looks like this will result in uppercase params in New? Can you change those back to lowercase?

sharwell:
The test failures may be caused by an incorrect definition of `StringType` in **Go.test.stg**. The default from Java is `String`, but if the equivalent type in Go is `string` you'll need to modify that template. The mistake would have been hidden previously by the incorrect application of `format="lower"` when rendering attribute types.

willfaught:
Ping @pboyer

willfaught:
Ignore my question about the New params.

wjkohnen:
The tests do pass with sharwell's suggested fix.

wxio:
@willfaught @sharwell @wjkohnen I just wanted to thank you all for your help.
Cheers
Gry

willfaught:
@wxio Looks like the first commit is a merge to your local branch? Can you rebase only these changes on top of upstream?

wxio:
@willfaught rebase done (I think)

willfaught:
I still see the merge commit. It's the first one: https://github.com/antlr/antlr4/pull/1594/commits

wxio:
@willfaught can you give me some help in doing this? Sorry I'm not a git guru, by any means.

willfaught:
You could create a new branch based on origin/master, then `git cherry-pick HASH` for the three commits you'll want to include.

wxio:
I've done this, create a new branch called `typecase2`.
How do I get it back to being the master branch?
Do I then do a 
`
git checkout typecase2
git merge -s ours master
git checkout master
git merge typecase2
`

or 

`
git branch -m master old-master
git branch -m typecase2 master
git push -f origin master
`

willfaught:
I think the second one will work.

PR #1599, opened 2017-01-11
Fix missing variable declarations

robertbrignull:
This PR adds the `var` keyword to three variable declarations that were missing it and implicitly declaring global variables. Note that actually in all of the cases there's a similar function nearby that does use a local variable. Also none of these changes will change the semantics of the functions.

I noticed this because when in `'use strict'` mode all of these lines will cause errors saying the variable in question is not defined. This is a problem because webpack will by default apply strict mode and this is the method of distribution that users are recommended to use.


ericvergnaud:
Blessed

Envoyé de mon iPhone

PR #1606, opened 2017-01-16
Tweak Repo to use SwiftPackageManager.

hanjoes:
This PR is meant for SPM integration. This will make future runtime test/build easier/faster and has better compatibility on Linux platforms.

Also this PR contains a fix for #1584 w/ unit tests.


parrt:
@hanjoes howdy! looks like a big change is it really just moving files around?

hanjoes:
yep! The purpose is to integrate Swift Package Manager which works based off some conventions.

parrt:
@janyou any issues with this?

hanjoes:
btw, i tweaked the old xcodeproj to work with the new directory structure. So everything should work as before with added SPM integration.

PR #1612, opened 2017-01-18
Fix wrong include path of antlr4cpp runtime in cmake/ExternalAntlr4Cpp.cmake

mshockwave:
The include path variable exported by `cmake/ExternalAntlr4Cpp.cmake` doesn't follow up the latest update of the include folder hierarchy. Which may raise compile error on generated cpp/header files since they try to `#include "antlr4-runtime.h"` but the correct path is `#include "antlr4-runtime/antlr4-runtime.h"`.
I modify `ExternalAntlr4Cpp.cmake` so that we don't need to touch the generation of grammar cpp/header files.

### DIFF
--- a/contributors.txt
+++ b/contributors.txt
@@ -129,3 +129,4 @@ YYYY/MM/DD, github id, Full name, email
 2016/12/11, Gaulouis, Gaulouis, gaulouis.com@gmail.com
 2016/12/22, akosthekiss, Akos Kiss, akiss@inf.u-szeged.hu
 2016/12/24, adrpo, Adrian Pop, adrian.pop@liu.se
+2017/01/18, mshockwave, Bekket McClane, yihshyng223@gmail.com
--- a/runtime/Cpp/cmake/ExternalAntlr4Cpp.cmake
+++ b/runtime/Cpp/cmake/ExternalAntlr4Cpp.cmake
@@ -155,9 +155,9 @@ ExternalProject_ADD(
 
 ExternalProject_Get_Property(antlr4cpp INSTALL_DIR)
 
-list(APPEND ANTLR4CPP_INCLUDE_DIRS ${INSTALL_DIR}/include)
+list(APPEND ANTLR4CPP_INCLUDE_DIRS ${INSTALL_DIR}/include/antlr4-runtime)
 foreach(src_path misc atn dfa tree support)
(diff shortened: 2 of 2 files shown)


parrt:
@mike-lischke Is this correct?

JPGygax68:
What is the fix for those not using `ExternalAntlr4Cpp.cmake`? Is there no better way than to find the absolute path of the runtime headers and add it to the compiler's include path?

mike-lischke:
That `ExternalAntlr4Cpp.cmake` is just an additional make file for people who prefer to use this approach. There's another CmakeLists.txt file for building the runtime in a more straight forward way, which should work as is.

JPGygax68:
Thank you @mike-lischke. My issue however is not with building the runtime, but with the include statement that antlr4 produces:
```
// Generated from Scene.g4 by ANTLR 4.7.1
#pragma once
#include "antlr4-runtime.h"
...
```
The file `antlr4-runtime.h` in turn includes a lot of header files, some of which having names that are at high risk of clashing with include files from other libraries (e.g. `Token.h`). 

Is there any way to make antlr4 generate `#include <antlr4-runtime/antlr4-runtime.h>` instead?

mike-lischke:
That include is generated via templates - a pretty complicated setup. Instead of forcing a specific include path there (and possibly produce problems in other environments) it's better to control the inclusion behavior with project header search paths (and put all the parsing code into an own library). That worked well for me even in cases where headers clash (e.g. Python also has a Token.h file IIRC).

JPGygax68:
Thanks! This is what I ended up doing, and indeed it works ok. The preprocessor inclusion rules will make sure that `antlr4-runtime.h` will find its own sub-headers first.

Still, I was wondering why not more people were complaining about this, and found the answer when I got the usual C++ linker errors: because this is C++ and not C, the libraries must match perfectly (release/debug, static or dll runtime - under Windows), so you have to find and set the specific library path anyway.

Ralith:
It's very confusing for the runtime to install the headers in a place the generated source files won't find. @JPGygax68's suggestion to use `#include <antlr4-runtime/antlr4-runtime.h>` instead makes a great deal of sense to me. It's not forcing a specific include path; it's referencing the location relative to global include paths that matches the one installed to by default. Surely if any environment should work out of the box, it should be the one produced by installing the runtime normally?

PR #1616, opened 2017-01-19
fix #1584 for C#, JavaScript and Python targets

### LINKED ISSUE #1584: BufferedTokenStream cannot be reused after EOF

ericvergnaud:

### DIFF
--- a/runtime/CSharp/runtime/CSharp/Antlr4.Runtime/BufferedTokenStream.cs
+++ b/runtime/CSharp/runtime/CSharp/Antlr4.Runtime/BufferedTokenStream.cs
@@ -371,6 +371,7 @@ namespace Antlr4.Runtime
             this._tokenSource = tokenSource;
             tokens.Clear();
             p = -1;
+			this.fetchedEOF = false;
         }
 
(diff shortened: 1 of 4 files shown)


runtime/CSharp/runtime/CSharp/Antlr4.Runtime/BufferedTokenStream.cs
```diff
@@ -371,6 +371,7 @@ public virtual void SetTokenSource(ITokenSource tokenSource)
             this._tokenSource = tokenSource;
             tokens.Clear();
             p = -1;
+			this.fetchedEOF = false;
```
KvanTTT: Please use spaces instead of tabs here.

PR #1623, opened 2017-01-23
Use tabs for code formatting in Emacs (except for Python)

bhamiltoncx:
Most of the ANTLR code base (except for Python) uses tabs for indentation, so this sets up Emacs `.dir-locals.el` files to ensure contributors use the correct style for each directory.

### DIFF
--- a/.dir-locals.el
+++ b/.dir-locals.el
@@ -1,0 +1,6 @@
+; Basic ANTLR code style for Emacs: Indents are tabs and represent 4 columns
+(
+ (nil . ((indent-tabs-mode . t)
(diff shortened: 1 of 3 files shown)


sharwell:
I'm going to vote :-1: in favor of continuing to use the EditorConfig open standard, for which [at least one](https://github.com/editorconfig/editorconfig-emacs) plug-in is available for Emacs.

bhamiltoncx:
Fair enough! Seems worth adding a mention to EditorConfig to the ANTLR development docs. I'll abandon this in favor of that.

PR #1624, opened 2017-01-24
New doc 'Contributing to ANTLR'

bhamiltoncx:
In https://github.com/antlr/antlr4/pull/1623#issuecomment-274801628 , @sharwell mentioned using `EditorConfig` is preferred to set up editors and IDEs for ANTLR's code base, but it wasn't mentioned in any of the docs I saw.

This adds a basic "Contributing to ANTLR" doc which mentions setting up `EditorConfig` before contributing.


doc/contributing.md
```diff
@@ -0,0 +1,10 @@
+# Contributing to ANTLR 4
```
sharwell: :question: Do we want to make this a **CONTRIBUTING.md** file in the root of the repository? GitHub recognizes that as a special file that's presented to the user in some parts of the UI.

(I really mean this as a question and not action item since it's subjective/preference either way)

bhamiltoncx: Sure! Done.

Issue #276, opened 2013-06-04
Support UTF-32 escape sequences


jbclements:
The pasted file contains a definition for the unicode XID_START character set. Processing it with the current git master causes this error:

```
jclements-09740:/tmp clements> java -jar /usr/local/lib/antlr-4.0.1-dev-complete.jar ./Bad.g4
Exception in thread "main" java.lang.UnsupportedOperationException: Serialized ATN data element out of range.
    at org.antlr.v4.automata.ATNSerializer.serialize(ATNSerializer.java:324)
    at org.antlr.v4.automata.ATNSerializer.getSerialized(ATNSerializer.java:458)
    at org.antlr.v4.codegen.model.SerializedATN.<init>(SerializedATN.java:46)
    at org.antlr.v4.codegen.model.Lexer.<init>(Lexer.java:72)
    at org.antlr.v4.codegen.OutputModelController.lexer(OutputModelController.java:175)
    at org.antlr.v4.codegen.OutputModelController.buildLexerOutputModel(OutputModelController.java:128)
    at org.antlr.v4.codegen.CodeGenerator.generateLexer(CodeGenerator.java:147)
    at org.antlr.v4.codegen.CodeGenPipeline.process(CodeGenPipeline.java:69)
    at org.antlr.v4.Tool.processNonCombinedGrammar(Tool.java:422)
    at org.antlr.v4.Tool.process(Tool.java:376)
    at org.antlr.v4.Tool.processGrammarsOnCommandLine(Tool.java:343)
    at org.antlr.v4.Tool.main(Tool.java:190)
```

... and, here's the file Bad.g4. It's 430 lines, just in case my copy buffer lost part of it:

```
grammar Bad;


XIDSTART :
              '\u0041' .. '\u005a'
            | '\u0061' .. '\u007a'
            | '\u00aa'
            | '\u00b5'
            | '\u00ba'
            | '\u00c0' .. '\u00d6'
            | '\u00d8' .. '\u00f6'
            | '\u00f8' .. '\u01ba'
            | '\u01bb'
            | '\u01bc' .. '\u01bf'
            | '\u01c0' .. '\u01c3'
            | '\u01c4' .. '\u0293'
            | '\u0294'
            | '\u0295' .. '\u02af'
            | '\u02b0' .. '\u02c1'
            | '\u02c6' .. '\u02d1'
            | '\u0ab5' .. '\u0ab9'
            | '\u0abd'
            | '\u0ad0'
[...]

PR #1625, opened 2017-01-24
New method IntegerList.toCharArray()

bhamiltoncx:
While working on #276 , I found that `Utils.toCharArray(IntegerList)` was silently corrupting `IntegerList`s containing values larger than `0xFFFF`.

The root cause was a cast from `int` to `char`, which silently truncates the (possibly 32-bit input) to 16 bits.

Currently, this doesn't cause issues, because there is no way to add an integer larger than 0xFFFF to an `IntegerList`, but fixing #276 will do this for the first time (serializing a grammar with Unicode literals > `0xFFFF` to an ATN needs to encode `IntegerList`s with values > `0xFFFF`).

In this diff, I fix the issue by:

1. Adding a new method `IntegerList.toCharArray()` and making `Utils.toCharArray(IntegerList)` call that
2. Treating the contents of the `IntegerList` as a series of Unicode code points when serializing to a `char` array
3. Encoding each Unicode code point as one or two UTF-16 code units and appending them to the `char` array

I added new unit tests to confirm the desired behavior.

### DIFF
--- a/runtime-testsuite/test/org/antlr/v4/test/runtime/java/TestIntegerList.java
+++ b/runtime-testsuite/test/org/antlr/v4/test/runtime/java/TestIntegerList.java
@@ -1,0 +1,70 @@
+/*
+ * Copyright (c) 2012-2017 The ANTLR Project. All rights reserved.
+ * Use of this file is governed by the BSD 3-clause license that
+ * can be found in the LICENSE.txt file in the project root.
+ */
+package org.antlr.v4.test.runtime.java;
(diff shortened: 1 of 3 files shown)


parrt:
excellent! good catch.

parrt:
I added a comment on method. hope it's correct. https://github.com/antlr/antlr4/blob/master/runtime/Java/src/org/antlr/v4/runtime/misc/IntegerList.java#L275

PR #1627, opened 2017-01-24
New class UTF8CodePointDecoder

bhamiltoncx:
In order to use the new `CodePointCharStream` from #1626 without causing perf regressions, we need a way to efficiently convert UTF-8 input data (from disk or from network) to Unicode code points.

Java's built-in functionality to do this requires an intermediate conversion to UTF-16, which will be a lot of overhead (both memory and CPU).

This PR introduces a new class `UTF8CodePointDecoder`, modeled on Java's [CharsetDecoder](https://docs.oracle.com/javase/7/docs/api/java/nio/charset/CharsetDecoder.html), which decodes UTF-8 data in a `ByteBuffer` to Unicode code points stored in an `IntBuffer`.

To avoid lots of comparisons in a hot codepath, I used a lookup table to map the UTF-8 leading byte to the number of needed trail bytes.

(We could use the same lookup table for trailing bytes, but it made the code hard to read, and I figured optimizing the 7-bit US-ASCII path was the most important for ANTLR.)

I included lots of unit tests for success and error cases.

The idea is the resulting `IntBuffer` code points will be ready to use with `CodePointCharStream`. (This class allows us to build a streaming version of `CodePointCharStream` in the future if desired.)

This is part of the fix for #276.

### DIFF
--- a/runtime-testsuite/test/org/antlr/v4/test/runtime/java/TestUTF8CodePointDecoder.java
+++ b/runtime-testsuite/test/org/antlr/v4/test/runtime/java/TestUTF8CodePointDecoder.java
@@ -1,0 +1,162 @@
+/*
+ * Copyright (c) 2012-2017 The ANTLR Project. All rights reserved.
+ * Use of this file is governed by the BSD 3-clause license that
+ * can be found in the LICENSE.txt file in the project root.
+ */
+package org.antlr.v4.test.runtime.java;
+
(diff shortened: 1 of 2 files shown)


parrt:
Damn you do nice work, @bhamiltoncx !

bhamiltoncx:
Thanks very much! Can't wait to put this stuff into practice.

PR #1628, opened 2017-01-25
JavaScript runtime: Optionally support decoding UTF-16 sequences to Unicode code points

bhamiltoncx:
As part of #276, we need JavaScript support for Unicode code points.

JavaScript exposes this via the ECMAScript 6 method [`String.codePointAt()`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/codePointAt), but it appears ANTLR 4 is intended to support ECMAScript 5.1 browsers:

https://github.com/antlr/antlr4/blob/fed8c97bda1a80ce62828c25083a45e8070d7108/doc/javascript-target.md#L5

So, this PR adds an MIT-licensed polyfill for `String.codePointAt()` from https://github.com/mathiasbynens/String.prototype.codePointAt which I will use in a subsequent PR to implement Unicode code point support in the JavaScript runtime.

I updated `FileStream` and `InputStream` to optionally support decoding UTF-16 code unit sequences to Unicode code points. The API should remain compatible with existing clients, which will not be affected.

### DIFF
--- a/runtime/JavaScript/src/antlr4/InputStream.js
+++ b/runtime/JavaScript/src/antlr4/InputStream.js
@@ -10,25 +10,20 @@ require('./polyfills/codepointat');
 
 // Vacuum all input from a string and then treat it like a buffer.
 
-function _loadString(stream) {
+function _loadString(stream, decodeToUnicodeCodePoints) {
 	stream._index = 0;
 	stream.data = [];
-	var decodeDataAt;
-	if (stream.decodeToUnicodeCodePoints) {
-		decodeDataAt = function(i) {
+	if (decodeToUnicodeCodePoints) {
+		for (var i = 0; i < stream.strdata.length; ) {
 			var codePoint = stream.strdata.codePointAt(i);
 			stream.data.push(codePoint);
-			return codePoint <= 0xFFFF ? 1 : 2;
-		};
+			i += codePoint <= 0xFFFF ? 1 : 2;
+		}
 	} else {
-		decodeDataAt = function(i) {
+		for (var i = 0; i < stream.strdata.length; i++) {
 			var codeUnit = stream.strdata.charCodeAt(i);
 			stream.data.push(codeUnit);
-			return 1;
-		};
-        }
-	for (var i = 0; i < stream.strdata.length; ) {
-		i += decodeDataAt(i);
+		}
 	}
 	stream._size = stream.data.length;
 }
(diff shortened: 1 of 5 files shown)


ericvergnaud:
Hi Ben,

thanks, but it’s difficult to understand how this would integrate with the current implementation to enable support of unicode chars > 0xFFFF.
can you provide a usage sample?

Eric

bhamiltoncx:
Will do! You can see #276 for the work in progress; I'm splitting it up
here to make the PRs easy to review.

bhamiltoncx:
@ericvergnaud : Here's an example of how I intend to use it:

https://github.com/bhamiltoncx/antlr4/commit/ec4926be2cfd1f83dccb876305ff1b373683d64f

Once we have full Unicode support in JS `InputStream`, we can then update ANTLR's grammar to support Unicode literals > U+FFFF.

bhamiltoncx:
@ericvergnaud : I updated this PR to include both the polyfill and updates to `InputStream.js` and `FileStream.js` to optionally decode UTF-16 sequences to Unicode code points.

By default, existing API clients should not be affected, since the new parameters will be `undefined` and we will not decode UTF-16 to Unicode code points.

runtime/JavaScript/src/antlr4/InputStream.js
```diff
@@ -6,21 +6,43 @@
...
+// If decodeToUnicodeCodePoints is true, the input is treated
+// as a series of Unicode code points.
+//
+// Otherwise, the input is treated as a series of 16-bit UTF-16 code
+// units.
+function InputStream(data, decodeToUnicodeCodePoints) {
 	this.name = "<empty>";
 	this.strdata = data;
+	this.decodeToUnicodeCodePoints =
+		decodeToUnicodeCodePoints===undefined ? false : decodeToUnicodeCodePoints;
```
ericvergnaud: js style would be: decodeToUnicodeCodePoints || false

runtime/JavaScript/src/antlr4/InputStream.js
```diff
@@ -6,21 +6,43 @@
...
+		};
+	} else {
+		decodeDataAt = function(i) {
+			var codeUnit = stream.strdata.charCodeAt(i);
+			stream.data.push(codeUnit);
+			return 1;
+		};
+        }
+	for (var i = 0; i < stream.strdata.length; ) {
+		i += decodeDataAt(i);
```
ericvergnaud: given the simplicity of the 2 functions, and they are not reused, it might be faster to inline them

bhamiltoncx:
Will fix the style, thanks!

For the inline issue, I was trying to avoid an extra comparison on each
call. I can store the anonymous functions as a member if that's cleaner.

runtime/JavaScript/src/antlr4/InputStream.js
bhamiltoncx: Fixed, thanks.

runtime/JavaScript/src/antlr4/InputStream.js
bhamiltoncx: Inlined logic.

bhamiltoncx:
Thanks, made those fixes.

ericvergnaud:
blessed

parrt:
thanks guys!

PR #1629, opened 2017-01-25
Fix writing Unicode values in runtime tests

bhamiltoncx:
When writing tests for #276, I ran into an issue where Unicode values passed via `TestDescriptor`s randomly failed for some languages' runtimes, not not others.

When I investigated, I found many copy-pasted versions of the "write a Java `String` to disk" routine. Some of these explicitly converted to the String UTF-8, but many did not, which meant they didn't work if the system property `file.encoding` wasn't set to `utf-8` (and, let's be honest, nobody really sets this).

This PR addresses the issue by consolidating all runtime tests to use `BaseRuntimeTest.writeFile()`.


bhamiltoncx:
I don't know the best way to stack PRs on Github, but I sent out https://github.com/bhamiltoncx/antlr4/pull/1 as a dependent PR on this one.

PR #1630, opened 2017-01-26
Use UTF-8 for Python runtime tests, allow specifying error handling

bhamiltoncx:

When I went to write tests for the Python runtimes using Unicode values, I found the existing tests failed whenever the runtimes emitted non-ASCII values:

```
  TestSets>BaseRuntimeTest.testOne:119->BaseRuntimeTest.testParser:170 expected:<null> but was:<Traceback (m
ost recent call last):
  File "/tmp/BasePython3Test-pool-1-thread-219-1485454367240/Test.py", line 35, in <module>
    main(sys.argv)   
  File "/tmp/BasePython3Test-pool-1-thread-219-1485454367240/Test.py", line 26, in main
    input = FileStream(argv[1])
  File "/data/users/beng/src/antlr4/runtime-testsuite/target/classes/Python3/src/antlr4/FileStream.py", line
 20, in __init__
    super().__init__(self.readDataFrom(fileName, encoding))
  File "/data/users/beng/src/antlr4/runtime-testsuite/target/classes/Python3/src/antlr4/FileStream.py", line
 27, in readDataFrom 
    return codecs.decode(bytes, encoding)
UnicodeDecodeError: 'ascii' codec can't decode byte 0xc3 in position 1: ordinal not in range(128)
>
```

This PR fixes the error by defaulting Python2/3 streams to UTF-8.

I also allowed specifying the error handling mode, and defaulted to substitute invalid UTF-8 sequences with `U+FFFD`.


ericvergnaud:
Hi,
sorry but this change would break code for existing consumers.
To fix the issue I recommend modifying the Python code template in BasePython_x_Test. java 
Eric

bhamiltoncx:
Hi @ericvergnaud ! Changing the code template is definitely possible to fix only the tests.

Since pretty much every Python client will have to remember to pass `utf-8` (and the existing API provides no way to specify error handling), I think that's pretty error-prone, which is why I sent this PR.

Another possibility is a separate `InputStream` class which uses UTF-8 by default and allows specifying the error mode.

However, I do feel like changing the default encoding from `ascii` to `utf-8` is unlikely to break code for existing consumers. Here's my reasoning:

1. UTF-8 is a strict superset of US-ASCII, so all existing valid inputs passed to our `InputStream` or `FileStream` will continue to be valid.
2. Non-US-ASCII inputs passed to `InputStream` previously would raise an exception. With this PR, we relax that to allow valid UTF-8, but non-UTF-8 inputs will replace invalid byte sequences with `U+FFFD`. (Then, the parser should reject those as invalid.)

Let me know which part concerns you the most for existing consumers.

ericvergnaud:
Hi
in practice, many parsers require ascii only, and can simply reject invalid input in a try/catch block so these would definitely break by no longer rejecting unsupported characters.
Re your proposal, not sure I follow your reasoning. The current API throws an error on the first non-ascii character, whatever the actual encoding is. That is not "error prone". On the contrary, the proposed API would hide those errors, thus postponing the detection of the encoding issue.
I'm happy to add `errors=None`as default parameter to enable silent error handing but will not make it the default, because that would go I believe against antlr's general goal of accurate parsing, and the fail fast principle.
Eric

parrt:
hi guys. maybe we go with a new input stream object (derived or not from existing)?

bhamiltoncx:
Happy to do that!

bhamiltoncx:
@ericvergnaud : I just updated this PR to:

1. Restore the default encoding to `ascii`
2. Expose a new optional `error` parameter (defaulting to `strict`) so clients can choose what to do with invalid input
3. Specify `encoding='ascii'` and `error='replace'` in the base python2 / python3 tests

ericvergnaud:
blessed

PR #1632, opened 2017-01-26
New C# runtime class CodePointCharStream

bhamiltoncx:
As part of #276, the C# runtime needs to be able to treat input files as a series of Unicode code points.

Similar to #1626, this PR adds a new class `CodePointCharStream` to the C# runtime to allow parsing input as a series of Unicode code points.

I'm not a C# expert, so for now, this class only supports construction from a C# `string`. To optimize this, we'll probably want a way to deserialize UTF-8 bytes directly to a `CodePointCharStream`, but I'll leave that for another PR.


parrt:
@ericvergnaud Are you ok with this reorg by @bhamiltoncx to get CodePointCharStream? I just merged similar into Java runtime.

ericvergnaud:
@parrt yes

PR #1637, opened 2017-01-30
Make BaseJavaTest and TestRig work with Unicode values > U+FFFF

bhamiltoncx:
As part of #276 , we need to make `BaseJavaTest` and the command-line `TestRig` work with Unicode values > U+FFFF.

To do this, I refactored the logic for creating `CodePointCharStreams` from Unicode data into a runtime utility class `CharStreams`. This class is a collection of static methods to connect `CodePointCharStream` with `UTF8CodePointDecoder`. It knows how to create `CodePointCharStreams` from:

1. `String`
2. `InputStream`
3. `ReadableByteChannel`

I updated `BaseJavaTest` and `TestRig` to use the new logic (the latter only when `-encoding UTF-8` is specified on the command line — we'll probably want to make this the default at some point).

### DIFF
--- a/runtime-testsuite/test/org/antlr/v4/test/runtime/java/BaseJavaTest.java
+++ b/runtime-testsuite/test/org/antlr/v4/test/runtime/java/BaseJavaTest.java
@@ -952,11 +952,12 @@ public class BaseJavaTest implements RuntimeTestSupport {
 			"import org.antlr.v4.runtime.*;\n" +
 			"import org.antlr.v4.runtime.tree.*;\n" +
 			"import org.antlr.v4.runtime.atn.*;\n" +
+			"import java.nio.file.Paths;\n"+
 			"import java.util.Arrays;\n"+
 			"\n" +
 			"public class Test {\n" +
 			"    public static void main(String[] args) throws Exception {\n" +
-			"        CharStream input = new ANTLRFileStream(args[0]);\n" +
+			"        CharStream input = CharStreams.createWithUTF8(Paths.get(args[0]));\n" +
 			"        <lexerName> lex = new <lexerName>(input);\n" +
 			"        CommonTokenStream tokens = new CommonTokenStream(lex);\n" +
 			"        <createParser>\n"+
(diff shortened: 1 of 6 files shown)


bhamiltoncx:
This would be nice to land.. :)

parrt:
i had a couple of questions about the code. can you check 'em out? thanks!

bhamiltoncx:
Absolutely! Where can I see the questions you had? Sorry if I missed them.

parrt:
yeah, i think i messed up and started a "review" rather than comments. sorry. snapshot of conversation above here:
screen shot 2017-02-10 at 1 06 27 pm

bhamiltoncx:
Ah hah, yeah, I think that lets you send all the review comments in a batch.

> Is this just an efficiency thing?

Partly (because it converts UTF-8 directly to Unicode code points without the intermediate UTF-16 `String` stage), but that's not the main goal.

The main goal is that `CharStreams.createWithUTF8Stream()` creates a `CodePointCharStream` from an `InputStream`, so the resulting `CharStream` contains Unicode code points instead of UTF-16 code units.

I can rename it to something else (`CharStreams.createCodePointStreamForUTF8InputStream()` maybe?) to make it more clear.

> What about UTF-16? Are all of the tests we generate using UTF-8 not UTF-16?

For test input and output, I highly recommend we only use UTF-8, which is what I've implemented. Here's why.

1. UTF-16 and UTF-8 are both encodings which can represent any Unicode code point value (all the way up to `U+10FFFF`).
2. However, only UTF-16 has little- and big-endian versions, so if we were to use it, we'd have to either negotiate endianness or define it between the test and test runner.
3. Also, UTF-16 contains `NUL` bytes (`0x00`) so the input and output cannot be read by most text editors or C/C++ APIs.

So, UTF-8 is definitely the way to go for test input and output.

parrt:
Cool.  So the tests read and write is under our control so utf-8 is no big deal. Got it. createWithUTF8Stream() is fine. merging!

PR #1647, opened 2017-02-09
initial pass at a tree rewriter mechanism based up on existing visitors.

parrt:


runtime/Java/src/org/antlr/v4/runtime/tree/TreeRewriteSupport.java
```diff
@@ -0,0 +1,157 @@
...
+ *  override two key methods:
+ * <pre>
+	 public class MyTreeRewriter extends MyParserNameBaseVisitor<ParseTree> {
+	    @Override
+	    public ParseTree visitChildren(RuleNode node) {
+	        return TreeRewriteSupport.rewriteChildren(this, node);
+	    }
+
+	    @Override
+	    public ParseTree visitTerminal(TerminalNode node) {
```
sharwell: :bulb: Probably want to override `visitErrorNode` as well.

runtime/Java/src/org/antlr/v4/runtime/tree/TreeRewriteSupport.java
```diff
@@ -0,0 +1,157 @@
...
+ *
+ *  The start and stop token parse tree fields are not altered in any way.
+ *  They are meaningless after altering the tree.  Use node's getText()
+ *  to get text of rewrite trees, not getText(RuleContext) from token stream,
+ *  which relies on node start/stop fields.
+ *
+ *  @since 4.6.1
+ */
+public class TreeRewriteSupport {
+	public static ParseTree rewriteChildren(AbstractParseTreeVisitor<ParseTree> visitor, RuleNode tree) {
```
sharwell: :question: Why is the argument here not `ParseTreeVisitor<ParseTree>`?

runtime/Java/src/org/antlr/v4/runtime/tree/TreeRewriteSupport.java
```diff
@@ -0,0 +1,157 @@
...
+	}
+
+	/** Hook replacement in as tree's ith child, set parent of new child properly.
+	 *  If replacement is a token node, make copy of it before setting parent
+	 *  if not a clone already (parent is null).
+	 *
+	 *  This is not meant as general API for users. It is destructive to tree
+	 *  argument, which we assume is a copy made from original incoming tree.
+	 */
+	protected static void _setChild(RuleNode tree, int i, ParseTree replacement) {
```
sharwell: :question: Do any of the members below here need to be public? If you make them private it will be less risky should you need to alter their behavior or refactor somehow in the future.

runtime/Java/src/org/antlr/v4/runtime/tree/TreeRewriteSupport.java
```diff
@@ -0,0 +1,157 @@
...
+import org.antlr.v4.runtime.ParserRuleContext;
+
+import java.util.ArrayList;
+import java.util.List;
+
+/** This class encapsulates code necessary to support tree rewriting
+ *  using visitors. To rewrite trees using ANTLR-generated visitors,
+ *  override two key methods:
+ * <pre>
+	 public class MyTreeRewriter extends MyParserNameBaseVisitor<ParseTree> {
```
sharwell: 💭 Currently you are using a return value of `null` to identify nodes that you want to remove. Another option would be not allowing node deletion in this way but instead require the parent be visited in order to remove one of its children. The advantage of this approach is you could avoid overriding `visitTerminal` and `visitErrorNode` every time, but it of course comes with the downside of making node removal potentially more difficult.

Roslyn's approach is more abstracted. Rather than provide the entire visitor as input, you only need to supply a transformation function (basically a `BiFunction<ParseTree, ParseTree, ParseTree>` for Java). The first argument to the transformation function is the original `ParseTree` that needs to be rewritten, and the second argument is the result of calling `rewriteChildren` (different from the first argument if and only if the children were modified). One could choose to implement this function with a visitor, or even something as simple as looking up values in a `Map<ParseTree, ParseTree>`.

Roslyn also provides a utility class similar to what we use `TokenStreamRewriter` for. This augments the rewriter functionality and simplifies it for certain uses, such as the following:

```java
ParseTree originalTree = ...;
TreeEditor editor = new TreeEditor(originalTree);
editor.replaceTerminal(someTerminal, newTerminal);
editor.replaceTree(someTree, newTree);
editor.remove(nodeToRemove);
ParseTree rewritten = editor.rewriteTree(); // Or perhaps editor.run()
```

:bulb: I would recommend at minimum adding such a `TreeEditor` to simplify many common cases for rewriting trees, so users don't have to worry about getting the visitor semantics exactly right.

runtime/Java/src/org/antlr/v4/runtime/tree/Trees.java
```diff
@@ -16,14 +16,23 @@
...
+import java.util.Map;
+import java.util.concurrent.ConcurrentHashMap;
 
 /** A set of utility routines useful for all kinds of ANTLR trees. */
 public class Trees {
+	/** A reflection cache tracking constructor methods for various tree nodes
+	 *  so we can clone nodes. If this starts to get too big,
+	 *  call {@see resetCtorCache}.
+	 */
+	protected static Map<Class<? extends ParserRuleContext>, Constructor<? extends ParserRuleContext>> ctorCache;
```
sharwell: :bulb: Make this `final` and initialize it here.

runtime/Java/src/org/antlr/v4/runtime/tree/Trees.java
```diff
@@ -189,6 +198,107 @@ else if ( !findTokens && t instanceof ParserRuleContext ) {
...
+	public static ParserRuleContext shallowCopy(ParserRuleContext t) {
+		try {
+			Class<? extends ParserRuleContext> cl = t.getClass();
+			Class<?> sup = cl.getSuperclass();
+			Class<?> supSup = sup.getSuperclass();
+			if ( ctorCache==null ) { // in a multithreaded environment, some thread will win to set ctorCache
+				ctorCache =	new ConcurrentHashMap<>();
+			}
+			boolean
+				isAltLabelNode = sup!=ParserRuleContext.class && // usual case
```
sharwell: ⚠️ This makes potentially invalid assumptions regarding the context super class option.

runtime/Java/src/org/antlr/v4/runtime/tree/Trees.java
```diff
@@ -189,6 +198,107 @@ else if ( !findTokens && t instanceof ParserRuleContext ) {
...
+		if ( t instanceof TerminalNode ) return shallowCopy((TerminalNode)t);
+		return shallowCopy((ParserRuleContext)t);
+	}
+
+	/** Make a complete copy of a parse tree rooted at t.
+	 *  The original tree, t, is not altered.
+	 *
+	 *  @since 4.6.1
+	 */
+	public static ParseTree deepCopy(ParseTree t) {
```
sharwell: :bulb: I would leave this out for now.

1. It's not needed to get rewriting working
2. It fails to assign deep-copied nodes back to fields created for labels, which could lead to us fielding bug reports for it

runtime-testsuite/test/org/antlr/v4/test/runtime/java/api/TreeRootForTesting.java
```diff
@@ -0,0 +1,110 @@
+package org.antlr.v4.test.runtime.java.api;
```
sharwell: :memo: Missing headers for a few files

runtime/Java/src/org/antlr/v4/runtime/tree/Trees.java
parrt: Yeah, I had to find a generic way to dup a regular node vs Alt label node. Easier if I generate factory methods in parser as shown in https://github.com/antlr/antlr4/pull/1675

runtime/Java/src/org/antlr/v4/runtime/tree/TreeRewriteSupport.java
parrt: Yeah, originally I was going to just generate a new base class called XRewriteVisitor and thereby avoid all this overriding of visitChildren etc... but it means generating a new file, a big burden on all target maintainers and a biggish change to documentation.

Intrigued by BiFunction idea. How would it do matching?  By root classname? user has to do the matching in BiFunction?

TreeEditor sounds interesting too for common replacements.

Re `null` return.  I supply a set/delete child but detecting when to delete could require significant parsing of the child, which means visiting functionality.  Perhaps not `null` but a special `DEADTREE`?

runtime/Java/src/org/antlr/v4/runtime/tree/TreeRewriteSupport.java
parrt: oops

runtime/Java/src/org/antlr/v4/runtime/tree/TreeRewriteSupport.java
parrt: oops

runtime/Java/src/org/antlr/v4/runtime/tree/TreeRewriteSupport.java
parrt: I think they could be handy and they are needed to alter child list at a specific subtree.

runtime/Java/src/org/antlr/v4/runtime/tree/TreeRewriteSupport.java
```diff
@@ -0,0 +1,157 @@
...
+	 *  transformations, one might want the parent pointers to be
+	 *  consistent with the final rewritten tree.
+	 */
+	public static void updateParentPointers(ParseTree tree) {
+		// TODO
+	}
+
+	public static TerminalNode copy(TerminalNode t) { return Trees.shallowCopy(t); }
+
+	private TreeRewriteSupport() { }
```
parrt: ctor probably clear.

runtime/Java/src/org/antlr/v4/runtime/tree/Trees.java
parrt: okidoki.

runtime/Java/src/org/antlr/v4/runtime/tree/Trees.java
parrt: good idea.

runtime-testsuite/test/org/antlr/v4/test/runtime/java/api/TreeRootForTesting.java
parrt: thanks. damn intellij

pureconfig:
But... unless I miss something, this means an entire pass over potentially huge input is required for even the smallest change, such as changing a token value based on pattern matching.

I'd love to see the ability to mutate the generated context objects during visitation. More than enough for many cases, with no loss in overall performance. With a "use at your own risk" warning, of course :)

parrt:
This is a visitor that alters as it goes. :)   A faster solution would be creating an index of nodes for fast XPath look ups then specific replacements.

pureconfig:
> This is a visitor that alters as it goes

@parrt Yes, but can I use it from my code generation visitor? If not, I must do two passes. Or do I miss something? :)

parrt:
In principle, one could combine a tree rewriting phase and codegeneration but because the rewriting affects codegeneration, I would expect that to be done completely before codegeneration starts. Imagine generating code for a sub expression only to find above that it is multiplied by zero and hence unnecessary code.

pureconfig:
I agree that a separate pass is sometimes necessary. But for other cases, you need to do very simple stuff, like maybe inject a node or change some literals. Mutators in generated context objects would solve many cases - of course with a big fat warning to use on your own risk ;)

pureconfig:
@parrt Is this 4.8 material, or even later?

parrt:
Not 4.7 for sure. I need to see such a beast in action before incorporating.

marcospassos:
@parrt, any news on this? We're looking forward to this feature :)

marcospassos:
May I help in some way?

parrt:
The reality is that I'm focused on something else these days. Can you just build your own visitor and use with your app?

thron7:
@parrt I don't know if I can provide "such a beast in action", and am aware that your focus has shifted, but would like to weigh in my 2c here.

- I'm also concerned with source-to-source transformation, so probably in the transpiler/compiler camp.
- My ideal approach to code transformation is a series of mutating visitors that each do a dedicated, well-defined change, ranging from re-writing class definitions, inlining methods, refactoring try-catch clauses, inserting statements at specific places, to initializing fields and changing access modifiers. I'm less concerned about speed, more about controlling a pipeline of such micro-transformations, which are easier to develop and to debug.
- I know of one person who exemplified this approach, a former member of ETH Zurich's Program Transformation Group, Marco Trudel, who implemented C-to-Eiffel and Java-to-Eiffel Translators and has [various papers about this](http://se.inf.ethz.ch/people/trudel/).
- What I would wish for would be an immutable parse tree with lenses-like accessors (could use XPath) that on mutation return a new tree that is structurally shared with the old one. Creating an entire new tree with every transformation I think would not be feasible.

parrt:
Sorry guys. no progress on this on my end.

kaby76:
Although this is an old topic, I just thought to throw in my 2 cents here as well. This summer I've been working on my Antlr4 editor tool, Antlrvsix, to include a couple of dozen parse tree transformations over [ANTLRv4Parser.g4](https://github.com/antlr/grammars-v4/blob/master/antlr/antlr4/ANTLRv4Parser.g4). I wrote the transformations in C# using the current ANTLR runtime, but the plan has always been to replace this low-level C# code with some kind of a term rewrite system. While the problem I'm working on isn't a transpiler, you might still be interested in what I learned.

(a) There's a lot of boilerplate code to traverse the parse tree to identify when a rewrite pattern applies. Visiting a single node without context is not enough in order to verify the pattern for the tree rewrite.

(b) There a lot of boilerplate code to create new subtrees and insert them into the existing tree to modify parent/child relationships, token start/stop, etc, of the tree.

For (a), I found that XPath works well in the recognition of patterns for rewrites. For example, suppose I have a grammar rule with parentheses that can be removed as in `r : a ( b ) c;`. The XPath expression to find all `altList` in `block` where the parentheses can be removed is `"//(altList | labeledAlt)/alternative/element/ebnf[not(child::blockSuffix)]/block/altList[not(@ChildCount > 1)]"`. This pattern makes sure that `blockSuffix` does not exist, otherwise the parentheses are required. Another example is finding parser rules that contain direct left recursion, which is required in order to remove indirect left recursion. The XPath expression is `"//parserRuleSpec[RULE_REF/text() = ruleBlock/ruleAltList/labeledAlt/alternative/*[name()='element'][1]/atom/ruleref/*[1]/text()]"`. It's kind of handy to just use `"../../../.."` to go up a parse tree rather than `var p = n.Parent.Parent.Parent.Parent;`. The XPath engine in the ANTLR 4.8 runtime was missing filters, so I ported and extended an XPath v2 engine to C#. Obviously, a pattern like `"//altList"`, the equivalent of a visitor method, is insufficient for these complex patterns. Also, XPath doesn't have variables, so I'm thinking of extending the XPath syntax with some kind of dynamic binding of partial results to C# variables.

For (b), I couldn't find a system that specifies trees, so I just wrote code to convert an s-expression into a partial tree that I can hook up to the existing parse tree. In order to know what node names mean, I require the tree type system, which is contained in the generated Antlr parser. For example, suppose I want to replace an ANTLR lexer rule for a case sensitive keyword with a case insensitive rule, e.g., `OR: 'or';` with `OR: [Oo][Rr];`. To construct the new `lexerElement` subtree, the s-expression is `"( lexerElement ( lexerAtom {id} ))"`, where `id` is a dynamic look-up of a tree node to insert a child (or children) of the `lexerAtom`.

In order to connect all this together, I still need C#. But, it is less C# code and more reliable. I have only a few transforms replaced with XPath and s-expressions, but so far it's looking good. This is still a rather verbose syntax for tree rewrites, so something like TXL might be a better way to go.

--Ken

PR #1650, opened 2017-02-10
Fix JavaScript InputStream.getText() when input contains Unicode values > U+FFFF

bhamiltoncx:
This is a follow-up to https://github.com/antlr/antlr4/pull/1628 . 

When I added optional support to read in code points > U+FFFF to `InputStream`, I forgot to update `InputStream.getText()` to handle converting those code points back to UTF-16 for the Javascript `String`.

That meant the code which enabled the optional `decodeToUnicodeCodePoints` feature and then invoked `InputStream.getText()` would fail.

The root cause was that `this.strdata.slice(start, stop + 1)` was using the wrong units for the index; `this.strdata` uses UTF-16 for its indexes, but `start` and `stop` would be in Unicode code points.

This updates `InputStream.getText()` to work correctly if the `decodeToUnicodeCodePoints` flag is truthy. I imported the `String.fromCodePoint()` polyfill for backwards compatibility to convert Unicode code points to UTF-16 code units.

### DIFF
--- a/LICENSE.txt
+++ b/LICENSE.txt
@@ -28,6 +28,7 @@ THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
 =====
 
 MIT License for codepointat.js from https://git.io/codepointat
+MIT License for fromcodepoint.js from https://git.io/vDW1m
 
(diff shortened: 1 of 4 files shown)


ericvergnaud:
blessed

PR #1651, opened 2017-02-10
Python 2 and Python 3 support for lexer and parser output to a file

bhamiltoncx:

Until Python 3.6 (https://www.python.org/dev/peps/pep-0528/), writing Unicode to `sys.stdout` on Windows just doesn't work correctly.

To work around this, this PR adds optional support to `Lexer` and `Parser` in the Python 2 and Python 3 runtimes to write output to a specified `file`-like object, instead of hard-coding `sys.stdout`.

We will still default to `sys.stdout` if nothing is specified.

I'll send a subsequent diff after this lands to use this in the Python2/3 runtime tests.


ericvergnaud:
blessed

PR #1652, opened 2017-02-10
C# support for lexer and parser output to a file

bhamiltoncx:

This is the same as https://github.com/antlr/antlr4/pull/1651 , but for C#.

It turns out C# on Windows writing to `Console.Out` has just as many problems with Unicode as Python < 3.6 does.

So, to work around these issues, this diff adds optional support to the C# `Lexer` and `Parser` for writing output to a file instead of hard-coding `Console.Out`.

A subsequent diff will use this work from the C# runtime tests.


ericvergnaud:
blessed

PR #1653, opened 2017-02-10
Emit language-specific Unicode escapes in codegen

bhamiltoncx:
This is part of the fix for #276.

Previously, if an input grammar definition included non-ASCII text or escaped Unicode literals, we would write out non-ASCII source files which then needed to be compiled.

While it's possible to configure many compilers to support UTF-8 source files, it's pretty challenging to do so across the board.

So, instead, this PR ensures all source which is code-generated by ANTLR4 to be 7-bit US-ASCII.

We do this by emitting language-specific Unicode literal escapes (`\uABCD` for Java and Javascript, `\uABCD` or `\U0001ABCD` for Python, and `\u{ABCD}` for Swift).

Note that this diff includes support for parsing the new extended `\u{ABCD}` ANTLR grammar escapes I plan to add to ANTLR in a subsequent diff. If you'd like, I can split that out and send it in separately, but I don't think it hurts here, since the grammar can never include that syntax yet.

### DIFF
--- a/tool-testsuite/test/org/antlr/v4/test/tool/TestUnicodeEscapes.java
+++ b/tool-testsuite/test/org/antlr/v4/test/tool/TestUnicodeEscapes.java
@@ -1,0 +1,78 @@
+/*
+ * Copyright (c) 2012-2017 The ANTLR Project. All rights reserved.
+ * Use of this file is governed by the BSD 3-clause license that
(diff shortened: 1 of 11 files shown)


parrt:
@sharwell can you take a peek? Looks like a nice refactoring of Target subclasses to central conversion. You are more familiar with unicode stuff so pinging you here before I merge.

PR #1655, opened 2017-02-13
Fix #1298 for CSharp

### LINKED ISSUE #1298: Single-token deletion error node not in parse tree

lionelplessis:
Port of the java fix for #1298 to C#.

### DIFF
--- a/contributors.txt
+++ b/contributors.txt
@@ -133,3 +133,4 @@ YYYY/MM/DD, github id, Full name, email
 2017/01/13, marcelo-rocha, Marcelo Rocha, mcrocha@gmail.com
 2017/01/23, bhamiltoncx, Ben Hamilton, bhamiltoncx+antlr@gmail.com
 2017/01/18, mshockwave, Bekket McClane, yihshyng223@gmail.com
+2017/02/10, lionelplessis, Lionel Plessis, lionelplessis@users.noreply.github.com
--- a/runtime-testsuite/test/org/antlr/v4/test/runtime/descriptors/ParseTreesDescriptors.java
+++ b/runtime-testsuite/test/org/antlr/v4/test/runtime/descriptors/ParseTreesDescriptors.java
@@ -112,7 +112,7 @@ public class ParseTreesDescriptors {
 
 		@Override
 		public boolean ignore(String targetName) {
-			return !targetName.matches("Java|Python2|Python3|Node|Swift");
+			return !targetName.matches("Java|Python2|Python3|Node|Swift|CSharp");
 		}
 	}
 
--- a/runtime/CSharp/runtime/CSharp/Antlr4.Runtime/ParserRuleContext.cs
+++ b/runtime/CSharp/runtime/CSharp/Antlr4.Runtime/ParserRuleContext.cs
@@ -116,7 +116,18 @@ namespace Antlr4.Runtime
             }
         }
 
-        /// <summary>COPY a ctx (I'm deliberately not using copy constructor)</summary>
+        /// <summary>
+        /// COPY a ctx (I'm deliberately not using copy constructor) to avoid
+        /// confusion with creating node with parent. Does not copy children.
+        /// 
+        /// This is used in the generated parser code to flip a generic XContext
+        /// node for rule X to a YContext for alt label Y. In that sense, it is
+        /// not really a generic copy function.
+        /// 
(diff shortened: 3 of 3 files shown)


parrt:
Thanks!

PR #1656, opened 2017-02-14
Import parser rule extension

millergarym:
These changes add functionality to the grammar import feature for the Go target.
The most significant is the adding of an `extends=<importedrule>` options to rules.
The Alternatives in the root grammar rule, with this option set, are added to the rule specified from the imported grammar.

Also for overridden rule the base listener delegates to the base listener from the imported grammar.

A command line argument `-importParams=<grammarName>:<package>:<importPath>` is need to stick the thing together.

There are another minor feature; a command line param to change the runtime import package.

Please let me know what you think & what is required to get this accepted.
Test-case, fixes, documentation etc.
Currently only testing with named alternatives.

Cheers
Gary

Example grammars and command line
```
parser grammar E0Parser;

tokens { 
	VIRTUAL1,
	VIRTUAL2
}

options {
  tokenVocab = E0Lexer;
}
 
start : cell EOF 
;

cell : Eq=EQ X+=WS* expr Y+=WS*  
     | .*? 
;

expr
    : Pre=('+'|'-') Ws+=WS* E=expr                          #Prefix
    | E1=expr L+=WS* T='^' R+=WS* E2=expr        #Infix 
    | E1=expr L+=WS* T='*' R+=WS* E2=expr         #Infix
    | Fname=id_a1 '(' Args+=expr* ')'                    	#Func
    | '(' elems  ')'                                                        #Block
    | ref                                                                      #Reference
    | errorType                                                           #Error
;

elems
	: A=expr (',' B+=expr)* 
;

ref 
    : R1=ref COLON R1=ref                       			#Range
    | A1=id_a1                                  			#A1Ref
;

id_a1
	: First=(CHAR | UNDERSCORE | BSLASH) Rest+=(CHAR | UNDERSCORE | BSLASH | INT )*
															#NamedRange
;

errorType
    : HASH A=CHAR B=(CHAR | SLASH | INT )* C=(QMARK | XMARK)
;
```

```
parser grammar E1Parser;

tokens { 
	ENHANCEDREF
}
options {
  tokenVocab = E0Parser;
}
import E0Parser;
 

ref 
    : R1=ref COLON R1=ref                       			#Range
    | A1=id_a1                                  			#A1Ref
	| E='[' INT ']'					#EHRef
;

id_r1c1
options{ 
	extends=id_a1;
}
	: hereIam+=INT+ #NumberRange
;
```

Command line args
``` 
E1Parser.g4 
-lib ../ep0
-package ep1parser
-o parser 
-Dlanguage=Go 
-runtimeImport github.com/wxio/antlr4-go 
-importParams E0Parser:parser:wx.io/antlr/ep0/parser
```


parrt:
Hmm...seems a massive change in number of files etc... widespread which makes me nervous. Also i can't add command-line options willy nilly. Would these perhaps be better as grammar options like contextSuperClass etc...?

millergarym:
Hey Terence (@parrt)

It was one of those changes where I only really figured out what I was doing whilst doing it.
There are a number of changes that are useful (eg isList in ContextGetterDecl) that I didn't land up using.
Should I remove them?

I totally agree about the grammar options verse command line and will change these.
Also makes testing a lot easier.

I have just pushed changes that get the tests working again, hopefully starts to give some comfort.

Still need to include some tests for this feature.

Should I start an issue to discuss this functionality? (i.e. the extending of rules via the import mechanism)

Cheers
Gary

parrt:
Can you start by telling me what problem you're trying to solve?

millergarym:
Sure,

For a number of reasons, mostly related to academic research & teaching, I want a number of variants of a language (and variants of variants). Any changes in the base language should propagate to the variants. 

For a more concrete use-case take the example of empirical (language usability) tests on language features. I would like to create a base language (& implementation) & then extend it, but with multiple different variants. This could possible be done with predicates (or maybe by leaving extension rules in the base grammar), but this would quickly get very messy. I also want to use these languages for academic purposes & really want to keep them clean.

The extends option is currently very simple & naive, it simply copies its alts into the specified rule. The complexity comes from generating a base listener that delegates correctly, as there are eight cases based on the combination rule v named alt, import v local and extended v regular. With Go's type system * all eight cases involve different generated code.

In general I have wanted a grammar inheritance mechanism in Antlr for years, since doing some quite extensive work with 3.2, including starting a project to target Go (got lexing and basing parsing working, gave up as there was a lot of instability between dot releases after 3.2). My interest has re-invigorated with Antlr 4, the Go target and my current. There is still work to be done. For example what should happen with non-alternative elements of an extension rule (params, returns, init and after etc.).

This change took me some effort. Please let me know what needs to happen to get rule extension/inheritance (or what every it gets called) into Antlr.
Here are a couple of questions that might help;
- Where is the best place to discuss this (eg Git issue, mailing list)?
- Is a Java implementation mandatory before inclusion? 
- What level of testing (unit, code coverage etc.) should be aimed for?
- What are the alternatives (can the current import mechanism support this)?
- Does this interfere with any other features on the roadmap?

My intend is to work on tests, document and hopefully a working Java implementation.

Hope this helped.
Regards
Gary

p.s. Can you look at the failing test in java (TestGrammarParserInterpreter), I don't know if I introduced it.

* Constrains of Go's type system. No struct inheritance (actually not inheritance at all, embedding in not inheritance), not overriding in interface and no overloading.

parrt:
ok, now I got it. While I recognize the potential use-case for multiple variants, it's not that common among users (anecdotally).  This is a major addition of complexity, allowing rules to alter imported rules. Many systems do this in academia but I'm not sure how many folks need it in practice. Now, if you have the variants problem to solve, ANTLR doesn't really help you much I admit.  @sharwell is currently examining a special predicate to gate rules in/out but I don't know if it'll end up in the official version.

parrt:
Good chattin with you @millergarym I'll close per our conversation but am happy to answer questions. thanks!

PR #1657, opened 2017-02-14
Python 2/3 tests: Write test output to a file

bhamiltoncx:

This builds on #1651 to write Python test output to a file instead of `sys.stdout`.

This works around bugs with Python < 3.6 when writing Unicode to `sys.stdout` on Windows.


parrt:
@ericvergnaud can you take a peek?

ericvergnaud:
Yes I’m fine with this.

PR #1661, opened 2017-02-16
Change MAX_CHAR_VALUE to 0x10FFFF

bhamiltoncx:

Previously, `Lexer.MAX_CHAR_VALUE` was set to `0xFFFF` (or sometimes `0xFFFE`).

This meant some lexer transitions (like `NotSet`) would not support Unicode values > `U+FFFF`.

This changes `Lexer.MAX_CHAR_VALUE` to `0x10FFFF` (the largest possible Unicode code point). I replaced all the range-checks which hard-coded `0xFFFF` or `0xFFFE` with this new constant.

### DIFF
--- a/runtime/CSharp/runtime/CSharp/Antlr4.Runtime/Atn/LexerATNSimulator.cs
+++ b/runtime/CSharp/runtime/CSharp/Antlr4.Runtime/Atn/LexerATNSimulator.cs
@@ -380,7 +380,7 @@ namespace Antlr4.Runtime.Atn
 
 		protected ATNState GetReachableTarget(Transition trans, int t)
 		{
-			if (trans.Matches(t, char.MinValue, char.MaxValue))
+			if (trans.Matches(t, Lexer.MinCharValue, Lexer.MaxCharValue))
 			{
 				return trans.target;
 			}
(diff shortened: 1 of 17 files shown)


parrt:
LOVE IT!  One concern. the min/max value for chars is used in NotSet transitions. Should we augment those to avoid the invalid range used for UTF-16 spec? E.g., https://github.com/antlr/antlr4/blob/master/runtime/Java/src/org/antlr/v4/runtime/atn/NotSetTransition.java#L22

bhamiltoncx:
> LOVE IT! One concern. the min/max value for chars is used in NotSet transitions. Should we augment those to avoid the invalid range used for UTF-16 spec? E.g., https://github.com/antlr/antlr4/blob/master/runtime/Java/src/org/antlr/v4/runtime/atn/NotSetTransition.java#L22

It's a good question! Technically, standalone or "dangling" surrogates in the special UTF-16 range (`U+D800` to `U+DFFF`) are not valid Unicode code points, so we could reject them at various stages (probably in `CharStream` would be the right place).

Note that these surrogate values are *not* actually legal to encode in UTF-8, so in reality they're not likely to occur in real life.

The only place you're likely to see a dangling surrogate in the wild is when someone arbitrarily truncates a string in a UTF-16-internal language (like Java or JS) then passes it to ANTLR for parsing. If they truncate it between two halves of a surrogate pair, ANTLR will need to figure out what to do when parsing it.

However, I think to support existing clients, we'll have to be agnostic of dangling surrogates. This is how Java and JavaScript behave today.

I've added a few tests in the WIP branch to explicitly test behavior of standalone (dangling) surrogates, but I'll add a few more to ensure all the runtimes handle them explicitly.

parrt:
ok, sounds good. merging.

PR #1663, opened 2017-02-16
Python 2.7 InputStream: Properly support Unicode code points on Windows and macOS

bhamiltoncx:

Until Python 3.3, Python could be compiled in one of two modes (`sys.maxunicode == 0x10FFFF` or `sys.maxunicode == 0xFFFF`).

In particular, Python 2.7 on macOS and on Windows uses `0xFFFF`, where Python 2.7 on Linux as well as Python >= 3.3 globally use `0x10FFFF`.

The two modes behave VERY differently when looking at Unicode strings with code points > `U+FFFF`.

Python 2.7 on Linux:

```
>>> import sys
>>> hex(sys.maxunicode)
'0x10ffff'
>>> len(u'\U0001F40D')
1
>>> [c for c in u'\U0001F40D']
[u'\U0001f40d']
```

Python 2.7 on Windows:

```
>>> import sys
>>> hex(sys.maxunicode)
'0xffff'
>>> len(u'\U0001F40D')
2
>>> [c for c in u'\U0001F40D']
[u'\ud83d', u'\udc0d']
```

To handle this inconsistency, this PR introduces a new module `antlr4.CodePoints`, which I've separately open-sourced here:

https://pypi.python.org/pypi/codepoints/

and uses it as part of `InputStream` on Python 2.

Note that Python 3 doesn't need this, since we require 3.5 or later (which always uses `sys.maxunicode == 0x10FFFF`).

### DIFF
--- a/runtime/Python2/src/antlr4/InputStream.py
+++ b/runtime/Python2/src/antlr4/InputStream.py
@@ -10,7 +10,8 @@ import unittest
 #  Vacuum all input from a string and then treat it like a buffer.
 #
 from antlr4.Token import Token
-
+from antlr4.CodePoints import from_unicode, to_unicode
+import sys
 
 class InputStream (object):
 
@@ -21,7 +22,7 @@ class InputStream (object):
 
     def _loadString(self):
         self._index = 0
-        self.data = [ord(c) for c in self.strdata]
+        self.data = list(from_unicode(self.strdata))
         self._size = len(self.data)
 
     @property
@@ -79,9 +80,9 @@ class InputStream (object):
         if stop >= self._size:
             stop = self._size-1
         if start >= self._size:
-            return ""
+            return u""
         else:
-            return self.strdata[start:stop+1]
+            return to_unicode(self.data[start:stop+1])
 
     def __str__(self):
         return unicode(self)
--- a/runtime/Python2/src/antlr4/CodePoints.py
+++ b/runtime/Python2/src/antlr4/CodePoints.py
@@ -1,0 +1,73 @@
+#
+# Copyright (c) 2012-2016 The ANTLR Project. All rights reserved.
+# Use of this file is governed by the BSD 3-clause license that
+# can be found in the LICENSE.txt file in the project root.
+#
+
+from __future__ import unicode_literals
+import sys
+
+def is_leading_surrogate(code_unit):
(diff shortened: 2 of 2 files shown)


parrt:
@ericvergnaud can you check this for backward compatibility and such? thanks!

bhamiltoncx:
For what it's worth, I wrote some unit tests here:

https://github.com/bhamiltoncx/pycodepoints/blob/pycodepoints/codepoints/tests/test_codepoints.py

and confirmed they pass on Python 2.7 and 3.5.

runtime/Python2/src/antlr4/InputStream.py
```diff
@@ -21,7 +21,7 @@ def __init__(self, data):
 
     def _loadString(self):
         self._index = 0
-        self.data = [ord(c) for c in self.strdata]
+        self.data = list(from_unicode(self.strdata))
```
ericvergnaud: why list()? from_unicode already returns a list, no?

ericvergnaud:
I believe the compatibility should be ok, just added a comment re a specific line of code.

runtime/Python2/src/antlr4/InputStream.py
bhamiltoncx: Actually, `from_unicode` returns a generator, not a list.

Since existing clients of `self.data` expect it to be a list, this eagerly evaluates the generator and converts the resulting items to a list.

bhamiltoncx: (Specifically, `_from_utf16()` returns a generator.)

ericvergnaud: Yes I had only looked at _from_unicode, not great that these 2 don't return the same type, maybe you want to make _from_unicode return a generator too, then in the future we can benefit from lazy conversion if clients of self.data support it

bhamiltoncx:
Agreed! Actually, I forgot that list comprehension aren't generators. :)

parrt:
Ok, so I should wait on a tweak from you @bhamiltoncx ?

bhamiltoncx:
> Ok, so I should wait on a tweak from you @bhamiltoncx ?

Sure, done. I just pushed a tweak to consistently use generators throughout.

PR #1665, opened 2017-02-16
 factor out the creation of error nodes and terminal nodes in the parser

parrt:
The new methods act like a built-in factory during parse tree construction. In the end it is a fairly minor change, but since I moved some methods around in Parser, it looks bigger than it is. All I did was to convert `new TerminalNodeImpl` and `new ErrorNodeImpl` code to calls to the factory methods. I also cleaned up ParserRuleContext add child stuff.

@antlr/antlr-targets should consider updating their targets as it adds flexibility that will be handy in a future PR where I support tree construction that contains hidden text before/after nodes.

### DIFF
--- a/runtime/Java/src/org/antlr/v4/runtime/ANTLRErrorStrategy.java
+++ b/runtime/Java/src/org/antlr/v4/runtime/ANTLRErrorStrategy.java
@@ -6,6 +6,8 @@
 
 package org.antlr.v4.runtime;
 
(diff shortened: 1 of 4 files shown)


hanjoes:
will take a look.

willfaught:
@parrt Can you create issues to track this work for the targets? Not able to work on it at the moment.

parrt:
I updated this branch with a continue PR. https://github.com/antlr/antlr4/pull/1674

PR #1667, opened 2017-02-16
Create parse trees whose leaves have left/right hidden token text

parrt:
I'm leaving this open and not merging until I have a use case that makes sure it works. Could be useful for those wanting to get a head start.

### DIFF
--- a/runtime/Java/src/org/antlr/v4/runtime/tree/TerminalNodeWithHidden.java
+++ b/runtime/Java/src/org/antlr/v4/runtime/tree/TerminalNodeWithHidden.java
@@ -76,6 +76,19 @@ public class TerminalNodeWithHidden extends TerminalNodeImpl {
 	// TODO: how to collect hidden on error nodes (deleted, inserted, during recovery)
 	protected void collectHiddenTokens(BufferedTokenStream tokens, int channel, Token symbol) {
 		List<Token> left  = tokens.getHiddenTokensToLeft(symbol.getTokenIndex(), channel);
+
+		// check the EOF special case separately
+		if ( symbol.getType()==Token.EOF ) {
+			int prevRealIndex = tokens.previousTokenOnChannel(symbol.getTokenIndex()-1, symbol.getChannel());
+			if ( prevRealIndex<0 ) { // EOF only gets hidden stuff if it's only token
+				if ( left!=null ) {
+					List<Token> allBefore = tokens.get(0, symbol.getTokenIndex()-1);
+					leading = allBefore.toArray(new Token[allBefore.size()]);
+				}
+			}
+			return;
+		}
+
 		if ( left!=null ) {
 			Token firstHiddenLeft = left.get(0);
 			Token prevReal = null;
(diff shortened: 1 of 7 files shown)


parrt:
@antlr/antlr-targets This is a feature addition that you might want to consider.

mike-lischke:
What is a left/right hidden token text? What would be the use of storing that in the parse tree (given that you can always get that from the token stream)?

parrt:
In the next PR, you'll see ability to rewrite parse trees. :) Makes it much easier to keep track of comments during rewrites if they are in the tree.

KvanTTT:
@parrt why did you decide to use text but not a sequence of hidden or maybe even not default channel tokens? This is a more flexible solution on my opinion.

parrt:
I changed my mind a few times; Sam concurs keeping tokens is better. I'll make Token[] and update.

mike-lischke:
Sorry, I'm still not getting the full picture. If you collect both left and right hidden tokens won't they be duplicated? The hidden tokens on the right of a visible token are the same as the left hidden tokens to the next visible one, no?

KvanTTT:
Association of hidden and significant tokens is a really good conception because of it allows to build a full-fidelity parse tree for any text even with errors. Also it brings more convenient handling of spaces, comments and other hidden tokens. Left and right tokens for every significant token has an empty intersection if an implementation is correct. Roslyn, .NET Compiler Platform, have the following definition of leading and trailing trivia:

> You can access trivia by inspecting a token’s LeadingTrivia or TrailingTrivia collections. When source text is parsed, sequences of trivia are associated with tokens. In general, a token owns any trivia after it on the same line up to the next token. Any trivia after that line is associated with the following token. The first token in the source file gets all the initial trivia, and the last sequence of trivia in the file is tacked onto the end-of-file token, which otherwise has zero width.

See more detail description in [Roslyn Overview](https://github.com/dotnet/roslyn/wiki/Roslyn%20Overview#syntax-trivia). On my opinion **leading** and **trailing** tokens are more appropriate names instead of **left** and **right** tokens.

mike-lischke:
I see, thanks @KvanTTT.

parrt:
I believe I have implemented the Roslyn mechanism with caveat of EOF only gets hidden tokens if there is no other token.  I just started testing the cases where the input is invalid. all hidden tokens should be attached somewhere in new augmented tree on error nodes or normal leaves.

KvanTTT:
@parrt, it's cool!

parrt:
:) @sharwell will review tomorrow.

mike-lischke:
Playing devil's advocate: what if the whitespaces are not hidden :-) ? Maybe leave out the "hidden" part and call those leading and trailing trivia as @KvanTTT pointed out?

parrt:
see recent commit ;)

parrt:
Ok, lots of cleanup. @sharwell can you take a sniff?

runtime/Java/src/org/antlr/v4/runtime/tree/TerminalNodeWithHidden.java
```diff
@@ -0,0 +1,174 @@
+package org.antlr.v4.runtime.tree;
```
mike-lischke: Copyright info missing.

runtime/Java/src/org/antlr/v4/runtime/tree/TerminalNodeWithHidden.java
```diff
@@ -0,0 +1,174 @@
...
+
+	/** Construct a node with left/right hidden tokens on a channel,
+	 *  or all hidden tokens if channel==-1.
+	 */
+	public TerminalNodeWithHidden(BufferedTokenStream tokens, int channel, Token symbol) {
+		super(symbol);
+		collectHiddenTokens(tokens, channel, symbol);
+	}
+
+	protected void collectHiddenTokens(BufferedTokenStream tokens, int channel, Token symbol) {
```
mike-lischke: Why do we pass in a token here? The terminal node already has a symbol member. Shouldn't that be used here?

hanjoes:
Still trying to get the whole picture, but is overriding the "createTerminalNode" of Parser the "official" way of enabling this hidden token mechanism? I'm asking since Swift has this field access modifier that controls overriding of members for subclasses (not like Java in which "open" is the default behavior unless you mark it as "final").

hanjoes:
it should be, saw it in a later comment:)

parrt:
I decided not to add this functionality, at least for now.

PR #1674, opened 2017-02-18
augment TerminalNode with setParent()

parrt:
An addendum to https://github.com/antlr/antlr4/pull/1665

Technically, this is not backward compatible as it changes the interface but no one was able to create custom TerminalNodes anyway so I'm adding as it improves internal code quality. addChild now sets the parent rather than create. MUCH better.

### DIFF
--- a/runtime/Java/src/org/antlr/v4/runtime/Parser.java
+++ b/runtime/Java/src/org/antlr/v4/runtime/Parser.java
@@ -20,7 +20,6 @@ import org.antlr.v4.runtime.misc.IntegerStack;
 import org.antlr.v4.runtime.misc.IntervalSet;
 import org.antlr.v4.runtime.tree.ErrorNode;
 import org.antlr.v4.runtime.tree.ErrorNodeImpl;
-import org.antlr.v4.runtime.tree.ParseTree;
 import org.antlr.v4.runtime.tree.ParseTreeListener;
 import org.antlr.v4.runtime.tree.ParseTreeWalker;
 import org.antlr.v4.runtime.tree.TerminalNode;
@@ -594,35 +593,21 @@ public abstract class Parser extends Recognizer<Token, ParserATNSimulator> {
 	}
 
 	/** How to create a token leaf node associated with a parent.
-	 *  Typically, the terminal node to create is not a function of the parent
-	 *  but this method must still set the parent pointer of the terminal node
-	 *  returned. I would prefer having {@link ParserRuleContext#addAnyChild(ParseTree)}
-	 *  set the parent pointer, but the parent pointer is implementation dependent
-	 *  and currently there is no setParent() in {@link TerminalNode} (and can't
-	 *  add method in Java 1.7 without breaking backward compatibility).
+	 *  Typically, the terminal node to create is not a function of the parent.
 	 *
 	 * @since 4.6.1
 	 */
 	public TerminalNode createTerminalNode(ParserRuleContext parent, Token t) {
-		TerminalNodeImpl node = new TerminalNodeImpl(t);
-		node.parent = parent;
-		return node;
+		return new TerminalNodeImpl(t);
 	}
 
 	/** How to create an error node, given a token, associated with a parent.
-	 *  Typically, the error node to create is not a function of the parent
-	 *  but this method must still set the parent pointer of the terminal node
-	 *  returned. I would prefer having {@link ParserRuleContext#addAnyChild(ParseTree)}
-	 *  set the parent pointer, but the parent pointer is implementation dependent
-	 *  and currently there is no setParent() in {@link ErrorNode} (and can't
-	 *  add method in Java 1.7 without breaking backward compatibility).
+	 *  Typically, the error node to create is not a function of the parent.
 	 *
 	 * @since 4.6.1
 	 */
 	public ErrorNode createErrorNode(ParserRuleContext parent, Token t) {
-		ErrorNodeImpl node = new ErrorNodeImpl(t);
-		node.parent = parent;
-		return node;
+		return new ErrorNodeImpl(t);
 	}
 
 	protected void addContextToParseTree() {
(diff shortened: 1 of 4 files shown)


runtime/Java/src/org/antlr/v4/runtime/Parser.java
```diff
@@ -594,35 +593,21 @@ public Token consume() {
...
-	 *  add method in Java 1.7 without breaking backward compatibility).
+	 *  Typically, the terminal node to create is not a function of the parent.
 	 *
 	 * @since 4.6.1
 	 */
 	public TerminalNode createTerminalNode(ParserRuleContext parent, Token t) {
-		TerminalNodeImpl node = new TerminalNodeImpl(t);
-		node.parent = parent;
-		return node;
+		return new TerminalNodeImpl(t);
```
mike-lischke: The parameter `parent` should go. It's no longer used and produces warnings e.g. in the C++ target.

runtime/Java/src/org/antlr/v4/runtime/Parser.java
```diff
@@ -594,35 +593,21 @@ public Token consume() {
...
-	 *  add method in Java 1.7 without breaking backward compatibility).
+	 *  Typically, the error node to create is not a function of the parent.
 	 *
 	 * @since 4.6.1
 	 */
 	public ErrorNode createErrorNode(ParserRuleContext parent, Token t) {
-		ErrorNodeImpl node = new ErrorNodeImpl(t);
-		node.parent = parent;
-		return node;
+		return new ErrorNodeImpl(t);
```
mike-lischke: Same here.

runtime/Java/src/org/antlr/v4/runtime/ParserRuleContext.java
```diff
@@ -101,9 +102,8 @@ public void copyFrom(ParserRuleContext ctx) {
 			this.children = new ArrayList<>();
 			// reset parent pointer for any error nodes
 			for (ParseTree child : ctx.children) {
-				if ( child instanceof ErrorNodeImpl ) {
-					this.children.add(child);
-					((ErrorNodeImpl) child).parent = this;
+				if ( child instanceof ErrorNode ) {
+					addChild((ErrorNode)child);
```
mike-lischke: This is a bit weird code. You are not copying the error node, but actually create references from 2 contexts and only set the new context as parent. However, it's still in the source context, since you don't remove it from there. What is the real intention here? Really move the node or create a duplicate (like the method name suggests).

runtime/Java/src/org/antlr/v4/runtime/ParserRuleContext.java
```diff
@@ -139,12 +139,18 @@ public RuleContext addChild(RuleContext ruleInvocation) {
 		return addAnyChild(ruleInvocation);
```
mike-lischke: You don't want to set the parent of the ruleInvocation context like you do for terminal and error nodes?

mike-lischke:
Also, what is the meaning of a setter for a property which is directly accessbile? The `parent` field is public in TerminalNodeImpl. I fail to see how this makes the code better. A setter only make sense if you have side effects when changing a member,

runtime/Java/src/org/antlr/v4/runtime/Parser.java
parrt: the leaf might be a function of the parent though, right?

parrt:
@mike-lischke TerminalNode is an interface not a class so no field is available.

runtime/Java/src/org/antlr/v4/runtime/ParserRuleContext.java
parrt: no, that is bound up in the parsing functionality; i know it's odd.

runtime/Java/src/org/antlr/v4/runtime/ParserRuleContext.java
parrt: The error nodes should be *moved* to the new context in `copyFrom` as it's used only for flipping contexts per the comment.

runtime/Java/src/org/antlr/v4/runtime/Parser.java
mike-lischke: Hmm, in what sense? The parent is set by the caller, so why pass it into this method?

runtime/Java/src/org/antlr/v4/runtime/ParserRuleContext.java
mike-lischke: As I said, a bit weird. You have a copy function which moves something :-) At least I have it then like you want. But the Java code is still buggy. The error node is still in the children list of the source context, since you don't remove it from there.

runtime/Java/src/org/antlr/v4/runtime/Parser.java
parrt: Well, one might want to create a different *kind* of leaf (i.e., type) if the parent is an expression versus a declaration.  For example, an ID token might become VarRef vs VarDef leaf? Just leaving it flexible but I can see removing it if you guys feel strongly.

runtime/Java/src/org/antlr/v4/runtime/ParserRuleContext.java
parrt: The node copied from is candidate for GC as all refs to it disappear; figured it's a waste to alter it when it's dead.  Maybe a better comment? :)

mike-lischke: Well, that is implicit knowledge, which is not visible at this point in code and for some targets like C++ it is very important to know what is managing the memory. And if you one day decide to keep the source context around you certainly forgot by that time that it contains nodes which are supposed to be in a different context.

runtime/Java/src/org/antlr/v4/runtime/Parser.java
mike-lischke: Ok, I see your point. Have to manually take care about that parameter then in C++ (not the first time, there are many other places with the same kind of issue).

PR #1675, opened 2017-02-19
Tree factory

parrt:
Not sure about this one at all but creating PR for discussion. Some pieces might be useful.


mike-lischke:
That looks partially like the tracker I had to add in the C++ target to manage parser rule contexts. I'm thinking a while already about a way to keep parse trees around while the parser already had new runs, so that you can assign these trees e.g. to symbol tables. Keeping the parser rule contexts in memory is probably not a big deal (just need to think how to manage the memory in C++) but what about the token references?

parrt:
The token refs are held in the token stream. the terminal nodes (pointing to them) are added as children of rule context so we shouldn't lose those refs.

mike-lischke:
From a memory efficiency standpoint we should probably not keep the full parse trees around including duplicates of all token refs. A more compact representation would be desirable, but maybe that's beyond ANTLR4's scope.

PR #1676, opened 2017-02-19
remove unnecessary installs in travis

ericvergnaud:


parrt:
thanks @ericvergnaud !!

.travis/before-install-linux-javascript.sh
```diff
@@ -3,7 +3,7 @@
 set -euo pipefail
 
 sudo apt-key adv --keyserver hkp://keyserver.ubuntu.com:80 --recv-keys 3FA7E0328081BFF6A14DA29AA6A19B38D3D831EF
-sudo add-apt-repository ppa:fkrull/deadsnakes -y
-sudo add-apt-repository ppa:rwky/nodejs -y
 sudo apt-get update -qq
+curl -sL https://deb.nodesource.com/setup_0.12 | sudo -E bash -
```
bhamiltoncx: Just curious, but would these instructions work as well?

http://austinpray.com/ops/2015/09/20/change-travis-node-version.html

In particular, it seems like what we really want is to add to `.travis.yml`:

```
env:
  - TRAVIS_NODE_VERSION="0.12"
```

.travis/before-install-linux-python3.sh
```diff
@@ -4,7 +4,6 @@ set -euo pipefail
 
 sudo apt-key adv --keyserver hkp://keyserver.ubuntu.com:80 --recv-keys 3FA7E0328081BFF6A14DA29AA6A19B38D3D831EF
 sudo add-apt-repository ppa:fkrull/deadsnakes -y
```
bhamiltoncx: Perhaps this would work more reliably?

http://stackoverflow.com/a/37660551/7572003

```
addons:
  apt:
    sources:
      - deadsnakes # source required so it finds the package definition below
    packages:
      - python3.5
```

ericvergnaud:
It’s the first time it breaks, so not sure there is much to improve.
I haven’t tried the TRAVIS_NODE_VERSION option but I suspect it will behave the same as the Python 3.5 one, i.e. will not operate because our builds are Java builds.

the root cause for our break was ppa suddenly removing some web pages.
Travis suffers from the same as can be seen with https://github.com/travis-ci/travis-ci/issues/7332 <https://github.com/travis-ci/travis-ci/issues/7332>

parrt:
Seems like the deadsnakes one looks more generic so i'll change that one.  The node js also seems to have a default spec rather than needing special download:

```
language: node_js
node_js:
  - "0.12"
```

I'll try this out and see if it works.

parrt:
Looks like travis lint has other issues. i'll fix those too:

```
in matrix.include section: unexpected key dist, dropping
in matrix.include section: unexpected key osx_image, dropping
in matrix.include section: specified compiler, but setting is not relevant for java
in matrix.include section: unexpected key addons, dropping
in matrix.include.env section: does not support multiple values, dropping CXX=g++-5
```

parrt:
working on PR https://github.com/antlr/antlr4/pull/1678

PR #1679, opened 2017-02-20
Issue 1666

Thomasb81:
Hello

Here is a proposed fix.
I pass a pointer on program to ToSingleOperationPerIndex. So this method can remove the instruction from initial structure. This avoid to delete an operation already deleted.

The coding style of this fix is probably poor because of code conservative approach : ToSingleOperationPerIndex method content is not very clear to me.


runtime/Cpp/runtime/src/TokenStreamRewriter.cpp
```diff
@@ -251,7 +251,7 @@ std::string TokenStreamRewriter::getText(const Interval &interval) {
 }
 
 std::string TokenStreamRewriter::getText(const std::string &programName, const Interval &interval) {
-  std::vector<TokenStreamRewriter::RewriteOperation*> rewrites = _programs[programName];
+  std::vector<TokenStreamRewriter::RewriteOperation*> *rewrites = &_programs[programName];
```
mike-lischke: Good find! However, I'd prefer to use a reference here, instead of a pointer, e.g.

`std::vector<TokenStreamRewriter::RewriteOperation*> &rewrites = _programs[programName];`

runtime/Cpp/runtime/src/TokenStreamRewriter.cpp
```diff
@@ -263,7 +263,7 @@ std::string TokenStreamRewriter::getText(const std::string &programName, const I
     start = 0;
   }
 
-  if (rewrites.empty() || rewrites.empty()) {
+  if (rewrites->empty() || rewrites->empty()) {
```
mike-lischke: Would not be unnecessary when using a reference.

runtime/Cpp/runtime/src/TokenStreamRewriter.cpp
```diff
@@ -305,7 +305,9 @@ std::string TokenStreamRewriter::getText(const std::string &programName, const I
 }
 
 std::unordered_map<size_t, TokenStreamRewriter::RewriteOperation*> TokenStreamRewriter::reduceToSingleOperationPerIndex(
-  std::vector<TokenStreamRewriter::RewriteOperation*> rewrites) {
+  std::vector<TokenStreamRewriter::RewriteOperation*> *rewrites2) {
+
+  std::vector<TokenStreamRewriter::RewriteOperation*> rewrites = *rewrites2;
```
mike-lischke: Is this really necessary? The entries in the `rewrite` vector are properly cleaned up. In the Java implemenation the null elements stay in the array, while you now remove them completely. I'm not 100% sure this won't produce other trouble. In fact I believe all the other changes beside the reference above is not really needed. The key issue was that a copy of the vector was used instead of the original vector in the program.

runtime/Cpp/runtime/src/TokenStreamRewriter.cpp
```diff
@@ -305,7 +305,9 @@ std::string TokenStreamRewriter::getText(const std::string &programName, const I
 }
 
 std::unordered_map<size_t, TokenStreamRewriter::RewriteOperation*> TokenStreamRewriter::reduceToSingleOperationPerIndex(
-  std::vector<TokenStreamRewriter::RewriteOperation*> rewrites) {
+  std::vector<TokenStreamRewriter::RewriteOperation*> *rewrites_ptr) {
+
+  std::vector<TokenStreamRewriter::RewriteOperation*> &rewrites = *rewrites_ptr;
```
mike-lischke: This is not needed. Just restore the original parameter list, but make `rewrites` a reference as well and that's it.

Thomasb81:
I guess it should be now, right ? :)

parrt:
@mike-lischke blessed?

PR #1683, opened 2017-02-21
New tool utility class Unicode

bhamiltoncx:
In https://github.com/antlr/antlr4/pull/1633/#issuecomment-280778324 , @mike-lischke suggested adding support for Unicode property escapes like `\p{L}` in ANTLR grammars.

This PR adds a new utility class `Unicode` to the tool which uses the ICU library (which provides all the up-to-date Unicode data we need) to generate various `IntervalSet` maps to trivially provide these property escapes.

We can use these so users can generate Unicode lexer grammars using Unicode escapes to represent sets of code points with general categories, binary properties, and scripts:

```
UPPER_CASE = '\p{Lu}'
LATIN = '\p{Latin}'
CJK = ('\p{Han}' | '\p{Hiragana}' | '\p{Katakana}' | '\p{Hangul}')
EMOJI = '\p{Emoji}'
```

### DIFF
--- a/tool/pom.xml
+++ b/tool/pom.xml
@@ -42,6 +42,11 @@
 			<artifactId>javax.json</artifactId>
 			<version>1.0.4</version>
 		</dependency>
+		<dependency>
+			<groupId>com.ibm.icu</groupId>
+			<artifactId>icu4j</artifactId>
+			<version>58.2</version>
+		</dependency>
 	</dependencies>
 
   <build>
--- a/tool-testsuite/test/org/antlr/v4/test/tool/TestUnicode.java
+++ b/tool-testsuite/test/org/antlr/v4/test/tool/TestUnicode.java
@@ -1,0 +1,80 @@
+/*
+ * Copyright (c) 2012-2017 The ANTLR Project. All rights reserved.
+ * Use of this file is governed by the BSD 3-clause license that
+ * can be found in the LICENSE.txt file in the project root.
+ */
+
+package org.antlr.v4.test.tool;
+
+import java.util.Map;
+
+import org.antlr.v4.codegen.Unicode;
+import org.antlr.v4.runtime.misc.IntervalSet;
+
+import org.junit.Test;
+
+import static org.junit.Assert.assertFalse;
+import static org.junit.Assert.assertEquals;
+import static org.junit.Assert.assertTrue;
+
+public class TestUnicode {
+	@Test
+	public void testUnicodeCategoryCodes() {
+		Map<String, IntervalSet> unicodeCategoryCodesToCodePointRanges = Unicode.getUnicodeCategoryCodesToCodePointRanges();
+		assertTrue(unicodeCategoryCodesToCodePointRanges.get("Lu").contains('X'));
+		assertFalse(unicodeCategoryCodesToCodePointRanges.get("Lu").contains('x'));
+		assertTrue(unicodeCategoryCodesToCodePointRanges.get("Ll").contains('x'));
+		assertFalse(unicodeCategoryCodesToCodePointRanges.get("Ll").contains('X'));
+		assertTrue(unicodeCategoryCodesToCodePointRanges.get("L").contains('X'));
+		assertTrue(unicodeCategoryCodesToCodePointRanges.get("L").contains('x'));
+		assertTrue(unicodeCategoryCodesToCodePointRanges.get("N").contains('0'));
+		assertTrue(unicodeCategoryCodesToCodePointRanges.get("Z").contains(' '));
+	}
+
+	@Test
+	public void testUnicodeCategoryCodesToNames() {
+		Map<String, String> unicodeCategoryCodesToNames = Unicode.getUnicodeCategoryCodesToNames();
+		assertEquals("Lowercase_Letter", unicodeCategoryCodesToNames.get("Ll"));
+		assertEquals("Letter", unicodeCategoryCodesToNames.get("L"));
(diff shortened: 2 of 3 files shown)


tool/pom.xml
```diff
@@ -42,6 +42,11 @@
 			<artifactId>javax.json</artifactId>
 			<version>1.0.4</version>
 		</dependency>
+		<dependency>
+			<groupId>com.ibm.icu</groupId>
```
parrt: @bhamiltoncx how big is this lib, btw? I make a big jar of all dependencies for deployment.

bhamiltoncx:
According to https://mvnrepository.com/artifact/com.ibm.icu/icu4j/58.2 ,
it's 2.1 MB.

This is only for the tool, not the runtime, so hopefully space-conscious
folks will use the runtime-only JAR.

bhamiltoncx:
Hmm, that website doesn't seem to be telling the truth. Seems like it's actually 11 MB.

```
% ls -lh icu4j*.jar
-rw-r--r--@ 11M Feb 21 10:44 icu4j-58.2.jar
```

Hopefully that's not a problem. If it is, we can write our own parser for the raw data:

http://source.icu-project.org/repos/icu/icu/tags/latest/source/data/unidata/ppucd.txt

which shouldn't be too hard. What do you think?

bhamiltoncx:
Thinking about the `\p{...}` escape, we probably want to require it to be inside a `LEXER_CHAR_SET`:

```
[\p{L}\p{N}]
```

because if we allow it inside an arbitrary string literal like `'foo \p{Lu} bar'`, it'll mean string literals now become a bit more complicated, since they can contain sets.

parrt:
excellent point about `\p`. let's make the change.

concerning the size, the current jar is only 1.6M so we would effectively be making it an order of magnitude bigger. it's still small but it seems a hefty price to pay for `\p` feature.

 can we extract just the class we need from that jar?

bhamiltoncx:
> can we extract just the class we need from that jar?

It's probably easier to just write a parser for `ppucd.txt`. I'll take care of that.

parrt:
Ok, are you sure that this feature is worth the effort?

bhamiltoncx:
> Ok, are you sure that this feature is worth the effort?

If I don't do it as part of ANTLR, I'll likely have to do it separately anyway. I'd rather more people be able to make full use of the Unicode feature, so I'll take care of it.

bhamiltoncx:
So, the parser itself isn't too tricky, but there are a number of special cases which ICU handles only in source:

http://source.icu-project.org/repos/icu/icu4j/tags/latest/main/classes/core/src/com/ibm/icu/impl/UCharacterProperty.java

```
    BinaryProperty[] binProps={
        /*
         * Binary-property implementations must be in order of corresponding UProperty,
         * and there must be exactly one entry per binary UProperty.
         */
        new BinaryProperty(1, (1<<ALPHABETIC_PROPERTY_)),
        new BinaryProperty(1, (1<<ASCII_HEX_DIGIT_PROPERTY_)),
        new BinaryProperty(SRC_BIDI) {  // UCHAR_BIDI_CONTROL
            @Override
            boolean contains(int c) {
                return UBiDiProps.INSTANCE.isBidiControl(c);
            }
        },
        new BinaryProperty(SRC_BIDI) {  // UCHAR_BIDI_MIRRORED
            @Override
            boolean contains(int c) {
                return UBiDiProps.INSTANCE.isMirrored(c);
            }
        },
        new BinaryProperty(1, (1<<DASH_PROPERTY_)),
        new BinaryProperty(1, (1<<DEFAULT_IGNORABLE_CODE_POINT_PROPERTY_)),
        new BinaryProperty(1, (1<<DEPRECATED_PROPERTY_)),
        new BinaryProperty(1, (1<<DIACRITIC_PROPERTY_)),
        new BinaryProperty(1, (1<<EXTENDER_PROPERTY_)),
        new BinaryProperty(SRC_NFC) {  // UCHAR_FULL_COMPOSITION_EXCLUSION
            @Override
            boolean contains(int c) {
                // By definition, Full_Composition_Exclusion is the same as NFC_QC=No.
                Normalizer2Impl impl=Norm2AllModes.getNFCInstance().impl;
                return impl.isCompNo(impl.getNorm16(c));
            }
        },

        (snip)
```

This isn't something that looks very fun to implement in ANTLR.

Another option is to have a new Maven target which depends on ICU and generates either Java code or a very simple resource file which the Tool depends upon.

bhamiltoncx:
Heh. I like the Maven example on generating sources:

https://maven.apache.org/guides/mini/guide-generating-sources.html

> Let's run though a short example to try and help. To generate sources you must first have a plugin that participates in the generate-sources phase like the ANTLR4 Maven Plugin.
> 
> So this is all fine and dandy, we have a plugin that wants to generate some sources from a Antlr4 grammar but how do we use it. You need to specify that you want to use it in your POM:

parrt:
:) Yeah, we can easily add a maven profile that generates a Java class as an initial step before the real build. We used to do that to generate the unit tests from templates, rather than the possibly too clever descriptors that I changed it to :)

bhamiltoncx:
OK, I'll add a step to use a template to generate the source file. That way the tool won't depend on ICU.

mike-lischke:
I'm fine with the decision to allow Unicode char classes only in lexer char sets. That may also teach people to define their string literals in the lexer, not in the parser.

bhamiltoncx:
OK, sent a PR to remove the ICU dependency: https://github.com/antlr/antlr4/pull/1687

PR #1685, opened 2017-02-21
Fix C# Pair.cs, Right arrow escaping in XML Comments

### LINKED ISSUE #1618: [CSharp] &rarr; (Right arrow) escaping in XML Comments
Currently if we build CSharp runtime with either VS or dotnet cli, we get these warning messages:

```
/home/dong/github/fork/antlr4/runtime/CSharp/runtime/CSharp/Antlr4.Runtime/Atn/ParserATNSimulator.cs(260,57): warning CS1570: XML comment has badly formed XML -- 'Reference to undefined entity 'rarr'.'
/home/dong/github/fork/antlr4/runtime/CSharp/runtime/CSharp/Antlr4.Runtime/Atn/ParserATNSimulator.cs(261,63): warning CS1570: XML comment has badly formed XML -- 'Reference to undefined entity 'rarr'.'
/home/dong/github/fork/antlr4/runtime/CSharp/runtime/CSharp/Antlr4.Runtime/Atn/ParserATNSimulator.cs(2050,51): warning CS1570: XML comment has badly formed XML -- 'Reference to undefined entity 'rarr'.'
```

In the code comment we are trying to use the escaping sequence to display a Right arrow (as in Java, Python target). The problem is, we get this in the generated xml file after build:
```
<!-- Badly formed XML comment ignored for member "F:Antlr4.Runtime.Atn.ParserATNSimulator.mergeCache" -->
```

Apparently msbuild can't recognize this escaping sequence at all, so it give up on trying to generate doc for this member. I have also tested with CDATA, seems the whole CDATA chunk was output as is, not just the content within as I expected.

The XML Comment has only limited ability to understand some escaping sequence, e.g. the most popular ```&lt; &gt;``` thing, but it can't recognize ```&rarr;``` if we use it, we can see a red tilde under the & char.

On the other hand, if we do 
```csharp
/// <summary>
///  test &#x2192;
/// </summary>
public class MergeCache
```
(http://right-arrow.net/)

Then we can have the arrow showing in Intellisence.

Not really sure what is the intended purpose of those comments, if it's for generating xml doc, then better replace ```&rarr;``` with ```&#x2192;```

If it's for people to read it directly in the code for better understanding, I guess it's better to replace that with a simple ->

### LINKED ISSUE #1634: C#Antlr4.6 runtime class Pair<>contains a couple errors.
I have been porting codebuff to C#. In doing so, I noticed several problems with Pair<> in the lastest version on the master branch of https://github.com/antlr/antlr4/blob/master/runtime/CSharp/runtime/CSharp/Antlr4.Runtime/Misc/Pair.cs.

(1) The ToString() method contains a call to String.Format(). The format parameter is not correct. It uses "%s,%s" which is Java/C++/etc. Please use "{0}, {1}" and read https://msdn.microsoft.com/en-us/library/system.string.format(v=vs.110).aspx#Format_Brief.

(2) When used with HashSet<Pair<int,int>>, there should not be duplicates. However, the current version does. In other words, when I compare the type between Java and C#, they don't function identically. Consider the following test code:
`            HashSet<Pair<int, int>> the_hash_set = new HashSet<Pair<int, int>>();
            the_hash_set.Add(new Pair<int, int>(1, 2));
            the_hash_set.Add(new Pair<int, int>(1, 2));
            the_hash_set.Add(new Pair<int, int>(2, 1));
            foreach (var p in the_hash_set)
                System.Console.WriteLine(p);
`
I should see: "(1,2) (2,1)" (as in Java, which I tested), but I see instead "(1,2) (1,2) (2,1)" using this C# code. Looking at the C# source code Pair.cs, and comparing with the Java source, the test in the C# should be fixed. One can either implement the equivalent comparison classes and code in C#, or use something like this:

`/* Copyright (c) 2012-2016 The ANTLR Project. All rights reserved.
 */

using System;

namespace ConsoleApplication16
{

    public class Pair<A, B>
    {
        public readonly A a;
        public readonly B b;

        public Pair(A a, B b)
        {
            this.a = a;
            this.b = b;
        }

        public override bool Equals(object obj)
        {
            if (obj == this)
            {
                return true;
            }
            else if (!(obj is Pair<A, B>))
            {
                return false;
            }

            Pair<A, B> other = (Pair<A, B>)obj;
            if (a != null)
            {
                if (b != null)
                    return a.Equals(other.a) && b.Equals(other.b);
                else
                    return a.Equals(other.a) && other.b == null;
            }
            else
            {
                if (b != null)
                    return other.a == null && b.Equals(other.b);
                else
                    return other.a == null && other.b == null;
            }
        }

        public override int GetHashCode()
        {
            int hash = MurmurHash.Initialize();
            hash = MurmurHash.Update(hash, a);
            hash = MurmurHash.Update(hash, b);
            return MurmurHash.Finish(hash, 2);
        }

        public override String ToString()
        {
            return String.Format("({0}, {1})", a, b);
        }
    }

}

`

KvanTTT:
fixes #1634.
fixes #1618.

### DIFF
--- a/runtime/CSharp/runtime/CSharp/Antlr4.Runtime/Atn/ParserATNSimulator.cs
+++ b/runtime/CSharp/runtime/CSharp/Antlr4.Runtime/Atn/ParserATNSimulator.cs
@@ -1,4 +1,4 @@
-/* Copyright (c) 2012-2016 The ANTLR Project. All rights reserved.
+﻿/* Copyright (c) 2012-2016 The ANTLR Project. All rights reserved.
  * Use of this file is governed by the BSD 3-clause license that
  * can be found in the LICENSE.txt file in the project root.
  */
@@ -257,8 +257,8 @@ namespace Antlr4.Runtime.Atn
 		 *  Don't keep around as it wastes huge amounts of memory. DoubleKeyMap
 		 *  isn't synchronized but we're ok since two threads shouldn't reuse same
 		 *  parser/atnsim object because it can only handle one input at a time.
-		 *  This maps graphs a and b to merged result c. (a,b)&rarr;c. We can avoid
-		 *  the merge if we ever see a and b again.  Note that (b,a)&rarr;c should
+		 *  This maps graphs a and b to merged result c. (a,b)→c. We can avoid
+		 *  the merge if we ever see a and b again.  Note that (b,a)→c should
 		 *  also be examined during cache lookup.
 		 */
 		protected MergeCache mergeCache;
@@ -2047,7 +2047,7 @@ namespace Antlr4.Runtime.Atn
 		 we don't consider any conflicts that include alternative 2. So, we
 		 ignore the conflict between alts 1 and 2. We ignore a set of
 		 conflicting alts when there is an intersection with an alternative
-		 associated with a single alt state in the state&rarr;config-list map.
+		 associated with a single alt state in the state→config-list map.
 
 		 It's also the case that we might have two conflicting configurations but
 		 also a 3rd nonconflicting configuration for a different alternative:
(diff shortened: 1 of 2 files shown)


ericvergnaud:
blessed

parrt:
Thanks!

PR #1687, opened 2017-02-22
Use code generation in new tool-codegen build step instead of making tool depend on ICU

bhamiltoncx:
In f384ef5fb49ad5e7d6b9b5f4d44e70021fd2c3d4, I made the tool depend on ICU at runtime to generate the Unicode ranges needed to implement `\p{Letter}` style escapes supporting the full Unicode range.

However, this caused the tool jar to bloat enormously in size:

# Before f384ef5fb49ad5e7d6b9b5f4d44e70021fd2c3d4

```
% mvn -DskipTests install && ls -lh tool/target/*.jar
-rw-r--r--. 1 beng users 1.6M Feb 22 10:55 antlr4-4.6.1-SNAPSHOT-complete.jar
-rw-r--r--. 1 beng users 783K Feb 22 10:55 antlr4-4.6.1-SNAPSHOT.jar
-rw-r--r--. 1 beng users 519K Feb 22 10:55 antlr4-4.6.1-SNAPSHOT-sources.jar
```

# After f384ef5fb49ad5e7d6b9b5f4d44e70021fd2c3d4

```
% mvn -DskipTests install && ls -lh tool/target/*.jar
-rw-r--r--. 1 beng users  13M Feb 22 10:54 antlr4-4.6.1-SNAPSHOT-complete.jar
-rw-r--r--. 1 beng users 786K Feb 22 10:54 antlr4-4.6.1-SNAPSHOT.jar
-rw-r--r--. 1 beng users 521K Feb 22 10:54 antlr4-4.6.1-SNAPSHOT-sources.jar
```

This diff fixes the issue by introducing a new Maven build step `tool-codegen` which uses a `StringTemplate` to code-generate a new class `UnicodeData` which does not depend on ICU.

I also improved the interface somewhat, and made the property names and aliases case-insensitive.

To work around the Java 64k bytecode limit, I split up each Unicode property into a separate static method.

The end result gets us down from a **12 MB** size increase to a **300 KB** size increase in the complete tool jar.

```
% mvn -DskipTests install && ls -lh tool/target/*.jar
-rw-r--r--. 1 beng users 1.9M Feb 22 10:45 antlr4-4.6.1-SNAPSHOT-complete.jar
-rw-r--r--. 1 beng users 1.1M Feb 22 10:45 antlr4-4.6.1-SNAPSHOT.jar
-rw-r--r--. 1 beng users 646K Feb 22 10:45 antlr4-4.6.1-SNAPSHOT-sources.jar
```

I can probably get this down further by being clever with a few of the general category properties which are defined as unions of other categories.


parrt:
Awesome.  I can also see that your mvn kung fu is much stronger than mine.

Issue #1540, opened 2016-12-21
Performance tips


ashishnegi:
We are planning to replace our hand written parser with ANTLR4 in production for parsing our graph database language spec. ANTLR grammar looks elegant and precise.
Our language spec is a variant of GraphQL. The user queries that we will have to parse look like :

```
// List of directors with whom Tom Hanks has worked
{
  debug(_xid_: m.0bxtg) {
    type.object.name.en
    film.actor.film {
      film.performance.film {
        film.film.directed_by {
          type.object.name.en
        }
      }
    }
  }
}
```

We started benchmarking from the simplest subset [grammar](https://github.com/dgraph-io/dgraph/blob/2477b83fdf90287768293f146f9e4b4ca1eb2ca9/antlr4go/graphqlpm/GraphQLPM.g4).
[Benchmarks](https://github.com/dgraph-io/dgraph/blob/2477b83fdf90287768293f146f9e4b4ca1eb2ca9/antlr4go/graphqlpm/main_test.go#L133) :
```
dgraph/antlr4go/graphqlpm$ gotb -test.run=XXX -benchtime=2s -v
BenchmarkQuery/q1-4                 5000       1221967 ns/op
BenchmarkQuery/q2-4                 2000       1710850 ns/op
BenchmarkQuery/q3-4                 3000       1230518 ns/op
BenchmarkQuery/q4-4                 3000       1568564 ns/op
BenchmarkQuery/q5-4                 2000       1803392 ns/op
PASS
ok      github.com/dgraph-io/dgraph/antlr4go/graphqlpm    22.179s
```

We expected these numbers to be under 0.05 ms. They are currently around 1.5 ms.

Here are comparisons with handwritten parser and antlr golang parser over practical [queries](https://github.com/dgraph-io/dgraph/blob/8ccf5cb5246099df4a01918536da9a46b25e6f75/query/query_test.go#L2263) :
[Benchmarks](https://github.com/dgraph-io/dgraph/blob/8ccf5cb5246099df4a01918536da9a46b25e6f75/query/query_test.go#L2516) :
```
query$ gotb -test.run=XXX -benchtime=2s -v
BenchmarkQueryParse/spielberg:handwitten:-4               100000         46918 ns/op
BenchmarkQueryParse/spielberg:antlr:-4                      2000       1741364 ns/op
BenchmarkQueryParse/tomhanks:handwitten:-4                100000         25982 ns/op
[...]

Issue #192, opened 2013-03-21
Poor performance of expression parsing


agacek:
Using Antlr 4.0 I've run into a performance problem with parsing a simple expression language. Here is the grammar:

```
grammar Expr;

program: expr;

expr: ID
    | 'not' expr
    | expr 'and' expr
    | expr 'or' expr
    ;

ID: [a-zA-Z_][a-zA-Z_0-9]*;
WS: [ \t\n\r\f]+ -> skip;
ERROR: .;
```

And here is the sample file I'm trying to parse:

```
not X1 and not X2 and not X3 and not X4 and not X5 and not X6 and not X7 and not X8 and not X9 and not X10 and not X11 and not X12 or
    X1 and not X2 and not X3 and not X4 and not X5 and not X6 and not X7 and not X8 and not X9 and not X10 and not X11 and not X12 or
not X1 and     X2 and not X3 and not X4 and not X5 and not X6 and not X7 and not X8 and not X9 and not X10 and not X11 and not X12 or
not X1 and not X2 and     X3 and not X4 and not X5 and not X6 and not X7 and not X8 and not X9 and not X10 and not X11 and not X12 or
not X1 and not X2 and not X3 and     X4 and not X5 and not X6 and not X7 and not X8 and not X9 and not X10 and not X11 and not X12 or
not X1 and not X2 and not X3 and not X4 and     X5 and not X6 and not X7 and not X8 and not X9 and not X10 and not X11 and not X12 or
not X1 and not X2 and not X3 and not X4 and not X5 and     X6 and not X7 and not X8 and not X9 and not X10 and not X11 and not X12 or
not X1 and not X2 and not X3 and not X4 and not X5 and not X6 and     X7 and not X8 and not X9 and not X10 and not X11 and not X12 or
not X1 and not X2 and not X3 and not X4 and not X5 and not X6 and not X7 and     X8 and not X9 and not X10 and not X11 and not X12 or
not X1 and not X2 and not X3 and not X4 and not X5 and not X6 and not X7 and not X8 and     X9 and not X10 and not X11 and not X12 or
not X1 and not X2 and not X3 and not X4 and not X5 and not X6 and not X7 and not X8 and not X9 and     X10 and not X11 and not X12 or
not X1 and not X2 and not X3 and not X4 and not X5 and not X6 and not X7 and not X8 and not X9 and not X10 and     X11 and not X12 or
[...]

Issue #1620, opened 2017-01-21
XPathLexer not updated in release process


davesisson:
Similar to #796 the XPathLexer.class was built with 4.5.3 instead of 4.6.  As a result the following error message will result whenever the XPath libraries are used:

"ANTLR Tool version 4.5.3 used for code generation does not match the current runtime version 4.6"

sharwell:
Should definitely include a regression test as well to make sure this doesn't happen in the future.

davesisson:
Turns out the problem here is the way we're building it (in a different way).  In order to avoid a cyclic dependency we're building XPathLexer with an older version of the ANTLR tool:

https://github.com/antlr/antlr4/blob/master/runtime/Java/pom.xml#L75

If we can fix this somehow (presumably by not using antlr4-maven-plugin to build the XPathLexer.g4 file) we can make sure it's always built the right way.

But yes, a regression test seems warranted given that this has happened twice so far.

juhakivekas:
What is currently the best solution to overcome this problem? Is there a workaround that may include cyclic dependencies. I would like to use this feature of the antlr tool and I haven't compiled any part of the project yet.

Given an example of regression tests (I can only find unit tests) I could create the needed test.

juhakivekas:
For anyone looking for a quick fix, reverting to 4.5.3 works fine.

    git checkout 4.5.3
    mvn -DskipTests install

davesisson:
An alternative I thought of last night was to modify the antlr4-maven-plugin to allow it to point to a smaller dependency.  The maven plugin presumably depends on the whole antlr4 target (which this is a part of) where we only need the tool to complete this particular action.

Maybe the plugin should always be modified to only require the tool.  (That would mean we'd have to give the runtime dependency to the generated code separately.)

parrt:
I wish I knew mvn better. I'm afraid to touch the plug-in for fear of screwing it up. I'm not even sure of how to make this work. @sharwell  I think we had this solved in the past right? How do we get XPath.g4 generated to Java and compiled after a complete build so that we can generated with the same version as the software?

parrt:
Ok, way to solve this once and for all is to do away with the .g4 based lexer. I will build one by hand quickly.

parrt:
Any comments before I accept the [pull](https://github.com/antlr/antlr4/pull/1748)?

PR #1689, opened 2017-02-23
Go micro performance. lexer atn config using murmur hash 32.

wxio:
The Go target needs some TLC at both a micro and macro level. A a micro level the lexer (and probably the parser) is slow, around 40x slower than a hand rolled one. Where the cpp target is only around 40% slower (see https://github.com/antlr/antlr4/issues/1540 & https://github.com/pboyer/antlr4/issues/92). At a macro level the LR optimisation (see https://github.com/antlr/antlr4/issues/192) have not been ported to the Go runtime.

This PR changed the LexerATNConfig hash code from a string that was hashed with fnv to hashcode ints that are murmur32 hashed. This gives about a 30% (benchmark number below) improvement (if implemented correctly), so there is still low hanging fruit in the 40x slow down. I intend to revisit this when I have some time or performance becomes an issue. My currently environment doesn't support profiling, which is probably an effective way of finding micro issues. Anyone interested in helping with this, I would be more than happy to collaborate.

@parrt and @sharwell any gotchas in implementing the LR changes?

**Comments**
**Benchmarking** - At this point in Antlr4's maturity it would be nice to have a benchmarking framework to complement  the unit testing framework. Go's benchmarking is an extension of testing and by the looks of it was influenced by [caliper](https://github.com/google/caliper), a Java benchmarking framework. Any suggestion of similar tools for the other targets and what such a framework might look like? Do any travis add-ins support reporting on benchmark data?

**Code coverage** - Whilst looking at the Go and Java code I came across another micro issue. In the Java runtime the `boolean add(T t)` in `Array2DHashSet` class will always return true (ie new element added), this behaviour also exists in the Go runtime. This doesn't effect behaviour, but without benchmarking it is difficult to tell the benefit from fixing it or replacing it with a hash set from a thirdparty library (eg [guava](https://github.com/google/guava) r19 is java 1.7 compatible). It wouldn't be a surprise if many micro issues exist across the runtime code bases. In this case the code in question dates back to 2012. So a deeper question is how to find them. The first answer that springs to mind is code coverage. Has anyone setup code coverage on Antlr4 using one of the travis add-ins? What other tools and techniques might be of interest?

Cheers
Gary

Go benchmarking numbers for this PR (see [modified dgraph benchmark](https://github.com/millergarym/dgraph/blob/bench-antlr4/antlr4go/lexer/benchmark_test.go))
@ashishne thought this might interest you.
before
```
bash-3.2$ go test -test.run=XXX -v  -benchtime=2s -bench . -cpuprofile cprof.prof
BenchmarkQuery/q1.1-12             10000            398395 ns/op
BenchmarkQuery/q1.2-12             10000            432751 ns/op
BenchmarkQuery/q2.1-12             10000            446644 ns/op
BenchmarkQuery/q2.2-12              5000            514398 ns/op
BenchmarkQuery/q3.1-12              5000            490114 ns/op
BenchmarkQuery/q3.2-12              5000            504340 ns/op
BenchmarkQuery/q4.1-12              5000            537611 ns/op
BenchmarkQuery/q4.2-12              5000            552284 ns/op
BenchmarkQuery/q5.1-12              5000            538648 ns/op
BenchmarkQuery/q5.2-12              5000            552259 ns/op
```

after
```
bash-3.2$ go test -test.run=XXX -v  -benchtime=2s -bench .
BenchmarkQuery/q1.1-12             10000            277898 ns/op
BenchmarkQuery/q1.2-12             10000            299240 ns/op
BenchmarkQuery/q2.1-12             10000            307816 ns/op
BenchmarkQuery/q2.2-12             10000            351049 ns/op
BenchmarkQuery/q3.1-12             10000            336241 ns/op
BenchmarkQuery/q3.2-12             10000            352299 ns/op
BenchmarkQuery/q4.1-12             10000            374569 ns/op
BenchmarkQuery/q4.2-12             10000            384309 ns/op
BenchmarkQuery/q5.1-12             10000            374470 ns/op
BenchmarkQuery/q5.2-12             10000            386929 ns/op
```

From https://github.com/antlr/antlr4/issues/1540
```
query$ gotb -test.run=XXX -benchtime=2s -v
BenchmarkQueryParse/spielberg:handwitten:-4               100000         46918 ns/op
BenchmarkQueryParse/spielberg:antlr:-4                      2000       1741364 ns/op
BenchmarkQueryParse/tomhanks:handwitten:-4                100000         25982 ns/op
BenchmarkQueryParse/tomhanks:antlr:-4                       2000       1654579 ns/op
BenchmarkQueryParse/nestedquery:handwritten:-4             30000         73053 ns/op
BenchmarkQueryParse/nestedquery:antlr:-4                    1000       3385005 ns/op
PASS
ok      github.com/dgraph-io/dgraph/query    21.922s
```

### DIFF
--- a/runtime-testsuite/test/org/antlr/v4/test/runtime/go/BaseGoTest.java
+++ b/runtime-testsuite/test/org/antlr/v4/test/runtime/go/BaseGoTest.java
@@ -165,6 +165,9 @@ public class BaseGoTest implements RuntimeTestSupport {
 	}
 
 	private static void copyFile(File source, File dest) throws IOException {
+		if ( source.isDirectory() ) {
+			return;
+		}
 		InputStream is = new FileInputStream(source);
 		OutputStream os = new FileOutputStream(dest);
 		byte[] buf = new byte[4 << 10];
--- a/runtime/Go/antlr/atn_config.go
+++ b/runtime/Go/antlr/atn_config.go
@@ -247,17 +247,17 @@ func NewLexerATNConfig1(state ATNState, alt int, context PredictionContext) *Lex
 	return &LexerATNConfig{BaseATNConfig: NewBaseATNConfig5(state, alt, context, SemanticContextNone)}
 }
 
-func (l *LexerATNConfig) Hash() string {
-	var f string
-
-	if l.passedThroughNonGreedyDecision {
-		f = "1"
-	} else {
-		f = "0"
-	}
-
-	return fmt.Sprintf("%v%v%v%v%v%v", l.state.GetStateNumber(), l.alt, l.context, l.semanticContext, f, l.lexerActionExecutor)
-}
+// func (l *LexerATNConfig) Hash() string {
+// 	var f string
+//
+// 	if l.passedThroughNonGreedyDecision {
+// 		f = "1"
+// 	} else {
+// 		f = "0"
+// 	}
+//
+// 	return fmt.Sprintf("%v%v%v%v%v%v", l.state.GetStateNumber(), l.alt, l.context, l.semanticContext, f, l.lexerActionExecutor)
+// }
 
 func (l *LexerATNConfig) equals(other interface{}) bool {
 	var othert, ok = other.(*LexerATNConfig)
@@ -283,6 +283,24 @@ func (l *LexerATNConfig) equals(other interface{}) bool {
 	}
 
 	return l.BaseATNConfig.equals(othert.BaseATNConfig)
+}
+
+func (l *LexerATNConfig) HashCode() int {
+	var f int
+	if l.passedThroughNonGreedyDecision {
+		f = 1
+	} else {
+		f = 0
+	}
+	h := initHash(7)
+	h = update(h, l.state.HashCode())
+	h = update(h, l.alt)
+	h = update(h, l.context.HashCode())
+	h = update(h, l.semanticContext.HashCode())
+	h = update(h, f)
+	h = update(h, l.lexerActionExecutor.HashCode())
+	h = finish(h, 6)
+	return h
 }
 
 func checkNonGreedyDecision(source *LexerATNConfig, target ATNState) bool {
--- a/runtime/Go/antlr/atn_state.go
+++ b/runtime/Go/antlr/atn_state.go
@@ -49,6 +49,7 @@ type ATNState interface {
 	AddTransition(Transition, int)
 
 	String() string
+	HashCode() int
 }
 
 type BaseATNState struct {
@@ -120,6 +121,10 @@ func (as *BaseATNState) GetNextTokenWithinRule() *IntervalSet {
 
 func (as *BaseATNState) SetNextTokenWithinRule(v *IntervalSet) {
 	as.NextTokenWithinRule = v
+}
+
+func (as *BaseATNState) HashCode() int {
+	return as.stateNumber
 }
 
 func (as *BaseATNState) String() string {
--- a/runtime/Go/antlr/lexer_action.go
+++ b/runtime/Go/antlr/lexer_action.go
@@ -22,6 +22,7 @@ type LexerAction interface {
 	getIsPositionDependent() bool
 	execute(lexer Lexer)
 	Hash() string
+	HashCode() int
 	equals(other LexerAction) bool
 }
 
@@ -49,6 +50,10 @@ func (b *BaseLexerAction) getActionType() int {
 
 func (b *BaseLexerAction) getIsPositionDependent() bool {
 	return b.isPositionDependent
+}
+
+func (b *BaseLexerAction) HashCode() int {
+	return b.actionType
 }
 
 func (b *BaseLexerAction) Hash() string {
--- a/runtime/Go/antlr/lexer_action_executor.go
+++ b/runtime/Go/antlr/lexer_action_executor.go
@@ -14,6 +14,7 @@ package antlr
 type LexerActionExecutor struct {
 	lexerActions     []LexerAction
 	cachedHashString string
+	cachedHash       int
 }
 
 func NewLexerActionExecutor(lexerActions []LexerAction) *LexerActionExecutor {
@@ -30,8 +31,10 @@ func NewLexerActionExecutor(lexerActions []LexerAction) *LexerActionExecutor {
 	// of the performance-critical {@link LexerATNConfig//hashCode} operation.
 
 	var s string
+	l.cachedHash = initHash(57)
 	for _, a := range lexerActions {
 		s += a.Hash()
+		l.cachedHash = update(l.cachedHash, a.HashCode())
 	}
 
 	l.cachedHashString = s // "".join([str(la) for la in
@@ -159,6 +162,13 @@ func (l *LexerActionExecutor) execute(lexer Lexer, input CharStream, startIndex 
 
 func (l *LexerActionExecutor) Hash() string {
 	return l.cachedHashString
+}
+
+func (l *LexerActionExecutor) HashCode() int {
+	if l == nil {
+		return 61
+	}
+	return l.cachedHash
 }
 
 func (l *LexerActionExecutor) equals(other interface{}) bool {
--- a/runtime/Go/antlr/lexer_atn_simulator.go
+++ b/runtime/Go/antlr/lexer_atn_simulator.go
@@ -10,7 +10,7 @@ import (
 )
 
 var (
-	LexerATNSimulatorDebug    = false
+	LexerATNSimulatorDebug = false
 	LexerATNSimulatorDFADebug = false
 
 	LexerATNSimulatorMinDFAEdge = 0
@@ -311,7 +311,7 @@ func (l *LexerATNSimulator) accept(input CharStream, lexerActionExecutor *LexerA
 }
 
 func (l *LexerATNSimulator) getReachableTarget(trans Transition, t int) ATNState {
-	if trans.Matches(t, 0, LexerMaxCharValue) {
+	if trans.Matches(t, 0, 0xFFFE) {
 		return trans.getTarget()
 	}
 
@@ -461,7 +461,7 @@ func (l *LexerATNSimulator) getEpsilonTarget(input CharStream, config *LexerATNC
 		trans.getSerializationType() == TransitionRANGE ||
 		trans.getSerializationType() == TransitionSET {
 		if treatEOFAsEpsilon {
-			if trans.Matches(TokenEOF, 0, LexerMaxCharValue) {
+			if trans.Matches(TokenEOF, 0, 0xFFFF) {
 				cfg = NewLexerATNConfig4(config, trans.getTarget())
 			}
 		}
--- a/runtime/Go/antlr/prediction_context.go
+++ b/runtime/Go/antlr/prediction_context.go
@@ -28,6 +28,7 @@ var (
 
 type PredictionContext interface {
 	Hash() string
+	HashCode() int
 	GetParent(int) PredictionContext
 	getReturnState(int) int
 	equals(PredictionContext) bool
@@ -204,6 +205,23 @@ func (b *BaseSingletonPredictionContext) Hash() string {
 	return b.cachedHashString
 }
 
+func (b *BaseSingletonPredictionContext) HashCode() int {
+	up := 0
+	if b.parentCtx != nil {
+		up = b.parentCtx.HashCode()
+	}
+
+	if up == 0 {
+		if b.returnState == BasePredictionContextEmptyReturnState {
+			up += 1 //TODO what should this be
+		}
+		up += b.returnState
+		return up
+	}
+	up += b.returnState
+	return up
+}
+
 func (b *BaseSingletonPredictionContext) String() string {
 	var up string
 
@@ -320,6 +338,26 @@ func (a *ArrayPredictionContext) equals(other PredictionContext) bool {
 		otherP := other.(*ArrayPredictionContext)
 		return &a.returnStates == &otherP.returnStates && &a.parents == &otherP.parents
 	}
+}
+
+func (a *ArrayPredictionContext) HashCode() int {
+	if a.isEmpty() {
+		return 19 //TODO what to return here, some prime is probably ol
+	}
+	v := 0
+	for i := 0; i < len(a.returnStates); i++ {
+		if a.returnStates[i] == BasePredictionContextEmptyReturnState {
+			v += (i * 23) //TODO what #
+			continue
+		}
+		v += a.returnStates[i]
+		if a.parents[i] != nil {
+			v += (27 * a.parents[i].HashCode())
+		} else {
+			v += 31
+		}
+	}
+	return v
 }
 
 func (a *ArrayPredictionContext) String() string {
--- a/runtime/Go/antlr/semantic_context.go
+++ b/runtime/Go/antlr/semantic_context.go
@@ -22,6 +22,8 @@ type SemanticContext interface {
 
 	evaluate(parser Recognizer, outerContext RuleContext) bool
 	evalPrecedence(parser Recognizer, outerContext RuleContext) SemanticContext
+
+	HashCode() int
 	String() string
 }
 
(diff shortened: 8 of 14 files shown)


parrt:
verrrry interesting. Can Go people from @antlr/antlr-targets take a look at this?

runtime/Go/antlr/prediction_context.go
```diff
@@ -322,6 +340,26 @@ func (a *ArrayPredictionContext) equals(other PredictionContext) bool {
 	}
 }
 
+func (a *ArrayPredictionContext) HashCode() int {
+	if a.isEmpty() {
+		return 19 //TODO what to return here, some prime is probably ol
```
willfaught: Incomplete comment

willfaught: Where does 19 come from?

runtime/Go/antlr/prediction_context.go
```diff
@@ -322,6 +340,26 @@ func (a *ArrayPredictionContext) equals(other PredictionContext) bool {
...
+	}
+	v := 0
+	for i := 0; i < len(a.returnStates); i++ {
+		if a.returnStates[i] == BasePredictionContextEmptyReturnState {
+			v += (i * 23) //TODO what #
```
willfaught: 23?

willfaught: Incomplete comment

runtime/Go/antlr/prediction_context.go
```diff
@@ -322,6 +340,26 @@ func (a *ArrayPredictionContext) equals(other PredictionContext) bool {
...
+			continue
+		}
+		v += a.returnStates[i]
+		if a.parents[i] != nil {
+			v += (27 * a.parents[i].HashCode())
```
willfaught: "

runtime/Go/antlr/prediction_context.go
```diff
@@ -322,6 +340,26 @@ func (a *ArrayPredictionContext) equals(other PredictionContext) bool {
...
+		} else {
+			v += 31
```
willfaught: "

runtime/Go/antlr/semantic_context.go
```diff
@@ -109,6 +111,10 @@ func (p *Predicate) equals(other interface{}) bool {
 	}
 }
 
+func (p *Predicate) HashCode() int {
+	return p.ruleIndex*43 + p.predIndex*47
```
willfaught: "

runtime/Go/antlr/semantic_context.go
```diff
@@ -155,6 +161,10 @@ func (p *PrecedencePredicate) equals(other interface{}) bool {
 	}
 }
 
+func (p *PrecedencePredicate) HashCode() int {
+	return p.precedence * 51
```
willfaught: "

runtime/Go/antlr/semantic_context.go
```diff
@@ -295,6 +305,21 @@ func (a *AND) evalPrecedence(parser Recognizer, outerContext RuleContext) Semant
 	return result
 }
 
+func (a *AND) HashCode() int {
+	v := 37
```
willfaught: "

runtime/Go/antlr/semantic_context.go
```diff
@@ -295,6 +305,21 @@ func (a *AND) evalPrecedence(parser Recognizer, outerContext RuleContext) Semant
...
+	for _, o := range a.opnds {
+		v += o.HashCode()
+	}
+	return v
+}
+func (a *OR) HashCode() int {
+	v := 41
```
willfaught: "

runtime/Go/antlr/token.go
```diff
@@ -14,6 +14,10 @@ type TokenSourceCharStreamPair struct {
 	charStream  CharStream
 }
 
+func NewTokenSourceCharStreamPair(source TokenSource, stream CharStream) *TokenSourceCharStreamPair {
```
willfaught: Why is this needed?

runtime/Go/antlr/utils.go
```diff
@@ -49,23 +49,34 @@ func (s *IntStack) Push(e int) {
 }
 
 type Set struct {
-	data           map[string][]interface{}
-	hashFunction   func(interface{}) string
-	equalsFunction func(interface{}, interface{}) bool
+	setData map[int][]interface{}
+	// stringhashFunction func(interface{}) string
```
willfaught: Dead code

runtime/Go/antlr/utils.go
```diff
@@ -49,23 +49,34 @@ func (s *IntStack) Push(e int) {
...
+	equalsFunction func(interface{}, interface{}) bool) *Set {
 
 	s := new(Set)
 
-	s.data = make(map[string][]interface{})
+	s.setData = make(map[int][]interface{})
 
-	if hashFunction == nil {
-		s.hashFunction = standardHashFunction
+	// could be nil
```
willfaught: Does this add anything that the following if statement doesn't?

runtime/Go/antlr/utils.go
```diff
@@ -49,23 +49,34 @@ func (s *IntStack) Push(e int) {
...
+	if hashcodeFunction != nil {
+		s.hashcodeFunction = hashcodeFunction
 	} else {
-		s.hashFunction = hashFunction
+		s.hashcodeFunction = standardHashFunction
 	}
 
+	// if hashFunction == nil {
```
willfaught: Dead code

runtime/Go/antlr/utils.go
```diff
@@ -87,11 +98,16 @@ func standardEqualsFunction(a interface{}, b interface{}) bool {
...
+	if h, ok := a.(HashCodeer); ok {
+		return h.HashCode()
+	}
 
-	if ok {
-		return h.Hash()
+	if h, ok := a.(Hasher); ok {
+		s := h.Hash()
+		ha := fnv.New32a()
+		ha.Write([]byte((s)))
```
willfaught: Extra parens

Can we replace s with the expression?

runtime/Go/antlr/utils.go
```diff
@@ -110,48 +126,47 @@ func standardHashFunction(a interface{}) string {
 type Hasher interface {
 	Hash() string
 }
-
-func hashCode(s string) string {
-	h := fnv.New32a()
-	h.Write([]byte((s)))
-	return fmt.Sprint(h.Sum32())
+type HashCodeer interface {
```
willfaught: Should be HashCoder (like Writer).

runtime/Go/antlr/utils.go
```diff
@@ -110,48 +126,47 @@ func standardHashFunction(a interface{}) string {
...
+		s.setData[key] = append(s.setData[key], value)
 		return value
 	}
 
-	s.data[key] = []interface{}{value}
+	v := make([]interface{}, 1, 10)
+	v[0] = value
+	s.setData[key] = v
+
+	// s.setData[key] = []interface{}{value}
```
willfaught: Dead

runtime/Go/antlr/utils.go
```diff
@@ -110,48 +126,47 @@ func standardHashFunction(a interface{}) string {
...
 			}
 		}
 
-		s.data[key] = append(s.data[key], value)
+		s.setData[key] = append(s.setData[key], value)
 		return value
 	}
 
-	s.data[key] = []interface{}{value}
+	v := make([]interface{}, 1, 10)
```
willfaught: Why cap 10?

runtime/Go/antlr/utils.go
```diff
@@ -164,10 +179,10 @@ func (s *Set) contains(value interface{}) bool {
 func (s *Set) values() []interface{} {
 	l := make([]interface{}, 0)
 
-	for key := range s.data {
-		if strings.Index(key, "hash_") == 0 {
-			l = append(l, s.data[key]...)
-		}
+	for key := range s.setData {
+		// if strings.Index(key, "hash_") == 0 {
```
willfaught: Dead

runtime/Go/antlr/utils.go
```diff
@@ -385,3 +400,36 @@ func TitleCase(str string) string {
...
+	c1_32 = 0xCC9E2D51
+	c2_32 = 0x1B873593
+	n1_32 = 0xE6546B64
+)
+
+func initHash(seed int) int {
+	return seed
+}
+
+func update(h1 int, k1 int) int {
```
willfaught: Can we combine these into `h1, k1 int`?

runtime/Go/antlr/utils.go
```diff
@@ -385,3 +400,36 @@ func TitleCase(str string) string {
...
+	k1 *= c1_32
```
willfaught: So many unexplained things. Is there a murmur hash already implemented that we can use? Something like https://gowalker.org/github.com/spaolacci/murmur3? Otherwise, this will need a lot of comments to explain what's going on.

willfaught:
Are both commits related? One is so small, and the other so large, it'd be nice if they could be squashed if so.

Can we simplify the "Go micro performance. Changed lexer atn config from Hash() string to …
…HashCode() int. Using murmur hash 32." commit summary to something like: "Change lexer ATN config hash to murmur 32"?

runtime/Go/antlr/token.go
```diff
@@ -208,7 +208,7 @@ func (c *CommonToken) String() string {
 		ch = ""
 	}
 
-	return "[" + "channel:" + strconv.Itoa(c.channel) + " @" + strconv.Itoa(c.tokenIndex) + "," + strconv.Itoa(c.start) + ":" + strconv.Itoa(c.stop) + "='" +
```
millergarym: Hey @willfaught. I changed this local and pushed it without realising that the format needs to be consistent across all targets. Therefore reverted it.

runtime/Go/antlr/token.go
millergarym: As far and I remember this is an old change in an already accepted PR. I'll check, a rebase might give ride of it. If memory servers this was need to enable a custom lexer (eg a block based lexer)

runtime-testsuite/test/org/antlr/v4/test/runtime/go/BaseGoTest.java
```diff
@@ -165,6 +165,9 @@ private static void cacheGoRuntime(File tmpPackageDir) throws Exception {
 	}
 
 	private static void copyFile(File source, File dest) throws IOException {
+		if ( source.isDirectory() ) {
```
millergarym: In general can you point me at a style guide. The bigger style issue with this in the exist of paren around the test. To much swapping between Java and Go :-)

millergarym:
If possible can some one with hashing / crypto experience to do a high level look at this. I'm not at all confident at the approach I've taken.

@parrt any suggestions as to who is expert in this area?

davesisson:
I second the import an existing implementation approach.  To get you started I found this particular implementation (there's a murmur hash in there) but haven't vetted it for capability:

https://github.com/leemcloughlin/gofarmhash

runtime/Go/antlr/atn_state.go
```diff
@@ -122,6 +123,10 @@ func (as *BaseATNState) SetNextTokenWithinRule(v *IntervalSet) {
 	as.NextTokenWithinRule = v
 }
 
+func (as *BaseATNState) HashCode() int {
+	return as.stateNumber
```
millergarym: See comment as the end of this thread.

runtime/Go/antlr/lexer_action.go
```diff
@@ -51,6 +52,10 @@ func (b *BaseLexerAction) getIsPositionDependent() bool {
 	return b.isPositionDependent
 }
 
+func (b *BaseLexerAction) HashCode() int {
+	return b.actionType
```
millergarym: ditto

runtime/Go/antlr/atn_config.go
```diff
@@ -285,6 +285,24 @@ func (l *LexerATNConfig) equals(other interface{}) bool {
 	return l.BaseATNConfig.equals(othert.BaseATNConfig)
 }
 
+func (l *LexerATNConfig) HashCode() int {
+	var f int
+	if l.passedThroughNonGreedyDecision {
```
millergarym: Again, can you point me at a style guide. Vscode fmt's the code on every save. As for white space between lines inside blocks, imo they should be used really sparingly. Is there a style guide for the Java antlr code?

runtime/Go/antlr/lexer_action_executor.go
```diff
@@ -30,8 +31,10 @@ func NewLexerActionExecutor(lexerActions []LexerAction) *LexerActionExecutor {
 	// of the performance-critical {@link LexerATNConfig//hashCode} operation.
 
 	var s string
+	l.cachedHash = initHash(57)
```
millergarym: random primes

runtime/Go/antlr/utils.go
millergarym: This isn't a cap,  slice = append( slice, element) with expand the slice as needed.
It was an attempt at a performance optimisation. There are so few clashes in the lexer being benchmarked that it made no difference.

runtime/Go/antlr/utils.go
millergarym: @willfaught, @davesisson 
There are no current dependancies in the runtime outside of the standard library. There should be a really good reason to introduce one. Murmur hash is small, not going chancing (at least not in this universe, but lets not get meta-physical) and was copied into the Java implementation.

Looking into using something like the guava collection in the Java code is an interesting discussion. Imo, importing murmur isn't.

millergarym:
@willfaught thanks for the feedback, much appreciated. This PR was more about generating discussing than being merged and I've learnt a lot from your comments. Hopefully some hashing expert will provide some high level feedback. Time permitting I'll revisit this.

Any comments on including benchmarking as a standard part of the build?

Cheers
Gary

runtime/Go/antlr/lexer_atn_simulator.go
```diff
@@ -311,7 +311,7 @@ func (l *LexerATNSimulator) accept(input CharStream, lexerActionExecutor *LexerA
 }
 
 func (l *LexerATNSimulator) getReachableTarget(trans Transition, t int) ATNState {
-	if trans.Matches(t, 0, LexerMaxCharValue) {
+	if trans.Matches(t, 0, 0xFFFE) {
```
parrt: @millergarym you need to pull from master as it looks like you are undoing our unicode 32 stuff.

runtime-testsuite/test/org/antlr/v4/test/runtime/go/BaseGoTest.java
pboyer: Hey @millergarym, The best thing to do is to run `gofmt` on the file. BTW, in Go, you don't need parens around the conditional expression. `if expr { ... }` should work.

runtime/Go/antlr/atn_config.go
pboyer: Again, just run `gotfmt` and we're good.

runtime/Go/antlr/utils.go
pboyer: Yes, let's fix this typo.

runtime/Go/antlr/utils.go
pboyer: @willfaught I'd prefer to copy paste an existing implementation, then to add a dependency for something as small as this.

pboyer:
@millergarym Thanks, this is a good start. Do you plan to follow up on this? If not, I could probably start a new PR incorporating some of these ideas. You're right that there's a bunch of performance improvement to be done on the Go runtime.

millergarym:
@pboyer, (cc: @parrt, @willfaught) lets close this PR and start a new one. I'm happy to help and think the best use of my time is looking into automated benchmarking. Travis doesn't seem to support this, however Jenkins does or can be made to via a combination of plugins. I'm happy to look into a Docker based Jenkins solution for non Travis CI stuff (benchmarking, cover coverage, todos and maybe even releases eg the XPath cyclic dependencies issue https://github.com/antlr/antlr4/issues/1620). Thoughts?

PR #1688, opened 2017-02-22
New \p{Letter} Unicode property escape

bhamiltoncx:
@mike-lischke asked for this in his review of https://github.com/antlr/antlr4/pull/1633 . I think it's a great way to show the power of the new full Unicode functionality in ANTLR4.

This PR adds two new lexer escapes suitable for use in a charset, so:

`[a-z]`

could become:

`[\p{Ll}]`

or, equivalently:

`[\p{Lowercase_Letter}]`

to match both a-z as well as exciting Unicode code points like 𝐚 (`U+1D41A`) through 𝐳 (`U+1D433`).

I included both matching and non-matching variants:

1. `\p{Letter}`: Include all Unicode code points with the general category "Letter"
2. `\P{Letter}`: Include all Unicode code points which do not have the general category "Letter"

This also works for:

1. Scripts (`\p{Latin}`, `\p{Hiragana}`, `\p{Cyrillic}`)
2. Binary properties (`\p{Emoji}`, `\p{Changes_When_Uppercased}`, `\p{Quotation_Mark}`)
3. Blocks prefixed with "In" (since their names collide with scripts): (`\p{InHiragana}`, `\p{InArabic_Ext_A}`, `\p{InGreek}`)

The names of properties are **case-insensitive**. In addition, `-` and `_` are treated the same.

PR #1693, opened 2017-02-24
New class EscapeSequenceParsing with \p{UnicodeProperty} support

bhamiltoncx:
This is split off from the WIP PR #1688.

The current logic to parse lexer charset escapes is a bit iffy. This introduces a new class `EscapeSequenceParsing` which we can use to add `\p{UnicodeProperty}` escape sequences, along with tests.


parrt:
thanks @bhamiltoncx !!

PR #1694, opened 2017-02-24
C++ implementation of issue #1665 and issue #1674

mike-lischke:
Second attempt at that patch. The first patch exists but seems to be dangling in space. Could not create a PR from that.


runtime/Cpp/runtime/src/Parser.h
```diff
@@ -376,6 +377,30 @@ namespace antlr4 {
...
+
+    /** How to create an error node, given a token, associated with a parent.
+       *  Typically, the error node to create is not a function of the parent
+       *  but this method must still set the parent pointer of the terminal node
+       *  returned. I would prefer having {@link ParserRuleContext#addAnyChild(ParseTree)}
+       *  set the parent pointer, but the parent pointer is implementation dependent
+       *  and currently there is no setParent() in {@link ErrorNode} (and can't
+       *  add method in Java 1.7 without breaking backward compatibility).
+       *
+       * @since 4.6.1
```
parrt: Could you change the 4.6.1 to 4.7?

runtime/Cpp/runtime/src/tree/TerminalNode.h
```diff
@@ -21,7 +21,7 @@ namespace tree {
      *  TerminalNodes anyway so I'm adding as it improves internal
      *  code quality.
      *
-     *  @since 4.6.1
+     *  @since 4.7
```
parrt: you might do a search for the other 4.6.1s :)

mike-lischke: done

mike-lischke:
@parrt ready for merge?

Issue #1682, opened 2017-02-21
A new indent option for better code style compliance in generated code


KvanTTT:
I decided to create a new issue separated from this one: [Add a new option for naming convention of generated Visitor and Listener methods](https://github.com/antlr/antlr4/issues/1615).

I suggest to add a new option `indent` with the following syntax: `-Dindent space4`. This value causes to generate code with four spaces indentation.

Default value is `tab1`.

ericvergnaud:
Mmmmm...
Indentation is more or less hard coded in the generator templates and  cannot change at runtime.
I'm not supportive

Envoyé de mon iPhone

KvanTTT:
Tabs can be replaced with spaces in template itself.

ericvergnaud:
That would mean it is no longer an option.

sharwell:
@KvanTTT The templates are intended to be self-consistent within a particular language, but there is no intent to make them consistent with particular coding styles. The use of tabs vs. spaces in the generated code is a file size issue. Some editors will refuse to open files over a certain size (varies by editor), and changing from tabs to spaces will substantially increase the number of ANTLR generated files which exceed these limits, posing a usability problem which we're currently avoiding.

KvanTTT:
> but there is no intent to make them consistent with particular coding styles.

Why not if it's possible and easy to implement?

> The use of tabs vs. spaces in the generated code is a file size issue

At present time C++ runtime uses 2 spaces as an indent, Python runtimes use 4 spaces. According to this reason, we have to replace all spaces with tabs.

> changing from tabs to spaces will substantially increase the number of ANTLR generated files which exceed these limits, posing a usability problem which we're currently avoiding.

I generated a parser for big [PL/SQL grammar](https://github.com/antlr/grammars-v4/blob/master/plsql/plsql.g4) grammar. The parser contains approximately ~50K lines and has size 1.63MB. After it I replaced all tabs on spaces and file size increased up to 1.99 MB. So, the ratio 1.99 / 1.63 ≈ **1.2** is not a critical on my opinion but good for common code style. Moreover the size can be decreased if replace tabs with empty strings (with option `-Dindent spaces0` or just `-Dindent 0`): 1.51 / 1.99  ≈ **0.76**. But zero-indent won't work with indent-sensitive languages (Python).

KvanTTT:
@ericvergnaud, it's an option. We can use the following code for template loading and changing:

```Java
[...]

PR #1695, opened 2017-02-24
Consistent indents (tabs) for all string templates

KvanTTT:
Also it decreases the size of generated code.

### DIFF
--- a/tool/resources/org/antlr/v4/tool/templates/codegen/Cpp/Cpp.stg
+++ b/tool/resources/org/antlr/v4/tool/templates/codegen/Cpp/Cpp.stg
@@ -30,14 +30,14 @@
 import "Files.stg" // All file specific stuff.
 
 cppTypeInitMap ::= [
-    "int":"0",
-    "long":"0",
-    "float":"0.0f",
-    "double":"0.0",
-    "bool":"false",
-    "short":"0",
-    "char":"0",
-    default: "nullptr" // anything other than a primitive type is an object
+		"int":"0",
+		"long":"0",
+		"float":"0.0f",
+		"double":"0.0",
+		"bool":"false",
+		"short":"0",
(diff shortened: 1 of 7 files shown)


bhamiltoncx:
It'd be amazing if we could introduce automatic code formatters here.

ericvergnaud:
Hi,
Sorry but I'm against this PR. Especially the current Python templates use spaces to comply with pep 8.
There is no such thing as consistency across targets when each language has its own "consistency" rules.

mike-lischke:
The title of this PR is a bit misleading. It's not about formatting all runtime code, but just the string templates. The intention is very good, but the changes have to mimic what is currently used in the handwritten code.

And tabs shouldn't be used at all, as their presentation in various tools vary way too much and more often than not produce terrible output.

mike-lischke:
@bhamiltoncx I could imagine a clang-format file in the repo which contains settings specific for each used language, to which the runtimes must conform. We just have to agree a base formatting style. I'm happy to contribute C++ settings for the Google coding style (the current C++ runtime is already formatted in that style).

KvanTTT:
@bhamiltoncx I'm afraid that's impossible for stg files at present time. This is a specific format.

@ericvergnaud current Python2 template uses mixed indents. See tabs at https://github.com/antlr/antlr4/blob/4.6/tool/resources/org/antlr/v4/tool/templates/codegen/Python2/Python2.stg#L169 for example. Python3 and JavaScript templates have the same problem.

@mike-lischke my issue about customizable indents has been closed: https://github.com/antlr/antlr4/issues/1682. But I like this idea and I also like whitespaces. But I choose tabs because of less generated file size.

parrt:
Yeah, i'm gonna close as I don't want to monkey with the generated code at this point.

Issue #1537, opened 2016-12-20
Escapes sequence recognition failure in character sets


renatahodovan:
After the 4.6 release, the behaviour of escape sequences in the lexer's character sets changed compared to 4.5.3. I'm not sure whether they are bugs or features but I think it's worth to mention:

1) `[\[]` this worked in 4.5.3 but in 4.6 it's an invalid escape sequence (instead `[[]` can be used)
2) `[\v]` vertical tabs worked in 4.5.3 but they are invalid in 4.6
3) In 4.5.3, arbitrary characters could be escaped like `[\j\k\l]` but now they are invalid.
4) `[d-a]`: although its a bit weird, but was valid in 4.5.3, but it's not in 4.6

If these changes are expected, then I'm going to create some PR-s in grammars-v4, since several of them are failing now.

parrt:
Yeah, basically it's saying now that the escape is unnecessary. I didn't realize the PR made it an error rather than warning. sorry about that.  Please do create a PR for grammars-v4! :)

renatahodovan:
Does it mean that vertical tabs are not allowed anymore (i.e., they should be removed)? And a second thing, grammars-v4 still uses 4.5.3. Will it be bumped up?

parrt:
Hmm...well, \v can be done with \u000B so let's leave it. it's rare.  Yeah, we should bump to 4.6

KvanTTT:
It seems to me that this pull request https://github.com/antlr/antlr4/pull/1517 brought such behavior. In my opinion these cases are not clear. So, @renatahodovan make pull request to grammars if you want :)

parrt:
@renatahodovan we could add `\v` if you want. Also let's drop from error to warning on `[\[]` type stuff. You ok with that @KvanTTT ?

renatahodovan:
@parrt yes, I think it would be useful for the sake of backward compatibility. Thanks!

KvanTTT:
Yes, warning instead of error is a good idea. I'll try to fix it at the beginning of January.

parrt:
Any pull request for this coming? If not, I can do it.

KvanTTT:
Not ready for now :(

Nulleye:
Just to help on testing.
This: 
ID : [a-zA-Z&/][a-zA-Z0-9\.@\-_\*/><]*;
fails on 4.6 but not on 4.5.3

KvanTTT:
@Nulleye thank you for feedback!

[...]

PR #1696, opened 2017-02-25
Useless escape in charset warning

KvanTTT:
Fixes https://github.com/antlr/antlr4/issues/1537.

### DIFF
--- a/tool-testsuite/test/org/antlr/v4/test/tool/TestSymbolIssues.java
+++ b/tool-testsuite/test/org/antlr/v4/test/tool/TestSymbolIssues.java
@@ -389,12 +389,13 @@ public class TestSymbolIssues extends BaseJavaToolTest {
 				"TOKEN_RANGE:      [aa-f];\n" +
 				"TOKEN_RANGE_2:    [A-FD-J];\n" +
 				"TOKEN_RANGE_3:    'Z' | 'K'..'R' | 'O'..'V';\n" +
-				"TOKEN_RANGE_4:    'g'..'l' | [g-l];\n",             // Handling in ATNOptimizer.
+				"TOKEN_RANGE_4:    'g'..'l' | [g-l];\n" +             // Handling in ATNOptimizer.
+				"TOKEN_RANGE_WO_COLLISION: '_' | [a-zA-Z];",
 
 				"warning(" + ErrorType.CHARACTERS_COLLISION_IN_SET.code + "): L.g4:2:18: chars \"a-f\" used multiple times in set [aa-f]\n" +
 				"warning(" + ErrorType.CHARACTERS_COLLISION_IN_SET.code + "): L.g4:3:18: chars \"D-J\" used multiple times in set [A-FD-J]\n" +
 				"warning(" + ErrorType.CHARACTERS_COLLISION_IN_SET.code + "): L.g4:4:13: chars \"O-V\" used multiple times in set 'Z' | 'K'..'R' | 'O'..'V'\n" +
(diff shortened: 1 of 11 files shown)


KvanTTT:
@parrt don't merge it until I check it on our grammars.

parrt:
@KvanTTT seems like a major bit of surgery when I was kind of expecting ERROR->WARNING on a few things.

KvanTTT:
I don't know a simple way to fix it. `CharSupport` does not have a `public final Grammar g;` property, therefore it's impossible to throw errors and warnings via `CharSupport`.

KvanTTT:
I fixed all discovered errors and check ANTLR 4.7 on our grammars repository.
Also I fixed all warnings in grammars: https://github.com/antlr/grammars-v4/pull/566

So, this PR is ready to merge for now.

parrt:
I still don't think I can agree with this PR. For one it is altering @bhamiltoncx 's new unicode stuff. what is the `intChar` in:

```java
int t1 = CharSupport.getCharValueFromGrammarCharLiteral(a.getText()).intChar
```

for example? that already is an int.  I think we just need to convert some errors to warnings. nothing else. Maybe I'll respond to #1537 myself to show what I mean.

parrt:
See https://github.com/antlr/antlr4/pull/1709

parrt:
It's ok to just give a warning:

```
A : [\[] ;
B : [\v] ;
C : [\j\k\l] ;
```

gets

```
warning(156): T.g:5:4: invalid escape sequence
warning(156): T.g:6:4: invalid escape sequence
warning(156): T.g:7:4: invalid escape sequence
```

KvanTTT:
I created a new pull request with my improvements: https://github.com/antlr/antlr4/pull/1715
`'\"'` string literal also should throw an "invalid escape sequence" warning.

Disadvantage of warning: such char sets and string literals `[\u24\uA2]` will be treated as warnings despite of they are errors clearly.

parrt:
oh right. we should have **unnecessary** vs **invalid** escape. First one is warning, second is error. Does that sound right?

KvanTTT:
Yes, you are right.

parrt:
Your new PR is probably good enough. :)

KvanTTT:
Just now found another issue with lexer (during testing our grammars repository) :( Will fix it ASAP.

KvanTTT:
It's ok! Just a error in grammar [UCBLogo.g4]  :)
We should escape a first minus sign in this rule:
```
NAME
 : ~[-+*/=<> \t\r\n[\]()":{}] ( ~[-+*/=<> \t\r\n[\](){}] | LINE_CONTINUATION | '\\' [-+*/=<> \t\r\n[\]();~{}] )*
 ;
```

But I want to improve error messages anyway. Now they are too confused for such rule:
```
warning(180): UCBLogo.g4:699:4: chars "42" used multiple times in set [-+*/=<> \t\r\n[\]()":{}]
```

KvanTTT:
> oh right. we should have unnecessary vs invalid escape. First one is warning, second is error. Does that sound right?

What do you think about moving grammar instance `Grammar g` to `CharSupport.java` for handling errors and warnings there?

parrt:
Can't. Grammar is not in runtime.

parrt:
Those are just static support functions. pass in g if you need I guess. I'd prefer we have a central place to check all this. i was uncomfortable with your change to ATNOptimizer to check ranges so...

KvanTTT:
> i was uncomfortable with your change to ATNOptimizer to check ranges so...

Actually there is no another place where some kind of ranges can be checked.

parrt:
I was thinking maybe a separate checking phase over the ATN but no biggie.

PR #1710, opened 2017-03-01
Update docs for new Unicode literal and property escapes

bhamiltoncx:
This updates the lexer, lexicon, and C++ documentation so 4.7 users know all about the new `\u{XXXXXX}` and `\p{Letter}` Unicode literal and property escapes.

PR #1712, opened 2017-03-02
Using new Unicode syntax for C++ demo project.

mike-lischke:
A small change to make the C++ demo project work with the new Unicode support.


bhamiltoncx:
BTW, I updated the C++ docs in #1710 , so you might want to look at them.

PR #1713, opened 2017-03-02
C++ implementation for #1667

mike-lischke:


parrt:
Cool. I'll leave this open like the Java version until we know more about whether the approach is good/correct.

PR #1718, opened 2017-03-02
Fixes #1703. 

### LINKED ISSUE #1703: IntervalSet.getSingleElement() returns Token.INVALID_VALUE (0) on failure
I ran into this while implementing #1688 .

Currently, a number of methods in `IntervalSet` (`getSingleElement()`, `getMaxElement()`, etc.) return `Token.INVALID_VALUE` on failure.

Unfortunately, `Token.INVALID_VALUE` is `0`, which is a perfectly valid value to hold in an `IntervalSet`, which claims to support everything from `Integer.MIN_VALUE` to `Integer.MAX_VALUE`.

In fact, `Java.g4` depends on an `IntervalSet` containing `0`:

```
fragment                                                                                                    
JavaLetter                                                                                                  
       :      [a-zA-Z$_] // these are the "java letters" below 0xFF                                       
       |      // covers all characters above 0xFF which are not a surrogate                               
              ~[\u0000-\u00FF\uD800-\uDBFF]
(snip)
       ;                                                                                                   
```

There are actually no callers of `IntervalSet.getSingleElement()`, so we can probably freely change its signature, but `IntervalSet.getMaxElement()` and friends have the same problem and do have lots of callers.

I thought we might just be able to change the error result to `-1`, but `IntervalSet` supports negative values.

I can think of a few possibilities:

1. Return a boxed `Integer` instead of `int` and require callers to check for `null` (error-prone)
2. Return `long` instead of `int` and use `Long.MIN_VALUE` to signify error
3. Throw an exception on error
4. Introduce something like Java 8 / Guava's `Optional<T>` and require callers to check for the result being present

Thoughts?

parrt:
Updated comments to ensure people test isNil() or size() before requesting min, max, any single element.

### DIFF
--- a/runtime/Java/src/org/antlr/v4/runtime/misc/IntSet.java
+++ b/runtime/Java/src/org/antlr/v4/runtime/misc/IntSet.java
@@ -6,8 +6,6 @@
 package org.antlr.v4.runtime.misc;
 
 
-import org.antlr.v4.runtime.Token;
-
 import java.util.List;
 
 /**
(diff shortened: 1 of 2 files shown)


parrt:
Updated: https://github.com/antlr/antlr4/pull/1720  error handler was calling with empty set. you were right!

PR #1724, opened 2017-03-03
More accurate error messages

KvanTTT:


tool-testsuite/test/org/antlr/v4/test/tool/Java.g4
```diff
@@ -1140,7 +1140,7 @@ SignedInteger
 
 fragment
 Sign
-	:	[+-]
+	:	[+\-]
```
mike-lischke: So this is no longer possible without escaping the minus? It worked when the minus was at the end of the set list.

KvanTTT: Yes, not possible. Old syntax was a bit incorrect.

tool/src/org/antlr/v4/automata/ATNOptimizer.java
```diff
@@ -147,6 +150,29 @@ private static void optimizeSets(Grammar g, ATN atn) {
 //		System.out.println("ATN optimizer removed " + removedStates + " states by collapsing sets.");
 	}
 
+	private static String getIntervalSetEscapedString(IntervalSet intervalSet) {
```
parrt: not sure this belongs in this class

KvanTTT: Ok, I'll move it to `CharSupport`.

KvanTTT:
Much better :)

tool-testsuite/test/org/antlr/v4/test/tool/Java.g4
sharwell: @KvanTTT Old syntax was pretty common. I expect it to remain valid.

parrt:
@sharwell oh, yeah, didn't we long time ago say anything not terminated was just the '-' char? Dang. @KvanTTT how big is the change to flip `a-` and `-z` to be two char not a range.

KvanTTT:
Pretty easy.

parrt:
ok, let's do that. thanks!

PR #1726, opened 2017-03-03
Reverted unterminated ranges: [+-], [-+], [-].

KvanTTT:
But on my opinion such syntax a bit confused despite of it's common...


parrt:
thanks!

PR #1728, opened 2017-03-04
Fixed crash with multi threaded parsers warming up at the same time, fixes #1601

### LINKED ISSUE #1601: [Cpp] Separate parser/lexer instances not threadsafe
Continued from #1435 

Still seeing issues when using separate parser instances concurrently from several threads.

Repostory with example code: https://github.com/Pursuit92/antlr_cpp_threads

I'm currently working on getting a test case into the runtime test suite, but I'm having difficulty even getting the existing ones to run. Seeing clang liker errors a la
```
./tmp/TLexer-5cde0b.o:(.rodata+0x208): undefined reference to `non-virtual thunk to antlr4::Lexer::getSourceName[abi:cxx11]()'
clang-3.9: error: linker command failed with exit code 1 (use -v to see invocation)
```

Is there anything I might be doing wrong/steps I've missed? I'll keep poking at it in the meantime.

mike-lischke:
The lock for the shared DFA state needs to protect a few more operations than just the addDFAState stuff and had to move up one call level. This in turn requires now 2 places where the lock must be aquired.

### DIFF
--- a/runtime/Cpp/runtime/src/atn/ATNSimulator.cpp
+++ b/runtime/Cpp/runtime/src/atn/ATNSimulator.cpp
@@ -32,8 +32,7 @@ PredictionContextCache& ATNSimulator::getSharedContextCache() {
 }
 
 Ref<PredictionContext> ATNSimulator::getCachedContext(Ref<PredictionContext> const& context) {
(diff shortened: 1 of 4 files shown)


mike-lischke:
Fixes #1601

PR #1731, opened 2017-03-06
Remove generated files from repository and add testRig for Swift target

hanjoes:
Current version of swift package manager doesn't support shell command
or any mechanism that we can leverage to generate parser files. Adding
a python script to kick off the unit tests.

This PR also removes those generated files from code base.


parrt:
thanks, Joe!

PR #1739, opened 2017-03-08
Also support Unicode enumerated properties via \p{Foo=Bar} syntax

bhamiltoncx:
When I went to implement [Unicode TR29 grapheme cluster parsing](http://www.unicode.org/reports/tr29/), I found I forgot to include enumerated Unicode properties in the code-generated `UnicodeData.java`.

Since enumerated properties have multiple values (as opposed to the binary properties we supported previously), I used the semi-standard syntax:

`\p{PropertyName=PropertyValue}`

All the existing tests pass and I added a few more for the new syntax.


parrt:
sounds good. Thanks!!

bhamiltoncx:
Thanks for the über-quick merge!

parrt:
You have the hard part. all i do is click!  I'm hoping we can finish off 4.7 and ship it soon. anything you can think of to do on your end?

bhamiltoncx:
Well, the devil is in the details. Once I finish the TR29 grapheme cluster parser and test perf, I'm sure I'll have at least one fix.

parrt:
cool. ok, it'll probably be a few weeks on my side.

PR #1723, opened 2017-03-03
	Improve error message in exceptions

chund:
Improve error messages given by exceptions.
This is necessary to get useful error messages in following case: when using bail error strategy and compile with VS2013 which does not support nested exceptions.

PR #1740, opened 2017-03-09
fix for compile error

teverett:
Currently the project tool-codegen generates the file UnicodeData.java into the /target directory of the project tool.  Running “mvn clean package” in tool erases this file and generates a compile error since UnicodeData.java can’t be found.

This patch changes tool-codegen to generate UnicodeData.java into the /src directory of tool.


jimidle:
It should really go in target/generated-files/antlr (or unicode or
something) and the generation should be tied in to the appropriate phase.

teverett:
@jimidle essentially I agree with you.  The problem is that the file is being generated by the "tool-codegen" maven project, onto the /target of the "tool" maven project.

bhamiltoncx:
Let's make a jar instead so the deps work properly.

teverett:
@bhamiltoncx I'm not sure what the advantage of making it a jar is, over simply combining tools and tools-codegen?

bhamiltoncx:
The reason the targets were separate were to keep the ICU dependency out of
tools. I'm sure there's a better way to do it.

teverett:
@bhamiltoncx Right, that makes sense.  In my commit I've excluded the ICU dependency from maven-shade-plugin, so that the large ICU jar doesn't appear in the antlr4 jar.

parrt:
Cool. I think that solves it. We can keep it in tool, not a subproject, and then simply not included in the shaded jar at the end.  excellent. Thanks guys!

Fixes #1742

parrt:
howdy @teverett looks like it causes [C++ to fail](https://travis-ci.org/antlr/antlr4/jobs/211029771). I think this was first commit where we see:

```
.................................................................................................Tests run: 98, Failures: 1, Errors: 0, Skipped: 0, Time elapsed: 481.32 sec <<< FAILURE! - in org.antlr.v4.test.runtime.cpp.TestLeftRecursion
testOne[Cpp:SemPredFailOption](org.antlr.v4.test.runtime.cpp.TestLeftRecursion)  Time elapsed: 4.895 sec  <<< FAILURE!
junit.framework.ComparisonFailure: 
expected:<...ule a custom message[]
> but was:<...ule a custom message[ near token z in line 1]
>
```

it always [times out](https://travis-ci.org/antlr/antlr4/jobs/211029780) for C++ on mac so don't worry about that part.

@mike-lischke any thoughts?

mike-lischke:
This is probably related to patch #1723. Users want better default error messages. Obviously this is incompatible with the test suite. We really need to get the CI tests passing all the time and not time out, so that breaking changes are detected. We got already used to the fact that Travis CI fails for the C++ target :-(

We have now 2 options: revert the patch or apply similar ones for all targets and update the tests.

parrt:
@mike-lischke ah! yeah, looks like first time that errored is merging #1723. https://travis-ci.org/antlr/antlr4/jobs/209910002  It adds `"Input mismatch"` to the error messages etc...  Should I just back this out?

mike-lischke:
Undoing it is the quicker fix, but is it the better one? If that gets undone we should tell the author about it.

seb314:
Could it be that while the shade plugin exclude removes icu from the jar, it does not remove the declared dependency from the pom? (see also https://github.com/antlr/antlr4/issues/4900)

PR #1748, opened 2017-03-10
Fixes #1620. Make handbuilt lexer to avoid cyclic dependence of tool and plugin.

### LINKED ISSUE #1620: XPathLexer not updated in release process

parrt:


runtime/Java/src/org/antlr/v4/runtime/tree/xpath/XPathLexer.java
```diff
@@ -0,0 +1,174 @@
+package org.antlr.v4.runtime.tree.xpath;
```
sharwell: :bulb: copyright header

sharwell:
I prefer the approach of not having `RuntimeMetaData` to break every release, but given that limitation this approach seems reasonable.

KvanTTT:
Where is `channelNames`?

parrt:
hahah. not needed for this little lexer.

Issue #1743, opened 2017-03-10
Go runtime - Performance - static fields in Java currently implemented as struct fields - should be package vars


millergarym:
see https://github.com/antlr/antlr4/issues/1540

PR #1750, opened 2017-03-11
Initialize decisionToDFA in init in Go.stg

pboyer:
Fixes: https://github.com/antlr/antlr4/issues/1743

### DIFF
--- a/tool/resources/org/antlr/v4/tool/templates/codegen/Go/Go.stg
+++ b/tool/resources/org/antlr/v4/tool/templates/codegen/Go/Go.stg
@@ -151,9 +151,7 @@ var parserATN = <atn>
 var parserATN []uint16
 <endif>
 
-
 var deserializer = antlr.NewATNDeserializer(nil)
-
 var deserializedATN = deserializer.DeserializeFromUInt16(parserATN)
 
 <if(parser.literalNames)>
@@ -163,7 +161,6 @@ var literalNames = []string{
 <else>
 var literalNames []string
 <endif>
-
 
 <if(parser.symbolicNames)>
 var symbolicNames = []string{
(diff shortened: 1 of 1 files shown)


pboyer:
@millergarym I've extended this to both the parser template as well.

millergarym:
@pboyer looks good to me.
Are there any places in the runtime that might have similar issues?
When I have some time I'll benchmark this and let you know the result.

pboyer:
@millergarym It's not obvious to me, at first glance, what parts of the system need more optimization right now. I covered some obvious issues in #1755 found in profiling.

I'm not entirely caught up with your benchmarking work, but it seems to me that we'd greatly benefit from profiling the parser against more realistic, large grammars.

millergarym:
@pboyer @parrt 

Woo-hoo. Between 6 and 12 times speed up. 

There is test user if you would like to look around and play
http://jenkins.wx.io:8080/
antlr4-trial
qIkE@3Urc@$W

I haven't tested it against the dgraph hand written lexer, as it requires a different tool chan. But since it was 40x faster, I would guess this would still be ~10x slower. The standard idiom of hand written lexers in Go is basically a combinator, my guess it that there is no heap allocation. From previous profiling a lot of the time is taken up with memory alloc and you can't beat zero.

I will leave the Jenkins instance up for another week, after which the below images will break.

Re: microbenching for Java, I came across http://microbenchmarks.appspot.com/ looks cool, but haven't tried it.

Cheers
Gary

Benchmarking
Benchmarking
Benchmarking
Benchmarking
Benchmarking
Benchmarking

pboyer:
Awesome! Thanks so much for doing this and getting the ball rolling on the
perf improvements @millergarym!

PR #1751, opened 2017-03-11
Use xml comments instead of [cut][/cut] for section clearing in TEMPLATE files

KvanTTT:


.github/ISSUE_TEMPLATE.md
```diff
@@ -3,6 +3,6 @@ Before submitting an issue to ANTLR, please check off these boxes:
 - [ ] I am not submitting a question on how to use ANTLR; instead, go to [antlr4-discussion google group](https://groups.google.com/forum/#!forum/antlr-discussion) or ask at [stackoverflow](http://stackoverflow.com/questions/tagged/antlr4)
 - [ ] I have done a search of the existing issues to make sure I'm not sending in a duplicate
```
KvanTTT: Can we also completely comment out this section for getting rid of noise in issues?

parrt: Hmm...yeah, we are getting noise. i guess so

PR #1752, opened 2017-03-11
Use murmur hash in Go target

pboyer:
This pull request continues the work that @millergarym started with some refinement and cleanup.

I generally followed the pattern established by @sharwell in the Java runtime. That is, I often adopted the same initialization number as he did.

In this PR, I started some work to reduced the visibility of various internal data structures in the `antlr` package, so this PR is technically API breaking. I plan a follow up PR that takes the API scope reduction further.

### DIFF
--- a/runtime/Go/antlr/atn_config.go
+++ b/runtime/Go/antlr/atn_config.go
@@ -6,7 +6,6 @@ package antlr
 
 import (
 	"fmt"
-	"strconv"
 )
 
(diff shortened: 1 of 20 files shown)


pboyer:
Closing this. It looks like I caused some regressions.

sharwell:
Let me know (with an @-reference) if you have questions about how/why I did something. :smile:

PR #1755, opened 2017-03-14
Go Runtime: Murmur Hash and various optimizations

pboyer:
This, combined with #1750 makes the runtime significantly faster. The work borrowed some changes from @millergarym, but takes a slightly different route. I have run this against a relatively large grammar with a fuzzer with no regressions.

In many cases, the hash initialization constants were pulled from the Java runtime.

Technically speaking, this PR breaks the Go runtime API. All of the `hash` methods are made package private. They should have been package private from the outset.

@sharwell This is my second attempt at this - this time with no regressions. :)


millergarym:
@pboyer where are the changes to Go.stg?

pboyer:
@millergarym Do you mean https://github.com/antlr/antlr4/pull/1750?

parrt:
Thanks guys!

PR #1758, opened 2017-03-15
Fixed portability problems in C# target.  Also cleaned up some XML do…

RobertvanderHulst:
Fixed portability problems in C# target.  Also cleaned up some XML doc comments


parrt:
Looks ok to me. @ericvergnaud ?

ericvergnaud:
blessed

PR #1760, opened 2017-03-15
tweak wildcard

parrt:


parrt:
@ericvergnaud did I do this right?

ericvergnaud:
seems ok to me, but would be great to have a test that triggers this template and prove correctness.

parrt:
Indeed. weird that we don't test wildcard...oh must be in lexer.

parrt:
well crap. worked locally but fails at travis. must be whitespace thing. pushing expanded tabs.

PR #1762, opened 2017-03-15
Fixes #1745. Added `prototype` to token, channel, etc... refs.

### LINKED ISSUE #1745: javascript target: lexer generator - inaccessible tokens, modes and channels(4.7) constants
The generated lexer exports the token, modes and channels(as of 4.7) on function level.
Later they are not accessible in the super class via this.
Better export them via the function prototype.

This is very useful when having a super class like this:
```
options
   { superClass = LexerAdaptor; }
```

parrt:


parrt:
@ericvergnaud did I get this right?

parrt:
Actually, no, I broke test `PositionAdjustingLexer`. We get this:

```
[@0,0:5='tokens',<6>,1:0]
[@1,7:14='tokens {',<4>,2:0]
...
```

Instead of

```
[@0,0:5='tokens',<6>,1:0]
[@1,7:12='tokens',<4>,2:0]
[@2,14:14='{',<3>,2:7]
...
```

v-leon:
The node test template should also be aligned to this change:

diff --git a/runtime-testsuite/resources/org/antlr/v4/test/runtime/templates/Node.test.stg b/runtime-testsuite/resources/org/antlr/v4/test/runtime/templates/Node.test.stg
index 92cb1db..2c80a42 100644
--- a/runtime-testsuite/resources/org/antlr/v4/test/runtime/templates/Node.test.stg
+++ b/runtime-testsuite/resources/org/antlr/v4/test/runtime/templates/Node.test.stg
@@ -117,10 +117,10 @@ PositionAdjustingLexer.prototype.nextToken = function() {

 PositionAdjustingLexer.prototype.emit = function() {
        switch(this._type) {
-       case PositionAdjustingLexer.TOKENS:
+       case this.TOKENS:
                this.handleAcceptPositionForKeyword("tokens");
                break;
-       case PositionAdjustingLexer.LABEL:
+       case this.LABEL:
                this.handleAcceptPositionForIdentifier();
                break;
        }

parrt:
@ericvergnaud does this break any backward compatibility or anything? Is it blessed if I make the other @v-leon changes?

parrt:
@ericvergnaud is this blessed?

ericvergnaud:
I'm not comfortable with EOF becoming an instance member
I'll give it more closer look tonight

Issue #1125, opened 2016-03-04
Rules named 'values' generates invalid csharp code


gadunga:
I am using the antlr snapshot included in the Antlr4.4.5-alpha003 package.

Using the following grammar: 

```
grammar hello;
values: 'hello' ;
```

Running the following command:
`java -jar antlr4-csharp-4.5-SNAPSHOT-complete.jar -Dlanguage=CSharp_v4_0 hello.g4`

I get the following invalid csharp code (line 112 in a text editor): 

``` C#
//------------------------------------------------------------------------------
// <auto-generated>
//     This code was generated by a tool.
//     ANTLR Version: 4.5-SNAPSHOT
//
//     Changes to this file may cause incorrect behavior and will be lost if
//     the code is regenerated.
// </auto-generated>
//------------------------------------------------------------------------------

// Generated from hello.g4 by ANTLR 4.5-SNAPSHOT

// Unreachable code detected
#pragma warning disable 0162
// The variable '...' is assigned but its value is never used
#pragma warning disable 0219
// Missing XML comment for publicly visible type or member '...'
#pragma warning disable 1591
// Ambiguous reference in cref attribute
#pragma warning disable 419

using Antlr4.Runtime;
using Antlr4.Runtime.Atn;
using Antlr4.Runtime.Misc;
using Antlr4.Runtime.Tree;
using System.Collections.Generic;
using DFA = Antlr4.Runtime.Dfa.DFA;

[System.CodeDom.Compiler.GeneratedCode("ANTLR", "4.5-SNAPSHOT")]
[System.CLSCompliant(false)]
public partial class helloParser : Parser {
    public const int
        T__0=1;
    public const int
        RULE_values = 0;
    public static readonly string[] ruleNames = {
        "values"
    };

    private static readonly string[] _LiteralNames = {
        null, "'hello'"
    };
    private static readonly string[] _SymbolicNames = {
    };
    public static readonly IVocabulary DefaultVocabulary = new Vocabulary(_LiteralNames, _SymbolicNames);

    [System.Obsolete("Use Vocabulary instead.")]
    public static readonly string[] tokenNames = GenerateTokenNames(DefaultVocabulary, _SymbolicNames.Length);

    private static string[] GenerateTokenNames(IVocabulary vocabulary, int length) {
        string[] tokenNames = new string[length];
        for (int i = 0; i < tokenNames.Length; i++) {
            tokenNames[i] = vocabulary.GetLiteralName(i);
[...]

PR #1764, opened 2017-03-15
Added "default" and "values" to bad words C# set

KvanTTT:


KvanTTT:
I fixed #1125 bug, but actually "symbol conflicts with generated code in target language or runtime" is not appropriate name of this error. We can move all [keywords](https://github.com/antlr/antlr4/blob/4.6/tool/resources/org/antlr/v4/tool/templates/codegen/CSharp/CSharp.stg#L1061) to badWords but I don't like this idea.

parrt:
Ah.  sorry I didn't see this.  Anyway, it looks like C# target does something different and so I just added the appropriate `values` keyword.

KvanTTT:
Can we add "default" keyword to badWords set? It's not working properly for C# for now...

parrt:
I saw it in the C# stg file. Does it not get converted properly to have `@`?

KvanTTT:
Yes, see this: https://github.com/antlr/stringtemplate4/issues/116

parrt:
So, you mean "no", right? it is *not* properly converted because of the string template issue. Am I understanding it fully now?

KvanTTT:
You are right: not properly converter. Sorry for my language 😕

sharwell:
@parrt That's correct. StringTemplate bug prevents the template from converting the word correctly, instead treating it as a magic StringTemplate value.

parrt:
crap. maybe it's time we push out another ST release?

KvanTTT:
It will be great!

ericvergnaud:
Since this is a ST bug, I would recommend fixing the bug rather than work around it i.e. Don't add to C# badwords

Envoyé de mon iPhone

KvanTTT:
@ericvergnaud, completely agree.

PR #1766, opened 2017-03-16
Go runtime: Protect shared DFA state with mutexes

pboyer:
In #1750, I made the `decisionToDFA` array passed to the `*ATNSimulator` shared between recognizers. This greatly speeds up initialization, but introduces shared state between concurrent recognizers. Specifically, the `DFA`'s in this array can be mutated concurrently.

In this PR, I've surrounded the `DFA` `s0` and `states` variables with mutexes. In my own highly concurrent system, this appears to fix this issue. I could not find any other shared state for this case.


parrt:
Hi @pboyer can you pull from master again? I fixed the C# thing causing failure. thanks!

pboyer:
@parrt Done.

PR #1771, opened 2017-03-18
A few updates to the Unicode documentation.

mike-lischke:
It should be made clear that the recommended use of CharStreams.fromPath() is a Java-only solution. The other targets just have their ANTLRInputStream class extended to support full Unicode.

### DIFF
--- a/doc/unicode.md
+++ b/doc/unicode.md
@@ -11,12 +11,12 @@ or the equivalent method for your runtime's language.
 # Unicode Code Points in Lexer Grammars
 
 To refer to Unicode [code points](https://en.wikipedia.org/wiki/Code_point)
-in lexer grammars, use the `\u` string escape. For example, to create
+in lexer grammars, use the `\u` string escape plus up to 4 hex digits. For example, to create
 a lexer rule for a single Cyrillic character by creating a range from
 `U+0400` to `U+04FF`:
 
 ```ANTLR
-CYRILLIC = ('\u0400'..'\u04FF');
+CYRILLIC = ('\u0400'..'\u04FF'); // or [\u0400-\u04FF] without quotes
 ```
 
 Unicode literals larger than U+FFFF must use the extended `\u{12345}` syntax.
@@ -24,10 +24,10 @@ For example, to create a lexer rule for a selection of smiley faces
 from the [Emoticons Unicode block](http://www.unicode.org/charts/PDF/U1F600.pdf):
 
 ```ANTLR
-EMOTICONS = ('\u{1F600}' | '\u{1F602}' | '\u{1F615}');
+EMOTICONS = ('\u{1F600}' | '\u{1F602}' | '\u{1F615}'); // or [\u{1F600}\u{1F602}\u{1F615}]
 ```
 
-Finally, lexer char sets can include Unicode properties:
+Finally, lexer char sets can include Unicode properties. Each Unicode code point has at least one property that describes the type group to which it belongs (e.g. alpha, number, punctuation). Other properties can be the language script or special binary properties and Unicode code blocks. That means however, that a property specifies a group of code points, hence they are only allowed in lexer char sets.
 
 ```ANTLR
 EMOJI = [\p{Emoji}];
@@ -40,6 +40,7 @@ escapes in lexer rules.
 
 # CharStreams and UTF-8
 
+## Java Target
 If your lexer grammar contains code points larger than `U+FFFF`, your
 lexer client code must open the file using `CharStreams.fromPath()` or
 equivalent in your runtime's language, or input values larger than
@@ -51,7 +52,10 @@ For backwards compatibility, the existing `ANTLRInputStream` and
 The existing `TestRig` command-line interface supports all Unicode
 code points.
 
-# Example
+## Other Targets
+Other language targets usually have their `ANTLRInputStream` extended to support the full Unicode range. See the target documentation for supported input encodings (e.g. UTF-8) and other related details.
+
+# Java Example
 
 If you have generated a lexer named `UnicodeLexer`:
 


parrt:
@bhamiltoncx should we take a wack at proposing a similar API for other targets or leave as Java only?

bhamiltoncx:
Let's standardize the other runtimes if we can!

parrt:
Ok, if it's not a huge hassle, could you make a PR the @antlr/antlr-targets people can examine?

bhamiltoncx:
Will do! We're having a bit of a wildfire here today, so hopefully we won't
be evacuated. :)

mike-lischke:
Uhm, why do you want to add extra code just for handling full Unicode? To me it makes much more sense to just have ANTLRInputStream handle it.

bhamiltoncx:
Mike: The worry was that changing ANTLRInputStream's default behavior would
break backwards compatibility.

parrt:
Read about that! Be safe!

parrt:
Well in the case of C plus plus, you don't have any legacy code so I guess it doesn't make sense for that target. To avoid breaking anything that exists we are leaving the old streams in place for other target by providing a new interface to access the new Unicode functionality

mike-lischke:
Sorry for my ignorance, but why can't you have backward compatibility and still use  `ANTLRInputStream`? Just because this class would allow now values > 0xFFFF doesn't mean it would break any existing application. They don't use such values so there is no behavior change for them. Sounds like fixing a problem where none is, to me.

parrt:
Once they are 32 bit code points it's no problem. It's properly decoding the input stream that is encoded as utf-x not 32-bit Ints

bhamiltoncx:
@mike-lischke: So, the situation is that in Java, C#, and JavaScript, the runtime for `ANTLRInputStream` (or its language equivalent) actually exposed **16-bit UTF-16 code units** — not Unicode code points — to the lexer.

So, for example, the Unicode input `U+1F600` (😀) when sent through `ANTLRInputStream` in Java, C#, or JavaScript, would actually expose *two* values to the lexer (`\uD83D\uDE00`), not one.

Changing that now would be a fairly serious backwards-compatibility issue, so we punted and I made sure to require clients to call a different API in those languages.

You're correct that the C++ runtime didn't have this issue.

In my analysis on https://github.com/antlr/antlr4/issues/276#issuecomment-275171258 , I confirmed that Go, Swift, and Python 3 were also OK.

Later, I found Python 2 was OK on Linux, but not Mac or Windows.

Since more than one language had issues (and the API to the `ANTLRInputStream` equivalent of each language is pretty different), I figured it'd be a good idea to consolidate everything into `CharStreams`.

mike-lischke:
> So, for example, the Unicode input U+1F600 (😀) when sent through ANTLRInputStream in Java, C#, or JavaScript, would actually expose two values to the lexer (\uD83D\uDE00), not one.

Yes, sure, that's why the ANTLRInputStream must be enhanced to return the 32bit values.

> Changing that now would be a fairly serious backwards-compatibility issue (it was perfectly legal to write a lexer for those two values \uD83D\uDE00),

Not at all if you would add encoding support to `ANTLRInputStream`. You can support UTF-16 (and make that the default to stay compatible) and also support UTF-8 (what is now in CharStreams) and even UTF-32 (which doesn't require any conversion). That would be a much better implementation because users don't have to use a new API and don't have to choose between the two just because **their input can be so or so** (that's just crazy). Seriously, I see that as unnecessarily confusing to have to use different stream implementations.

bhamiltoncx:
There are two encodings to think about here.

First is the input encoding: when reading streams of bytes, how do we
convert those bytes to Unicode?

Second is the internal storage encoding used inside ANTLRInputStream: which
Unicode encoding should we use to store the Unicode code points in memory?

ANTLRInputStream already supports many input encodings via subclasses like
ANTLRFileStream.

It's true that ANTLRInputStream only supports a single internal storage
encoding (UTF-16) and we could add support for other storage encodings.

However, ANTLRInputStream is a very leaky abstraction, as it exposes its
internals to subclasses via protected members. It is fairly common for code
outside the ANTLR runtime itself to manipulate these internals via
subclasses.

Changing the internal storage encoding would mean changing not just ANTLR
but all external subclasses as well. Not impossible, but when I proposed
it, others suggested just avoiding breaking backwards compatibility by
using a separate class entirely.

Ben

parrt:
For long-lived libraries, it's common to change the interface and leave the old; e.g., job ahead to fix the streams so that they handled Unicode / utf-8 properly and created the Reader interface. Let's just see what @bhamiltoncx  comes up with for the various targets and let those target authors discuss. It sounds like there's no point in creating something for C++ as it has no legacy (yet!) :)

bhamiltoncx:
OK, sent out https://github.com/antlr/antlr4/pull/1775 and https://github.com/antlr/antlr4/pull/1776 as examples of how the `CharStreams` API would look for C# and JavaScript.

C# and JS were the other two languages besides Java whose `ANTLRInputStream` equivalents exposed UTF-16 code units instead of Unicode code points. We wouldn't really need to provide this for other languages unless we wanted consistency.

mike-lischke:
OK, seems I'm fighting on a lost battle here. I'm sorry but in the case I believe you are not doing the right thing. There were so many changes since 4.5.x without that you created alternative classes to keep "backward compatibility". Backward compatibility is a nice thing, but you can take it too far and I believe this happens here. Version 4.7 is not a point release. Nobody expects that you just drop it in and forget about it. People have to and will check if their software still works after upgrade. I'm sure it's not asked too much to also check if their solution still works with the enhanced input handling. Most people won't notice anyway.

Now we have half of the targets using a different input solution than the others, which makes documentation more complicated and examples harder to convert. I really disagree with the approach taken here. I'm sorry for that.

bhamiltoncx:
I hear you! Personally, I agree and think we should break compatibility
here as well, but I understand the concerns of Terence and Sam.

parrt:
What say the other @antlr/antlr-targets authors?

parrt:
Ok, NOW I remember: the new stream has to use 32-bits per symbol, which literally doubles the size requirements to hold a document in memory. Given that I regularly process large corpora and hold them all in memory at once, this is a nontrivial burden. For a parser, ANTLR-generated parsers are already memory pigs.

I therefore cannot simply replace ANTLRFileStream with one that takes twice as much memory. I think that the best solution is to simply add a new ANTLRFileStream32 or some such and users that are willing to take the hit in order to do 32-bit Unicode, can do so. Existing users will not experience any increase in memory or speed reduction or backward compatibility.  Given this, it doesn't make sense to have a new interface `CharStreams.XXX`; we should stick with the ctor-based mechanism and do `new ANTLRFileStream32(...)` or whatever.

bhamiltoncx:
We could keep the memory use down (or even improve it for ASCII text) in
several ways without introducing a new class:

1) Use UTF-8 encoding and build a lookup table of code point byte offsets
following any non-ASCII code points
2) Use UTF-16 encoding and build a lookup table of code point offsets >
U+FFFF

parrt:
yeah, I was wondering about encoding it but it seems like the lookup table would either add a lot of memory or slowdown access for `LT(i)`.

I think we should just introduce the new `CodePointCharStream` class with appropriate constructors. I am open to that class name or `ANTLRInputStream32` or some such.

mike-lischke:
I was thinking about how to make the input stream decode utf-8 on the fly. No need to keep the entire document in memory then. The token stream will keep the start/stop indexes of the chars for seeking in the input.

parrt:
Well, we also have the unbuffered stream. I'd rather not mix the two.

Also I noticed that the new `CodePointCharStream` can only handle UTF-8 versus any other encoding. Ben indicated that that is more or less what everyone uses. Can people comment about other locales, such as India/Japan/Laos? In other words, do people need the ability to read non-unicode char points and possibly with some non-UTF-X related encoding? In other words, people surely had file formats for Japan before Unicode came around... do we need to worry about this? I ask this out of ignorance.

bhamiltoncx:
It's extremely unlikely that anyone using ANTLR will be parsing non-UTF-8
input, except on Windows.

Lookup tables are fine for applications like ANTLR, since non-ASCII input
is pretty unlikely.

We could also dynamically change between 8/16/32 bit storage depending on
the input.

parrt:
Wouldn't it complicate the code tremendously to switch between 3 different fields, or at least decreased performance quite a bit?

The lookup table to me seems like it either cost even more memory than just having 32-bit characters or it would drop performance from O(1) to O(n) for indexed lookup.

bhamiltoncx:
Since we read the entire contents of the stream into memory before
constructing the object, we can choose the optimal width for storage once,
so there would be no impact at runtime.

A lookup table can be made in a number of ways to avoid perf impact.

mike-lischke:
@parrt Speaking of unbuffered char streams, do all targets support 32bit unbuffered streams? From what I see the C++ target does not. There is no UTF-8 decoding going on. Seems I totally missed that.

parrt:
that's if we go to the new factory based instead of constructor based style.

 can you point me at something that indicates how those lookup tables can be done?

@mike-lischke  On not sure if the other targets have created the unbuffered versions. Maybe it's only Java. I have no idea whether it's actually used are not

bhamiltoncx:
Even with the existing constructor, we pass in the contents of the text
data, so we could convert there to 8, 16, or 32-bit storage then discard
the input.

But yeah, for optimal storage, we would need the factory type initializer
to avoid the extra temporary copy.

parrt:
Hmm...yeah, I still think it would complicate the code dramatically to choose between byte, short, int fields or using just `byte[]` and then indexing per how wide chars are.

parrt:
Unless of course we have separate classes, which is what we have now ;)

bhamiltoncx:
I bet I could abstract it pretty nicely if we wanted to go that way!

parrt:
Well, I'll sit on it for a day or so to see who else pops up but I'm leaning towards just using ctor-style `ANTLRInputStream32`, declaring victory on moving on :)

bhamiltoncx:
My main problem with that is I feel pretty strongly that:

1. The main API should Do The Right Thing (code points) by default
2. Introducing an API with numbers at the end almost guarantees people will not use or understand them
3. We want some kind of path to deprecate the UTF-16 API eventually for Java, JS, and C#

I don't think introducing an `ANTLRInputStream32` really gets us there.

parrt:
I mostly agree with 1 and 2 but I'm not sure using 50% less RAM will go out of style for me. ;)  Maybe we make ANTLRFileStream / ANTLRInputStream be the 32-bit code point stuff and shift 16 bit stuff to a new name. Getting those streams to work correctly for chars, tokens, etc... was fiddly. I hesitate to try for an all-in-one class.  I don't see how to get single interface w/o factory which means changing API from ctor style.  That means C++/Python/etc... will be different from Java and the other ones we have PRs for.

bhamiltoncx:
I hear you on improving RAM usage! I can send a diff to update `CodePointCharStream` to choose one of `IntBuffer` / `CharBuffer` / `ByteBuffer` for storage based on the contents of the input.

I would be OK with shifting the existing 16-bit logic in the runtimes to a legacy name if that's good with everyone else. However, we'd have to rename all the subclasses as well.

bhamiltoncx:
Basically, my strategy would be to assume US-ASCII only, and start with `ByteBuffer` until we see a code point > `U+007F`. If we see a 16-bit code point between `U+007F` and `U+FFFF`, we upgrade to `CharBuffer`. If we see a surrogate or code point larger than `U+FFFF`, we upgrade to `IntBuffer`.

millergarym:
I'm not sure if this useful, as I'm not fully across the issue. Here a perspective from Go land.

In Go all strings are utf8 encoded. Not supprising since the authors of utf8 authored Go. All built in functions & controls know this. len( somestring ) return the number if 'runes' not bytes, range somestring returns runes, somestring[i] is a rune. For ascii a rune is a byte and bigger for codepoints higher the ascii (look up the tech details if necessary). Strings are really stored as a byte array but from the programmers perspective are an array of runes.

I don't know how Java handles this. Might get expensive if every byte needs to be boxes.

Does talking about ascii vs utf8 make a difference than talking about 8 or 32 bits?

bhamiltoncx:
@millergarym : Right. The logic to transparently choose between 8, 16, and 32-bit units for Go `runes` is basically what I'm suggesting.

ericvergnaud:
We should NOT read the entire contents of the stream.
We've seen use cases where the consumer fetches a construct repeatedly on a potentially infinite stream

Envoyé de mon iPhone

ericvergnaud:
Sorry but this is not unlikely at all.
In banks we have all sorts of mature systems with all sorts of encodings.
An ANTLR parser running on a Linux box consuming data from these is much more frequent than what people might think.

Envoyé de mon iPhone

ericvergnaud:
We definitely do!

Envoyé de mon iPhone

bhamiltoncx:
@ericvergnaud : I'm just talking about the default behavior of `ANTLRInputStream` / `ANTLRFileStream` in Java, which is to read the entire stream into memory. We can make better-behaved ones of course.

The `CharStreams` code I added doesn't require UTF-8, of course. It supports other encodings, but the Java lgoic assumes they will never have code points > `U+FFFF`.

bhamiltoncx:
@parrt: I sent out https://github.com/antlr/antlr4/pull/1781 to improve memory usage of the new Java runtime.

It should actually be ~ 2x better than the existing `ANTLRInputStream` for US-ASCII-only content, depending on the initial buffer size we choose.

bhamiltoncx:
> In banks we have all sorts of mature systems with all sorts of encodings.

And yes, @ericvergnaud is totally correct — I just meant it might not be as important to worry about making the non-UTF-8 code path performance-optimal. (It of course needs to be fully functional and bug-free!)

parrt:
@ericvergnaud  regarding reading entire input; that's exactly what we've been doing for years! hahaha

bhamiltoncx:
@parrt : I do see @ericvergnaud 's point, though — for the factory-based APIs which take in a stream, it makes sense to have a version which also takes a callback which we invoke after each read to ask the client if we should continue reading.

That allows the client to parse multiple objects directly from a stream with some kind of separator (imagine a streaming JSON parser which separates JSON objects by `\n`).

parrt:
The problem is that the parser does all sorts of lookahead and itself must control how much of the input the buffer, possibly the entire thing.

bhamiltoncx:
Sure, but the socket client can have some other kind of inexpensive delimiter like `\n` which guarantees the parser will have a full buffer available, after which another full buffer will occur.

@ericvergnaud 's point is that without a way for the client to say "hold off on parsing more", there's no way to implement these types of streaming parsers.

parrt:
There is no guarantee that the parser does not need to see beyond `\n`. It all depends on the skill of the grammar writer, unfortunately.    The best way is to have a reader, outside of antlr stuff, that collects one string at a time and sends that to the parser.

Whatever we have now appears to be working, unless there's a case people are simply not telling me about. The guys that Apple specifically asked me if they could do unbuffered input for reading sockets and that sort of thing and without building a tree. I made sure that is possible.

bhamiltoncx:
I suspect the missing connection here is that the new `CharStreams` factory API doesn't have an unbuffered input option from which clever parser authors can create such a client.

parrt:
ah. heh, i just found something i wrote down (in the book):

13.8 Unbuffered Character and Token Streams
Because ANTLR recognizers buffer up the entire input character stream and all input tokens by default, they can’t handle input files that are bigger than a computer’s memory and can’t handle infinite streams like socket connec- tions. To overcome this, you can use unbuffered versions of the character and token streams, which keep just a small sliding window into the streams: UnbufferedCharStream and UnbufferedTokenStream.

```java
CharStream input = new UnbufferedCharStream(is);
CSVLexer lex = new CSVLexer(input);
// copy text out of sliding buffer and store in tokens lex.setTokenFactory(new CommonTokenFactory(true));
TokenStream tokens = new UnbufferedTokenStream<CommonToken>(lex);
```

ericvergnaud:
c’m'on!!!

parrt:
@ericvergnaud  what do you mean?

parrt:
Per @mike-lischke, we really should make the API consistent across language targets. @bhamiltoncx says:

> C# and JS were the other two languages besides Java whose ANTLRInputStream equivalents exposed UTF-16 code units instead of Unicode code points. We wouldn't really need to provide this for other languages unless we wanted consistency.

This would mean a factory style `CharStreams.fromFileName(...)` type thing for all targets. Is this even meaningful in C++? I suppose it could be just a static function scoped inside `CharStreams::fromFileName(...)`?  It would appear that @ericvergnaud  has no fundamental disagreement with adding this interface for his targets. What about the other @antlr/antlr-targets? I am warming to the new factory style interface.

@bhamiltoncx  has removed my size objection to the new Unicode code point streams by making it 50% the size that he used to be. haha. kudos. It of course is using the new factory interface, though.

The new interface would be described in the documentation as the way to get input into the lexer.

I think the only remaining question is what to do with existing ANTLRInputStream and friends. choices:

1. we leave the ANTLRInputStream and friends around as-is so that existing code compiles and runs.
2. we leave those class names around but replace their guts with the new code point stream. This has the advantage of having just one internal mechanism to handle character streams; the class names would hang around just for backward compatibility.

In light of the fact that the new stream will be 8-bit optimized, it seems like option number 2 is the right choice.

bhamiltoncx:
I like that proposal (adopting the factory-style interface, which could just be namespaced functions for C++). We can make the existing `ANTLRInputStream` a simple wrapper if we think the perf and memory usage of the new code is good.

I agree the factory-style interface doesn't do a ton for C++ today, but it opens up possibilities for more types of input streams in the future.

If we're cool with getting rid of backwards compatibility with UTF-16 input in Java, C#, and JS, I think it'd be a big step forward.

parrt:
Re UTF-16: oh right. There were a number of grammars that were manually looking for UTF-16 code units in an effort to simulate 32-bit code points right? Hmm... probably okay.

I forgot to ask about different encodings, like the legacy ones Eric mentioned from bank systems. How do we incorporate different encodings than UTF-X?  Will this explode the number of methods we need in the common character stream factory interface?

bhamiltoncx:
> Re UTF-16: oh right. There were a number of grammars that were manually looking for UTF-16 code units in an effort to simulate 32-bit code points right? Hmm... probably okay.

Right. So, we have three options:

1. Provide no support for such grammars
2. Keep `ANTLRInputStream` around as-is for such grammars but mark it and its equivalents as deprecated in Java, C#, and JavaScript
3. Provide a new, supported API to create `CharStreams` which expose UTF-16 code units

> I forgot to ask about different encodings, like the legacy ones Eric mentioned from bank systems. How do we incorporate different encodings than UTF-X? Will this explode the number of methods we need in the common character stream factory interface?

Mostly done. I made the new factory APIs take an optional `Charset`:

https://github.com/antlr/antlr4/blob/1f6a3296926f79cf31a52c15cd72d3ff6f4cd623/runtime/Java/src/org/antlr/v4/runtime/CharStreams.java#L59

Right now, these APIs delegate to `ANTLRInputStream` which converts from those legacy encodings to UTF-16, but I'll follow up to https://github.com/antlr/antlr4/pull/1781 and make them directly convert to `CodePointCharStream`.

mike-lischke:
I see there's really a lot going on here, great, with @bhamiltoncx giving 100% for a good solution. Still to me that entire thing looks very Java-centric. Language have streams and ANTLR should support those (I mean their native streams), via a simple interface (CharStream probably). The idea should be: "user, give me a stream that returns 32bit code points to me, period". The user then has to take care to convert other encodings to UTF-32, if required; ANTLR cannot cope with all used encodings in this world. The factory approach doesn't allow to select the stream class that fits a specific purpose. IMO ANTLR shouldn't care what native stream is used under the hood. This is also the most flexible approach for us target authors.

For convenience ANTLR provides one such stream out of the box (the one that takes UTF-8). If backwards compatibility is such a big concern then we can indeed deprecate the current ANTLRInputStream class and provide a new AntlrUtf8InputStream. The new targets just rename their ANTLRInputStream classes and the older ones add a new class with that name. This would be my preferred solution here.

Just a few more thoughts for the general handling as I imagine it:

If a stream loads everything into memory or not should **not** be the decision of the ANTLR runtime (users might have a [native] stream interface for memory mapped files that allow to deal with gigabyte sized input, which is both: not buffered and allows random seeking). Streams can indicate if they support random seek or not. Most of the time users will probably use a string stream, which allows to handle in-memory strings like a stream (some targets like C++ already have a class for this). This approach will also end the buffered vs unbuffered separation.

ericvergnaud:
@parrt not sure what I had in mind... best to forget about it :-)

ericvergnaud:
Hi,

I’m sure all of us are already comfortable with the below, but just in case it may be worth restating the difference between character set and encoding

ASCII, Unicode 3.0 (U+0000 to U+FFFF) and Unicode 3.1 (U+0000 to U+10FFFF) are character sets, not encodings.
Technically, ANTLR is character set agnostic i.e. it makes no assumption about the meaning of a byte or series of bytes coming from the input stream.
But since Java chars were defined from Unicode 3.0 and are 16 bits wide, there is a strong inclination to use the Unicode 3.0 character set, and UTF-16, even without knowing.

UTF-8, UTF-16 and UTF-32 are encodings of Unicode 3.1. i.e. they all support the widest character set.
UTF-8 encodes each character using varying lengths (1 to 4 bytes). Great for space, horrible for string manipulation. UTF-16 uses 1 or 2 shorts. UTF-32 uses 1 32-bit int.
Other examples of encodings are MBCS, DBCS, Shift-JIS etc.. they are a way to encode a character set, not the character set itself.

I’m personally comfortable with the idea that ANTLR itself would internally only support Unicode 3 and Unicode 3.1 character sets, since Unicode 3.1 encompasses all glyphs ever imagined, and upcoming ones such as emoticons.
For binary streams, character sets have no meaning so it would have no impact.

Practically, it means that a user of the « factory » should only have to:
 - specify the technical source (file, string, buffer, socket, stdin…) and its encoding (currently utf-8 by default, but can be specified)
 - specify the character set to use internally i.e. Unicode 3 (using 16-bit code points) or Unicode 3.1(using 32-bit code points) 

I’m unable to state what exact conversions are supported by each language, but I’d be surprised if at minimal conversion from all widely used encodings to the language’s native encoding was not supported by native utilities.
In Java: InputStreamReader, in C#: StreamReader, in Python: codecs.decode, in Node: fs.readFileSync, in Swift: Utils.readFile, all these accept an encoding as parameter (and currently used by ANTLR)
This shows that while the conversion is specified through an ANTLR API, it is performed outside ANTLR.

So not sure why ANTLR would have to cope with more than the 2 proposed encodings, since the conversion itself is already handled by the native decoders.
In short, all ANTLR has to provide and maintain is conversion from a target’s native encoding (coming out of the native conversion utility) to the one to be used internally i.e. 16 or 32 bits code points
From there, it doesn’t seem unachievable to expose the same API across all targets, especially now that @bhamiltoncx <https://github.com/bhamiltoncx> has done all the hard work.

I notice that Go and Cpp do not yet support file stream encoding.
On the short term, I would recommend adding the parameter, and throw if it is not UTF-8. This would I believe clarify the requirement, until conversion is supported.
in Go:golang.org/x/text/encoding <http://golang.org/x/text/encoding>, in Cpp: ICU4C provide the equivalent functionality.

Eric

n.b. I also agree that incoming data should not be buffered totally.

PR #1775, opened 2017-03-20
New C# CharStreams static factory class

bhamiltoncx:
Based on the discussion at https://github.com/antlr/antlr4/pull/1771#issuecomment-287627618 , this brings the Java runtime's `CharStreams` factory API to C#.

I updated `BaseCSharpTest` to make use of the new API to illustrate how it works.

### DIFF
--- a/runtime-testsuite/test/org/antlr/v4/test/runtime/csharp/BaseCSharpTest.java
+++ b/runtime-testsuite/test/org/antlr/v4/test/runtime/csharp/BaseCSharpTest.java
@@ -725,13 +725,11 @@ public class BaseCSharpTest implements RuntimeTestSupport /*, SpecialRuntimeTest
 				"\n" +
 				"public class Test {\n" +
 				"    public static void Main(string[] args) {\n" +
-				"        string inputData = File.ReadAllText(args[0], Encoding.UTF8);\n" +
+				"        var input = CharStreams.fromPath(args[0]);\n" +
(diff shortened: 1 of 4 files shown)


parrt:
@ericvergnaud Can you comment on this new API?

ericvergnaud:
Overall looks good except FromStream which unlike other APIs should NOT 'ReadToEnd'.
If you're parsing from a socket, or stdin, EOF could be far away, or even unreachable.

bhamiltoncx:
Makes sense. But how do we know how much to parse? Add length args?

ericvergnaud:
You don't need to know.
You let the grammar author del with it.
i.e. with stdin they can decide to stop when encountering 'exit' token
with socket they can detect a pattern to reset

parrt:
I've added important comments [here](https://github.com/antlr/antlr4/pull/1771#issuecomment-288007686)

bhamiltoncx:
> You don't need to know.
> You let the grammar author deal with it.
> i.e. with stdin they can decide to stop when encountering 'exit' token
> with socket they can detect a pattern to reset

Hmm. Currently, the design of `ICharStream` doesn't really allow for reading less than the entire input; all known implementations read the entire contents before constructing `ICharStream`.

Could you give me an example of what you mean?

parrt:
@bhamiltoncx  are these consistent now with our tweaks to the Java stream API?

bhamiltoncx:
Yes, everything is consistent, although I didn't do any optimization passes yet, so we don't have a version that takes a size.

PR #1781, opened 2017-03-22
Improve memory usage and perf of CodePointCharStream: Use 8-bit, 16-bit, or 32-bit buffer

bhamiltoncx:
This greatly improves the memory usage and performance of `CodePointCharStream` by ensuring the internal storage uses either an 8-bit buffer (for Unicode code points <= `U+00FF`), 16-bit buffer (for Unicode code points <= `U+FFFF`), or a 32-bit buffer (Unicode code points > `U+FFFF`).

I split out the internal storage into a class `CodePointBuffer` which has a `CodePointBuffer.Builder` class which has the logic to upgrade from 8-bit to 16-bit to 32-bit storage.

I found the perf hotspot in `CodePointCharStream` on master was the virtual method calls from `CharStream.LA(offset)` into `IntBuffer`.

Refactoring it into `CodePointBuffer` didn't help (in fact, it added another virtual method call).

To fix the perf, I made `CodePointCharStream` an `abstract` class and made three concrete subclasses: `CodePoint8BitCharStream`, `CodePoint16BitCharStream`, and `CodePoint32BitCharStream` which directly access the array of underlying code points in the `CodePointBuffer` without virtual method calls.

PR #1787, opened 2017-03-23
Fix TimeLexerSpeed when run from jar

bhamiltoncx:
This fixes the issue where we were passing URLs to APIs which expected filenames:

https://github.com/antlr/antlr4/pull/1781#issuecomment-288807273

I had to use the `Reader` API of `ANTLRInputStream` since the `InputStream` constructor constructs an `InputStreamReader()` without specifying a `Charset`.

That API sadly uses the default `Charset` from the test runner's environment, which is not necessarily UTF-8.

I also removed the ASCII time tests, since they were functionally identical to the UTF-8 test.

The timings are a bit different now, since we include time to read the file consistently across the `CharStream` implementations.


parrt:
I forgot to hit send on an email to ya. i added mem info. can you pull again to resolve conflicts?

bhamiltoncx:
Will do.

PR #1791, opened 2017-03-26
use JOL to compute buffer sizes. Dumps footprint now too.

parrt:


parrt:
@bhamiltoncx When not loading from a file (using a stream instead), we don't know buffer size and it allocates way more than needed due to buffer size doubling during loads.

bhamiltoncx:
Should be easy to add a stream API that takes size and allow perf-sensitive
types to stat() the file beforehand to get its exact size in bytes.

We can also support a preflighting API so we can count the code points and
do an exact size, but usually the extra I/O isn't worth it.

parrt:
I fear we are exploding the interface size. How about if we say `.fromFileName` or `.fromPath`, it gets the size right by asking the object we are loading?

parrt:
I'm going to add some tests (back) that won't work from the jar, but directly load from a file to see what the sizes are.

bhamiltoncx:
If we're worried about the interface size, easy to consolidate it down with
a `CharStreams.Builder` which takes the params. Then we get down to a
single API!

parrt:
haha, yeah but an unreadable API like python libs tend to have. :)

parrt:
Ok, new tests for file load confirm proper char buffer sizes when we know the size. (It gives warning when run from jar.)

PR #1792, opened 2017-03-27
New micro-benchmarks using JMH

bhamiltoncx:
This uses the excellent JMH framework (http://openjdk.java.net/projects/code-tools/jmh/) to build a set of micro-benchmarks for the ANTLR 4 `ANTLRInputStream` and `CharStreams`, as well as for a few sample `Lexer` grammars (Java and Unicode code points).

I included a script in `benchmark` called `run.sh` to build and run the microbenchmarks.


parrt:
This is really cool but I think I'm going to leave it out...just cleaning up the PR list really.

PR #1793, opened 2017-03-28
Obsolete String.contains(_:) of Swift runtime

cmd-johnson:
Starting with [iOS 10, macOS 10.12, tvOS 10.0 and watchOS 3.0](https://developer.apple.com/reference/swift/string/1643241-contains), Foundation contains its own definition of String.contains(_:), which conflicts with the extension provided by antlr and breaks builds targeting iOS 10 and above.
This change shouldn't affect any versions below the ones stated above.


parrt:
@hanjoes  Does this look correct?

hanjoes:
I agree with this change. The current implementation marked the String extension as "public" which will be exposed to user of Antlr4 all the extended methods. We can first mark this particular method obsolete for later SDK versions and then remove "public" access modifier for the extension. @cmd-johnson would be great if you can do it in this single PR.

hanjoes:
btw, by removing "public" modifier of the extension, it runs the risk of breaking backward compatibility if any users of Antlr4 actually uses some of these  String extensions, but by all means it's not a part of Antlr4, @parrt are you okay with that?

cmd-johnson:
@hanjoes I'll take care of that later today.

Reading my PR again, with 'breaks builds targeting iOS 10 and above' I meant projects _using_ the Antlr runtime, not the runtime itself and looking for usages of the `String.contains(_:)` extension inside the runtime itself, I couldn't find any so far. I'll try and build the whole thing with the method removed altogether and if that works, add that change to the PR instead (together with removing the "public" modifier of the extension).
May I also propose renaming the source-file of the extension to "String.swift" instead of "Stirng.swift"?

By the way, great work with the Swift runtime and Antlr4 in general, it's really a charm working with it.
I'm relatively new to both Antlr4 and Swift, but I noticed the Antlr Swift-API style differs slightly from the styles of other Swift frameworks I've used (for example, you are using getter-methods instead of computed properties). Would it be worth putting my time into refactoring methods to make the API more 'swifty'? Obviously those changes wouldn't be backwards-compatible. Also I assume the Swift-API is mostly identical to the Java one, maybe that's something you want to keep.

hanjoes:
Great point about the API style, thats also my long term concern. the original impl was 100% mimicing the JAVA runtime (we even have a StringBuilder!) which is not even close to swiftly. I think we might want to deprecate the old API while adding support for the new ones.

parrt:
@cmd-johnson 4.7 will be going out the door in the next day or so. If you'd like this integrated, see if you can sneak in any changes you'd like to make to this PR.

cmd-johnson:
I double-checked the swift runtime, and from what I can tell, the method is never actually used, so I removed it. I removed the `public` modifier from the extension and fixed the spelling error of the extension's source file - this PR should be good to go.

parrt:
@hanjoes can you do a final check?

hanjoes:
I did a workspace search for the keyword "contains" and looked into each usage of that symbol. Didn't see any usage of that extension. This PR is good to go.

PR #1794, opened 2017-03-29
Tidy comments and add tests to ensure UTF-16 and UTF-32 support Unicode code points > U+FFFF

bhamiltoncx:
Following up to https://github.com/antlr/antlr4/pull/1781#issuecomment-290154707 , this removes the outdated comments that suggest UTF-8 is necessary to support code points > `U+FFFF`, and adds tests to confirm `CharStreams` supports those code points with UTF-16 and UTF-32.


parrt:
@bhamiltoncx just realized: how can InputStream support UTF-32 since it's `read()` only returns up to 0xFFFF? Do we need to comment on the various methods to indicate which can support > 0xFFFF?

PR #1796, opened 2017-03-29
Change UnbufferedCharStream to use code points

### LINKED ISSUE #1795: What to do with UnbufferedCharStream for 4.7?
@bhamiltoncx  Is there a way we can easily update `UnbufferedCharStream` so that it uses the new 32-bit functionality? it would be okay to get the implementation out and replace with a delegation to the new stuff. there would be no concern about buffer size in memory.

bhamiltoncx:
Fixes #1795 .

This changes the Java implementation of `UnbufferedCharStream` to use code points, not UTF-16 code units, following the changes to `CharStreams`.

For now, I didn't bother putting this in the `CharStreams` interface, but if we end up changing that to use a `Builder` style, we could make an `unbuffered()` option.

I updated the tests. Note the weirdness around `\uFFFF` — before, we were relying on `(char)IntStream.EOF` converting `-1` to `\uFFFF`, but now we no longer have that, so the test code explicitly appends `\uFFFF`.

### DIFF
--- a/runtime/Java/src/org/antlr/v4/runtime/UnbufferedCharStream.java
+++ b/runtime/Java/src/org/antlr/v4/runtime/UnbufferedCharStream.java
@@ -25,7 +25,7 @@ public class UnbufferedCharStream implements CharStream {
 	 * we keep adding to buffer. Otherwise, {@link #consume consume()} resets so
 	 * we start filling at index 0 again.
 	 */
-   	protected char[] data;
+	protected int[] data;
 
 	/**
 	 * The number of characters currently in {@link #data data}.
@@ -82,7 +82,7 @@ public class UnbufferedCharStream implements CharStream {
 	/** Useful for subclasses that pull char from other than this.input. */
 	public UnbufferedCharStream(int bufferSize) {
 		n = 0;
-		data = new char[bufferSize];
+		data = new int[bufferSize];
 	}
 
 	public UnbufferedCharStream(InputStream input) {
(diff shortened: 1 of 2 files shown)


runtime/Java/src/org/antlr/v4/runtime/UnbufferedCharStream.java
```diff
@@ -82,7 +82,7 @@ public UnbufferedCharStream() {
 	/** Useful for subclasses that pull char from other than this.input. */
 	public UnbufferedCharStream(int bufferSize) {
 		n = 0;
-		data = new char[bufferSize];
+		data = new int[bufferSize];
 	}
 
 	public UnbufferedCharStream(InputStream input) {
```
bhamiltoncx: Note this ends up calling `new InputStreamReader(input)` without specifying a `Charset`. That means this logic depends on the client's `$LANG` environment variable or equivalent, which is usually not what anyone wants.

runtime/Java/src/org/antlr/v4/runtime/UnbufferedCharStream.java
```diff
@@ -183,8 +206,8 @@ public int LA(int i) {
         int index = p + i - 1;
         if ( index < 0 ) throw new IndexOutOfBoundsException();
 		if ( index >= n ) return IntStream.EOF;
-        char c = data[index];
-        if ( c==(char)IntStream.EOF ) return IntStream.EOF;
+        int c = data[index];
```
bhamiltoncx: Oops, just realized this can be tidied up to just `return data[index];`.

runtime/Java/src/org/antlr/v4/runtime/UnbufferedCharStream.java
```diff
@@ -145,13 +145,36 @@ protected void sync(int want) {
...
+			if (this.n > 0 && data[this.n - 1] == IntStream.EOF) {
 				return i;
 			}
 
 			try {
 				int c = nextChar();
-				add(c);
+				if (c > Character.MAX_VALUE || c == IntStream.EOF) {
+					add(c);
+				} else {
```
parrt: reminder that our "style" is else starts a line ;)

runtime/Java/src/org/antlr/v4/runtime/UnbufferedCharStream.java
parrt: It seems like we should add an optional charset arg on one or more of the arguments. Also, doesn't the fill() method assume UTF-X? I.e., it would not work with some other encoding, right?

bhamiltoncx: Yes! We absolutely should add a `Charset` arg, or there's no way to use anything but the default in the calling environment. (We should probably also change to default to UTF-8..)

`fill()` uses the `InputStreamReader` to get UTF-16 encoded `char`s. It will work with any encoding, although there's no way to specify an encoding other than the environment default with the current `UnbufferedCharStream` API.

parrt: ok, great! could you please make the changes?  Does that mean we change this field to InputStream?

```
    protected Reader input;
```

bhamiltoncx: Sure, I'll make the changes! Nope, we'll still use a `Reader`, we'll just use a different `InputStreamReader` constructor which allows us to specify a `Charset`.

PR #1799, opened 2017-03-29
Small tidy-up for Java UnbufferedCharStream

bhamiltoncx:
Noticed this when I ported #1796 to C# in #1798.


parrt:
thanks! ok, i'm documenting that only C#/Java have unbuffered stream

PR #1805, opened 2017-03-31
Python 3 codegen probably didn't mean to limit ATN segment size to 2 XOR 31

bhamiltoncx:
Noticed this poking around in the Python 3 codegen. `2 ^ 31` is not 2<sup>31</sup>, it's 2 XOR 31, which is 29..


ericvergnaud:
blessed

PR #1808, opened 2017-04-02
Fixing Problems after Swift3.1 Release

hanjoes:
Swift 3.1 was [released](https://swift.org/blog/swift-3-1-released) on Mar. 27th.
The language change should be source compatible, but our implementation is using some language feature that's not guaranteed by the public API thus breaking the ANTLR4 Swift runtime build. This PR will fix those issues along with several warnings because of the language release.

The PR  #1793 removing StringExtension public modifier is breaking one unit test that uses the "length" method in the old string extension. The second commit will fix this issue by unwrapping the "length" method in the SwiftTest.stg file. We don't want to expose (those pretty generic) extensions inside Swift runtime but need to make sure our generated parsers (in a different module) still works.


parrt:
Ah. yeah, just saw the errors fail: https://travis-ci.org/antlr/antlr4/jobs/219371570

PR #1810, opened 2017-04-04
[dotnet] update to .netcore v1.0.1; add osx build

xied75:
This updates dotnet build to use .netcore v1.0.1. And we added an osx build for dotnet.

But:

1, the osx runs more than 47 mins, and always got terminated by Travis, but it definitely works.
2, there is an issue preventing linux dotnet target working at all, there is nothing we can do other than hope Microsoft fix it asap. https://github.com/dotnet/core-setup/issues/1947

@ericvergnaud @parrt


parrt:
Interesting that it takes so long on os x. I will let @ericvergnaud try it out and let me know how it goes. I can try it also

xied75:
ping ?

ericvergnaud:
Hi, sorry haven’t had much time to look into this yet. It’s on my todo list.

runtime-testsuite/test/org/antlr/v4/test/runtime/csharp/BaseCSharpTest.java
```diff
@@ -493,83 +493,55 @@ public boolean createProject() {
...
+            final ClassLoader loader = Thread.currentThread().getContextClassLoader();
+            final URL runtimeProj = loader.getResource("CSharp/runtime/CSharp/Antlr4.Runtime/Antlr4.Runtime.dotnet.csproj");
+            if ( runtimeProj==null ) {
+                throw new RuntimeException("C# runtime project file not found!");
+            }
+            File runtimeProjFile = new File(runtimeProj.getFile());
+            String runtimeProjPath = runtimeProjFile.getPath();
 
-        try {
+            // add Runtime project reference
```
ericvergnaud: Adding a reference to the project (runtimeProjPath) rather than the Assembly is likely to be the cause of the very long Travis build with both Linux and MacOSX. I'm guessing this tells dotnet to rebuild both the test project and the runtime project.
Have you tried simply referencing the Assembly here?

ericvergnaud:
Hi,

it seems that the Ubuntu issue is solved, see https://travis-ci.org/antlr/antlr4/builds/218514385. Great!
Indeed the MacOSX build fails due to Travis timeout.

Re build time, I looked into it, see my comments in BaseCSharpTest.java/buildDotNetProject.

I was also able to build on MacOSX, and currently updating 'releasing-antlr.md' accordingly.

Re Nuget I noticed that the target assembly has **_not_** been renamed i.e. it has the same name as the Standard one. Can you change this to be 'Antlr4.Runtime.Core.dll' ? Then I can publish to Nuget.

Good job!

parrt:
@ericvergnaud looks like it still times out on Mac. should I wait to merge?

ericvergnaud:
Yes, I believe the solution is at hand, + we need the assembly to be renamed 

Envoyé de mon iPhone

xied75:
Regarding test run, just curious, the C++ test on OSX is also timing out, what is the reason there? Are they not going to improve on this?

parrt:
unfortunately that is probably a limitation of C++ :(

xied75:
@ericvergnaud The PR you just commented on is the old one, so I'll reply here instead.

Sorry for the radio silent. I'm busy trying to get a StringTemplate4 working on Antlr4 runtime in C# now, I need that for my product urgently.

Regarding the timeout, I don't think the runtime was compiled for each test. I need to investigate further, being this slow is embarrassing when other languages able to finish in 4 or 5 minutes. But I did notice cpp on Mac is also timing out.

Regarding the naming of the dll, my opinion is that calling it **Antlr4.Runtime.Core** is confusing, as **Core** normally means the bare central pieces of functions for a library, e.g. https://www.nuget.org/packages/Castle.Core/
But we are trying to use it to mean **dotnet core**. So a clearer name could be:
**Antlr4.Runtime.netcore** or **Antlr4.Runtime.netstandard**, let me know what you think.

ericvergnaud:
Well we can’t use « netstandard » because the regular version uses Antlr4.Runtime.Standard
And all nuget is about .Net so really don’t need net in the suffix.
Plus we need to cater with all the stuff published by Sam
So please stick to Antlr4.Runtime.Core which expresses the difference with Antlr4.Runtime.Standard i.e. .Net Core vs .Net Standard

Issue #1592, opened 2017-01-09
[C++] Memory leaks in the C++ runtime related to ParserATNSimulator and SingletonPredictionContext


Naios:
Valgrind reports some memory leaks on revision aa909ddcb16ed74f571f5 related to `ParserATNSimulator` and `SingletonPredictionContext`.

Log: https://gist.github.com/Naios/041853b195dce6de43a2bdd66d4004f3

Im using the lexer and parser as following:
```c++
GeneratedLexer lexer(&input);
auto errorListener = createDiagnosticListener(this);
lexer.removeErrorListeners();
lexer.addErrorListener(errorListener.get());

auto tokens = std::make_shared<antlr4::CommonTokenStream>(&lexer);

...

GeneratedParser parser(tokens.get());
parser.removeErrorListeners();
parser.addErrorListener(errorListener.get());
auto compilationUnit = parser.compilationUnit();
LocalScopeVisitor localScopeVisitor(compilationUnit_, astContext);
compilationUnit->accept(&localScopeVisitor);
````

mike-lischke:
I see what I can do there...

pureconfig:
@mike-lischke memory leaks causes me pain in an interpreter where I re-eval repeatedly, so it adds up quickly. Any progress on this one?

mike-lischke:
To be honest, I can't say when I come around to look at this. So many things on the agenda. But I keep trying...

mike-lischke:
@pureformat Have you done any recent checks again? Could be this issue is solved now.

PR #1813, opened 2017-04-05
CPP target runtime: fixed a memleak

FloorGoddijn:


Naios:
Did you check whether this is a possible fix for the leaks reported in #1592 ?

FloorGoddijn:
Hi Naios,

I have looked at your valgrind output, my fix is not fixing your problem, but is an equivalent situation,  I will closemy current pull request and fix yours as wel in a new one.

Regards,

Floro

FloorGoddijn:
I am closing this one as it turns out there more locations where it leaks in a similar way, and will fix this issue: https://github.com/antlr/antlr4/issues/1592

FloorGoddijn:
Sorry was looking at old code, this fix seems indeed to be the fix for https://github.com/antlr/antlr4/issues/1592 as well

Naios:
@FloorGoddijn Thanks a lot :-)

parrt:
@mike-lischke shall I merge?

mike-lischke:
@parrt Yes, that is a good fix, however I doubt this is completely fixing #1592 (from a quick look at the memcheck log I see that there are other areas affected), but I'm happy to be proved wrong :-)

Issue #1780, opened 2017-03-21
[FR] make ANTLR only generate the serialized ATN and print it on stdout


mike-lischke:
The idea behind this feature request is that it would be very helpful if the ANTLR tool could be put in a mode (via a command line parameter) where it does not generate anything but the serialized ATN and print that to the console, so it can be read via redirection. That would allow e.g. editors to implement debugging and edit help for ANTLR grammars, even for targets that cannot just recompile code on the fly.

In order to use the ParserInterpreter class it is necessary to also have the rule and token names in the correct order. Hence the printed output should include that too (maybe with a super simple protocol like: `<heading><token names><separator><rule names><separator><serialized ATN>`, the names list are just space separated strings).

PR #1823, opened 2017-04-15
Make ANTLR generate interpreter data

mike-lischke:
Especially when you want to use LexerInterpreter and/or ParserInterpreter in any of the non-Java targets you have to provide the ATN and other data. The classes to generate these values are not in the runtime, however. Hence we need a way to tell ANTLR to produce that in a way that can be consumed by all targets.

This patch adds a new command line parameter (-interpreter) which causes ANTLR to parse the given grammars as usual and then let it generated a file for each grammar with the required interpreter values. A new InterpreterDataReader class has been added to the Java + C++ runtimes. This class can load the data file (a plain text file) and generate the structures that can directly be fed to the interpreters.


mike-lischke:
Closes FR #1780

mike-lischke:
@antlr/antlr-targets The InterpreterDataReader might be worth to be added to all runtimes once @parrt merged my patch (already in Java + C++ with this PR).

parrt:
Hi @mike-lischke , If java doesn't need the cmd-line arg, why do others?

mike-lischke:
Well, in Java you always can import the tool jar as well to get access to Grammar and other classes as described in the interpreter.md document. If you don't want that and only rely on the runtime jar then it would also be useful for the Java target. Other targets always need the generation (and hence the new parameter).

parrt:
Ah. Maybe time to revisit how antlr-java does interp mode as you say.

mike-lischke:
Yes, unfortunately the `Grammar` and `LexerGrammar` classes are not in the runtime (they depend on too many other classes, including antlr3 stuff). So instead of trying to port them to other targets we can simply setup the interpreters (which **are** in the runtime) with the ATN, vocabulary etc. generated by ANTLR via that new option.

parrt:
Would it make sense to generate a standard .atn data file (like we do with .tokens) that *any* target could read?

mike-lischke:
Well, the generated file is plain text and can serve as such a standard file. I contains however not only atn data, but also vocabulary/symbol info.

parrt:
One of my general principles is to avoid introducing commandline options.  I'm not sure that generating parsers versus that text file could even be measured in tens of milliseconds. There's no advantage to turning off the code generation. I can see generating target code with the data properly set up in arrays or whatever instead of a text file format...hmm... this will require some thinking.

mike-lischke:
Tbh I cannot really follow why you are so hesitant when it comes to command line options. It really does not matter if everything is documented well (btw, I probably should update the docs for that new parameter). If parameters matter so much for you, why do you have `-vistor`, `-no-visitor`, `-listener` and `-no-listener`? These serve the same purpose like the code generation skip for interpreter data. The patch I propose is supposed to have the least impact possible and is sparse with extra data, i.e. it does not generate anything that it is **not** asked for. All I need is the data for setting up interpreters - that is the primary goal.

mike-lischke:
@parrt Will you merge this patch? If so I can write up some description in the docs. Otherwise it would be wasted time.

parrt:
i'm swamped this week...still thinking...About the options: I'm trying to prevent further sins that I committed earlier ;)  I think just always generating the file would be useful. Every new option doubles the state space of the program for testing purposes. I also think that the visitor/ listener options are useless but I can't remove them now.

mike-lischke:
I can't see any sins here from your side ;-) For me the command line options make sense as this is the best way to configure the jar, but if you insist I can change the patch to always create the interpreter data.

mike-lischke:
@parrt Done as you wish. Doc updated (added a C++ example).

parrt:
Now only question is what is in that file? What's the format etc...this will require a lot of thought I think. can you describe what it is now in this PR?

mike-lischke:
Uhm, it's all in the patch actually. The .tokens file does not carry all necessary information. Keep also in mind that things should be simple, especially simple to use.

runtime/Cpp/runtime/src/UnbufferedCharStream.cpp
```diff
@@ -101,7 +101,7 @@ size_t UnbufferedCharStream::LA(ssize_t i) {
     return EOF;
   }
 
-  if (_data[(size_t)index] == (uint32_t)EOF) {
+  if (_data[(size_t)index] == 0xFFFF) {
```
parrt: Are you sure this is valid? anytime you touch anything dealing with characters we need to be very careful as Ben spent a lot of time on this. why are you converting to a 16-bit -1?

runtime/Java/src/org/antlr/v4/runtime/misc/InterpreterDataReader.java
```diff
@@ -0,0 +1,141 @@
...
+import org.antlr.v4.runtime.dfa.DFA;
+
+import java.util.Collection;
+import java.util.List;
+import java.util.ArrayList;
+import java.io.BufferedReader;
+import java.io.FileReader;
+
+// A class to read plain text interpreter data produced by ANTLR.
+public class InterpreterDataReader {
```
parrt: is not clear who uses this without using an IDE. does the ParserInterpreter know how to use this somehow?

parrt:
So looks like *&lt;a single line with comma separated int values> enclosed in a pair of squared brackets.* Why brackets? Why anything at all?

It looks to me that the only difference is the mode and channel names from .tokens. Oh and rule names if a parser.

Hmm....I'd hate to generate the same data that is already in a different file but there is a distinct advantage to having a single file to pass around.

I made a few comments in line.

Let me say that I am mainly in maintenance mode after more than 25 years and you should not get annoyed when I don't immediately just include something. I have to carefully think through every new feature; and, basically, I don't want to ;) .  My brain is stuck in a completely different problem at the moment.

runtime/Cpp/runtime/src/UnbufferedCharStream.cpp
mike-lischke: Oh damn, you are right. I somehow thought that was ATN related (and hence a 16bit value). But that's from the unbuffered char stream. Need to review that.

runtime/Java/src/org/antlr/v4/runtime/misc/InterpreterDataReader.java
mike-lischke: Sorry, I don't get your question. This class is a utility class to read the interpreter data file. You can do whatever you like with the returned values (e.g. I will not only use it for interpreters but also for ATN diagrams).

mike-lischke:
> So looks like <a single line with comma separated int values> enclosed in a pair of squared brackets. Why brackets? Why anything at all?

This is just a dump of an int list (the brackets are added by that toString() method). It's not a big deal to ignore them on load, but I wanted to make it simple.
> It looks to me that the only difference is the mode and channel names from .tokens. Oh and rule names if a parser.
> 
> 
:-D so in sum these files are totally different and only have one thing in common (the token list).
> Hmm....I'd hate to generate the same data that is already in a different file but there is a distinct advantage to having a single file to pass around.
> 
We could make the .tokens file handling to use the new interpreter data file instead, but that would require more work and testing. My patch has the least impact possible on  the existing code. It's just printing out data which is available anyway.
> I made a few comments in line.
> 
> Let me say that I am mainly in maintenance mode after more than 25 years and you should not get annoyed when I don't immediately just include something. I have to carefully think through every new feature; and, basically, I don't want to ;) . My brain is stuck in a completely different problem at the moment.
> 
Fully understand, you cannot only spend your time with ANTLR. Maybe you shouldn't take such simple patches like this one so serious? It would save you some time...

parrt:
Hi @mike-lischke  I think I figured out what was bothering me about this.  The advantage of an interpreter is primarily that you can go from the text of a grammar to parsing without a build step. In this case, converting the parser to an ATN requires ANTLR itself, which is in Java. Meaning, that no other target will be able to take advantage of this unless the entire tool is converted to that target. Or, they launch Java as a separate process to convert the grammar file.

The only minor advantage I can see is that a C++/C program's code segment would be smaller because it did not have a recursive descent parser in there. In the Lexer, it is almost entirely the serialized ATN so no advantage there. In the end, I'm not sure adding a major new feature is worth one target getting a smaller code segment. Are there other advantages?

mike-lischke:
The interpreter data export is target independent. It has nothing to do specifically with the C++ target. However, I added a Java interpreter data loader for other target writers, as a template. And I added the C++ port as well, as the first target implementation of that class (since it is part of the runtime then).

Of course, to generate the interpreter data you have to run ANTLR in a background process (or manually if you only want that once). Tbh, I don't understand why this simple patch needs so much discussion. All it does is to write out ATN + symbol data in a separate file so that other targets can load it (using a helper class I also added). Nothing more. It doesn't affect the existing functionality of ANTLR nor that of any target.

parrt:
My experience in software development is that adding features for little gain causes nothing but technical debt for no reason.   Adding software to a tool used by thousands and thousands of people around the world should have huge inertia for changes because of the widespread effects.  What problem are you trying to solve?  This is the first time in recent memory anyone has asked for this, except for Sam and me (for our plugins).

millergarym:
What about a separate class with a main, say ATNGen.java? That way it's not tied to Tool.java which is just there to generate code. ATNGen and Tool would probably deviate over time. It is easy to image ATNGen having a server flag and it return interpreter data over a socket.

@mike-lischke separately have you thought of encoding the data using something like flatbuffers or capnproto? The advantages are performance and encoders & decoders can be generated for the different target languages.

I was thinking a separate class for generate (and printing) boiler plate code for listeners and visitors might be a nice idea.

mike-lischke:
> My experience in software development is that adding features for little gain causes nothing but technical debt for no reason. Adding software to a tool used by thousands and thousands of people around the world should have huge inertia for changes because of the widespread effects.
> 
Well, it's not like there wouldn't be already classes which are almost never used and only have very special purposes like:

LexerInterpreter, ListTokenSource, ParserInterpreter, RuleContextWithAltNum, RuntimeMetaData, ContextSensitivityData, DecisionEventInfo, DecisionInfo, ParseInfo (maybe more).

How many of the thousands of thousands of users actually use any of these classes? I bet not that much and still they exist and are translated to all ANTLR targets. Also the tree pattern and xpath stuff is something I dare to say rather exotic. Still there might be use cases where these classes are useful.
> What problem are you trying to solve? This is the first time in recent memory anyone has asked for this, except for Sam and me (for our plugins).
> 
Yes, in my case it's also about a plugin (this time for vscode). But independent of that, how do you imagine the use of LexerInterpreter and ParserInterpreter in non-Java targets? Without getting the ATN + symbol data from the main ANTLR tool I see no chance to use them at all without that. So they are just dead code currently (and nobody discusses for days how they affect thousands and thousands of users).

I'm trying to open a new door to show what can be done with ANTLR. I'm trying to extend its usage horizon and provide new tools for ANTLR users. So much for my motivation.

> What about a separate class with a main, say ATNGen.java? That way it's not tied to Tool.java which is just there to generate code. ATNGen and Tool would probably deviate over time. It is easy to image ATNGen having a server flag and it return interpreter data over a socket.
> 
@millergarym You mean a separate ja? How would you transfer the in-memory ATN data from the main jar to the new jar then? Alone the transfer mechanism is more complicated than my simple export function.
> @mike-lischke <https://github.com/mike-lischke> separately have you thought of encoding the data using something like flatbuffers or capnproto? The advantages are performance and encoders & decoders can be generated for the different target languages.
> 
There is no performance problem I have to solve here. Key point is simplicity and ease of use. I need no additional software/package just to write out a bit string data. It has a simple, yet human readable format and we are not talking about big sizes. The loaders need to be written once and this can be done within an hour or 2. It would be just a big overkill to use a tool to do that. I don't even expect that the data format will ever change (and if it would it's no problem either, the interpreter data is just intermediate data).

millergarym:
> You mean a separate ja? How would you transfer the in-memory ATN data from the main jar to the new jar then? Alone the transfer mechanism is more complicated than my simple export function.

 I don't really understand Terence's objections and was just suggesting an alternative to modifying the Tool class. 

No, just a separate class. Just wrap the Tool class or copy and gut it. Instead of having the interpreter option and running `java -jar antlr.jar -interpreter ...` you would need to run `java -cp antlr.jar ATNGen`. The only really difference is the amount of coupling between generation and export.

mike-lischke:
Ah, now I got your idea. Yes, this would be a good option as well. It's almost like a new command line option, but it decouples the interpreter data generation from the standard processing.

tool/src/org/antlr/v4/Tool.java
```diff
@@ -117,24 +119,24 @@ public Option(String fieldName, String name, OptionArgType argType, String descr
 	public boolean longMessages = false;
 
     public static Option[] optionDefs = {
-        new Option("outputDirectory",	"-o", OptionArgType.STRING, "specify output directory where all output is generated"),
```
parrt: Is this file just whitespace changes? It's hard to see

parrt:
ok, In the final analysis, I think an -X... option it is okay to generate this intermediate file. I can see a use for it for delivery with an application.  I will merge this manually I guess to set the option

tool/src/org/antlr/v4/Tool.java
mike-lischke: Yes, while I was on this I also fixed formatting. This part was pretty much broken.

parrt:
Hi @mike-lischke I'm trying to integrate but I'm getting tool failures that looks like it's in the new generate interpreter stuff. I have merged the current master into your branch locally and I get the same error

```
java.lang.RuntimeException: set is empty

	at org.antlr.v4.runtime.misc.IntervalSet.getMaxElement(IntervalSet.java:421)
	at org.antlr.v4.runtime.atn.ATNSerializer.serialize(ATNSerializer.java:169)
	at org.antlr.v4.Tool.generateInterpreterData(Tool.java:740)
	at org.antlr.v4.Tool.processNonCombinedGrammar(Tool.java:395)
	at org.antlr.v4.Tool.process(Tool.java:364)
	at org.antlr.v4.Tool.processGrammarsOnCommandLine(Tool.java:323)
	at org.antlr.v4.test.runtime.BaseRuntimeTest.antlrOnString(BaseRuntimeTest.java:265)
	at org.antlr.v4.test.runtime.BaseRuntimeTest.antlrOnString(BaseRuntimeTest.java:227)
	at org.antlr.v4.test.tool.BaseJavaToolTest.testErrors(BaseJavaToolTest.java:26)
	at org.antlr.v4.test.tool.TestToolSyntaxErrors.testInvalidCharSetsAndStringLiterals(TestToolSyntaxErrors.java:568)
	at sun.reflect.NativeMethodAccessorImpl.invoke0(Native Method)
	at sun.reflect.NativeMethodAccessorImpl.invoke(NativeMethodAccessorImpl.java:62)
	at sun.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAccessorImpl.java:43)
	at java.lang.reflect.Method.invoke(Method.java:498)
	at org.junit.runners.model.FrameworkMethod$1.runReflectiveCall(FrameworkMethod.java:50)
	at org.junit.internal.runners.model.ReflectiveCallable.run(ReflectiveCallable.java:12)
	at org.junit.runners.model.FrameworkMethod.invokeExplosively(FrameworkMethod.java:47)
	at org.junit.internal.runners.statements.InvokeMethod.evaluate(InvokeMethod.java:17)
	at org.junit.internal.runners.statements.RunBefores.evaluate(RunBefores.java:26)
	at org.junit.runners.ParentRunner.runLeaf(ParentRunner.java:325)
	at org.junit.runners.BlockJUnit4ClassRunner.runChild(BlockJUnit4ClassRunner.java:78)
	at org.junit.runners.BlockJUnit4ClassRunner.runChild(BlockJUnit4ClassRunner.java:57)
	at org.junit.runners.ParentRunner$3.run(ParentRunner.java:290)
	at org.junit.runners.ParentRunner$1.schedule(ParentRunner.java:71)
	at org.junit.runners.ParentRunner.runChildren(ParentRunner.java:288)
	at org.junit.runners.ParentRunner.access$000(ParentRunner.java:58)
	at org.junit.runners.ParentRunner$2.evaluate(ParentRunner.java:268)
	at org.junit.runners.ParentRunner.run(ParentRunner.java:363)
	at org.junit.runner.JUnitCore.run(JUnitCore.java:137)
	at com.intellij.junit4.JUnit4IdeaTestRunner.startRunnerWithArgs(JUnit4IdeaTestRunner.java:68)
	at com.intellij.rt.execution.junit.IdeaTestRunner$Repeater.startRunnerWithArgs(IdeaTestRunner.java:47)
	at com.intellij.rt.execution.junit.JUnitStarter.prepareStreamsAndStart(JUnitStarter.java:242)
	at com.intellij.rt.execution.junit.JUnitStarter.main(JUnitStarter.java:70)
```

mike-lischke:
According to the stacktrace it crashes in this call: `IntegerList serializedATN = ATNSerializer.getSerialized(g.atn);`, so it's not the export function per se which has a problem but `getSerialized(g.atn)`. To me it looks more like a problem with the ATN serialization process. Somehow an empty set is added during this process. Not related to interpreter data export at all.

parrt:
Ok,I figured it was too much of a coincidence given that those tool tests have been running for a long time for me not to notice but I will go take a look at the master and see if I get the same errors.

parrt:
All tool tests passed in the master so either something you have pulled in or your changes have altered the tool I guess. can you take another look?

mike-lischke:
Hmm, I'd like to help here, but tbh I have no idea how to debug the tests. This is the result from the smoke test, right? How's that triggered (what script is used)?

When I load the ANTLR4 folder into Eclipse I get an endless number of errors. Probably a dependency setup problem, but Java + Eclipse ar not my native home and I have no idea how to solve all that in a reasonable amount of time. I wish there were a project I can just load somewhere that allows to build and debug a test.

mike-lischke:
Regarding the problem: obviously something with the ATN is not right at the point when execution breaks. Can't you just look what it is? You could just call `ATNSerializer.getSerialized(g.atn);` instead of `generateInterpreterData(g);` and it will probably crash too. There's nothing in the patch that alters the behavior of the existing code, except for this single function call.

My suspicion is rather that my interpreter code exposed a sleeping bug in the runtime. From what I see `ATNSerializer.getSerialized(g.atn);` has not been used so far.

parrt:
Ok, getting on airplane for 2 day trip but will have laptop. isn't ATNSerializer.getSerialized what we use to gen code though?

mike-lischke:
Yes, right, `getSerialized()` is called indirectly, so the only other idea I have is that the call to it might come too early (i.e. not everything is fully set up in the ATN for serialization). However it's used to generate the dot graphs, so I assumed it would be ok to use the ATN at this point.

parrt:
Merged manually as I needed to tweak.

parrt:
And....we're home, @mike-lischke ! Thanks for all the work on this :)

mike-lischke:
Good to hear! Out of curiousity: what was the reason for the crash actually?

parrt:
upon erroneous input, we shouldn't try to serialize an ATN :)

parrt:
anything else we really need to get in for 4.7.1?

mike-lischke:
Nothing from my side. I'm done with my changes so far.

Issue #753, opened 2014-11-01
Unexpected code generation location behavior 


grosenberg:
On Windows / Antlr 4.4  (might be related to #638)

The following works as expected: grammars exist in the parent directory (parser) & all generated files wind up in the current directory (gen)

cd /d D:\DevFiles\Java\JsonEditor\src\net\certiv\json\parser\gen\
java org.antlr.v4.Tool ../JsonLexer.g4 ../JsonParser.g4

The following fails: 

cd /d D:\DevFiles\Java\JsonEditor\src\net\certiv\json\parser\gen\
java org.antlr.v4.Tool ..\JsonLexer.g4 ..\JsonParser.g4

JsonLexer.java and JsonLexer.tokens are generated in the parent directory (parser) and the compile of the JsonParser.g4 fails saying that it cannot find the JsonLexer.tokens in the current directory (gen).

Note the difference in the slashes in the grammar paths - "../" works completely as expected, but "..\" works sufficiently for the Tool to find the grammars, but generates in the wrong directory.  Adding -o and -lib options in various combinations don't affect the result.

This problem was encountered while rebuilding an early Antlr4 project.  So using "..\" apparently worked as expected "way back then."

WalkerCodeRanger:
I also ran into this issue

mike-lischke:
I believe this is actually the same problem as #638. The different target folder creation wouldn't even be a problem if ANTLR would stop doing that and just use what it was given it as output path.

parrt:
Closed by https://github.com/antlr/antlr4/pull/2065

Issue #638, opened 2014-06-19
Compiling seperate lexer and parser in subdirectory fails


arthurfabre:
If one defines a seperate Parser and Lexer as such:

Lexer.g4:

```
lexer grammar Lexer;

tokens {INDENT, DEDENT}

INT     : [0-9]+ ;
```

Parser.g4:

```
parser grammar Parser;

options { tokenVocab=Lexer; }

main    
        : INT* EOF
        ;
```

With the following directory layout:

```
-project
    -src
        -Lexer.g4
        -Parser.g4
    -build
    -antlr4.2.2-complete.jar
```

Compiling from the `project` directory with the following command:
`java -jar antlr-4.2.2-complete.jar -o build src/*.g4`

Fails with: `error(3):  cannot find tokens file 'build/Lexer.tokens'`

`Lexer.tokens` is correctly generated, but is in `build/src/Lexer.tokens` as expected.

This works fine if the grammar files are in the current directory when compiling them (ie one runs `java -jar ../antlr-4.2.2-complete.jar -o ../build *.g4` or if an extra `-lib build/src/` option is used. 

It seems this shouldn't be required and antlr should know where to find the .tokens file it generates.
## UPDATE:

Here are some additional details about the proposed feature:

For the following, assume this directory layout:

```
-project
    -src
        -foo
            -lexer.g4
            -metarParser.g4
            -tafParser.g4
        -bar
            -notamParser.g4
    -build
```

With the following dependency between the grammars:

```
lexer.g4:
metarParser.g4: lexer.g4
notamParser.g4: lexer.g4
tafParser.g4: notamParser.g4
```

In all the following cases it is assumed that all of the grammar files are compiled in a single invocation of antlr, and that we want the generated files
to be in the build directory (optionally in sub folders).
### 1st case: All grammars in working directory.
- We wish to compile metarParser.g4
- Our working directory is project/src/foo/
- Run command: antlr *.g4 -o ../../build/
  
  Observed behavior:
  - Antlr compiles lexer.g4
  - lexer.g4 has a relative path of ., the generated token file is therefore in build/
  - Antlr adds the ../../build/ directory to it's "lib" search path
  - Antlr compiles meterParser.g4  
  - Antlr searches the "lib" directories for the required lexer.tokens
  - Antlr correctly finds the lexer.tokens file in ../../build/
  
  Desired behavior:
  - Same.
### 2nd case: All grammars are in a given subdirectory.
- We wish to compile metarParser.g4
[...]

PR #1829, opened 2017-04-19
Simplify output path generation

### LINKED ISSUE #1087: Tool path handling on Windows
`antlr-4.5.1-complete.jar` doesn't behave consistently with "/" and "\" on Windows.

Initial folder structure:

```
|   antlr-4.5.1-complete.jar
\---src
    \---main
        \---antlr
                example.g4
```

Case 1. After `java -jar antlr-4.5.1-complete.jar -o src\main\generated src\main\antlr\example.g4`:

```
|   antlr-4.5.1-complete.jar
\---src
    \---main
        +---antlr
        |       example.g4
        \---generated
            \---src
                \---main
                    \---antlr
                            example.tokens
                            exampleBaseListener.java
                            exampleLexer.java
                            exampleLexer.tokens
                            exampleListener.java
                            exampleParser.java
```

Case 2. After `java -jar antlr-4.5.1-complete.jar -o src/main/generated src/main/antlr/example.g4`:

```
|   antlr-4.5.1-complete.jar
\---src
    \---main
        +---antlr
        |       example.g4
        \---generated
                example.tokens
                exampleBaseListener.java
                exampleLexer.java
                exampleLexer.tokens
                exampleListener.java
                exampleParser.java
```

According to #683 the Case 1. behavior is correct, however in cross-platform scripts you are quite likely to use the path in Case 2.

Ideally both would behave exactly the same way.

mike-lischke:
The twisted way to create a javac like output path breaks many use cases, including very simple ones with lexer + parser grammar in a relative subdir (where parser generation doesn't find the lexer tokens). Additionally, not all projects prefer to have their generated files in an output folder which is comprised of the given output path and the subdir given for the grammar files. The check for absolute paths as decision criterion is a weak one anyway - it makes simple usage difficult and doesn't deal with path expansion on all platforms.

If a javac like output path is required it should be easy to construct this upfront and provide it as the output path parameter, instead of implicitely creating it. This patch simplifies things without any WTF moments because generated files end up at unexpected locations. It now can deal with simple cases (no output, no grammar subdir) up to complicated cases (nested and/or absolute output dir, multiple grammar subfolders).

If more than one output folder is wanted run the tool multiple times with different output path settings and add the lib option if you need to reference previously created files.

It might look like a disadvantage that now generation of parser files ends up at a different path when run with relative grammar subfolders, but actually that was broken anyway so I don't expect anyone having used that (but instead either absolute paths or no subfolders at all). 

Fixes #1087, #753,  #638, tunnelvisionlabs/antlr4ts#306, tunnelvisionlabs/antlr4ts#303


mike-lischke:
Is it possible to finally merge this patch? I unfortunately did the changes on the master branch and now I cannot open any pull request without the changes from this patch. But since I merged my master already into several other branches (partially also with a pull request) it's impossible to fix this situation without this patch being merged.

I'm sorry for this hassle - I should have done the patch on a different branch. However, Travis CI seems ok with the changes (Appveyor obviously has build problems), so I think it is safe to merge. As I said before, relative paths don't work anyway currently (which was the reason I actually rewrote that part) and absolute paths are not affected by this patch. Also the sheer list of bugs that get fixed by this patch should be a good reason to merge. I also explained in e.g. bug #638 why other solutions don't work.

mike-lischke:
OK, doesn't seem to be so "impossible" as I thought :-) I reverted the path changes in master and will do a separate pull request for them. Sorry for confusion.

PR #1836, opened 2017-04-24
Go Target - make token source all public

millergarym:
This change effects the Go Target only.
It makes all of the TokenSource signatures public so that an alternative, client code implementation can be provided.


millergarym:
will re-issue PR bundled with others

PR #1841, opened 2017-04-26
Go Target - Visit implementation - two other minor changes runtimeimport tokenstream

millergarym:
@cyberfox see example on the doc/go-target.md
Some differences to note.
- if VisitNext return false, the loop in VisitChildren continues instead of using a break.
- The Default and Aggregate methods aren't needed, neither is the "super"

This is a breaking change to the generated Listeners interface.
I'll added an option to generated this version with the default being current *struct signatures.

This is an alternative implementation to
https://github.com/antlr/antlr4/pull/1807

### DIFF
--- a/runtime/Go/antlr/parser_rule_context.go
+++ b/runtime/Go/antlr/parser_rule_context.go
@@ -181,8 +181,8 @@ func (prc *BaseParserRuleContext) GetRuleContext() RuleContext {
 	return prc
 }
 
-func (prc *BaseParserRuleContext) Accept(visitor ParseTreeVisitor) interface{} {
-	return visitor.VisitChildren(prc)
+func (prc *BaseParserRuleContext) Accept(delegate ParseTreeVisitor) interface{} {
+	return delegate.VisitChildren(prc, delegate)
 }
 
 func (prc *BaseParserRuleContext) SetStart(t Token) {
--- a/runtime/Go/antlr/parser_rule_context.go
+++ b/runtime/Go/antlr/parser_rule_context.go
@@ -181,8 +181,8 @@ func (prc *BaseParserRuleContext) GetRuleContext() RuleContext {
 	return prc
 }
 
-func (prc *BaseParserRuleContext) Accept(delegate ParseTreeVisitor) interface{} {
-	return delegate.VisitChildren(prc, delegate)
+func (prc *BaseParserRuleContext) Accept(delegate ParseTreeVisitor) {
+	delegate.VisitChildren(prc, delegate)
 }
 
 func (prc *BaseParserRuleContext) SetStart(t Token) {
--- a/runtime/Go/antlr/tree.go
+++ b/runtime/Go/antlr/tree.go
@@ -28,7 +28,7 @@ type SyntaxTree interface {
 type ParseTree interface {
 	SyntaxTree
 
-	Accept(Visitor ParseTreeVisitor) interface{}
+	Accept(visitor ParseTreeVisitor) interface{}
 	GetText() string
 
 	ToStringTree([]string, Recognizer) string
@@ -54,41 +54,44 @@ type ErrorNode interface {
 }
 
 type ParseTreeVisitor interface {
-	Visit(tree ParseTree) interface{}
-	VisitChildren(node RuleNode) interface{}
+	Init() interface{}
+	VisitNext(node Tree, resultSoFar interface{}) bool
+	Aggregate(resultSoFar, childResult interface{}) interface{}
 	VisitTerminal(node TerminalNode) interface{}
 	VisitErrorNode(node ErrorNode) interface{}
+	VisitChildren(node RuleNode, delegate ParseTreeVisitor) interface{}
 }
 
 type BaseParseTreeVisitor struct{}
 
-var _ ParseTreeVisitor = &BaseParseTreeVisitor{}
-
-func (v *BaseParseTreeVisitor) Visit(tree ParseTree) interface{}            { return nil }
-func (v *BaseParseTreeVisitor) VisitChildren(node RuleNode) interface{}     { return nil }
-func (v *BaseParseTreeVisitor) VisitTerminal(node TerminalNode) interface{} { return nil }
-func (v *BaseParseTreeVisitor) VisitErrorNode(node ErrorNode) interface{}   { return nil }
-
-// TODO
-//func (this ParseTreeVisitor) Visit(ctx) {
-//	if (Utils.isArray(ctx)) {
-//		self := this
-//		return ctx.map(function(child) { return VisitAtom(self, child)})
-//	} else {
-//		return VisitAtom(this, ctx)
-//	}
-//}
-//
-//func VisitAtom(Visitor, ctx) {
-//	if (ctx.parser == nil) { //is terminal
-//		return
-//	}
-//
-//	name := ctx.parser.ruleNames[ctx.ruleIndex]
-//	funcName := "Visit" + Utils.titleCase(name)
-//
-//	return Visitor[funcName](ctx)
-//}
+func (*BaseParseTreeVisitor) Init() interface{}                                 { return nil }
+func (*BaseParseTreeVisitor) VisitNext(node Tree, resultSoFar interface{}) bool { return true }
+func (*BaseParseTreeVisitor) Aggregate(resultSoFar, childResult interface{}) interface{} {
+	return childResult
+}
+func (*BaseParseTreeVisitor) VisitTerminal(node TerminalNode) interface{} { return nil }
+func (*BaseParseTreeVisitor) VisitErrorNode(node ErrorNode) interface{}   { return nil }
+func (*BaseParseTreeVisitor) VisitChildren(node RuleNode, delegate ParseTreeVisitor) interface{} {
+	result := delegate.Init()
+	for _, child := range node.GetChildren() {
+		if !delegate.VisitNext(child, result) {
+			continue
+		}
+		switch child := child.(type) {
+		case TerminalNode:
+			childResult := delegate.VisitTerminal(child)
+			result = delegate.Aggregate(result, childResult)
+		case ErrorNode:
+			delegate.VisitErrorNode(child)
+		case RuleNode:
+			childResult := child.Accept(delegate)
+			result = delegate.Aggregate(result, childResult)
+		default:
+			// can this happen??
+		}
+	}
+	return result
+}
 
 type ParseTreeListener interface {
 	VisitTerminal(node TerminalNode)
--- a/runtime/Go/antlr/tree.go
+++ b/runtime/Go/antlr/tree.go
@@ -28,7 +28,7 @@ type SyntaxTree interface {
 type ParseTree interface {
 	SyntaxTree
 
-	Accept(visitor ParseTreeVisitor) interface{}
+	Accept(visitor ParseTreeVisitor)
 	GetText() string
 
 	ToStringTree([]string, Recognizer) string
@@ -54,43 +54,33 @@ type ErrorNode interface {
 }
 
 type ParseTreeVisitor interface {
-	Init() interface{}
-	VisitNext(node Tree, resultSoFar interface{}) bool
-	Aggregate(resultSoFar, childResult interface{}) interface{}
-	VisitTerminal(node TerminalNode) interface{}
-	VisitErrorNode(node ErrorNode) interface{}
-	VisitChildren(node RuleNode, delegate ParseTreeVisitor) interface{}
+	VisitNext(next Tree) bool
+	VisitTerminal(node TerminalNode)
+	VisitErrorNode(node ErrorNode)
+	VisitChildren(node RuleNode, delegate ParseTreeVisitor)
 }
 
 type BaseParseTreeVisitor struct{}
 
-func (*BaseParseTreeVisitor) Init() interface{}                                 { return nil }
-func (*BaseParseTreeVisitor) VisitNext(node Tree, resultSoFar interface{}) bool { return true }
-func (*BaseParseTreeVisitor) Aggregate(resultSoFar, childResult interface{}) interface{} {
-	return childResult
-}
-func (*BaseParseTreeVisitor) VisitTerminal(node TerminalNode) interface{} { return nil }
-func (*BaseParseTreeVisitor) VisitErrorNode(node ErrorNode) interface{}   { return nil }
-func (*BaseParseTreeVisitor) VisitChildren(node RuleNode, delegate ParseTreeVisitor) interface{} {
-	result := delegate.Init()
+func (*BaseParseTreeVisitor) VisitNext(node Tree) bool        { return true }
+func (*BaseParseTreeVisitor) VisitTerminal(node TerminalNode) {}
+func (*BaseParseTreeVisitor) VisitErrorNode(node ErrorNode)   {}
+func (*BaseParseTreeVisitor) VisitChildren(node RuleNode, delegate ParseTreeVisitor) {
 	for _, child := range node.GetChildren() {
-		if !delegate.VisitNext(child, result) {
+		if !delegate.VisitNext(child) {
 			continue
 		}
 		switch child := child.(type) {
 		case TerminalNode:
-			childResult := delegate.VisitTerminal(child)
-			result = delegate.Aggregate(result, childResult)
+			delegate.VisitTerminal(child)
 		case ErrorNode:
 			delegate.VisitErrorNode(child)
 		case RuleNode:
-			childResult := child.Accept(delegate)
-			result = delegate.Aggregate(result, childResult)
+			child.Accept(delegate)
 		default:
 			// can this happen??
 		}
 	}
-	return result
 }
 
 type ParseTreeListener interface {
@@ -166,8 +156,8 @@ func (t *TerminalNodeImpl) GetChildCount() int {
 	return 0
 }
 
-func (t *TerminalNodeImpl) Accept(v ParseTreeVisitor) interface{} {
-	return v.VisitTerminal(t)
+func (t *TerminalNodeImpl) Accept(v ParseTreeVisitor) {
+	v.VisitTerminal(t)
 }
 
 func (t *TerminalNodeImpl) GetText() string {
@@ -206,8 +196,8 @@ func NewErrorNodeImpl(token Token) *ErrorNodeImpl {
 
 func (e *ErrorNodeImpl) errorNode() {}
 
-func (e *ErrorNodeImpl) Accept(v ParseTreeVisitor) interface{} {
-	return v.VisitErrorNode(e)
+func (e *ErrorNodeImpl) Accept(v ParseTreeVisitor) {
+	v.VisitErrorNode(e)
 }
 
 type ParseTreeWalker struct {
--- a/tool/resources/org/antlr/v4/tool/templates/codegen/Go/Go.stg
+++ b/tool/resources/org/antlr/v4/tool/templates/codegen/Go/Go.stg
@@ -12,9 +12,9 @@ package parser // <file.grammarName>
 <endif>
 
 import (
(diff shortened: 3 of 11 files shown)


tool/src/org/antlr/v4/codegen/model/ListenerFile.java
```diff
@@ -26,10 +26,18 @@
 	public String exportMacro; // from -DexportMacro cmd-line
 	public String grammarName;
 	public String parserName;
+
```
cyberfox: This (and below) is unrelated changes, right?

runtime/Go/antlr/tree.go
```diff
@@ -101,10 +94,10 @@ type BaseParseTreeListener struct{}
 
 var _ ParseTreeListener = &BaseParseTreeListener{}
 
-func (l *BaseParseTreeListener) VisitTerminal(node TerminalNode)      {}
```
cyberfox: I understand the impetus to change the listener implementation, but I think that a more focused pull request (just adding support for Visitor) would be better here.  Especially if the Listener change is going to be a breaking one.

The Visitor never worked, so implementing it should be straightforward, but breaking the way Listener worked (and the unrelated changes) seem like they'd be better as a separate pull request.

millergarym:
> I feel like this pull request tries to do too many different things, and at the same time doesn't do enough for the code that's trying to implement a Visitor. Losing return values from visit nodes and not being able to implement a subset of the Visitor are pretty serious losses.

See my response in https://github.com/antlr/antlr4/issues/1843#issuecomment-297849852

runtime/Go/antlr/tree.go
```diff
@@ -28,7 +28,7 @@ type SyntaxTree interface {
 type ParseTree interface {
 	SyntaxTree
 
-	Accept(Visitor ParseTreeVisitor) interface{}
+	Accept(visitor ParseTreeVisitor)
```
millergarym: Accept should only get called by the generate code, not by visitor implementations.

tool/resources/org/antlr/v4/tool/templates/codegen/Go/Go.stg
```diff
@@ -105,21 +135,26 @@ package <file.genPackage> // <file.grammarName>
...
 <if(header)>
 
 <header>
 <endif>
 
 // A complete Visitor for a parse tree produced by <file.parserName>.
 type <file.grammarName>Visitor interface {
-	antlr.ParseTreeVisitor
+    antlr.ParseTreeVisitor
+    <file.visitorNames:{lname | <lname; format="cap">ContextVisitor}; separator="\n">
```
millergarym: The complete visitor interface is actually not needed at all. I only left it in as a convenience.

tool/resources/org/antlr/v4/tool/templates/codegen/Go/Go.stg
```diff
@@ -746,8 +798,9 @@ InvokeRule(r, argExprsChunks) ::= <<
 	var _x = p.<r.name; format="cap">(<argExprsChunks>)
 	<endif>
 
+    <r.labels:{l | <labelRefStuff(l,"_x")> }; separator="\n">
```
millergarym: it is only styling. go fmt was fix it. it only matter if there are style checkers as part of commit code to a repo and you commit generated code.

tool/resources/org/antlr/v4/tool/templates/codegen/Go/Go.stg
```diff
@@ -131,16 +248,39 @@ package <file.genPackage> // <file.grammarName>
 package parser // <file.grammarName>
 <endif>
 
-import "github.com/antlr/antlr4/runtime/Go/antlr"
+// See base listener file for example implementations
```
millergarym: I am not generation a base visitor. The code generated is a template visitor that can be copied uncommented and used as a stating point.

tool/resources/org/antlr/v4/tool/templates/codegen/Go/Go.stg
```diff
@@ -1196,102 +1295,116 @@ func New<struct.name>(parser antlr.Parser, ctx antlr.ParserRuleContext) *<struct
...
+<endif>
 >>
 
 VisitorDispatchMethod(method) ::= <<
-func (s *<struct.name>) Accept(visitor antlr.ParseTreeVisitor) interface{} {
-	switch t := visitor.(type) {
-	case <parser.grammarName>Visitor:
-		return t.Visit<struct.derivedFromName; format="cap">(s)
-
+func (s *<struct.name>) Accept(delegate antlr.ParseTreeVisitor) {
```
millergarym: as noted before Accept should only get called by the generated code. Please provide an example of using the returned interface{}.

runtime/Go/antlr/tree.go
millergarym: I need to look into accept a bit more. It gets called to kick off the visitation, but shouldn't get called inside the visitor (afaik). Doing so a receipt for confusion.

cyberfox: My approach came out of creating an expression evaluator which had no mutable state of its own, which allows `.Accept(v)` to work fine.  If your visitor type has to carry changing state with it, then yes, you can't really call back into it as the state will get confused.  But you shouldn't HAVE to do that, and I didn't think your implementation forced that.

I also do this in a standard Java ANTLR-generated vistor, which I know shouldn't mean that's relevant for Go, but I mean it more as a point that it might be expected by ANTLR users.

If you imagine something which implements a simple expression parser:
```antlr
expr
    : '(' expr ')'                 #ParenExpr
    | a=expr op=(ADD|SUB) b=expr   #AddSubExpr
    | literal                      #LiteralExpr
    ;

literal
    : lit=INT                      #IntLiteral
    ;

SUB : '-' ;
ADD : '+' ;
INT :   [0-9]+ ;
```

An obvious implementation of `VisitIntLiteral(...)` is
```go
	result, err := strconv.ParseInt(ctx.GetText(), 10, 64)
	if err == nil {
		return result
	}
	return nil
```

That lets your `AddSubExpr` visitor do something like what I said at the top, and then (after a type check) add the numbers and return them.  Now you have a recursive visitor that evaluates expressions.  Essentially an interpreter, using the Visitor pattern to execute it.

(The one I'm working with in my current project is more complicated, but that's all extracted from pieces of it that are emblematic of the whole.  The interpreter and surrounding code is written in Java and Go, but both use the same grammar, and the same aspects of ANTLR.)

tool/resources/org/antlr/v4/tool/templates/codegen/Go/Go.stg
cyberfox: That's my bad, then.  I couldn't see how it was implementing a partial interface.  Sorry about that.  I'll try implementing a simple labeled expression evaluator in it, to see how it goes.

tool/resources/org/antlr/v4/tool/templates/codegen/Go/Go.stg
cyberfox: Does `go fmt` run on code that's auto-generated by `go generate`?

Anyway, right now I'm committing the generated code because I didn't know about `go generate`. 👍

tool/resources/org/antlr/v4/tool/templates/codegen/Go/Go.stg
cyberfox: I did above; I'll see if I can put together a more complete example to share.

tool/resources/org/antlr/v4/tool/templates/codegen/Go/Go.stg
millergarym: No, but you can always do
```
//go:generate go fmt 
```

runtime/Go/antlr/tree.go
cyberfox: Sure, I've put up https://github.com/cyberfox/expressions for now, and I'd be happy to mix it into a catalog of examples.  I did integrate returns and actions for a bit, while using the visitor at the same time.  (My QuotedLiteral sliced the quotes off in the grammar, but I realized it was easier and less prone to weird compile errors to just write it as a visitor.)

For what it's worth (as I bet you guessed), I first built a grammar for Java, because that's what I was targeting for work.  I started stripping out actions as I improved my Java visitors, and eventually I was left with only a few and a _bunch_ of returns, when...

I realized there was a Go target, and I've had a goal to improve my Go skills, and so I started making it work for both languages.  Doing that from one grammar forced me to strip out all the actions and returns.  It also forced me to make the Visitor ANTLR code work. :)  So 'prototype a grammar with actions, args, and returns' and then moving to visitors afterwards was pretty much exactly the process I used.

millergarym:
I was finding implementing small visitors a bit cumbersome as Go doesn't allow nested method definitions, but does allow nested anonymous function definitions . So I added a VisitFunc method that takes a struct of functions.

eg

```
grammar Scratch;

options {
  language = Go;
  newCallback = true;
  runtimeImport = 'github.com/wxio/antlr4-go';  
}

start 
  : a EOF # Start1
;

a  
  : (t='1'|t='0') b? a? # A1
;

b
  : t='x' # B1
  | t='y' # B1
;
``` 

```
func TestVisit(t *testing.T) {
	input := antlr.NewInputStream("101x0x01y11111")
	lexer := parser.NewScratchLexer(input)
	stream := antlr.NewCommonTokenStream(lexer, 0)
	p := parser.NewScratchParser(stream)
	p.AddErrorListener(antlr.NewDiagnosticErrorListener(true))

	tree := p.Start()

	indent := ""
	tree.VisitFunc(&parser.ScratchHandlers{
		EnterEveryRule: func(ctx antlr.RuleNode) { indent += "\t" },
		ExitEveryRule:  func(ctx antlr.RuleNode) { indent = indent[1:] },
		Start1: func(this *parser.ScratchHandlers, ctx parser.IStart1Context) {
			fmt.Printf("%s%T\n", indent, ctx)
			ctx.VisitChildrenFunc(this)
		},
		A1: func(this *parser.ScratchHandlers, ctx parser.IA1Context) {
			fmt.Printf("%s%T\n", indent, ctx)
			ctx.VisitChildrenFunc(this)
		},
//		B1: func(this *parser.ScratchHandlers, ctx parser.IB1Context) {
//			fmt.Printf("%s%T\n", indent, ctx)
//			ctx.VisitChildrenFunc(this)
//		},
	})

}
```

cyberfox:
I've been using my fork for a while now; I'm happy to use yours if it becomes the standard way, although I haven't tried changing my stuff over wholesale to it to see if it has holes.  How's the Visit implementation going?  Have you used it in anger to implement an interpreter?

Is there even anybody who can reasonably code review it?  It seems like a kinda specialized combination of knowledge, unfortunately.

davesisson:
I'm willing to review (and even implement a parser to do that).  Do we have agreement about which implementation we're going to go with?  (I'm good with either -- I'm eager to have anything usable in the main branch at this point.)

millergarym:
Hey @cyberfox and @davesisson,

I used the https://github.com/wxio/antlr4/commits/visit-runtimeimport-tokenstream implementation pretty extensively to implement a basic spreadsheet engine. 

There are still a few usability issues. The biggest of which is that the visitor & listener should generated code should use pointers to structs, as this is what most people expect. Generating code that uses interfaces should be an option.

I haven't used https://github.com/cyberfox/antlr4 since April and there are a number of feature is this fork that would stop me switching.

If it help evaluate things, I'll push the latest version to
https://github.com/wxio/antlr4-go . Please let me know. 

I should also post how to go from a full antlr4 repo to a Go specific one. Basically it's just a git prune. Let me know if you would find this useful.

Cheers
Gary

davesisson:
I'll try https://github.com/wxio/antlr4-go out then.  It's pretty much already in the form I need -- I end up transplanting the runtime into my build system.

millergarym:
@davesisson please use https://github.com/wxio/antlr4 the pruned repo is out of date.

cyberfox:
Sounds good to me; is https://github.com/millergarym/expressions up-to-date with the current version of the Go visitor changes?  I find it a really helpful example to wrap my head around how things are working, with most of the features I use.

millergarym:
@cyberfox no, but I'll try make some time to bring it up to date.

cyberfox:
@millergarym I've updated cyberfox/expressions to match the new code.  I also converted my own project, which was...complicated, but ultimately doable.

I don't necessarily like the long method declarations, but I get it.  It'd be good to get this merged...

cch123:
Any progress on this?

savvopoulos:
This change didn't work for me - the generated .go files didn't compile:
`parser/cypher_parser.go:1274:40: undefined: antlr.ParserTreeVisitorHandlers`

Grammar attached - as .txt because of github.
[Cypher.txt](https://github.com/antlr/antlr4/files/1308694/Cypher.txt)

savvopoulos:
Reviewed 1 of 3 files at r4, 16 of 19 files at r5, 3 of 3 files at r6.
Review status: all files reviewed at latest revision, 4 unresolved discussions, some commit checks broke.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/antlr/antlr4/1841)*

davesisson:
We tried https://github.com/wxio/antlr4/commits/visit-runtimeimport-tokenstream here as well and it meets our needs.  Could we convert that into a pull request and go with that approach?

cyberfox:
@savvopoulos Not sure if you already dealt with this, but the Cypher.g4 grammar does compile if you check out the `visit-runtimeimport-tokenstream` branch in `$GOPATH/src/github.com/wxio/antlr4`

This is how I tried it (presuming GOPATH is ~/go)
```
cd ~/go/src/github.com
mkdir -p cyberfox/cypher
curl -o cyberfox/cypher/Cypher.g4 -L https://github.com/antlr/antlr4/files/1308694/Cypher.txt
mkdir -p wxio
git clone git@github.com:wxio/antlr4 wxio/antlr4 --branch visit-runtimeimport-tokenstream
cd wxio/antlr4
mvn package -Dmaven.test.skip=true #this will give an error; it's okay.
pushd ../../cyberfox/cypher
java -jar ~/go/src/github.com/wxio/antlr4/tool/target/antlr4-4.7.1-SNAPSHOT-complete.jar -Dlanguage=Go -DruntimeImport=github.com/wxio/antlr4/runtime/Go/antlr -visitor -no-listener Cypher.g4
go build
```
The idea, as I understand it, is that you should check in the generated code, so that not everybody who checks it out needs to rebuild the antlr jar files, but I included the process to do that above, since it's the steps I went through to test it.

@davesisson This pull request *is* the pull request from `wxio:visit-runtimeimport-tokenstream`, so merging this would merge the right code, I believe.

It'd be really good if we could get this pull request merged; because `go get` only uses master, and doesn't have a way to specify branches, it's a bit of a pain to get it to build, as you can see from the script above.

davesisson:
@parrt I think we want to go with this implementation.  Who do we need to agree that this is the right way to go?

parrt:
@davesisson I think you were the primary Go guy ,right? can't remember. if so, I will merge

davesisson:
Ah, the main Go guy is @pboyer.

I was doing C++ but I've been helping out in Go lately.  :)

pboyer:
@davesisson @parrt Thanks for pinging me. I'm on it.

runtime/Go/antlr/parser_rule_context.go
```diff
@@ -181,8 +193,8 @@ func (prc *BaseParserRuleContext) GetRuleContext() RuleContext {
 	return prc
 }
 
-func (prc *BaseParserRuleContext) Accept(visitor ParseTreeVisitor) interface{} {
-	return visitor.VisitChildren(prc)
+func (prc *BaseParserRuleContext) Visit(delegate ParseTreeVisitor, args ...interface{}) (result interface{}) {
+	return delegate.VisitChildren(prc, delegate)
 }
```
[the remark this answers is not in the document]
cyberfox: So...IIRC, this is the complicated part.  Because Go doesn't actually have what you'd typically call 'inheritance', the `VisitChildren` method implemented in the `ParseTreeVisitor` will actually NOT call the right method if you just use the `*BaseParseTreeVisitor` pointer.  So IIRC he passes in `delegate` so that it can call the right methods on it.  I know that was necessary for my initial version, but I did it very differently, and without a lot of the Go experience that you and he bring to the table.  Using type assertions and type `switch` appears to have gone a long way.

I did a more 1:1 implementation of `VisitChildren`, using the Java code as the guideline, but I had to put it into the generated code in order for it to work and did a lot of ugly `if` checks.  He found a way to put it into the base code, and have it still call the right methods.  (Although ignoring default value, and a few other niceties.)

A lot of this is from memory, refreshed by looking at the various pieces; it's been a while, and I admit, I don't really understand the reason for `args`, for example. :-/

cyberfox:
@pboyer Implementing VisitChildren in a way suitable to be used generically was extremely hard for me, at least, mainly because of Go's lack of inheritance.  There needed to be a way to pass in the thing you wanted the child calls (`DefaultResult`, `ShouldVisitNextChild`, `AggregateResult`) to actually go to, for instance.

There's a repo, https://github.com/cyberfox/expressions which shows how (the current version, and earlier my own version) it's all used in a slightly less trivial example, if that'd help.

My initial [VisitChildren](https://github.com/antlr/antlr4/pull/1807/files#diff-4277976fc8a398928e64f4eacf4ec59fR166) stab at implementation was in the generated code, not the base code, and was very small (#1807; ignore the `import` changes) but depended on some sketchy choices on my part, (in retrospect) truly awful naming, and I was (and am) very new to Go code.

millergarym:
Sorry that I haven't been more responsive, this needs attention and I don't have the time at the moment. If anyone has time it would be appreciated.

The following is what I would like to do when I can make some time.

As noted this there are multiple things going on is this PR, some should be removed. The biggest one is the generation of used of interface instead of pointers to structures. I still think this might be a good idea for some circumstances, but imo it does belong in this PR.

@pboyer The functionality of the Java visitor was not rich enough of what I was trying to do, hence the deviation. It appears that the visitor hasn't had very attention payed to it. To back up this comment I would like to provide a couple of examples.

I find it very useful to pass and maintain state in the call stack. Yes, all the state can be maintained in the visitor implementation, but this get very cumbersome. 

Yes, Accept is the more common name the Go code should use it. I changed it while trying to understand the code, it took me a while to get my head around things and I never changed it back.

I second @cyberfox comment about Go and inheritance. There is not default receiver, to achieve implement inheritance (Java, C++, C#, Smalltalk, Javascript etc.) an extra argument needs to be passed.

 @savvopoulos and @cyberfox, I've merged the `antlr/antlr4` master head with `wxio:visit-runtimeimport-tokenstream` and pushed in to `wxio:master` (not `vistit-...`). `wxio:visit-runtimeimport-tokenstream` is still four months behind antlr4. Hope this help with your go gets. 

I would also like to bring `https://github.com/wxio/antlr4-go` up-to-date and automate it. This is currently behind `wxio:visit-runtimeimport-tokenstream`. It is a more go get able version of antlr4 for Go. I.e. just the Go runtime and build jar of Antlr. Anyone able to automate this, it has eluded my so to-date. This is an out of date version of the first set 
```
export PATH=~/dev/apache-maven-3.3.9/bin:$PATH
export JAVA_HOME=`/usr/libexec/java_home`
mvn compile
mvn install -DskipTests
mkdir runtime/Go/antlr/lib/
cp tool/target/antlr4-4.6-SNAPSHOT.jar runtime/Go/antlr/lib/
git filter-branch --prune-empty --subdirectory-filter runtime/Go/antlr master
``` 
After this one or two intermediate workspaces and repos are needs to merge the changes into the antlr4-go repo. Any git masters who could help automate this?

@pboyer I haven't addressed all the issues you have raised, but I hope this helps.

@cyberfox thanks for your input. Where do you think we should do from here?

nathanhack:
Any updates on this?

danaugrs:
Any updates on this?

PR #1845, opened 2017-04-28
[dotnet] speed up tests; rename runtime dll ready for nuget

xied75:
@parrt @ericvergnaud https://travis-ci.org/xied75/antlr4/builds/226855441

Renamed dll to Antlr4.Runtime.Core as Eric suggested.

dotnet test on Mac now able to finish successfully.

dotnet test on Linux runs faster.

The cmd to pack nuget package is ```dotnet pack Antlr4.Runtime.dotnet.csproj``` then we'll get a nupkg inside bin/

In fact we can push the nupkg automatically (nightly build would be great) if we want. Let me know.

### DIFF
--- a/.travis/before-install-linux-dotnet.sh
+++ b/.travis/before-install-linux-dotnet.sh
@@ -6,7 +6,7 @@ set -euo pipefail
 sudo sh -c 'echo "deb [arch=amd64] https://apt-mo.trafficmanager.net/repos/dotnet-release/ trusty main" > /etc/apt/sources.list.d/dotnetdev.list'
 sudo apt-key adv --keyserver hkp://keyserver.ubuntu.com:80 --recv-keys 417A0893
 sudo apt-get update
-sudo apt-get install dotnet-dev-1.0.1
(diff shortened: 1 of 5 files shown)


ericvergnaud:
mmm… from the below link, it seems to still to be too slow on Mac:
https://travis-ci.org/antlr/antlr4/builds/226877875 <https://travis-ci.org/antlr/antlr4/builds/226877875>

xied75:
That PR build could be a one-off, I've restarted my pull run 3 times, all finished much faster.

ericvergnaud:
Ok, seems to work when re-running it manually.
So I support moving on.
@parrt not sure we want to bump the version in Antlr4.Runtime.dotnet.csproj i.e. the runtime itself is pretty much the same, only the build/test is changed.
I will provide Nuget stuff in a separate PR once this one is merged.

xied75:
ping @ericvergnaud would be great to have your follow up PR so that this finally makes it to the NuGet source. :)

PR #1846, opened 2017-04-28
Fix for recompilation issue with the maven plugin

alimg:
The current version of the maven plugin places the generated files directly into the `generated-sources/antlr4` directory. Unfortunately Maven compiler plugin is unable to detect stale sources correctly if the directory structure doesn't match the package name (Java class files are recreated each time mvn is ran even if no changes were made in the code).

This change set places the files generated by the maven plugin to proper Java package structure when `-package` option is set.


parrt:
How is this PR affected by this PR we are likely to incorporate in some fashion? https://github.com/antlr/antlr4/issues/638

alimg:
In a limited way, this PR seems to be related to #638 in the case about the usage of the `-package` argument being neglected when deciding the output path of the generated code.

This PR attempts to add some sanitation on the Maven plugin side only. The tool behaviour has been kept same, in case there are others relying on the existing behaviour.

However, I've realized that you can get the correct package structure if don't use `-pacakge` and keep the grammar sources on the correct relative path instead. I think this is useful enough, so I'm closing the PR. If you have any suggestion of a practical addition this, we can update and reopen though.

PR #1847, opened 2017-04-29
Small improvement on getErrorDisplay

lys0716:
It would be better to show error always with single quotes in getErrorDisplay(int c). Remove getCharErrorDisplay(int c) since it is duplicate to getErrorDisplay(int c).

Signed-off-by: Yishuang Lu <luyscmu@gmail.com>


lys0716:
After thinking about it, probably I should do it in another way.

PR #1848, opened 2017-04-30
Fix tests for PR #1844.

davesisson:
Fixes tests for PR #1844 (and updates affected documentation).


davesisson:
@parrt can we pull this?  apparently the go code pulls from the master which means anyone attempting to use go right now is broken without this fix.

PR #1849, opened 2017-04-30
Fix the utf8 convertor for VS2017

shravanrn:
The UTF8 conversion functions are broken in a slightly different way in VS2017. Add functions to hide these differences


shravanrn:
The errors in the CI build process seem to be unrelated to this change

runtime/Cpp/runtime/src/support/StringUtils.h
```diff
@@ -10,12 +10,34 @@
 namespace antlrcpp {
   // For all conversions utf8 <-> utf32.
   // VS 2015 has a bug in std::codecvt_utf8<char32_t> (VS 2013 works fine).
-#if defined(_MSC_VER) && _MSC_VER == 1900
+#if defined(_MSC_VER) && _MSC_VER > 1900 && _MSC_VER < 2000
```
KindDragon: Shouldn't be `#if defined(_MSC_VER) && _MSC_VER >= 1900`?

mike-lischke: Yes, true. @shravanrn can you fix that yet?

KindDragon: and bellow

shravanrn:
Sorry about the delay. Based on comments I realized I had actually gotten this slightly incorrect the last. I have double checked that the following works for both VS2015 and VS 2017

shravanrn:
@mike-lischke @KindDragon - Just a friendly reminder :)

mike-lischke:
We need @parrt for a merge. Development on ANTLR4 stopped abruptly and everybody is now waiting for a new time slot where things begin to move again. However, it's not clear that "everybody" has time to move at the same moment as the ANTLR4 starts moving again...

parrt:
Sorry...here ya go.

PR #1854, opened 2017-05-04
Memoize prediction contexts

JWalker1995:
Memoizes `SingletonPredictionContext.toString` and `ArrayPredictionContext.toString`.

### DIFF
--- a/contributors.txt
+++ b/contributors.txt
@@ -144,3 +144,4 @@ YYYY/MM/DD, github id, Full name, email
 2017/03/15, robertvanderhulst, Robert van der Hulst, robert@xsharp.eu
 2017/03/28, cmd-johnson, Jonas Auer, jonas.auer.94@gmail.com
 2017/04/12, lys0716, Yishuang Lu, luyscmu@gmail.com
+2017/05/04, jwalker1995, Joel Walker, joelwalker1995@gmail.com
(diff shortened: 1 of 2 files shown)


ericvergnaud:
Hi,
What problem would this solve?
Eric

Envoyé de mon iPhone

JWalker1995:
This is an optimization. For example, using the JavaScript target to parse this java8 snippet:

```java
class HelloWorldApp {
    public static void main(String[] args) {
        System.out.println("Hello World!"); // Display the string.
    }
}
```

Without memoization: 137s
With memoization: 5s

ericvergnaud:
Wow!
Impressive but it would be interesting to understand why toString is called.
Since we introduced MurmurHash, toString should only be called for display.
I suspect something is not working as expected in the hashing mechanism.
Since you've obviously spent some time instrumenting this, would you mind providing a stack trace of a repeated toString call for the above example?
Then we can eventually provide together a more solid optimisation.

parrt:
@JWalker1995 any movement on this PR?

sharwell:
@JWalker1995 The real bug here is that `toString` is called *at all* in that example. The best way to fix this would be altering `toString` to throw an exception and then fixing anything in the core parsing code which called it.

parrt:
Per @sharwell's comment, I'm closing. thanks!

PR #1856, opened 2017-05-05
Add EOF check for recover method in Lexer

lys0716:
We have two recover() methods in some Lexer(Lexer.go, Lexer.py). But only one method check EOF for the input. Added EOF check for another one(Lexer.java, Lexer.cpp, Lexer.swift, Lexer.cs).

Signed-off-by: Yishuang Lu <luyscmu@gmail.com>

### DIFF
--- a/runtime/CSharp/runtime/CSharp/Antlr4.Runtime/Lexer.cs
+++ b/runtime/CSharp/runtime/CSharp/Antlr4.Runtime/Lexer.cs
@@ -632,7 +632,10 @@ outer_continue: ;
             //System.out.println("consuming char "+(char)input.LA(1)+" during recovery");
             //re.printStackTrace();
             // TODO: Do we lose character or line position information?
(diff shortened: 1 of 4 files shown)


lys0716:
errors seem to be unrelated to this change

parrt:
Hi. We need contributors.txt sign as well.

lys0716:
Hi Terence, I had signed before. Thanks.

parrt:
ah! thanks. Now that I'm looking at it, doesn't the CharStream do the checking so it avoids consuming past end of file?

lys0716:
I made this changes because I found function **recover(LexerNoViableAltException e)** had implemented use case for recover function. They probably both need to have the same if clause. My idea is proven after I looked into Lexer.py and Lexer.js

parrt:
I'm sorry but I'm not following what you say.

PR #1860, opened 2017-05-05
[CharStreams] Factor out Java 7 Path APIs into new PathCharStreams

bhamiltoncx:
I wanted to use the new ANTLR 4.7 runtime on Android, but when I tried to run Proguard on the result, I discovered that Android versions older than the "Android O" developer preview are missing the Java 7 `java.nio.file.Path` and related APIs.

Since these are just convenience APIs to wrap around the `Channel` APIs (which are present in all versions of Android), this PR refactors the Java 7-specific APIs into a new helper class `PathCharStreams` which is simple to exclude via a Proguard rule.

### DIFF
--- a/doc/faq/general.md
+++ b/doc/faq/general.md
@@ -82,7 +82,7 @@ Make sure to use two-stage parsing. See example in [bug report](https://github.c
 
 ```Java
 
-CharStream input = CharStreams.fromPath(Paths.get(args[0]));
+CharStream input = PathCharStreams.fromPath(Paths.get(args[0]));
 ExprLexer lexer = new ExprLexer(input);
 CommonTokenStream tokens = new CommonTokenStream(lexer);
 ExprParser parser = new ExprParser(tokens);
--- a/doc/interpreters.md
+++ b/doc/interpreters.md
@@ -30,7 +30,7 @@ public static ParseTree parse(String fileName,
     throws IOException
 {
     final Grammar g = Grammar.load(combinedGrammarFileName);
-    LexerInterpreter lexEngine = g.createLexerInterpreter(CharStreams.fromPath(Paths.get(fileName)));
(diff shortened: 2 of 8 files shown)


bhamiltoncx:
Anything I should do about the tests which are timing out?

parrt:
Hmm...that's a drag. I hesitate to change the API again.  Can we make a different API just for android, leaving current one in place?

bhamiltoncx:
> Hmm...that's a drag. I hesitate to change the API again. Can we make a different API just for android, leaving current one in place?

That would require duplicating the logic of the existing class, which seems like a shame. Since the API is brand-new, I think it's not too terrible to change it.

Alternatively, we could just delete the `Path` API entirely.

parrt:
dang. I really hesitate to change that API, particularly given we just asked people to change. At the very least it would have to come at another major release, 4.8, I guess. Hmm... What if we made a minimal android CharStreams that only has the things you need? Couldn't call back into the implementation goodies? Perhaps it would require splitting apart the existing implementation, but it would leave the interface the same hopefully.

bhamiltoncx:
The problem with calling back into the existing class from Android is the
Dalvik runtime optimizer will crash if it encounters references to unknown
classes, even if they are unused at runtime. :(

So, we really do need to move the Java 7-using code out into its own class.

ericvergnaud:
Not sure if that's supported by the Android JVM, but an alternative approach could be to provide an Android version of these missing Java 7 classes

parrt:
Yeah, the Dalvik (assuming that is what they still use) no doubt does a conversion from Java byte codes to dalvik, which means it has to sniff all of the classes.  I think the best solution, which avoids altering 4.7 API, is to create a minimal android only api that does not reference verboten classes; we can keep the code duplication to a minimum by using a minimal android API.

parrt:
hi guys, given my limited time, I think I'm going to close this one in favor of my "solution" from May 15.

PR #1867, opened 2017-05-11
Add `new` keyword to return proper InputStreams

jimallman:
Currently the JS runtime `CharStreams.fromString` sometimes returns (and mangles) the global `window` object instead of returning a proper InputStream. This is prevented by using the `new` keyword in this and similar helper methods.

@bhamiltoncx, please take a look. Perhaps I'm doing something wrong... if so, I'm happy to be mistaken!

### DIFF
--- a/contributors.txt
+++ b/contributors.txt
@@ -144,3 +144,4 @@ YYYY/MM/DD, github id, Full name, email
 2017/03/15, robertvanderhulst, Robert van der Hulst, robert@xsharp.eu
 2017/03/28, cmd-johnson, Jonas Auer, jonas.auer.94@gmail.com
 2017/04/12, lys0716, Yishuang Lu, luyscmu@gmail.com
+2017/05/11, jimallman, Jim Allman, jim@ibang.com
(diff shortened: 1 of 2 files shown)


bhamiltoncx:
Ahh, yikes. Yes, this looks totally kosher to me.

Can you also fix lines 33 and 56? They have the same error.

https://github.com/TreeViz/antlr4/blob/5e648f0867720ded8a251c0a5c69b506986e159c/runtime/JavaScript/src/antlr4/CharStreams.js#L33
https://github.com/TreeViz/antlr4/blob/5e648f0867720ded8a251c0a5c69b506986e159c/runtime/JavaScript/src/antlr4/CharStreams.js#L56

jimallman:
> Can you also fix lines 33 and 56? They have the same error.

Done! Thanks for taking a look.

bhamiltoncx:
LGTM. @parrt, anything we need before merging?

ericvergnaud:
Blessed

PR #1868, opened 2017-05-12
fix go runtime compile error

AlexanderChen1989:
go runtime can not compile with error
image


davesisson:
PR #1848 will also fix this issue (and fixes a few other places).

AlexanderChen1989:
@davesisson why not #1848 not merged?

davesisson:
I pinged Terrence on that PR.  I don't think he realized it was an issue
because apparently the recommended way of getting go is accessing the
master branch.  It should be to access the released code instead.

PR #1870, opened 2017-05-14
All error codes distinct

### LINKED ISSUE #1865: Duplicate error codes
In file [ErrorType.java](https://github.com/antlr/antlr4/blob/master/tool/src/org/antlr/v4/tool/ErrorType.java) there are 2 errors using the same error code 160, which is wrong. The first occurence also appears out of order, so I guess this a copy/paste problem and the error number should be 114 or so.

KvanTTT:
fixes #1865

### DIFF
--- a/tool-testsuite/test/org/antlr/v4/test/tool/TestToolSyntaxErrors.java
+++ b/tool-testsuite/test/org/antlr/v4/test/tool/TestToolSyntaxErrors.java
@@ -8,6 +8,7 @@ package org.antlr.v4.test.tool;
 
 import org.antlr.v4.Tool;
 import org.antlr.v4.tool.ErrorType;
+import org.junit.Assert;
 import org.junit.Before;
 import org.junit.Test;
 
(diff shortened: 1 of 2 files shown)


tool-testsuite/test/org/antlr/v4/test/tool/TestToolSyntaxErrors.java
```diff
@@ -60,6 +61,16 @@ public void testSetUp() throws Exception {
 		super.testSetUp();
 	}
 
+	@Test
+	public void AllErrorCodesDistinct() {
+		ErrorType[] errorTypes = ErrorType.class.getEnumConstants();
+		for (int i = 0; i < errorTypes.length; i++) {
```
parrt: maybe add error types to a set then compare length of the set and the number of error types?

KvanTTT: Ok, will fix it.

KvanTTT: I think my solution is better because of `assertNotEquals` fails on the certain error code. It's more clear that length comparison.

PR #1884, opened 2017-05-27
[dotnet] NuGet packaging for .NET Core

waf:
This PR updates the nuspec file so the NuGet package includes the .NET Core DLL. The NuGet package has the following structure to support both .NET 3.5 and up, as well as .NET Core:

```
lib
├── net35
│   └── Antlr4.Runtime.Standard.dll
└── netstandard
    └── Antlr4.Runtime.Core.dll
```

I've verified that when the NuGet package is installed in .NET Framework projects ("normal" .NET), it references Antlr4.Runtime.Standard.dll, and in .NET Core projects it references Antlr4.Runtime.Core.dll.

For the build process, ideally to do a cross-platform build we could use a multi-targeting build with `dotnet`, the new command-line tool from Microsoft. However, due to https://github.com/Microsoft/msbuild/issues/1333 (a bug with msbuild core) we need to continue using mono to build .NET3.5, and `dotnet` to build .NET core. I've included a script that makes this easier, and updated the documentation. 

Please see the conversation from https://github.com/antlr/antlr4/pull/1229#issuecomment-304002438 for more context. /cc @ericvergnaud

### DIFF
--- a/contributors.txt
+++ b/contributors.txt
@@ -145,3 +145,4 @@ YYYY/MM/DD, github id, Full name, email
 2017/03/28, cmd-johnson, Jonas Auer, jonas.auer.94@gmail.com
 2017/04/12, lys0716, Yishuang Lu, luyscmu@gmail.com
 2017/05/11, jimallman, Jim Allman, jim@ibang.com
+2017/05/26, waf, Will Fuqua, wafuqua@gmail.com
--- a/doc/releasing-antlr.md
+++ b/doc/releasing-antlr.md
@@ -247,16 +247,13 @@ popd
 
 *Publishing to Nuget from Linux/MacOSX*
 
-**Getting ready to run Nuget**
+**Install the pre-requisites**
 
 Of course you need Mono and `nuget` to be installed. On mac:
 
-```bash
-brew install mono
-brew install nuget
-```
-
-Or, you can [download nuget.exe](https://dist.nuget.org/win-x86-commandline/latest/nuget.exe).
+- mono - on mac, `brew install mono`
+- nuget - on mac, `brew install nuget` or you can [download nuget.exe](https://dist.nuget.org/win-x86-commandline/latest/nuget.exe)
+- dotnet - follow [the instructions here](https://www.microsoft.com/net/core)
 
 From the shell on mac, you can check all is ok by typing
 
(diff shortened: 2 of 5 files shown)


ericvergnaud:
Hey,
this looks great!
Thanks

ericvergnaud:
Hi,

published 4.7.1!
can you check if it works for you?

Eric
n.b. many thanks to all that contributed to this target :-)

waf:
I have a project that targets both the .NET framework and .NET Core. I upgraded both those projects to 4.7.1, and both worked great! Thank you for packaging it so quickly.

One small thing I noticed is that both DLLs are stamped with the "4.7.0.0", even though the NuGet package is is 4.7.1:



However, this doesn't appear to have any negative effect. Thanks again!

ericvergnaud:
Hi,

noted, we’ll ensure the next release is in sync!
tbh, the runtime hasn’t changed, so it’s 4.7.0, but you can’t republish on Nuget (which is good), so had to bump the nuget version.

Eric

parrt:
Hmm...i haven't released 4.7.1 antlr yet so won't 4.7.1 version confuse things?

ericvergnaud:
TBH I had no choice, given 4.7.0 was already published to Nuget.
It might indeed confuse people a bit, until you release 4.7.2 (skipping 4.7.1?)
Maybe we can add a note in the latest release docs?

parrt:
no problem but why release 4.7.1 now? is that only way to test it via nuget?

ericvergnaud:
Nuget has no such thing as a maven snapshot repo, so yes you’d have to publish to test - very bad imho
But my intent was simply to publish on Nuget the 4.7.0 .Net Core runtime which is long due
As mentioned, it’s the exact same 4.7.0 runtime, no code change, only change is for the Nuget packaging
FYI the Nuget package contains both the Standard and Core versions

KvanTTT:
You could name it as 4.7.0.1 instead of 4.7.1. Antlr version incremented at the third number as usual.

ericvergnaud:
@waf 
@xied75 
Hi, NetCore builds is now failing on Travis CI with error: JAVA_HOME is not defined correctly
Could you have a look?

xied75:
@ericvergnaud 
https://github.com/travis-ci/travis-ci/issues/7884#issuecomment-308451879
http://www.webupd8.org/2017/06/why-oracle-java-7-and-6-installers-no.html
https://blog.travis-ci.com/2017-06-21-trusty-updates-2017-Q2-launch

PR is done. https://github.com/antlr/antlr4/pull/1927

PR #1886, opened 2017-05-28
Set version to 4.7.1 because 4.7.0 already exists on Nuget

ericvergnaud:


parrt:
Note: nuget C# version is really a snapshot as we have not formally released 4.7.1. On the other hand, there will not be very many C# changes before we release 4.7.1 but keep in mind.

PR #1892, opened 2017-05-30
Fix double-free bug in example code.

kosak:
MyGrammarParser owns the pointer returned by parser.key(). That pointer is
ultimately deallocated by Parser::~Parser() via that method's call to
_tracker.reset(). Taking ownership via Ref<tree::ParseTree> would lead to
a double-free bug.

### DIFF
--- a/doc/cpp-target.md
+++ b/doc/cpp-target.md
@@ -63,7 +63,8 @@ int main(int argc, const char* argv[]) {
   CommonTokenStream tokens(&lexer);
   MyGrammarParser parser(&tokens);
 
-  Ref<tree::ParseTree> tree = parser.key();
+  // MyGrammarParser owns this pointer.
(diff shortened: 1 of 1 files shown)


kosak:
I don't understand how a documentation-only change could trigger a Travis CI build failure.

ericvergnaud:
Any PR triggers a build, and builds sometimes fail due to Travis timeouts

Envoyé de mon iPhone

kosak:
Should I fix this PR so there's no merge conflict?

(sorry for the dumb Q; I don't know what your process is)

parrt:
I think this conflicts because the fix was already pulled in, right @mike-lischke ?

mike-lischke:
Ah right, I thought I had seen this fix already. So this pull request can be closed.

kosak:
Ah, cool, yes, thanks.

PR #1902, opened 2017-06-10
C++ runtime changes for high warning levels

jm-mikkelsen:
These changes are for compiling the C++ runtime and code using the C++ runtime headers with high warning levels and -Werror. There are no functional changes in this commit. Compiled with gcc 5.4 and clang 3.8.

### DIFF
--- a/runtime/Cpp/runtime/src/CommonTokenStream.cpp
+++ b/runtime/Cpp/runtime/src/CommonTokenStream.cpp
@@ -12,8 +12,8 @@ using namespace antlr4;
 CommonTokenStream::CommonTokenStream(TokenSource *tokenSource) : CommonTokenStream(tokenSource, Token::DEFAULT_CHANNEL) {
 }
 
-CommonTokenStream::CommonTokenStream(TokenSource *tokenSource, size_t channel_in)
-: BufferedTokenStream(tokenSource), channel(channel_in) {
+CommonTokenStream::CommonTokenStream(TokenSource *tokenSource, size_t channel_)
+: BufferedTokenStream(tokenSource), channel(channel_) {
 }
 
 ssize_t CommonTokenStream::adjustSeekIndex(size_t i) {
--- a/runtime/Cpp/runtime/src/CommonTokenStream.cpp
+++ b/runtime/Cpp/runtime/src/CommonTokenStream.cpp
@@ -12,8 +12,8 @@ using namespace antlr4;
 CommonTokenStream::CommonTokenStream(TokenSource *tokenSource) : CommonTokenStream(tokenSource, Token::DEFAULT_CHANNEL) {
 }
 
-CommonTokenStream::CommonTokenStream(TokenSource *tokenSource, size_t channel)
-: BufferedTokenStream(tokenSource), channel(channel) {
+CommonTokenStream::CommonTokenStream(TokenSource *tokenSource, size_t channel_in)
+: BufferedTokenStream(tokenSource), channel(channel_in) {
 }
 
 ssize_t CommonTokenStream::adjustSeekIndex(size_t i) {
@@ -25,7 +25,7 @@ Token* CommonTokenStream::LB(size_t k) {
     return nullptr;
   }
 
-  ssize_t i = (ssize_t)_p;
+  ssize_t i = static_cast<ssize_t>(_p);
   size_t n = 1;
   // find k good tokens looking backwards
   while (n <= k) {
@@ -46,7 +46,7 @@ Token* CommonTokenStream::LT(ssize_t k) {
     return nullptr;
   }
   if (k < 0) {
-    return LB((size_t)-k);
+    return LB(static_cast<size_t>(-k));
   }
   size_t i = _p;
   ssize_t n = 1; // we know tokens[p] is a good one
--- a/runtime/Cpp/runtime/src/Exceptions.cpp
+++ b/runtime/Cpp/runtime/src/Exceptions.cpp
@@ -23,11 +23,42 @@ const char* IOException::what() const NOEXCEPT {
   return _message.c_str();
 }
 
-IllegalStateException::~IllegalStateException() = default;
-IllegalArgumentException::~IllegalArgumentException() = default;
-NullPointerException::~NullPointerException() = default;
-IndexOutOfBoundsException::~IndexOutOfBoundsException() = default;
-UnsupportedOperationException::~UnsupportedOperationException() = default;
-EmptyStackException::~EmptyStackException() = default;
-CancellationException::~CancellationException() = default;
-ParseCancellationException::~ParseCancellationException() = default;
+//------------------ IllegalStateException -----------------------------------------------------------------------------
+
+IllegalStateException::~IllegalStateException() {
+}
+
+//------------------ IllegalArgumentException --------------------------------------------------------------------------
+
+IllegalArgumentException::~IllegalArgumentException() {
+}
+
+//------------------ NullPointerException ------------------------------------------------------------------------------
+
+NullPointerException::~NullPointerException() {
+}
+
+//------------------ IndexOutOfBoundsException -------------------------------------------------------------------------
+
+IndexOutOfBoundsException::~IndexOutOfBoundsException() {
+}
+
+//------------------ UnsupportedOperationException ---------------------------------------------------------------------
+
+UnsupportedOperationException::~UnsupportedOperationException() {
+}
+
+//------------------ EmptyStackException -------------------------------------------------------------------------------
+
+EmptyStackException::~EmptyStackException() {
+}
+
+//------------------ CancellationException -----------------------------------------------------------------------------
+
+CancellationException::~CancellationException() {
+}
+
+//------------------ ParseCancellationException ------------------------------------------------------------------------
+
+ParseCancellationException::~ParseCancellationException() {
+}
--- a/runtime/Cpp/runtime/src/Exceptions.cpp
+++ b/runtime/Cpp/runtime/src/Exceptions.cpp
@@ -22,3 +22,12 @@ IOException::IOException(const std::string &msg) : std::exception(), _message(ms
 const char* IOException::what() const NOEXCEPT {
   return _message.c_str();
 }
+
+IllegalStateException::~IllegalStateException() = default;
+IllegalArgumentException::~IllegalArgumentException() = default;
+NullPointerException::~NullPointerException() = default;
+IndexOutOfBoundsException::~IndexOutOfBoundsException() = default;
+UnsupportedOperationException::~UnsupportedOperationException() = default;
+EmptyStackException::~EmptyStackException() = default;
+CancellationException::~CancellationException() = default;
+ParseCancellationException::~ParseCancellationException() = default;
--- a/runtime/Cpp/runtime/src/Exceptions.h
+++ b/runtime/Cpp/runtime/src/Exceptions.h
@@ -21,32 +21,51 @@ namespace antlr4 {
 
   class ANTLR4CPP_PUBLIC IllegalStateException : public RuntimeException {
   public:
-    IllegalStateException(const std::string &msg = "") : RuntimeException(msg) {};
+    IllegalStateException(const std::string &msg = "") : RuntimeException(msg) {}
+    IllegalStateException(IllegalStateException const&) = default;
+    ~IllegalStateException();
+    IllegalStateException& operator=(IllegalStateException const&) = default;
   };
 
   class ANTLR4CPP_PUBLIC IllegalArgumentException : public RuntimeException {
   public:
-    IllegalArgumentException(const std::string &msg = "") : RuntimeException(msg) {};
+    IllegalArgumentException(IllegalArgumentException const&) = default;
+    IllegalArgumentException(const std::string &msg = "") : RuntimeException(msg) {}
+    ~IllegalArgumentException();
+    IllegalArgumentException& operator=(IllegalArgumentException const&) = default;
   };
 
   class ANTLR4CPP_PUBLIC NullPointerException : public RuntimeException {
   public:
-    NullPointerException(const std::string &msg = "") : RuntimeException(msg) {};
+    NullPointerException(const std::string &msg = "") : RuntimeException(msg) {}
+    NullPointerException(NullPointerException const&) = default;
+    ~NullPointerException();
+    NullPointerException& operator=(NullPointerException const&) = default;
   };
 
   class ANTLR4CPP_PUBLIC IndexOutOfBoundsException : public RuntimeException {
   public:
-    IndexOutOfBoundsException(const std::string &msg = "") : RuntimeException(msg) {};
+    IndexOutOfBoundsException(const std::string &msg = "") : RuntimeException(msg) {}
+    IndexOutOfBoundsException(IndexOutOfBoundsException const&) = default;
+    ~IndexOutOfBoundsException();
+    IndexOutOfBoundsException& operator=(IndexOutOfBoundsException const&) = default;
   };
 
   class ANTLR4CPP_PUBLIC UnsupportedOperationException : public RuntimeException {
   public:
-    UnsupportedOperationException(const std::string &msg = "") : RuntimeException(msg) {};
+    UnsupportedOperationException(const std::string &msg = "") : RuntimeException(msg) {}
+    UnsupportedOperationException(UnsupportedOperationException const&) = default;
+    ~UnsupportedOperationException();
+    UnsupportedOperationException& operator=(UnsupportedOperationException const&) = default;
+
   };
 
   class ANTLR4CPP_PUBLIC EmptyStackException : public RuntimeException {
   public:
-    EmptyStackException(const std::string &msg = "") : RuntimeException(msg) {};
+    EmptyStackException(const std::string &msg = "") : RuntimeException(msg) {}
+    EmptyStackException(EmptyStackException const&) = default;
+    ~EmptyStackException();
+    EmptyStackException& operator=(EmptyStackException const&) = default;
   };
 
   // IOException is not a runtime exception (in the java hierarchy).
@@ -63,12 +82,18 @@ namespace antlr4 {
 
   class ANTLR4CPP_PUBLIC CancellationException : public IllegalStateException {
   public:
-    CancellationException(const std::string &msg = "") : IllegalStateException(msg) {};
+    CancellationException(const std::string &msg = "") : IllegalStateException(msg) {}
+    CancellationException(CancellationException const&) = default;
+    ~CancellationException();
+    CancellationException& operator=(CancellationException const&) = default;
   };
 
   class ANTLR4CPP_PUBLIC ParseCancellationException : public CancellationException {
   public:
-    ParseCancellationException(const std::string &msg = "") : CancellationException(msg) {};
+    ParseCancellationException(const std::string &msg = "") : CancellationException(msg) {}
+    ParseCancellationException(ParseCancellationException const&) = default;
+    ~ParseCancellationException();
+    ParseCancellationException& operator=(ParseCancellationException const&) = default;
   };
 
 } // namespace antlr4
--- a/runtime/Cpp/runtime/src/InputMismatchException.cpp
+++ b/runtime/Cpp/runtime/src/InputMismatchException.cpp
@@ -13,3 +13,7 @@ InputMismatchException::InputMismatchException(Parser *recognizer)
   : RecognitionException(recognizer, recognizer->getInputStream(), recognizer->getContext(),
                          recognizer->getCurrentToken()) {
 }
+
+InputMismatchException::~InputMismatchException()
+{
+}
--- a/runtime/Cpp/runtime/src/InputMismatchException.cpp
+++ b/runtime/Cpp/runtime/src/InputMismatchException.cpp
@@ -14,6 +14,5 @@ InputMismatchException::InputMismatchException(Parser *recognizer)
                          recognizer->getCurrentToken()) {
 }
 
-InputMismatchException::~InputMismatchException()
-{
+InputMismatchException::~InputMismatchException() {
 }
--- a/runtime/Cpp/runtime/src/antlr4-common.h
+++ b/runtime/Cpp/runtime/src/antlr4-common.h
@@ -77,7 +77,7 @@
 
   class ANTLR4CPP_PUBLIC std::exception; // Needed for VS 2015.
 
-#elif __APPLE__
+#elif defined(__APPLE__)
   typedef std::u32string UTF32String;
 
   #define GUID_CFUUID
--- a/runtime/Cpp/runtime/src/antlr4-common.h
+++ b/runtime/Cpp/runtime/src/antlr4-common.h
@@ -12,6 +12,7 @@
 #include <fstream>
 #include <iostream>
 #include <iterator>
+#include <limits>
 #include <limits.h>
 #include <list>
 #include <map>
@@ -76,7 +77,7 @@
   class ANTLR4CPP_PUBLIC std::exception; // Needed for VS 2015.
 #endif
 
-#elif __APPLE__
+#elif defined(__APPLE__)
   #define GUID_CFUUID
   #if __GNUC__ >= 4
     #define ANTLR4CPP_PUBLIC __attribute__ ((visibility ("default")))
@@ -120,5 +121,5 @@
 #undef EOF
 #endif
 
-#define INVALID_INDEX (size_t)-1
+#define INVALID_INDEX std::numeric_limits<size_t>::max()
 template<class T> using Ref = std::shared_ptr<T>;
--- a/runtime/Cpp/runtime/src/atn/LexerActionType.h
+++ b/runtime/Cpp/runtime/src/atn/LexerActionType.h
@@ -5,6 +5,8 @@
 
 #pragma once
 
+#include <cstddef>
+
 namespace antlr4 {
 namespace atn {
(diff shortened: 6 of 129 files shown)


jm-mikkelsen:
Just added changes to remove the fix the utf8_to_utf32 auto return type which broke the CI builds. While I was there I removed the static utfConverter instance; access to its {from,to}_bytes methods are not guaranteed thread safe.

runtime/Cpp/runtime/src/atn/LexerActionType.h
```diff
@@ -5,6 +5,8 @@
 
 #pragma once
 
+#include <stddef.h>
+
```
mike-lischke: What's that for?

runtime/Cpp/runtime/src/atn/LexerActionType.h
```diff
@@ -14,7 +16,7 @@ namespace atn {
   /// @author Sam Harwell
   /// @since 4.2
   /// </summary>
-  enum class ANTLR4CPP_PUBLIC LexerActionType : size_t {
+  enum class LexerActionType : size_t {
```
mike-lischke: Removed by accident?

runtime/Cpp/runtime/src/atn/SemanticContext.h
```diff
@@ -144,6 +144,8 @@ namespace atn {
    */
   class ANTLR4CPP_PUBLIC SemanticContext::Operator : public SemanticContext {
   public:
+    ~Operator();
```
mike-lischke: Shouldn't that be a virtual d-tor?

runtime/Cpp/runtime/src/misc/Interval.h
```diff
@@ -28,7 +28,9 @@ namespace misc {
     Interval();
     explicit Interval(size_t a_, size_t b_); // For unsigned -> signed mappings.
     Interval(ssize_t a_, ssize_t b_);
-    virtual ~Interval() {};
+    Interval(Interval const&) = default;
+    virtual ~Interval();
```
mike-lischke: Code formatting: d-tors after c-tors, but before rest of declarations/definitions.

runtime/Cpp/runtime/src/misc/Interval.h
```diff
@@ -28,7 +28,9 @@ namespace misc {
...
+    Interval& operator=(Interval const&) = default;
```
mike-lischke: Marked with `default` but has a body? How does this go together?

runtime/Cpp/runtime/src/misc/IntervalSet.cpp
```diff
@@ -46,6 +46,21 @@ IntervalSet::IntervalSet(int n, ...) : IntervalSet() {
...
 
+IntervalSet::~IntervalSet()
+{
+}
+
+IntervalSet& IntervalSet::operator=(const IntervalSet& other)
+{
+  if (_readonly) {
+    throw IllegalStateException("can't alter read only IntervalSet");
+  }
```
mike-lischke: I'm not sure this is the intention of the original code. If you copy an interval set the old one goes completely, so it doesn't matter if it was read only or not, but instead takes the entire state from the source interval set (including it's read only value).

runtime/Cpp/runtime/src/misc/IntervalSet.h
```diff
@@ -30,15 +31,17 @@ namespace misc {
   protected:
     /// The list of sorted, disjoint intervals.
     std::vector<Interval> _intervals;
-    bool _readonly;
+    std::atomic<bool> _readonly;
```
mike-lischke: Why do we have a mutex in the ATN for protecting the read-only state change **and** make the _readonly flag atomic? That's not necessary - one of the two can be removed. I'd prefer `atomic` to stay.

runtime/Cpp/runtime/src/support/Any.cpp
```diff
@@ -0,0 +1,10 @@
+#include "Any.h"
+
+antlrcpp::Any::~Any()
+{
+    delete _ptr;
+}
+
+antlrcpp::Any::Base::~Base()
+{
+}
```
mike-lischke: I'm usually fine with having the implementation in a cpp file, so that change is ok. But when you start such a clean up it should be done fully, i.e. move all definitions here from the header.

runtime/Cpp/runtime/src/support/CPPUtils.h
```diff
@@ -21,14 +21,14 @@ namespace antlrcpp {
   // Using RAII + a lambda to implement a "finally" replacement.
   struct FinalAction {
     FinalAction(std::function<void ()> f) : _cleanUp { f } {}
-    FinalAction(FinalAction &&other) {
-      _cleanUp = other._cleanUp;
-      _enabled = other._enabled;
+    FinalAction(FinalAction &&other) :
+	_cleanUp(std::move(other._cleanUp)), _enabled(other._enabled)
+    {
```
mike-lischke: Coding style: opening brace on the end of the line opening the block (as it was originally).

runtime/Cpp/runtime/src/support/StringUtils.h
```diff
@@ -11,31 +11,33 @@ namespace antlrcpp {
   // For all conversions utf8 <-> utf32.
   // VS 2015 and VS 2017 have different bugs in std::codecvt_utf8<char32_t> (VS 2013 works fine).
 #if defined(_MSC_VER) && _MSC_VER >= 1900 && _MSC_VER < 2000
-  static std::wstring_convert<std::codecvt_utf8<__int32>, __int32> utfConverter;
+  using UtfConverterType = std::wstring_convert<std::codecvt_utf8<__int32>, __int32>;
+  using UtfConverterWide = std::u32string;
 #else
-  static std::wstring_convert<std::codecvt_utf8<char32_t>, char32_t> utfConverter;
+  using UtfConverterType = std::wstring_convert<std::codecvt_utf8<char32_t>, char32_t>;
+  using UtfConverterWide = std::wstring_convert<std::codecvt_utf8<char32_t>, char32_t>::wide_string;
```
mike-lischke: Have you tested this on Windows? UtfConverterWide is defined as std::u32string here, while for non-Win it's a std::wstring_convert.

This part has already been improved and it's better not to include such changes in this patch.

runtime/Cpp/runtime/src/tree/xpath/XPathLexer.cpp
```diff
@@ -47,7 +47,7 @@ const atn::ATN& XPathLexer::getATN() const {
 
 void XPathLexer::action(RuleContext *context, size_t ruleIndex, size_t actionIndex) {
   switch (ruleIndex) {
-    case 4: IDAction(dynamic_cast<antlr4::RuleContext *>(context), actionIndex); break;
+    case 4: IDAction(context, actionIndex); break;
```
mike-lischke: This is generated code. You shouldn't change that manually.

mike-lischke:
And, btw. this patch of course has functional changes.

jm-mikkelsen:
Thanks for the review Mike -- I appreciate that it is a big patch and takes an effort!

I'll respond in detail to your comments inline.

runtime/Cpp/runtime/src/CommonTokenStream.cpp
```diff
@@ -12,8 +12,8 @@ using namespace antlr4;
 CommonTokenStream::CommonTokenStream(TokenSource *tokenSource) : CommonTokenStream(tokenSource, Token::DEFAULT_CHANNEL) {
 }
 
-CommonTokenStream::CommonTokenStream(TokenSource *tokenSource, size_t channel)
-: BufferedTokenStream(tokenSource), channel(channel) {
+CommonTokenStream::CommonTokenStream(TokenSource *tokenSource, size_t channel_in)
+: BufferedTokenStream(tokenSource), channel(channel_in) {
 }
```
jm-mikkelsen: Will update

runtime/Cpp/runtime/src/Exceptions.cpp
```diff
@@ -22,3 +22,12 @@ IOException::IOException(const std::string &msg) : std::exception(), _message(ms
...
 }
+
+IllegalStateException::~IllegalStateException() = default;
+IllegalArgumentException::~IllegalArgumentException() = default;
+NullPointerException::~NullPointerException() = default;
+IndexOutOfBoundsException::~IndexOutOfBoundsException() = default;
+UnsupportedOperationException::~UnsupportedOperationException() = default;
+EmptyStackException::~EmptyStackException() = default;
+CancellationException::~CancellationException() = default;
+ParseCancellationException::~ParseCancellationException() = default;
```
jm-mikkelsen: The problem with putting the "= default" in the header file is that the vtable is not generated as part of the translation unit. I kept the same order as the pattern used in the header file changes, the difference is that the destructor doesn't have an inline definition.

Would you prefer an empty destructor or "= default" in a .cpp file?

runtime/Cpp/runtime/src/Exceptions.h
```diff
@@ -21,32 +21,51 @@ namespace antlr4 {
 
   class ANTLR4CPP_PUBLIC IllegalStateException : public RuntimeException {
   public:
-    IllegalStateException(const std::string &msg = "") : RuntimeException(msg) {};
+    IllegalStateException(const std::string &msg = "") : RuntimeException(msg) {}
+    IllegalStateException(IllegalStateException const&) = default;
+    ~IllegalStateException();
```
jm-mikkelsen: As in the previous comment. To make sure that the vtable for the class with virtual functions is only generated in one translation unit.

runtime/Cpp/runtime/src/InputMismatchException.cpp
```diff
@@ -13,3 +13,7 @@ InputMismatchException::InputMismatchException(Parser *recognizer)
   : RecognitionException(recognizer, recognizer->getInputStream(), recognizer->getContext(),
                          recognizer->getCurrentToken()) {
 }
+
+InputMismatchException::~InputMismatchException()
```
jm-mikkelsen: As in the previous two comments. Happy make a choice between empty destructors and "= default"

runtime/Cpp/runtime/src/atn/LexerActionType.h
jm-mikkelsen: The definition of size_t.

runtime/Cpp/runtime/src/atn/LexerActionType.h
jm-mikkelsen: No, "enum class" defines compile time constants, not external symbols. Those symbols are not exported from a DLL like a real class, so specifying linker behaviour for them is meaningless.

runtime/Cpp/runtime/src/atn/SemanticContext.h
jm-mikkelsen: It is virtual because SemanticContext has a virtual destructor. In our codebase I would have written it is as "~Operator override;". Our reasoning on virtual functions is "virtual" keyword is superfluous when using "override" or "final", and "override" is superfluous when using "final", so our rule is to pick one and one only on a virtual function.

I was trying (not that it's obvious!) to stay consistent with the files I was touching, so happy to say "virtual ~Operator()" or "~Operator() override" as you prefer.

runtime/Cpp/runtime/src/misc/Interval.h
jm-mikkelsen: I think I'm missing something here ... The function after the destructor is the assignment operator.

runtime/Cpp/runtime/src/misc/IntervalSet.h
jm-mikkelsen: I comment on this in the bug report. The atomic<bool> alone is not sufficient to remove the race condition around updates. This is a pattern we use a lot.

jm-mikkelsen:
> That's quite a big patch and I haven't commented on all places where a change is needed. Let me rather summarize here and consider changing all places that are affected:
> 
> 1. Consistency: your change from {} to = default is ok, but it should be used consequently. Often you added an empty body instead. These should all be removed and replaced by default.

Is this how you would like things to look for destructors in .cpp files as well, or would you prefer empty bodies there?

> 2. You added several times an empty d-tor (not even virtual), which makes no sense. It would make sense for virtual d-tors. The other d-tors are just noise.

Which destructor is not virtual?

There are two types of warnings leading to the destructor changes:

1. A vtable for the class will be generated in each translation unit.
2. The class has pointer members but has compiler generated functions, eg. destructor or operator=.

For the first case, the answer is explicitly declare the destructor and put it into a .cpp file.

For the second case, the answer is to explicitly define the function, even if it is empty (for a destructor) or "= default". The intent here is to signal that you have considered what is going on, and say it's OK. 

Seeing the default constructed operator=() being significantly different to the copy constructor in IntervalSet came directly from dealing with this type of warning. Which in turn led to discovering that there were multiple assignments of cached interval sets in ATN, which led directly to a resolution of the race condition in issue #1826.

> 3. You added several times copy c-tor and copy assignment functions only to say they should use the default impl. Why, by all means, did you even add them if you rely on the default impl anyway. That looks so totally superfluous.

Please see above.

> 4. The change from #if to #if defined also seems like useless cosmetic change, or is there anything that requires the latter (I haven't seen any problems or warnings resulting from the original code)?

`#if defined(...)` is used throughout the code, I fixed a few cases where it was not used. The warning was around a construct like `#if SYMBOL > 0` rather than `#if defined(SYMBOL) && SYMBOL > 0`. 

> 5. Code formatting: in the code you added (mostly the new d-tor impls, but also others) you did not consider the coding style in the files -> opening brace at the end of the line, not an own line.

Will bow to the local curly brace convention!

runtime/Cpp/runtime/src/antlr4-common.h
```diff
@@ -76,7 +77,7 @@
   class ANTLR4CPP_PUBLIC std::exception; // Needed for VS 2015.
 #endif
 
-#elif __APPLE__
+#elif defined(__APPLE__)
```
jm-mikkelsen: Here's the message:

`antlr4/runtime/Cpp/runtime/src/antlr4-common.h:80:7: error: '__APPLE__' is not
      defined, evaluates to 0 [-Werror,-Wundef]
#elif __APPLE__
`

runtime/Cpp/runtime/src/misc/Interval.h
jm-mikkelsen: Not sure what you mean here? I don't see the body ...

runtime/Cpp/runtime/src/misc/IntervalSet.cpp
jm-mikkelsen: Hmm. That implies that the meaning of `_readonly` is different to my usage in the concurrency patch.

However, is that really true? This code detected the double assignment in `ATN::nextStates(ATNState*)`, and C++ operator= is not the same as Java reference assignment.

The possibilities I see:

1. Copy `_readonly` in the operator=(), have a different atomic field for use by ATN to detect races.
2. Leave the behaviour as in this patch, still add a different atomic field for use by ATN.
3. Leave the patch as-is.

There are probably other options I've missed. What do you think?

runtime/Cpp/runtime/src/support/StringUtils.h
jm-mikkelsen: Merged your changes.

runtime/Cpp/runtime/src/Exceptions.cpp
mike-lischke: OK, I see. In such cases make them empty d-tors in the cpp file - properly ordered with the rest of the code.

runtime/Cpp/runtime/src/Exceptions.h
mike-lischke: Weird, I never had problems with v-table creation so far. Is this something that changed in a newer compiler? Neither clang nor msc seem to have a problem with that either.

runtime/Cpp/runtime/src/atn/LexerActionType.h
mike-lischke: I see, maybe we should follow the pattern here I used in other header files. They either include antlr4-common.h or a file which includes that.

runtime/Cpp/runtime/src/atn/LexerActionType.h
mike-lischke: Aha, didn't think of that (maybe I got just confused by `enum *class*`. OK then.

runtime/Cpp/runtime/src/atn/SemanticContext.h
mike-lischke: Good reasoning. I prefer explicit code (and use that in all my projects). So for a base virtual method I use `virtual method();` and for overrides `virtual method() override;`. But for virtual d-tors I often missed to use **override**, which is a mistake. Have to take more care for that.

For the antlr runtime I'd like to use the same approach, so this (and similar places) should be changed to `virtual ~Operator() override;` even though `virtual` seems to be superfluous here (but aligns nicely visually with other virtuals).

runtime/Cpp/runtime/src/misc/Interval.h
mike-lischke: Ops, sorry, you are right. All fine here.

runtime/Cpp/runtime/src/misc/IntervalSet.cpp
mike-lischke: Hmm, tricky. In fact there is no copy function like this in Java code, so there is no real copy operation there. When they use an assignment they replace a reference. Maybe we should just **delete** the assignment operator to avoid any unintended behavior and so stay closer to the Java  code? If needed one can easily create a new interval set.

runtime/Cpp/runtime/src/misc/IntervalSet.h
mike-lischke: Hmm, now I get it. You have that check before the mutex (where atomic comes in) then the mutex lock. So this seems ok then. I'm just a bit nervous about performance impact. The interval set class is one of the lowest level classes and used everywhere in the antlr4 runtime.

mike-lischke:
For your discussion about my summary, I believe that's all clarified in the code review now and we are almost there with the patch.

runtime/Cpp/runtime/src/Exceptions.h
jm-mikkelsen: The vtables are created OK, and will work. The problem is that there are too many of them.

That has been part of my coding standard for so long, I assumed everyone did it that way. However, finding an online reference that gives this advice has been surprisingly difficult in about half an hour of searching.

The best origin reference I found with a quick look around the office from John Lakos's "Large-Scale C++ Software Design" (from 1996), which has the "minor design rule":

> "In every class that declares or is derived from a class that declares a virtual function, explicitly declare the destructor as the first virtual function in the class and define it out of line."

It goes on to describe the reasons, including an anecdote of thousands of instances of the same functions in different translation units caused by duplicated vtables. Locality of reference will probably also suffer (ie. different instances of the same function will lead to less code fitting into the CPU's instruction cache.)

(I read that in 1997 and pretty much adopted all of it, so I have probably been doing this for at least 20 years.)

Item 24 in Scott Meyers' "More Effective C++" also cautions against inline virtual function definitions with the quote "In large systems, this can lead to programs containing hundreds or thousands of copies of a class's vtbl!"

And, of course, Clang has a warning to let you know you're making your object too big:

> -Wweak-vtables
>
> Diagnostic text:
> 
> warning: A has no out-of-line virtual method definitions; its vtable will be emitted in every translation unit

runtime/Cpp/runtime/src/atn/LexerActionType.h
jm-mikkelsen: OK, will change.

runtime/Cpp/runtime/src/misc/IntervalSet.cpp
jm-mikkelsen: The problem is that IntervalSet assignment is used in ATN.cpp, LL1Analyzer.cpp and Parser.cpp, so they'd all need to change in some way. The only place that _readonly is really used in the code (as far as I can see) is in ATN.cpp where clearing the readonly flag on copy seems like the correct behaviour. This is also consistent with the copy constructor -- It clears the read only flag.

I argue that the copy constructor and the assignment operator should have the same behaviour; otherwise `IntervalSet s { other_set };` and `IntervalSet s; s = other_set;` will have different results when `other_set` is read only, and `IntervalSet s = other_set;` could go either way. This seems surprising.

Given C++'s value semantics, assigning to a read-only interval set seems like a bug. Honouring the flag like the copy constructor does makes intuitive sense to me.

runtime/Cpp/runtime/src/Exceptions.h
mike-lischke: Makes a lot of sense. Something new learned. Thanks a lot.

runtime/Cpp/runtime/src/misc/IntervalSet.h
jm-mikkelsen: The idea here is that you can check _readonly orders of magnitude more quickly than taking a lock on a mutex. That's pretty fast, and in the scheme of all the other stuff going on (eg. copying lots of IntervalSet) not too bad.

For optimising IntervalSet, adding `IntervalSet(IntervalSet&&)` and `IntervalSet& operator=(IntervalSet&&)` and using `std::move` is likely to be a big win. The other question is whether the copy constructor should really call `addAll()` at all because the source instance will have already done all the checking in `add(Interval const&)`. 

IntervalSets seem to be copied a lot in the code, this is likely to improve things.

runtime/Cpp/runtime/src/misc/IntervalSet.cpp
jm-mikkelsen: Thinking further, it seems to me that _readonly is a Java way of getting const references. `ATN::nextTokens()`, for example, returns a non-const reference to a cache IntervalSet, and there has to be a way to prevent callers from modifying the cache.

In this world-view, clearing the read only flag on copy seems correct. It also seems like the flag could be removed if the code started return const references rather than non-const references in that case. Making all this stuff const-correct seems like a big job, though.

jm-mikkelsen: A first pass seemed a bit too easy. Do you have a performance test you can run?

[movable-IntervalSet.txt](https://github.com/antlr/antlr4/files/1067896/movable-IntervalSet.txt)

mike-lischke: Hmm, I'm not sure if that's the whole story. Way too many side effects possible...

We only have the runtime tests (there are a few performance tests in them, but rather for measuring of the overall process, hard to use them as comparison). At least the runtime could indicate if something goes wrong in general. You can easily run the C++ runtime tests only on your local box to check this.

jm-mikkelsen:
After going further into the Interval/IntervalSet/readonly and the C++ pass-by-value vs. Java pass reference by value differences, I now see many issues in the code where classes with virtual functions are passed by value and assigned to instances of the base class.

For example, ParseTreeMatcher::split() creates an instance of a TextChunk and assigns it to a Chunk(). This is a bug.

https://en.wikipedia.org/wiki/Object_slicing

I'm looking at fixing this, at least to an extent. Not sure whether to add to this PR, or just start another branch ...

mike-lischke:
This is a separate issue and should be handled in an own patch. To be honest the parse tree matcher has not been tested so far and I'm not even sure there are runtime tests for it. So, everything below pattern/ is not fully complete, I'm afraid (because it's rarely used).

jm-mikkelsen:
OK, thanks Mike. After picking up Antlr again on Friday afternoon (after a few years), I've gone a lot further than I expected in looking at the C++ runtime.

You've answered my internal "how does this work for anyone" question about the pattern/ stuff with "it doesn't". I'll submit my patch once this patch is accepted.

Other pieces of work I haven't committed yet:

- A rework of Interval, IntervalSet, ATN and ATNState, removing the readonly status from IntervalSet, removing virtual functions from IntervalSet and Interval, moving the atomic<bool> into ATNState as a "now cached" flag, returning a const reference from ATN::nextStates(ATNState*) so the readonly status is enforced by the compiler not at runtime in the code and value semantics using std::move to reduce the number of copies performed.

- Fixes in tree/xpath for object slicing problems. Is this code expected to work?

- Remove virtual functions from Vocabulary, ATNDeserializationOptions and ParseInfo and make them final. They are copied by value throughout the code so virtual functions are meaningless. Also remove the unused readonly flag on ATNDeserializationOptions.

- Explicitly delete copy constructor and assignment operator on various classes with virtual functions where actually using them would be an error (as with the actual usage of XPathElement and Chunk). There are more instances but I'm not going to go after them all now. I've turned down warnings a lot for the Antlr C++ runtime code now.

- CMakeLists.txt update to support FreeBSD.

Unit tests pass for me with all changes.

Of these, the first one is the one that is (in my view) possibly worth adding to this pull request. 

I hope the changes are worthwhile to you -- I think they help improve things, happy to update in response to feedback.

(I'm in Sydney so I'm a little out of sync with your timezone.)

mike-lischke:
Ah, that sounds promising. Nobody else (other than the runtime authors) has put so much detail work into the C++ runtime yet like you. Please open PRs as you progress and we will review and merge them one by one.

mike-lischke:
@parrt This patch applies to the C++ runtime only and improves several aspects in it and fixes a threading issue. The runtime tests succeed, so I think it's worth to be merged.

PR #1907, opened 2017-06-11
[C++] Small optimizations

mike-lischke:
Found a better way to manage `s0` in the DFA if set from outside. And a few other small changes.


jm-mikkelsen:
The change to Cpp.stg, removing the "antlr4::" qualifier on RuleContext, breaks unit tests for me.

mike-lischke:
A pity that the Travis test constantly time out. That way you never see if you break the build. Thanks @jm-mikkelsen, I'll open a new PR to revert that change.

PR #1908, opened 2017-06-11
[C++] Build fix/correction for UTF32 conversion.

mike-lischke:
Adds a few fixes after the recent UTF32 conversion changes. Also fixes some warnings.


ericvergnaud:
TestCompositeParsers still fails with a compiler error, maybe you need to have a second look?

mike-lischke:
@ericvergnaud Where do you see a compiler error? What fails is what always fails since last fall, job runtime exceeded maximum: https://travis-ci.org/antlr/antlr4/jobs/241708892

ericvergnaud:
@mike-lischke look in https://travis-ci.org/antlr/antlr4/jobs/241757707, which is from the latest master + non cpp changes

mike-lischke:
That's a completely different branch. Why do you post the failure notice in this PR? I have a pull request (#1910)  for that failure, but there's not much moving in antlr4 these days.

mike-lischke:
Now I see what you mean. Still it's the wrong PR. The one where the built broke was #1907 which had been merged short before this one. Maybe @parrt can merge the fix in #1910 quickly, which is just a simple template change.

ericvergnaud:
Agreed

parrt:
Hi guys, i think i've caught up with merges related to bug fixes.

PR #1909, opened 2017-06-11
Corrected `.js` file endings to `.cs` in doc/chsharp-target.md

erikbra:
The file endings for `MyGrammarBaseListener`, `MyGrammarVisitor` and `MyGrammarBaseVisitor` was `.js`. Changed to `.cs`


erikbra:
I don't think the build error has anything to do with me changing the documentation...

PR #1917, opened 2017-06-15
added SVG export of parse tree

rlfnb:
Main driver for exporting to SVG was the ability to search in big parse trees. I wont add new dependencies to the pom.xml, but if its ok, I'll prepare the next commit for the SVG view, too.


parrt:
@rlfnb Can you change it so the whitespace doesn't show up as changed? I can't understand the difference with all those showing.

rlfnb:
rebased it and resolved the whitespace issues

parrt:
thanks!

neon12345:
It seems some special chars are not handled/translated? &amp; &quot; &apos;

Issue #1855, opened 2017-05-04
Support passing std::unique_ptr through antlrcpp::Any


Griffon26:
The C++ runtime for Antlr4.7 uses a class `Any` as return type of visitor methods. This type has copy constructors that require the contained type to be copyable. `std::unique_ptr` obviously doesn't satisfy this requirement.

The following fragment demonstrates this. It won't compile giving an error about a copy constructor of `std::unique_ptr` being used from the instantiation of the `clone()` function of the `Derived<T>` class in `Any.h`:
```
#include <memory>
#include "antlr4-runtime.h"

int main()
{
  std::unique_ptr<int *> a;
  antlrcpp::Any b(std::move(a));
  return 0;
}
```

As an experiment I explicitly deleted the copy constructors from the `Any` class and removed the `clone()` functions. I then also modified the `AbstractParseTreeVisitor::aggregateResult()` function to be:
```
    virtual antlrcpp::Any aggregateResult(antlrcpp::Any &/*aggregate*/, antlrcpp::Any &nextResult) {
      return std::move(nextResult);
    }
```

This function was the only one requiring copy constructors of `Any`. With these modifications my code is able to store objects containing `std::unique_ptr`s in `Any` instances.

My idea is that the purpose of `Any` in visitors is to pass ownership of an arbitrary object created in one function up to its caller. `std::unique_ptr` was specifically created to support this transfer of ownership. 

I would therefore like to request the possibility to store objects containing `std::unique_ptr`s in the `Any` class (possibly by making the modifications I showed above).

Griffon26:
I have created a fork at https://github.com/Griffon26/antlr4 with the proposed changes on both the master branch and the 4.7-unique_ptr branch. I think these changes should work just as well with copyable types, so I invite anyone to try it out with their existing code.

Griffon26:
See pull request #1919

PR #1919, opened 2017-06-16
C++ runtime: Add support for non-copyable objects to the Any and visitor classes

Griffon26:
These changes make it possible to use std::unique_ptr to manage ownership of objects created while parsing. I think it works just as well with copyable types, but I would appreciate anyone trying this with some real-world examples.

I tried to run the runtime tests, but that failed with this message and I don't have the Java knowledge to fix this: "Could not find artifact org.antlr:antlr4-maven-plugin:jar:4.7.1-SNAPSHOT"

### DIFF
--- a/contributors.txt
+++ b/contributors.txt
@@ -154,3 +154,4 @@ YYYY/MM/DD, github id, Full name, email
 2017/07/11, dhalperi, Daniel Halperin, daniel@halper.in
 2017/07/27, shirou, WAKAYAMA Shirou, shirou.faw@gmail.com
 2017/07/09, neatnerd, Mike Arshinskiy, neatnerd@users.noreply.github.com
+2017/06/16, Griffon26, Maurice van der Pot, griffon26@kfk4ever.com
--- a/runtime/Cpp/runtime/src/support/Any.h
+++ b/runtime/Cpp/runtime/src/support/Any.h
@@ -24,21 +24,17 @@ struct ANTLR4CPP_PUBLIC Any
   bool isNull() const { return _ptr == nullptr; }
   bool isNotNull() const { return _ptr != nullptr; }
 
-  Any() : _ptr(nullptr) {
-  }
+  Any() = delete;
 
-  Any(Any& that) : _ptr(that.clone()) {
-  }
+  Any(Any& that) = delete;
 
   Any(Any&& that) : _ptr(that._ptr) {
     that._ptr = nullptr;
   }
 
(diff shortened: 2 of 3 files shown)


Griffon26:
Ok then let's figure out systematically where we misunderstand each other, because I think the use case of wrapping a unique_ptr in the Any class is very valid and it is not currently possible.

Do you understand and agree with the usefulness of:
* constructing objects in the visitor class methods?
* using unique_ptrs to hold the ownership of these objects?
* getting unique_ptrs out of the Any objects that are returned from visit methods?

mike-lischke:
Using shared/unique pointers is one use case, yes. That doesn't mean you have to disallow copying an Any instance. Of course you cannot copy such an instance if it holds a unique pointer (which is the same as with say a `std::vectory<std::unique_ptr>`, but there can be other use cases where this is perfectly fine.

For getting out a unique pointer from `Any` you will have to move it in your code, not in `Any`. Then you also know your `Any` instance is no longer valid, which would be surprise for anyone else if that happened always.

Griffon26:
Ok, I agree with you that disallowing copying is not required.

However this patch also changes the aggregateResult method. As I described in issue #1855, it requires copy constructors and is a problem for unique_ptr.

mike-lischke:
Hmm, looks like we need a better tree visitor implementation. For instance go back to a templated class and get rid of the abstract base class (to avoid the need for templated virtual functions). What do you think?

Griffon26:
I don't know enough about this to see all consequences of such a change, but I do have one question. Would that mean that it requires the class hierarchy of constructed objects to mirror the grammar? Currently I have some code that checks if an instance of Any contains one of several classes and those classes do not share a base class.

mike-lischke:
Yes, a template would limit a specific walker to the class it is configured to. But that doesn't mean you cannot use multple unrelated classes in an evaluation. You could for instance use a common context class in the walker which in turn contains whatever you want (including your own Any implementation, if you want that).

But whatever we do with the code, this PR can be closed now.

Griffon26:
Closed? The visitor still doesn't support non-copyable objects.

mike-lischke:
True, but the PR doesn't provide a good solution, so why keep it open? You can always create a new PR if you have a better idea how to solve this, but I think this is something that's better solved in users code, not in the runtime.

Griffon26:
Ok, then let's close this pull request and keep the issue open.

PR #1921, opened 2017-06-19
Fix input mismatch errors not reported the same way as Java in Python

alimg:
I'm getting different error messages from Java and Python runtimes for the same grammar. Figured out a check from the Java version was missing in the Python one so here's it

### DIFF
--- a/contributors.txt
+++ b/contributors.txt
@@ -148,3 +148,4 @@ YYYY/MM/DD, github id, Full name, email
 2017/05/11, jimallman, Jim Allman, jim@ibang.com
 2017/05/26, waf, Will Fuqua, wafuqua@gmail.com
 2017/05/29, kosak, Corey Kosak, kosak@kosak.com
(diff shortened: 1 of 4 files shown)


ericvergnaud:
Hi,
while you're at it, and since you have a grammar to reproduce the issue, could you add a test case in:
runtime-testsuite/test/org/antlr/v4/test/runtime/descriptors/ParserErrorsDescriptors.java
Eric

alimg:
Hi Eric,
Added a test case which can reproduce the issue. Looks like we didn't notice before since the problems occurs when there are more terminal  symbols defined.

ericvergnaud:
Great!
Not sure what's wrong with the cpp tests, but clearly unrelated to this.
@parrt blessed

parrt:
@alimg you need to sign contributors.txt

alimg:
@parrt updated

PR #1926, opened 2017-06-25
Fix #1925

### LINKED ISSUE #1925: Python2 target cannot recognize unicode
I have a `ID` terminator

    ID      : ([A-Z_]|'\u0100'..'\uFFFE') ([A-Z_0-9]|'\u0100'..'\uFFFE')*;

and a .txt sample file to parse

    均60:=MA(C,60);

I generated Java and Python2 target and test each against sample file respectively. Java target can parse this file correctly. But Python2 target can't. It throws `token recognition error at: '均'`. And I tested Python2 target against other valid inputs, all works except which contains unicode characters. Did I miss something or python target don't support unicode parsing. I believe there is some bug in python2 target implementation.

java

    mkdir -p java
    java -jar /usr/local/lib/antlr-4.5.3-complete.jar TDX.g4 -o ./java
    cd ./java
    javac TDX*.java
    java org.antlr.v4.gui.TestRig TDX prog -gui ../samples/1.txt

python target generating command

    java -jar /usr/local/lib/antlr-4.5.3-complete.jar -Dlanguage=Python2 TDX.g4 -o ./tdx_py/antlrgen -visitor

python code

    import sys
    from antlr4 import *
    from tdx_py.antlrgen import TDXLexer, TDXParser

    def executefile(file):
        input = FileStream(file, encoding='utf-8')
        lexer = TDXLexer(input)
        parser = TDXParser(stream)
        tree = parser.prog()

    if __name__ == '__main__':
        executefile(sys.argv[1])

I found this bug in 4.5.3 and in version 4.7

ericvergnaud:
add test to check damage

### DIFF
--- a/runtime-testsuite/test/org/antlr/v4/test/runtime/descriptors/LexerExecDescriptors.java
+++ b/runtime-testsuite/test/org/antlr/v4/test/runtime/descriptors/LexerExecDescriptors.java
@@ -73,6 +73,31 @@ public class LexerExecDescriptors {
 
 	}
 
+	/* regression test for antlr/antlr4#1925 */
+	public static class UnicodeCharSet extends BaseLexerTestDescriptor {
+		public String input = "均";
+		/**
+		 ID
+		 [@0,0:0='均',<1>,1:0]
+		 [@1,1:0='<EOF>',<-1>,1:1]
+		 */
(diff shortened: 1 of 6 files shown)


ericvergnaud:
Fixes #1925

PR #1934, opened 2017-06-30
Support for osgi bundles

catull:
The distribution form is a still a jar file, which can be used as before.
The main advantage of a bundle is that it can be used way more easier in an OSGi container.

This PR introduces support for OSGi bundle packaging.

### DIFF
--- a/contributors.txt
+++ b/contributors.txt
@@ -151,3 +151,4 @@ YYYY/MM/DD, github id, Full name, email
 2017/06/11, erikbra, Erik A. Brandstadmoen, erik@brandstadmoen.net
 2017/06/10, jm-mikkelsen, Jan Martin Mikkelsen, janm@transactionware.com
 2017/06/25, alimg, Alim Gökkaya, alim.gokkaya@gmail.com
+2017/06/30, catull, Carlo Dapor, catull@gmail.com
--- a/pom.xml
+++ b/pom.xml
@@ -120,6 +120,18 @@
 	</scm>
 
 	<build>
+		<plugins>
+				<plugin>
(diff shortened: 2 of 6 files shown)


sharwell:
I thought we enabled OSGi bundle output a long time ago through a different build step?

catull:
@sharwell You are right, I downloaded antlr 4.5.1 (we have to use that one) and 4.7.
Both versions have the bundle meta-information generated with an older version of the bundler tool.
The one used in antrl is Bnd-2.4.1.201501161923.
A more recent version would be: Bnd-3.3.0.201609221906.

In that case, perhaps upgrading the bundler tool would be an improvement.

catull:
There is a commented line in `tool/pom.xml`, line 127; for better readability, it is rendered 1 tag per line here:

```xml
<!-- can't get this to add bundle items
<plugin>
	<groupId>org.apache.felix</groupId>
	<artifactId>maven-bundle-plugin</artifactId>
	<version>2.5.4</version>
	<executions>
		<execution>
			<id>bundle-manifest</id>
			<phase>process-classes</phase>
			<configuration>
				<instructions>
					<Bundle-SymbolicName>org.antlr.antlr4-tool</Bundle-SymbolicName>
				</instructions>
			</configuration>
			<goals>
				<goal>manifest</goal>
			</goals>
		</execution>
	</executions>
</plugin> -->
```

The ball is back in your court now.

parrt:
Hmm...I am ignorant in the ways of bundles but I'm afraid to make wide sweeping maven changes. @sharwell  does this look okay to you?

PR #1935, opened 2017-07-02
Build fixes after previous C++ patch

mike-lischke:
This patch is for the C++ target only and contains fixes for compilation problems in VS and XCode after merging the last C++ patch. There is no functional change.

### DIFF
--- a/runtime/Cpp/runtime/src/Recognizer.h
+++ b/runtime/Cpp/runtime/src/Recognizer.h
@@ -11,7 +11,7 @@ namespace antlr4 {
 
   class ANTLR4CPP_PUBLIC Recognizer {
   public:
-    static const size_t EOF = (size_t)-1; // std::numeric_limits<size_t>::max(); doesn't work in VS 2013.
+    static const size_t EOF = static_cast<size_t>(-1); // std::numeric_limits<size_t>::max(); doesn't work in VS 2013.
 
     Recognizer();
     Recognizer(Recognizer const&) = delete;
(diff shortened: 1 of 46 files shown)


runtime/Cpp/runtime/src/Recognizer.h
```diff
@@ -11,7 +11,7 @@ namespace antlr4 {
 
   class ANTLR4CPP_PUBLIC Recognizer {
   public:
-    static const size_t EOF = std::numeric_limits<size_t>::max();
+    static const size_t EOF = (size_t)-1; // std::numeric_limits<size_t>::max(); doesn't work in VS 2013.
```
jm-mikkelsen: Would it be possible to use `static_cast<size_t>(-1)` as an alternative? (here are elsewhere where there are old-style casts in this patch.)

Also: there is the `INVALID_INDEX` #define at the end of antlr-common.h, which uses the same construct. Does VS2013 choke on use of that as well?

mike-lischke: Certainly doable, I just like to avoid lengthy cast expressions in simple cases like this (it's probably optimized away by the compiler).

jm-mikkelsen: Neither type of cast will be have any runtime overhead. The reason I prefer static_cast over a C-style cast is that the compiler enforces that the cast is safe. That way, when examining code, you only need to look for occurrences of reinterpret_cast to check for potential unsafe casting. This is also a rule on one of Scott Meyer's "Effective C++" books.

mike-lischke:
@parrt Please merge this C++-only patch. People try to build from master and it fails due do missing project settings on Windows.

PR #1938, opened 2017-07-07
[Go] Add getter method to Interval.

shirou:
When I am creating custom `TokenStream`, `GetTextFromInterval` method needs start and stop of `Interval`. However, `Interval` does not export these values.

This PR addes getter and `Contains`  to `Interval` in order to implement custom TokenStream.

### DIFF
--- a/runtime/Go/antlr/interval_set.go
+++ b/runtime/Go/antlr/interval_set.go
@@ -10,41 +10,33 @@ import (
 )
 
 type Interval struct {
-	start int
-	stop  int
+	Start int
+	Stop  int
 }
 
 /* stop is not included! */
 func NewInterval(start, stop int) *Interval {
 	i := new(Interval)
 
-	i.start = start
-	i.stop = stop
+	i.Start = start
+	i.Stop = stop
 	return i
 }
 
-func (i *Interval) GetStart() int {
-	return i.start
-}
-
-func (i *Interval) GetStop() int {
-	return i.stop
-}
-
 func (i *Interval) Contains(item int) bool {
-	return item >= i.start && item < i.stop
+	return item >= i.Start && item < i.Stop
 }
 
 func (i *Interval) String() string {
-	if i.start == i.stop-1 {
-		return strconv.Itoa(i.start)
+	if i.Start == i.Stop-1 {
+		return strconv.Itoa(i.Start)
 	}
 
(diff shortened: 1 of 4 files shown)


runtime/Go/antlr/interval_set.go
```diff
@@ -23,6 +23,18 @@ func NewInterval(start, stop int) *Interval {
...
+func (i *Interval) GetStart() int {
+	return i.start
+}
+
+func (i *Interval) GetStop() int {
+	return i.stop
+}
+
+func (i *Interval) Contains(item int) bool {
+	return i.contains(item)
```
davesisson: Better to update the code calling contains() to call Contains() instead of having two versions.  Since contains() is internal to the package it should be rather easy to do.

shirou: Thank you comment. Will update soon.

shirou:
@davesisson I updated this PR. Would you check it again? Thank you.

runtime/Go/antlr/interval_set.go
```diff
@@ -23,7 +23,15 @@ func NewInterval(start, stop int) *Interval {
 	return i
 }
 
-func (i *Interval) contains(item int) bool {
+func (i *Interval) GetStart() int {
```
davesisson: Go style doesn't use the word Get in getters.  This would just be Start().

Looking at the implementation of this class in Java I see that the start/stop are just called a and b (and are public and not final).  Since we aren't using subclasses for this class we could probably follow suit and expose these values directly.

http://www.antlr.org/api/Java/org/antlr/v4/runtime/misc/Interval.html

shirou: I agree about Go style doesn't use `Get`, I just respect consistency among other codes. And other way is just expose fields, as you said.

So,

1. Change to `Start()` and `Stop()`
2. Expose directory (change `start int` to `Start int` in `Interval` struct

Which is better for this project? I will follow your comment. (personally, 2 is prefer for me)

davesisson:
I like exposing the variables because it matches the Java implementation
(and avoids having Get in the name).  But I'm not the final authority here.

shirou:
@davesisson and @parrt, I have updated this PR to expose variables same as Java implementation. Would you check it out? Thank you.

shirou:
@parrt ping. If you have any concern, please let me know. Thank you.

shirou:
Thank you!!

PR #1939, opened 2017-07-07
Fix expressions that have no effect

robertbrignull:
I spotted a couple of places where the comma operator was being used, but possibly wrongly as there were expressions which had no effect. See https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Comma_Operator

In the case of `ATNDeserializer.js` it looked like it was something like text substitution in to the log message, but in fact the whole first two parts were discarded and it would just print `SERIALIZED_UUID`. By removing the last two parts and the commas it'll actually print the whole mesage.

In the case of `Errors.js` I'm not sure what was meant to happen but this change in this PR is entirely behaviour preserving so hopefully no problem.

As a bonus, this will fix a few alerts on [lgtm.com](lgtm.com). See [ATNDeserializer.js](https://lgtm.com/projects/g/antlr/antlr4/snapshot/dist-1787830766-1498572200128/files/runtime/JavaScript/src/antlr4/atn/ATNDeserializer.js#V160) and [Errors.js](https://lgtm.com/projects/g/antlr/antlr4/snapshot/dist-1787830766-1498572200128/files/runtime/JavaScript/src/antlr4/error/Errors.js#V81).

### DIFF
--- a/runtime/JavaScript/src/antlr4/atn/ATNDeserializer.js
+++ b/runtime/JavaScript/src/antlr4/atn/ATNDeserializer.js
@@ -158,7 +158,7 @@ ATNDeserializer.prototype.checkUUID = function() {
     var uuid = this.readUUID();
     if (SUPPORTED_UUIDS.indexOf(uuid)<0) {
         throw ("Could not deserialize ATN with UUID: " + uuid +
-                        " (expected " + SERIALIZED_UUID + " or a legacy UUID).", uuid, SERIALIZED_UUID);
+                        " (expected " + SERIALIZED_UUID + " or a legacy UUID).");
     }
     this.uuid = uuid;
 };
--- a/runtime/JavaScript/src/antlr4/error/Errors.js
+++ b/runtime/JavaScript/src/antlr4/error/Errors.js
@@ -78,7 +78,7 @@ LexerNoViableAltException.prototype.constructor = LexerNoViableAltException;
 LexerNoViableAltException.prototype.toString = function() {
     var symbol = "";
     if (this.startIndex >= 0 && this.startIndex < this.input.size) {
-        symbol = this.input.getText((this.startIndex,this.startIndex));
+        symbol = this.input.getText(this.startIndex);
     }
     return "LexerNoViableAltException" + symbol;
 };


parrt:
@ericvergnaud is this blessed?

parrt:
ping @ericvergnaud

runtime/JavaScript/src/antlr4/atn/ATNDeserializer.js
```diff
@@ -158,7 +158,7 @@ ATNDeserializer.prototype.checkUUID = function() {
     var uuid = this.readUUID();
     if (SUPPORTED_UUIDS.indexOf(uuid)<0) {
         throw ("Could not deserialize ATN with UUID: " + uuid +
-                        " (expected " + SERIALIZED_UUID + " or a legacy UUID).", uuid, SERIALIZED_UUID);
+                        " (expected " + SERIALIZED_UUID + " or a legacy UUID).");
```
ericvergnaud: ok

runtime/JavaScript/src/antlr4/error/Errors.js
```diff
@@ -78,7 +78,7 @@ LexerNoViableAltException.prototype.constructor = LexerNoViableAltException;
 LexerNoViableAltException.prototype.toString = function() {
     var symbol = "";
     if (this.startIndex >= 0 && this.startIndex < this.input.size) {
-        symbol = this.input.getText((this.startIndex,this.startIndex));
+        symbol = this.input.getText(this.startIndex);
```
ericvergnaud: should this not be: get Text(new Interval(...

robertbrignull: Could you explain why it should be `getText(new Interval(...))`? I'm having trouble following the code but I think it should be `getText(this.startIndex, this.startIndex)`.

The only two places that define a `getText` method that I could find are `InputStream` and `BufferedTokenStream`. Of those two the former takes a start and end as separate parameters, and the latter takes an interval.

This function we're modifying is on `LexerNoViableAltException` and one of those is only constructed from `LexerATNSimulator`. In this class every function received an `input` parameter and they seem to be `InputStream`s as later on the `getText` version with two arguments is used (unless that call is wrong). I'm thinking it makes sense as well that you can't have a token stream yet as the lexer is in the process of making one and is if this exception is thrown then it failed.

ericvergnaud:
@robertbrignull would you mind checking this proposed change against 4.9 ?

robertbrignull:
@ericvergnaud, I'm afraid I'd completely forgotten about this change. Looking at `master` both places that this PR is changing appear to be the same, so the changes made here should still be valid. However apparently at some point I deleted my fork, so I'm not sure it's possible to salvage this PR. GitHub won't let me fix the conflicts in the UI unfortunately and I don't know if recreating the fork would work or not. To be honest I'm just going to close this PR.

PR #1942, opened 2017-07-09
[Go] Implemented TokenStreamRewriter

neatnerd:
TokenStreamRewriter implementation was missing
Ported code from Java version; however, there are couple of deviations due to difference between composition (Go) and inheritance (Java) concepts
Ported tests from Swift for LexerA

### DIFF
--- a/runtime/Go/antlr/tokenstream_rewriter.go
+++ b/runtime/Go/antlr/tokenstream_rewriter.go
@@ -9,11 +9,10 @@ import (
 )
 
 
-// Useful for rewriting out a buffered input token stream after doing some
-//
+// 
 // Useful for rewriting out a buffered input token stream after doing some
 // augmentation or other manipulations on it.
-//
+
 // <p>
 // You can insert stuff, replace, and delete chunks. Note that the operations
 // are done lazily--only if you convert the buffer to a {@link String} with
@@ -24,31 +23,31 @@ import (
 // operation is done and then normal {@link String} rendering continues on the
 // buffer. This is like having multiple Turing machine instruction streams
 // (programs) operating on a single input tape. :)</p>
-//
 // <p>
+
 // This rewriter makes no modifications to the token stream. It does not ask the
 // stream to fill itself up nor does it advance the input cursor. The token
 // stream {@link TokenStream#index()} will return the same value before and
 // after any {@link #getText()} call.</p>
-//
+
 // <p>
 // The rewriter only works on tokens that you have in the buffer and ignores the
 // current input cursor. If you are buffering tokens on-demand, calling
 // {@link #getText()} halfway through the input will only do rewrites for those
 // tokens in the first half of the file.</p>
-//
+
 // <p>
 // Since the operations are done lazily at {@link #getText}-time, operations do
 // not screw up the token index values. That is, an insert operation at token
 // index {@code i} does not change the index values for tokens
 // {@code i}+1..n-1.</p>
-//
+
 // <p>
 // Because operations never actually alter the buffer, you may always get the
 // original token stream back without undoing anything. Since the instructions
 // are queued up, you can easily simulate transactions and roll back any changes
 // if there is an error just by removing instructions. For example,</p>
-//
+
 // <pre>
 // CharStream input = new ANTLRFileStream("input");
 // TLexer lex = new TLexer(input);
@@ -57,10 +56,10 @@ import (
 // TokenStreamRewriter rewriter = new TokenStreamRewriter(tokens);
 // parser.startRule();
 // </pre>
-//
+
 // <p>
 // Then in the rules, you can execute (assuming rewriter is visible):</p>
-//
+
 // <pre>
 // Token t,u;
 // ...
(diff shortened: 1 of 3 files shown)


runtime/Go/antlr/tokenstream_rewriter.go
```diff
@@ -0,0 +1,722 @@
...
+// can be found in the LICENSE.txt file in the project root.
+package antlr
+
+import (
+"bytes"
+"fmt"
+)
+
+
+// Useful for rewriting out a buffered input token stream after doing some
```
davesisson: duplicate comment line

runtime/Go/antlr/tokenstream_rewriter.go
```diff
@@ -0,0 +1,722 @@
...
+// tokens in the first half of the file.</p>
+//
+// <p>
+// Since the operations are done lazily at {@link #getText}-time, operations do
+// not screw up the token index values. That is, an insert operation at token
+// index {@code i} does not change the index values for tokens
+// {@code i}+1..n-1.</p>
+//
+// <p>
+// Because operations never actually alter the buffer, you may always get the
```
davesisson: Line 47 has this same text.

runtime/Go/antlr/tokenstream_rewriter.go
```diff
@@ -0,0 +1,722 @@
...
+// System.out.println(rewriter.getText("pass2"));
+// </pre>
+//
+// <p>
+// If you don't use named rewrite streams, a "default" stream is used as the
+// first example shows.</p>
+// /augmentation or other manipulations on it.
+//
+// <p>
+// You can insert stuff, replace, and delete chunks. Note that the operations
```
davesisson: This looks like line 18.

runtime/Go/antlr/tokenstream_rewriter.go
```diff
@@ -0,0 +1,722 @@
...
+
+// Define the rewrite operation hierarchy
+
+type RewriteOperation interface {
+	// Execute the rewrite operation by possibly adding to the buffer.
+	// Return the index of the next token to operate on.
+	Execute(buffer *bytes.Buffer)	int
+	String()						string
+	GetInstructionIndex()			int
+	GetIndex()						int
```
davesisson: Just making these public is probably easier to use.  The added functionality of explicit getters/setters is overkill.

But I'll let @parrtt comment on whether there might be a use case for these to be overridden in the future.

runtime/Go/antlr/tokenstream_rewriter.go
```diff
@@ -0,0 +1,722 @@
...
+	}
+}
+
+func (tsr *TokenStreamRewriter) GetTokenStream() TokenStream{
+	return tsr.tokens
+}
+
+//  Rollback the instruction stream for a program so that
+//  the indicated instruction (via instructionIndex) is no
+//  longer in the stream. UNTESTED!
```
davesisson: Please write a test for this.  :)

runtime/Go/antlr/tokenstream_rewriter.go
neatnerd: thx, nice catch!

runtime/Go/antlr/tokenstream_rewriter.go
neatnerd: the problem was that I needed to make an interface and as far as I know i cannot put fields into interface, thats why I went with methods

neatnerd: the problem was that I needed to make an interface and as far as I know i cannot put fields into interface, thats why I went with methods

runtime/Go/antlr/tokenstream_rewriter.go
neatnerd: I took it from the Java source code, it is also untested there and in Swift implementation. In general, I must admit that runtime is not really tested

runtime/Go/antlr/tokenstream_rewriter.go
davesisson: Good point.  I've always disliked that bit about interfaces.  Leaving it as is fine.

An alternative approach to consider would be to allow construction of the underlying interface implementation with a New() method of some sort (which would eliminate the setters at least).  But given that it can't kill the getters I'm not sure it's worth the effort.  I have no strong opinion here.

neatnerd: Nah, the constructor is not gonna cut, Go composition vs Java inheritance. Thats the max you can get away with imho

runtime/Go/antlr/tokenstream_rewriter.go
neatnerd: I am just porting code, Java version - 0 test for this class, python - 0, C++ - 0. I did more than required for porting imho, maybe somebody else will be willing to venture to this untested zone

runtime/Go/antlr/tokenstream_rewriter_test.go
```diff
@@ -0,0 +1,392 @@
...
+
+var lexerRuleNames = []string{
+	"A", "B", "C",
+}
+
+type LexerA struct {
+	*BaseLexer
+	channelNames []string
+	modeNames    []string
+	// TODO: EOF string
```
davesisson: This TODO (and the other) is not actionable (i.e. it's not clear what needs to be done to finish whatever needs to be done).

neatnerd: This comes from code gen file right [here](https://github.com/antlr/antlr4/blob/master/tool/resources/org/antlr/v4/tool/templates/codegen/Go/Go.stg#L1418)

davesisson:
Ah, good point.  OK, I have no further issues with this particular
implementation.  LGTM

neatnerd:
Ok, thanks @davesisson. @parrt can you, please, merge?

PR #1945, opened 2017-07-12
Many fixes to antlr4-maven-plugin dependency analysis

dhalperi:
We were having some issues with getting grammars to recompile with changed imports. Found a few things, commit by commit.

1. The existing tests were bogus. Because the test grammars were edited and then restored in a try-with-resources, but then not actually recompiled after being restored, there were false-positives. Specifically the nth variant in the test suite could think that the checksum changed, but it might have changed because of undoing the prior test rather than the current test.
    So I fixed this by rerunning state after every try-with-resources and then asserting that the checksums were equal to how they were the first time.

2. This exposed bugs in the lexer-changes test. First, we were adding a newline to the file which did not change the output checksum. Second, the timestamp modification check (at least on my computer) failed because it was only second-level granularity. Fix both these issues in the next two commits.

3. For some reason, the dependency analysis was only checking the first import, or assuming only one imported grammar per import statement. It should be checking all imports!

4. Add tests of the latter.

and of course, add myself to the contributors.

This builds on the really nice work by @marcohu in #1353. Maybe you want to take a look and sanity check me / add color? In any case, thanks for starting this off!

Please take a look and let me know what I can improve!

Thanks! @parrt


parrt:
Thanks, I'll get to this as soon as I can. I started teaching a 5-week course this week :( for which I've been prepping like crazy.

dhalperi:
Hi @parrt – any chance you'll have time to look at this? I'd not like to lose context when you come back asking for changes :)

parrt:
As I am not a maven expert, @marcohu could you take a look at these changes by @dhalperi ? thanks!

dhalperi:
@marcohu @parrt any updates?

parrt:
hmm...no response. let me take another peek

parrt:
Ok, seems harmless, and useful if your comments are correct. haha. thanks!

parrt:
Hiya @dhalperi Can you check the build at travis-ci.com? it appears this PR broke the dotnet mac builds. weird. thanks!

dhalperi:
Hi @parrt ,

Can you send a link to an example failure? I did see this in one of the runs on `master`:

```
 java -Xmx32m -version
java version "1.8.0_112"
Java(TM) SE Runtime Environment (build 1.8.0_112-b16)
Java HotSpot(TM) 64-Bit Server VM (build 25.112-b16, mixed mode)
$ javac -J-Xmx32m -version
javac 1.8.0_112
51.25s$ ./.travis/before-install-$TRAVIS_OS_NAME-$TARGET.sh
error: RPC failed; curl 56 SSLRead() return error -36
fatal: The remote end hung up unexpectedly
fatal: early EOF
fatal: index-pack failed
Error: Fetching /usr/local/Homebrew/Library/Taps/homebrew/homebrew-core failed!
Updated 1 tap (caskroom/cask).
The command "./.travis/before-install-$TRAVIS_OS_NAME-$TARGET.sh" failed and exited with 1 during .
```

but that is clearly some sort of network error.

Do you have additional Travis runs that I can't see? (I can only access travis-ci.org, not travis-ci.com)

parrt:
hi. check this one https://travis-ci.org/antlr/antlr4/builds/257836911 I think it is the build for merge of your stuff. Hmm...error doesn't seem related but maybe. for dotnet says: "Error: openssl 1.0.2j is already installed" etc...

parrt:
hmm...maybe this is spurious travis crap again

dhalperi:
You can rerun an individual build on travis, if you're logged in and a repository owner. That often helps to check if errors are transient, assuming it's not a flaky unit test (which this does not seem to be).

parrt:
it does seem travis is flaky. ok, thanks!

BTW, @dhalperi is it reasonable to move the antlr4 mojo to a separate repo under antlr org do you think?

PR #1947, opened 2017-07-13
Local Docker development workflow

DHager:
The purpose of this PR is to make it easier for contributors to get started with building, testing, and debugging ANTLR. It contains instructions and scripts for initializing and running a local Docker container that is capable of:

* Reflecting user edits as they occur
* Executing Maven tasks and caching JAR/POM resources
* Running core tests
* Running the compiled ANTLR-Tool against a grammar
* Interactive debugging with Java IDEs

Out of scope:

* Additional Docker images for specific runtimes. (Eventually, perhaps.)
* Using Docker in a way which competes with existing Travis/CI tooling. The emphasis here is purely on local and offline development.


DHager:
I'd be interested if anyone has a better way to handle the initial Maven download. Right now it's a little bit on the kludgy side -- it's tied to a specific Apache mirror.

DHager:
Hi @ericvergnaud , would this fall into your area-of-interest for the project?

I could try to make some Docker scaffolding that runs Travis-CI itself locally, but my feeling is that would slow the debug/test loop and introduce more potential for things to fail in ways an average contributor can't easily diagnose.

ericvergnaud:
Not sure there is much value in running Travis within docker.
Rather, if you manage to set up a complete docker image with all prerequisites setup (Java 8, mono, nodejs...) and it just fetches the latest antlr code I would expect that to actually run faster and be very useful to newbies


Envoyé de mon iPhone

parrt:
Seems like we should just update the documentation. is there something other than what we have that would make things easier for contributors to get started?

DHager:
@parrt I feel that a docker image has several benefits:

1. It lowers the barrier-to-entry for new contributors.
1. Creates a *locally*-reproducible and consistent environment for building and testing.
1. Avoids potential conflict or inconvenience for users who are on other operating-systems or who don't want to endanger a fragile desktop environment they use for other projects.

parrt:
@DHager well,I think I would prefer to keep it as we have now by fixing whatever documentation is misleading. It should be fairly straightforward. I'd rather not make my life more complex by introducing a new concept and tool (docker).

DHager:
@parrt I understand and agree that it represents an extra step for anybody who focuses primarily on ANTLR development, since they will already have everything set up just the way they like it.

Ultimately I think this comes down to a non-technical decision about how you see the future of volunteer work on the project, and how you want to optimize for existing contributors versus lowering the bar for potential new ones.

parrt:
ok, yeah, i guess I'll make the command decision to close for now. thank you for your potential contribution however.  I will keep this around actually for future reference and might reactivate.

Issue #42, opened 2012-03-12
Explicitly resolving if/if/else ambiguity in a grammar


sharwell:
I'm working with an ANTLR 4 grammar to correct any construct which results in a call to `reportAmbiguity`. In many cases this is done by refactoring or combining rules, but the case for an `if`/`if`/`else` construct is proving more challenging.

Starting with the following grammar:

```
grammar T;
stmt : ifStmt | ID;
ifStmt : 'if' ID stmt ('else' stmt)?;

ID : [a-zA-Z]+;
WS : ' '+ -> skip;
```

Given the input `if x if x a else b`, I'm looking for a way to modify the grammar such that no call to `reportAmbiguity` is needed, while following the standard `if`/`else` rule of binding to the closest unmatched `if` statement.

In ANTLR 3, a syntactic predicate allowed the user to explicitly handle this situation. In ANTLR 4, the correct decision is made through min-alt resolution, but it always wants to emit a "warning" regarding the situation.

gkudva:
Hi,
   I have a grammar with if/elseif/else statements (C# code generation), and I've run into an issue whereby a deeply nested if/else block (24 levels deep) that parses successfully in ANTLR 4.0 (albeit very slowly), now hangs on ANTLR 4.6.1 - it ran for over 2 hours without completing before I killed it.  While trying to debug this issue, I ran across this thread and tried the sll = true option.  However, the parsers generated with and without the sll=true option are identical.  Is this option supported for C# code generation?

ericvergnaud:
Hi,
The place for support is the google discussion group
Eric
Envoyé de mon iPhone

parrt:
try 4.7?

gkudva:
OK, will post to google groups - thanks.

@parrt - Latest NuGet package for code-generator is 4.6.1 so figured I'd use 4.6.1 for runtime as well to be  in sync.  Also ANTLR Runtime 4.6.1 package was later dated 5/30/2017 than ANTLR Standard Runtime 4.7.1 published on 5/28/2017. Will try 4.7.1 and see if it works.

sharwell:
@gkudva The NuGet package (4.6.1) you referred to is part of a different repository. For issues with that (bugs or anything else) you should file an issue on GitHub here: https://github.com/tunnelvisionlabs/antlr4cs/

PR #1955, opened 2017-07-18
Avoid adding to closureBusy before all ATNConfig properties are set

sharwell:
Setting `ATNConfig` properties can change the hash code of the instance, leading to cases where the `closureBusy` set places objects in the wrong buckets. While this has not led to known cases of stack overflow, it has led to cases where one or more buckets contains a large number of duplicate objects, and the set's add operation goes from O(1) to O(n).

:link: This bug was found during investigation of a report by @gkudva in https://github.com/antlr/antlr4/issues/42#issuecomment-312299708.

### DIFF
--- a/runtime/Java/src/org/antlr/v4/runtime/atn/ParserATNSimulator.java
+++ b/runtime/Java/src/org/antlr/v4/runtime/atn/ParserATNSimulator.java
@@ -1541,11 +1541,6 @@ public class ParserATNSimulator extends ATNSimulator {
 			ATNConfig c = getEpsilonTarget(config, t, continueCollecting,
 										   depth == 0, fullCtx, treatEofAsEpsilon);
 			if ( c!=null ) {
-				if (!t.isEpsilon() && !closureBusy.add(c)) {
-					// avoid infinite recursion for EOF* and EOF+
-					continue;
-				}
-
 				int newDepth = depth;
 				if ( config.state instanceof RuleStopState) {
 					assert !fullCtx;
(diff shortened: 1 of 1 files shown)


danielsun1106:
Sam ( @sharwell ), when we profile the new parser of Apache Groovy, we find `closure` method costs a lot of time and the performance of parsing is not good. If possible, could you release a new version of your optimized antlr4(e.g. 4.7.0.1?), which includes the improvement `Avoid adding to closureBusy before all ATNConfig properties are set`.

tp1

Here are grammar files of Apache Groovy: 
https://github.com/apache/groovy/blob/master/src/antlr/GroovyLexer.g4
https://github.com/apache/groovy/blob/master/src/antlr/GroovyParser.g4

parrt:
@danielsun1106 can youGive it a try now with latest master?

sharwell:
@danielsun1106 I'll try to get a build published soon with this fix included. It didn't *seem* like it would have a big impact for your case, but it's a possibility.

danielsun1106:
@sharwell That's great! Could you please give me some advices to improve performance of the parser further? Thanks in advance!

@parrt It's hard to switch to the reference implementation of antlr4 because of different API...

parrt:
Which part of the API is different?

sharwell:
I was on a very tight schedule, but helped @danielsun1106 set up an experiment related to this change. As I expected, this change wasn't a notable contributor to the scenario. From what I can tell, the number of users who benefit from this change is relatively small, but the impact for those cases can be substantial.

I'll work with Daniel in the coming weeks to see if further optimizations can be made to the internals of the closure operation and/or improve the ways we help users refactor their grammar to improve performance when using the profiler.

parrt:
Excellent. thanks @sharwell As usual, you are the performance guru!

danielsun1106:
@parrt  Here is the commit to show the difference of API.
https://github.com/sharwell/groovy-parser/commit/27291391f3387e1aadbe6a34e287820294c6ad93

ewanmellor:
This change was ported to the Swift runtime as part of PR #2407.

PR #1956, opened 2017-07-18
Python: add public getter to parsers to access syntax error count.

renatahodovan:
ANTLR parsers in Java are allowed to access the number of encountered
syntax errors via the getNumberOfSyntaxErrors method. However, the
Python variants must use the protected _syntaxErrors member to get this
value. The patch defines the same getter for Python targets too.


ericvergnaud:
Blessed

Envoyé de mon iPhone

PR #1958, opened 2017-07-19
fix to the ATN deserialiser issue on big endian architecture

vaibhavaingankar09:
In ReadUUID `byte d[]` to be reversed only in case of little endian architecture. In the case of big endian  `Array.Reverse(d)` is not required since the bytes are stored in the order most significant to least significant.


ericvergnaud:
Out of curiosity, which system is exposing the problematic behaviour?
I'm a bit surprised that this would be the only issue with the deserialiser.

vaibhavaingankar09:
ATN deserialiser issue arises on s390x systems, with UUID 59627784-3be5-417a-**8960-28a73181ebb9** whereas the expected UUID is 59627784-3be5-417a-**b9eb-8131a7286089**. If we notice the highlighted section of UUID is the representation of 8 bit byte array which is reversed.

ericvergnaud:
Thanks
To your knowledge, does this only happen with the c# runtime?

vaibhavaingankar09:
Yes, it is specific to c# runtime.

ericvergnaud:
Thanks
@parrt blessed

PR #1966, opened 2017-07-22
Require superClass for Lexer in split mode for JavaScript target

venkatperi:
* In split mode grammars, the generated lexer.js wasn't requiring the superclass.

* Lexer ids such as token/mode/channel names were set on the generated lexer and not its prototype making them class/static variables. For grammar actions to be universal across Java and JavaScript, these names must be accessible on the lexer instance (via [this.]SOME_MODE in Java, or this.SOME_MODE in JavaScript since JS doesn't scope to the local class without this.).

* Finally, `Utils.escapeWhiteSpace` was replacing only the first instance of `\n` etc. Switched to a regex with global replace.


runtime/JavaScript/src/antlr4/Utils.js
```diff
@@ -401,11 +401,11 @@ DoubleDict.prototype.set = function (a, b, o) {
...
-    s = s.replace("\r", "\\r");
+    s = s.replace(/\t/g, "\\t")
+         .replace(/\n/g, "\\n")
+         .replace(/\r/g, "\\r");
     if (escapeSpaces) {
-        s = s.replace(" ", "\u00B7");
+        s = s.replace(/ /g, "\u00B7");
     }
     return s;
 }
```
ericvergnaud: ok with these

tool/resources/org/antlr/v4/tool/templates/codegen/JavaScript/JavaScript.stg
```diff
@@ -160,6 +160,7 @@ Object.defineProperty(<parser.name>.prototype, "atn", {
 });
 
 <parser.name>.EOF = antlr4.Token.EOF;
+<parser.name>.prototype.EOF = antlr4.Token.EOF;
```
ericvergnaud: sorry but this is superfluous

tool/resources/org/antlr/v4/tool/templates/codegen/JavaScript/JavaScript.stg
```diff
@@ -802,6 +803,9 @@ var antlr4 = require('antlr4/index');
 >>
 
 Lexer(lexer, atn, actionFuncs, sempredFuncs, superClass) ::= <<
+<if(superClass)>
+var <superClass> = require('./<superClass>').<superClass>;
+<endif>
```
ericvergnaud: ok

tool/resources/org/antlr/v4/tool/templates/codegen/JavaScript/JavaScript.stg
```diff
@@ -819,14 +823,18 @@ function <lexer.name>(input) {
 <lexer.name>.prototype.constructor = <lexer.name>;
 
 <lexer.name>.EOF = antlr4.Token.EOF;
+<lexer.name>.prototype.EOF = antlr4.Token.EOF;
```
ericvergnaud: superfluous

tool/resources/org/antlr/v4/tool/templates/codegen/JavaScript/JavaScript.stg
```diff
@@ -819,14 +823,18 @@ function <lexer.name>(input) {
...
 <lexer.tokens:{k | <lexer.name>.<k> = <lexer.tokens.(k)>;}; separator="\n", wrap, anchor>
+<lexer.tokens:{k | <lexer.name>.prototype.<k> = <lexer.tokens.(k)>;}; separator="\n", wrap, anchor>
```
ericvergnaud: superfluous

tool/resources/org/antlr/v4/tool/templates/codegen/JavaScript/JavaScript.stg
```diff
@@ -819,14 +823,18 @@ function <lexer.name>(input) {
...
 
 <if(lexer.channels)>
 <lexer.channels:{c| <lexer.name>.<c> = <lexer.channels.(c)>;}; separator="\n">
+<lexer.channels:{c| <lexer.name>.prototype.<c> = <lexer.channels.(c)>;}; separator="\n">
```
ericvergnaud: superfluous

tool/resources/org/antlr/v4/tool/templates/codegen/JavaScript/JavaScript.stg
```diff
@@ -819,14 +823,18 @@ function <lexer.name>(input) {
...
 
 <endif>
 <if(rest(lexer.modes))>
 <rest(lexer.modes):{m| <lexer.name>.<m> = <i>;}; separator="\n">
+<rest(lexer.modes):{m| <lexer.name>.prototype.<m> = <i>;}; separator="\n">
```
ericvergnaud: superfluous

ericvergnaud:
Hi,
thanks for this. 
Tokens are already members of the lexer and parser type, so should not be members of each instance. In java they are static, so should not be accessed through parser/lexer instances.
ANTLR does not support 'universal' grammar actions. Instead, you should use listener or visitor. 
Other proposals make sense.
Also, if not already done, you need to sign the license agreement.

ericvergnaud:
Hi 
I can't bleed this until you remove the prototype level stuff

ericvergnaud:
I meant bless not bleed

venkatperi:
So the downside of ids on the class vs instance is that superclasses will need to reach into the subclass (usually the generated Lexer) to access definitions. e.g. if I have a token:

```
tokens { TOKEN }
```

From my lexer's super class, I need to do this:

```
SuperLexer.prototype.someAction = function() {
  console.log(Object.getPrototypeOf(this).TOKEN);
}
```

I guess its the same with Java currrently:

```
//superclass member
  public void someAction() {
     int type = GeneratedLexer.TOKEN;
  }
```

Example: 
https://github.com/venkatperi/asciidoc-antlr/blob/master/src/main/js/AbstractAsciidocLexer.js
https://github.com/venkatperi/asciidoc-antlr/blob/master/src/main/java/AbstractAsciidocLexer.java

venkatperi:
Reversed proto changes.

ericvergnaud:
@parrt blessed

PR #1969, opened 2017-07-26
Report InputMismatchException with original context information

### LINKED ISSUE #1922: Worse expected token output in ANTLR 4.7
Using the following grammar
```
member : 'a';
body: member*;
file : body EOF;
B: 'b';/* ensure we can lex b */
```
on input `baa` 

we get `mismatched input 'b' expecting <EOF>` with antlr 4.7 but `extraneous input 'b' expecting {<EOF>, 'a'}` with antlr 4.6

The output of 4.6 is preferable as 'a' was a valid input at that position. 

The recovered parse tree is also worse with 4.6 recovering the later members but 4.7 generating an empty body and error nodes for the later members. 

This happens in general when there is a rule that is optionally empty. In this case we can get better results by inlineing `body` into `file` but that significantly increases the size of the grammar in the non reduced example and makes the java code for converting the tree more complex.

sharwell:
Fixes #1922 

:memo: This needs more test cases, particularly for LL(k) failures where k>1. However, my current understanding is these cases have been impacted by this bug for much longer (since 4.2.2) than the issue being addressed for the k=1 case.

### DIFF
--- a/runtime-testsuite/test/org/antlr/v4/test/runtime/descriptors/ParserErrorsDescriptors.java
+++ b/runtime-testsuite/test/org/antlr/v4/test/runtime/descriptors/ParserErrorsDescriptors.java
@@ -618,4 +618,28 @@ public class ParserErrorsDescriptors {
 		public String grammar;
 
 	}
+
+	public static class ExtraneousInput extends BaseParserTestDescriptor {
+		public String input = "baa";
+		public String output = null;
+		public String errors = "line 1:0 mismatched input 'b' expecting {<EOF>, 'a'}\n";
+		public String startRule = "file";
+		public String grammarName = "T";
+
+		/**
+		 grammar T;
+
+		 member : 'a';
+		 body : member*;
+		 file : body EOF;
+		 B : 'b';
+		 */
+		@CommentHasStringValue
+		public String grammar;
+
+		@Override
+		public boolean ignore(String targetName) {
+			return !"Java".equals(targetName);
+		}
+	}
 }
--- a/runtime-testsuite/test/org/antlr/v4/test/runtime/descriptors/SemPredEvalParserDescriptors.java
+++ b/runtime-testsuite/test/org/antlr/v4/test/runtime/descriptors/SemPredEvalParserDescriptors.java
@@ -283,11 +283,16 @@ public class SemPredEvalParserDescriptors {
 		public String input = "s\n\n\nx\n";
 		public String output = "(file_ (para (paraContent s) \\n \\n) (para (paraContent \\n x \\n)) <EOF>)\n";
 		/**
-		line 5:0 mismatched input '<EOF>' expecting '
-		'
+		line 5:0 mismatched input '<EOF>' expecting {'s', '
+		', 'x'}
 		 */
 		@CommentHasStringValue
 		public String errors;
+
+		@Override
+		public boolean ignore(String targetName) {
+			return !"Java".equals(targetName);
+		}
 	}
 
 	public static class PredFromAltTestedInLoopBack_2 extends PredFromAltTestedInLoopBack {
--- a/runtime/Java/src/org/antlr/v4/runtime/DefaultErrorStrategy.java
+++ b/runtime/Java/src/org/antlr/v4/runtime/DefaultErrorStrategy.java
@@ -37,6 +37,21 @@ public class DefaultErrorStrategy implements ANTLRErrorStrategy {
 	protected IntervalSet lastErrorStates;
 
 	/**
+	 * This field is used to propagate information about the lookahead following
+	 * the previous match. Since prediction prefers completing the current rule
+	 * to error recovery efforts, error reporting may occur later than the
+	 * original point where it was discoverable. The original context is used to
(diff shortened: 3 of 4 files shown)


sharwell:
@Et999 @michaelpj @PayneSun @marcospassos It should be possible to apply this change to your code by extending `DefaultErrorStrategy` and overriding `sync` and `reportInputMismatch`. Can one of you try this and let me know if it resolves the issue in your test suites? Since the `InputMismatchException` constructor used in this code is not available, you'll have to do a bit of trickery to make this work right:

```java
int priorState = recognizer.getState();
ParserRuleContext priorContext = recognizer.getContext();
try {
  recognizer.setState(nextTokensState);
  recognizer._ctx = nextTokensContext;
  throw new InputMismatchException(recognizer);
} finally {
  recognizer.setState(state);
  recognizer._ctx = priorContext;
}
```

⚠️ Note that the message does still change in some cases relative to 4.6 - you get "mismatched input" instead of "extraneous token". However, the expected set should be reporting good information again.

michaelpj:
Thanks enormously for looking into this! I'll give this a try and see how it looks.

michaelpj:
Built locally, tested. Fixes the regressions in 4.7, and actually improves the errors in several other cases. Big :+1: from me.

parrt:
@Et999 and @marcospassos can you build locally and confirm this works for you?

sharwell:
@parrt After confirmation, I'll start the work to implement this for other targets.

marcospassos:
I'll test it by tomorrow :)

parrt:
@antlr/antlr-targets please note Sam's improvements to the error reporting in this PR.

parrt:
ugh. we appear to be getting random travis fluctuations

marcospassos:
@sharwell we caught 160+ tests failing due to the BC break introduced in 4.7, even applying the patch you provided. 

The errors can be grouped into the following cases:

- `No viable alternative at <EOF>` is now `Mismatched input '<EOF>'` (no big deal)
- `Missing token ')'` is now `Mismatched input '<EOF>'` (worst error reporting)
- `Mismatched input '<EOF>'` is now `Extraneous token '<EOF>'` (really weird)

Besides this, we found several cases where error recovery got worst. Given the following grammar and invalid input, notice how the error recovery produces worst guesses:

```
root: variableDeclarationList? expression;

variableDeclarationList
    : variableDeclaration + ;

variableDeclaration
    : 'let' variableDeclarator (',' variableDeclarator)* ';' ;

variableDeclarator
    : identifier '=' expression ;

expression: ...;
```

Invalid input:
```
let a = , b = 2; true
```

In 4.6, the previously input was recognized as:
```
let a = b; true
``

While in 4.7 the same input is recognized as:

```
let a = b;  2
```

**Comment by marcospassos** [comment]
@sharwell we caught 160+ tests failing due to the BC break introduced in 4.7, even applying the patch you provided. 

The errors can be grouped into the following cases:

- `No viable alternative at <EOF>` is now `Mismatched input <EOF>` (no big deal)
- `Missing token )` is now `Mismatched input <EOF>` (worst error reporting)
- `Mismatched input <EOF>` is now `Extraneous token <EOF>` (really weird)

Besides this, we found several cases where error recovery got worst. Given the following grammar and invalid input, notice how the error recovery produces worst guesses:

```
root: variableDeclarationList? expression;

variableDeclarationList
    : variableDeclaration + ;

variableDeclaration
    : 'let' variableDeclarator (',' variableDeclarator)* ';' ;

variableDeclarator
    : identifier '=' expression ;

expression: ...;
```

Invalid input:
```
let a = , b = 2; true
```

In 4.6, the previously input was recognized as:
```
let a = b; true
```

While in 4.7 the same input is recognized as:

```
let a = b;  2
```

This difference is probably related to the same issue, right?

sharwell:
@marcospassos If you are testing error messages, I would not expect this change to reduce the number of "regressing tests", since the messages will still change. Is your project open source so I can help triage the differences?

marcospassos:
@sharwell the project is not open source. However, we're not testing only error messages, but also the generated parse tree. Why don't you consider the error message a BC break? This changes not only changed the error messages but also makes the parser produce different parse trees, which is a serious BC break.

Btw, is "Extraneous token <EOF>" supposed to be a valid error message?

sharwell:
@marcospassos I sent you an email 👍

sharwell:
After talking with @marcospassos, it appears the negative outcomes were likely influenced by uncommon characteristics of the language being parsed. In particular:

1. Input text is typically short, so error handling strategies which favor error locality to error recovery are favorable
2. Input is closer to natural language than most, so distinctive tokens that tend to yield good error recovery in many languages (e.g. the braces in Java) don't help for this one

For this case, I suggested creating a type derived from `DefaultErrorStrategy` using the implementation of `sync` prior to my change, since it yields better results for this language.

marcospassos:
Thank you @sharwell!

parrt:
Ok, sounds like we have no objections to this improvement.

parrt:
Thanks, @sharwell !!

PR #1973, opened 2017-07-28
Ignore cmake policy 0054

matthauck:


matthauck:
Prior to this, there are a slew of warnings printed out (cmake 3.3.2) that look like:

```
CMake Warning (dev) at CMakeLists.txt:85 (if):
  Policy CMP0054 is not set: Only interpret if() arguments as variables or
  keywords when unquoted.  Run "cmake --help-policy CMP0054" for policy
  details.  Use the cmake_policy command to set the policy and suppress this
  warning.

  Quoted variables like "MSVC" will no longer be dereferenced when the policy
  is set to NEW.  Since the policy is not set the OLD behavior will be used.
This warning is for project developers.  Use -Wno-dev to suppress it.
```

PR #1974, opened 2017-07-28
Add MSVC support to c++ cmake build

matthauck:


matthauck:
I understand that the windows builds are currently setup to use checked-in vcproj files. I personally would like to be able to build antlr with cmake on windows and non-windows to keep our build definition simpler and not have to follow different paths for building on different platforms. 

I have tried to add the MSVC compiler flags corresponding to the gcc flags where relevant, and tried to add most of the settings i could find in the project file. I think it would be cool to move away from checked in project files; imho, it is easier to maintain and automate. It also enables the `install` target, which is pretty cool. :)

matthauck:
(Note: contributors.txt signed in #1973. I can add that here too if needed)

mike-lischke:
I dare to disagree. Using cmake in addition to checked-in VS project files is acceptable (even though I still find this rather strange), however the project files must stay in the repo. You *always* need them. Why would you want to generate them first? And you would force people to use cmake even though it's not needed at all in this case.

Unfortunately, I cannot review this patch. I don't use cmake.

matthauck:
I think the reasons for using cmake instead of checked-in project files are:
* Ability to use the `install` target instead of manually copying files around
* Easier to automate 
* Ability to unify configuration of source files across all platforms, and even MSVC settings across versions. 
* Frees up the antlr project from creating and clicking around in MSVC UI whenever a new compiler needs to be supported (or going back and updating things w/ old MSVC versions when bugs are found). 

CMake is a great tool and having people install it before they *build* antlr doesn't seem like a burdensome requirement to me, and it makes the project easier to build, imo.

PR #1978, opened 2017-08-02
[Go] Interval

neatnerd:
Added additional interval math functionality

### DIFF
--- a/runtime/Go/antlr/interval_set.go
+++ b/runtime/Go/antlr/interval_set.go
@@ -27,6 +27,68 @@ func (i *Interval) Contains(item int) bool {
 	return item >= i.Start && item < i.Stop
 }
 
+// Does this start completely before other? Disjoint
+func (i *Interval) StartsBeforeDisjoint(other *Interval) bool{
+	return i.Start<other.Start && i.Stop < other.Start
+}
+
+// Does this start at or before other? Nondisjoint
+func (i *Interval) StartsBeforeNonDisjoint(other *Interval) bool{
+	return i.Start <= other.Start && i.Stop >= other.Start
+}
+
+// Does this.a start after other.b? May or may not be disjoint
+func (i *Interval) StartsAfter(other *Interval) bool{
+	return i.Start > other.Start
+}
+
+// Does this start completely after other? Disjoint
+func (i *Interval) StartsAfterDisjoint(other *Interval) bool{
+	return i.Start > other.Stop
+}
+
+// Does this start after other? NonDisjoint
+func (i *Interval) StartsAfterNonDisjoint(other *Interval) bool{
+	return i.Start>other.Start && i.Start <= other.Stop // i.Stop>=other.Stop implied
+}
+
+// Are both ranges disjoint? I.e., no overlap?
+func (i *Interval) Disjoint(other *Interval) bool{
+	return i.StartsBeforeDisjoint(other) || i.StartsAfterDisjoint(other)
+}
+
+// Are two intervals adjacent such as 0..41 and 42..42?
+func (i *Interval) Adjacent(other *Interval) bool{
+	return i.Start == other.Stop+1 || i.Stop == other.Start-1;
+}
+
+func (i *Interval) ProperlyContains(other *Interval) bool{
+	return other.Start >= i.Start && other.Stop <= i.Stop
+}
+
+// Return the interval computed from combining this and other
+func (i *Interval) Union(other *Interval) *Interval{
+	return NewInterval(intMin(i.Start, other.Start), intMax(i.Stop, other.Stop))
+}
+
+// Return the interval in common between this and other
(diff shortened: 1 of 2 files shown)


neatnerd:
Yo @davesisson I will write the test tonight then, but only for Go target and only this section, ok?

davesisson:
Thanks!  That would be great!

neatnerd:
There is one issue that I have stumble across during testing
Why does Go target actually not include Stop position of token? as mention here:

```go
/* stop is not included! */
func NewInterval(start, stop int) *Interval {
	i := new(Interval)
	i.Start = start
	i.Stop = stop
	return i
}
```

Actually the calculation of length then is also adjusted accordingly:
Go:
```go
func (i *Interval) length() int {
	return i.Stop - i.Start
}
```
vs
Java:
```java
/** return number of elements between a and b inclusively. x..x is length 1.
 *  if b &lt; a, then length is 0.  9..10 has length 2.
 */
public int length() {
	if ( b<a ) return 0;
	return b-a+1;
}
```
vs
C++:
```cpp
size_t Interval::length() const {
  if (b < a) {
    return 0;
  }
  return size_t(b - a + 1);
}
```
I did the blame and seems it has not changed since initial commit. @pboyer @davesisson @parrt do you, guys, know why?

Reason I am asking is cause it is a fundamental difference in logic in many parts of application imho

davesisson:
I'm not sure why it is this way -- perhaps it was thought that it was simpler.

But I guess it really comes down to -- are you really going to normalize the behavior at this point?  I would expect that changing it to match (if that happens) would happen as a separate PR.

neatnerd:
Ok, that means, I have to adjust logic for Interval position that put into this PR, ok?

parrt:
Sounds like a bug in Go target. length() makes no sense as 1 less than the length I guess.

neatnerd:
@parrt I dont think it is a bug, it just uses half-opened intervals instead of closed interval, thus calculation is ok. There are in general two options: adjust this logic or adjust everything around it

parrt:
Oh right. yeah, i like a..b inclusive but [a,b) is more typical. hmm... as long as the doc/API is clear which way it goes, maybe we leave for compatibility?

neatnerd:
To be honest, this part is pretty undocumented, the only reference I found in official docs is [here](http://www.antlr.org/api/Java/index.html) with 

> public int length()
return number of elements between a and b inclusively. x..x is length 1. if b < a, then length is 0. 9..10 has length 2.

So I would say the call how we want to keep it is up to you [a,b] or [a, b)
Also maybe it would be beneficial to add it to Interval class docu

neatnerd:
@parrt any thoughts on this maybe?

parrt:
To be consistent with other targets I suggest using a..b inclusive `[a,b]`

parrt:
@davesisson there are tests for interval: TestIntervalSet.java :)
Ter

pboyer:
I'm liking this change! :D

runtime/Go/antlr/interval_set.go
```diff
@@ -27,6 +27,68 @@ func (i *Interval) Contains(item int) bool {
...
+
+// Return the interval in common between this and other
+func (i *Interval) Intersection(other *Interval) *Interval{
+	return NewInterval(intMax(i.Start, other.Start), intMin(i.Stop, other.Stop))
+}
+//  Return the interval with elements from this not in other;
+//  other must not be totally enclosed (properly contained)
+//  within this, which would result in two disjoint intervals
+//  instead of the single one returned by this method.
+func (i *Interval) DifferenceNotProperlyContained(other *Interval) *Interval{
```
pboyer: It's Go style to prefix any method comment with the name of the function. So, this should read `DifferenceNotProperlyContained returns the interval...`

neatnerd:
@parrt Regarding tests, yes there are tests for IntervalSet, but there are no tests for Interval.

parrt:
@neatnerd ah! ok.

parrt:
Unfortunately I assumed you had tested this. It breaks all of the unit tests so I'm going to have to revert.

neatnerd:
Yeah, sorry, wanted to fix it this weekend. Shall I make a new PR?

parrt:
yes, please. seems like it'll change many many locations in the API, right?

neatnerd:
Well... There were couple of inconsistencies as to how intervals are handled. For example, I have assumed that if I delete from interval [10;20], point 15, result shall be [10;14] and [16;20]. Seems like it has to be [10;15] and [16;20]. Is there a good set of tests/documentation somewhere what I can rely upon for intervals? I assume no, right?

parrt:
nope. We made the assumption that all intervals were inclusive so anything that deviates from that is likely to be wrong or inconsistent.

neatnerd:
Ok, thx. Will research it a bit more in detail and make a separate PR

PR #1979, opened 2017-08-05
Updated swift-target.md.

hanjoes:
@parrt This is to put in performance note for the Swift target.


parrt:
Does this close that issue?

hanjoes:
Hmm, I will close it.

PR #1983, opened 2017-08-07
[Go] Add CommonTokenStream test.

shirou:
This PR add tests for `CommonTokenStream`. These tests is ported from [Java test](https://github.com/antlr/antlr4/blob/master/tool-testsuite/test/org/antlr/v4/test/tool/TestCommonTokenStream.java).

To testing purpose, these two files are also added. I think these files can be used for future tests.

- testing_lexer_b_test.go
- testing_util_test.go
- testing_assert_test.go  <-- assert library which is almost same as Java assertions.

(updated: as thankful comment, I removed dependency to assert library)

### DIFF
--- a/runtime/Go/antlr/common_token_stream_test.go
+++ b/runtime/Go/antlr/common_token_stream_test.go
@@ -6,8 +6,6 @@ package antlr
 
 import (
 	"testing"
-
-	"github.com/stretchr/testify/assert"
 )
 
 type commonTokenStreamTestLexer struct {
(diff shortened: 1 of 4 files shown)


neatnerd:
Is it really necessary to introduce assert lib here? You can rewrite the code to use plain vanilla testing library included in std lib. This would break testing, builds (sort of, if testing is part of it).

Im in general against such "hacks"

shirou:
@neatnerd I understand your concern. I have updated this PR about re-implementing assert functions to remove dependency.  (I edited this PR description)

davesisson:
This may be my workplace's style talking but do we need the assertion library at all?

shirou:
I believe Java style assertion functions are useful for porting Java tests.

davesisson:
Good point.

neatnerd:
But anyways for now, it is done without asserts, I would recommend for merge

PR #1986, opened 2017-08-14
Added ParserRuleContext.getFullText()

RedTailedHawk:

### DIFF
--- a/runtime/Java/src/org/antlr/v4/runtime/Parser.java
+++ b/runtime/Java/src/org/antlr/v4/runtime/Parser.java
@@ -623,7 +623,7 @@ public abstract class Parser extends Recognizer<Token, ParserATNSimulator> {
 	 * {@link #_ctx} get the current context.
 	 */
 	public void enterRule(ParserRuleContext localctx, int state, int ruleIndex) {
-		if (localctx.tokenStream == null)
+		if (localctx.parent == null && localctx.tokenStream == null)
 			localctx.tokenStream = _input;
 		
 		setState(state);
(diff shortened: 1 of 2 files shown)


sharwell:
With ANTLR 4.7, it should be possible to accomplish these goals without making changes to the runtime:

1. Create a class derived from `ParserRuleContext`
2. Set the new class as the base class for all rule context classes created for your grammar
3. Override `RuleContext.setParent`, and set `this.tokenStream = parent.tokenStream` (this way after setting it in the root it propagates to the entire tree).
4. In your root rule, set the token stream for the root node:

    ```antlr
    compilationUnit
    @before {
      $ctx.tokenStream = getTokenStream();
    }
      : ...
      ;
    ```

RedTailedHawk:
@sharwell `Set the new class as the base class for all rule context classes created for your grammar` -- how do I do this?

sharwell:
You can use the `contextSuperClass` option:

```
options {
  contextSuperClass=com.your.qualified.CustomParserRuleContext;
}
```

RedTailedHawk:
Thanks, `contextSuperClass` is working great!  But I'm not having success with `@before`.  Here's my rule:

```
script
@before {
  $ctx.tokenStream = getTokenStream();
}
    : statements EOF
    ;
```

I regenerated the class, but I'm not seeing the above code in the generated `ScriptContext` class.  Am I looking in the wrong place?

sharwell:
It's supposed to be `@init`. Sorry about that!

RedTailedHawk:
OK, that works great!

Is there any way to skip these Java-specific annotations in my grammar when I also generate my classes for JavaScript?  If not, I can comment them out when I generate JavaScript, and then uncomment for Java.  Not too big a deal, but I do have to remember to do that whenever my grammar changes.

parrt:
Thanks to @sharwell  I believe we have a solution so I am closing.

Issue #1858, opened 2017-05-05
GO runtime: Can not compile on ARM


milanaleksic:
As presented in the reproduction scenario, current code that constitutes Go runtime the can't be compiled if target is linux + ARM architecture (no problem if I remove `GOARCH=arm` from the request below). Currently reproducible in my project wallabago, commit 48a5e22d2336c12fe24ab410ab260c79809623b4.

```bash
milan@dell /opt/go/src/github.com/milanaleksic/wallabago/cmd/main → GOARCH=arm go build -o wallabago
# github.com/milanaleksic/wallabago/vendor/github.com/antlr/antlr4/runtime/Go/antlr
../../vendor/github.com/antlr/antlr4/runtime/Go/antlr/utils.go:368: constant 3432918353 overflows int
../../vendor/github.com/antlr/antlr4/runtime/Go/antlr/utils.go:374: constant 3864292196 overflows int
../../vendor/github.com/antlr/antlr4/runtime/Go/antlr/utils.go:381: constant 2246822507 overflows int
../../vendor/github.com/antlr/antlr4/runtime/Go/antlr/utils.go:383: constant 3266489909 overflows int
```

whrool:
It is the same on windows + 386

neatnerd:
I think the error is pretty clear. The murmur hash implementation is 64-bit (i guess?) in code, so ARM cannot handle it. Can you research the ARM murmur hash implementation, then some1 can PR that

milanaleksic:
All I've done in my case is change undefined type (probably resolved to int) to uint and I didn't encounter any issues but didn't have time to verify via comparing with some other implementation, since there are no explicit tests for this code (murmurInit etc) that I could find. I don't like sending PR code I didn't have time to test

milanaleksic:
I've sent a PR of the change that worked for me

exander77:
Seems to be broken, see bottom: https://github.com/antlr/antlr4/issues/2152

PR #1990, opened 2017-08-21
Allowing Go runtime compilation to succeed under ARM

milanaleksic:
Discussion already present in #1858 - effectively the deduced type wasn't big enough to allow compilation and execution under my ARM server (Radxa Rock Pro). This change allowed compilation to succeed and I didn't find any issues so far

If tests are needed please hold the PR until I have time to add them.


runtime/Go/antlr/utils.go
```diff
@@ -353,36 +353,37 @@ func PrintArrayJavaStyle(sa []string) string {
...
 
 func murmurInit(seed int) int {
 	return seed
 }
 
 func murmurUpdate(h1 int, k1 int) int {
-	k1 *= c1_32
-	k1 = (k1 << 15) | (k1 >> 17) // rotl32(k1, 15)
-	k1 *= c2_32
+	var k1u uint
```
davesisson: You can avoid defining the var if you use the := operator.  (Here and elsewhere.)

runtime/Go/antlr/utils.go
```diff
@@ -353,36 +353,37 @@ func PrintArrayJavaStyle(sa []string) string {
...
-	h1 = (h1 << 13) | (h1 >> 19) // rotl32(h1, 13)
-	h1 = h1*5 + 0xe6546b64
-	return h1
+	var h1u = uint(h1) ^ k1u
+	h1u = (h1u << 13) | (h1u >> 19) // rotl32(h1u, 13)
+	h1u = h1u*5 + 0xe6546b64
+	return int(h1u)
 }
 
 func murmurFinish(h1 int, numberOfWords int) int {
```
davesisson: Probably should fix the input type instead of relying on a cast (enforcing that there are no negative numbers passed to this function).  Also might be good to document that numberOfWords is limited to be 4x less than the normal uint range.

runtime/Go/antlr/utils.go
milanaleksic: OK

runtime/Go/antlr/utils.go
milanaleksic: I started doing this but changing of the input arguments to uint percolates to many places, I don't have time to fix it at this moment, will push when it's ready

neatnerd: i dont think it is needed. I did consider suggesting this option, then i checked alg specification, to my understanding h1 and k1 cannot be negative

parrt:
hi gang. Are we ready to merge this one?

parrt:
@milanaleksic @davesisson @neatnerd Are we ready to merge this one?

davesisson:
Yes, I think this one is good to go.

PR #1996, opened 2017-08-29
Export Go lexer methods

ereyes01:
I've exported several methods in the Go runtime. I found that these methods were used in some grammar actions in grammars I was porting to Go. The analogues of these methods in the Java runtime also appear to be public methods, so this change brings the Go runtime closer to parity with the Java runtime.


pboyer:
Thanks @ereyes01! LGTM.

PR #2014, opened 2017-09-16
Docs: Clarify that visitors are unavailable under go.

savvopoulos:
---
This change is [Reviewable](https://reviewable.io/reviews/antlr/antlr4/2014)


savvopoulos:
@cyberfox

cyberfox:
@savvopoulos It makes sense to document, but if we could get a pull request through, I'd rather it be the one that actually implements it.  I'm also not sure if, paradoxically, this change would conflict with pull/1841 and make it harder to land. :-/

PR #2033, opened 2017-09-28
Improve MurmurHash

ewanmellor:
Improve MurmurHash in a number of ways:

Use Swift's overflowing operators rather than multipliedReportingOverflow etc.

Use UInt32 for the hash values.  This matches how MurmurHash3 is generally defined (e.g. on Wikipedia).

Add support for decoding Strings (UTF-8, then little-endian) and hashing the resultant UInt32 values.

Add a test set, using test patterns from Ian Boyd (public domain).

This is dependent on PR #2032 (I found these problems when working on the Swift 4 migration).

### DIFF
--- a/runtime/Swift/Sources/Antlr4/misc/MurmurHash.swift
+++ b/runtime/Swift/Sources/Antlr4/misc/MurmurHash.swift
@@ -6,20 +6,28 @@
 
 
 /// 
+/// https://en.wikipedia.org/wiki/MurmurHash
 /// 
 /// - Author: Sam Harwell
 /// 
 
 public final class MurmurHash {
 
-    private static let DEFAULT_SEED: Int = 0
+    private static let DEFAULT_SEED: UInt32 = 0
+
+    private static let c1 = UInt32(0xCC9E2D51)
+    private static let c2 = UInt32(0x1B873593)
+    private static let r1 = UInt32(15)
+    private static let r2 = UInt32(13)
+    private static let m = UInt32(5)
+    private static let n = UInt32(0xE6546B64)
 
     /// 
     /// Initialize the hash using the default seed value.
     /// 
     /// - Returns: the intermediate hash value
     /// 
-    public static func initialize() -> Int {
+    public static func initialize() -> UInt32 {
         return initialize(DEFAULT_SEED)
     }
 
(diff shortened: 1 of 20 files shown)


ewanmellor:
Rebased.

parrt:
@hanjoes Would you have time to take a look at this PR?

parrt:
@hanjoes Can you take a quick peek at this one?

hanjoes:
Sorry @parrt didn’t see this this morning. I’m outside now, will take a look later today.

runtime/Swift/Sources/Antlr4/misc/MurmurHash.swift
```diff
@@ -111,14 +112,55 @@ public final class MurmurHash {
...
+    /// little-endian values, giving UInt32 values for the update call.
+    ///
+    /// If the bytes do not evenly divide by 4, the final bytes are treated
+    /// slightly differently (not doing the final rotate / multiply / add).
+    ///
+    /// This matches the treatment of byte sequences in publicly available
+    /// test patterns (see MurmurHashTests.swift) and the example code on
+    /// Wikipedia.
+    ///
+    public static func hashString(_ s: String, _ seed: UInt32) -> UInt32 {
```
hanjoes: I don't see where this function is used?

ewanmellor: It's only used by the unit tests.  The test patterns that I found were built around UTF-8 strings as input.  The UTF-8 strings are then decoded to UInt32 arrays before being passed to the hash function.  (See `MurmurHashTests.testMurmurHash`.)

In our codebase proper, we're only using MurmurHash with UInt32 (or lists of them) as input.  We don't have anywhere where the input is passed as a UTF-8 string.  I added hashString so that we can check that our MurmurHash matches the public definition of MurmurHash, but other than for that, we're not using it.

hanjoes: I see. We need it to verify the correctness of the implementation.

parrt:
ok, so I can merge?

hanjoes:
Yeah, let’s merge.

PR #2044, opened 2017-10-06
Common token stream

davesisson:
Make getHiddenTokensToRight/Left public to enable rewriting tasks.  Also fixes a bug in GetTokens (should return all tokens including the specified stop token).


runtime/Go/antlr/common_token_stream.go
```diff
@@ -154,7 +154,7 @@ func (c *CommonTokenStream) GetTokens(start int, stop int, types *IntervalSet) [
 		stop = len(c.tokens) - 1
 	}
 
-	for i := start; i < stop; i++ {
+	for i := start; i <= stop; i++ {
```
neatnerd: are you sure? is it not related to discussion of open vs. closed interval? #1978

davesisson: I think you're right.  In general it's better for the start/stop to be inclusive.  All of the code I've seen using this is always doing a +1 here to make the stop work correctly.

parrt: My API functions all assume inclusive intervals so `<=` seems right, doesn't it?

davesisson: This line is wrong if you're looking at the rest of the Go Interval implementation.  This change would be right in context of the Java API implementation (and the docs for this function above).

parrt:
Seems we need to resolve that interval issue... @davesisson  if that line is wrong, looking at the rest of the implementation, how should we change it? Note that I did not merge the change to the intervals from inclusive to exclusive that was submitted earlier.

parrt:
hmm...errors from travis. it says:

```
 This job was configured to run on an OS X image that was retired on November 28, 2017. New jobs will route to our Xcode 8.3 image.
```

@ericvergnaud or @ewanmellor have any ideas?

ewanmellor:
@parrt If you're looking at the tests on this PR right now (https://travis-ci.org/antlr/antlr4/builds/284022552) then that test ran two months ago. That macOS image wasn't retired at the time, though it is now, but that is beside the point.  The actual test failures in that run were 1) the OracleJDK 7 image was retired, failing all the JDK7 tests, and 2) a transient network issue which hit one of the C++ runs and one of the .NET macOS runs.

You (or the author) should rebase this change onto latest master to get the new Travis config.  You will need the JDK 7 and macOS image upgrades for sure, and there are reliability improvements too.

davesisson:
Let's drop this one.  We need to fix the interval behavior but existing code works (as long as they do offsets).  I have a separate PR for the exporting of these functions.

davesisson:
(I'm thinking we should have a more comprehensive realignment for intervals in a later Antlr release.)

parrt:
@ewanmellor I thought we had switched already from oracle to openjdk. hmm...ah we did. ok, that was old test as you say. new build appears to be working for openjdk7 https://travis-ci.org/antlr/antlr4/builds/312611861

PR #2046, opened 2017-10-06
Add a new CharStream that converts the symbols to upper or lower case.

bramp:
This is useful for many of the case insensitive grammars found at https://github.com/antlr/grammars-v4/ which assume the input would be all upper or lower case. Related discussion can be found at https://github.com/antlr/antlr4test-maven-plugin/issues/1

It would be used like so:
```go
input, _ := antlr.NewFileStream("filename")

in = antlr.NewCaseChangingStream(is, true) // true forces upper case symbols, false forces lower case.

lexer := parser.NewAbnfLexer(in)
```

While writing this, I found other people have written their own similar implementations ([go](https://github.com/KvanTTT/DAGE/blob/master/AntlrGrammarEditor/Runtimes/Go/AntlrCaseInsensitiveInputStream.go), [java](https://github.com/antlr/antlr4test-maven-plugin/blob/master/src/main/java/com/khubla/antlr/antlr4test/AntlrCaseInsensitiveFileStream.java)). It makes sense to place this in the core, so everyone can use it.

I would love for the grammar to have a option that says the lexer should upper/lower case all input, and then this code could be moved into the generated Lexer, and no user would need to explicitly use a CaseChangingStream (similar to what's discussed in #1002).


runtime/Go/antlr/case_changing_stream.go
```diff
@@ -0,0 +1,37 @@
...
+// from the underlying CharStream and converts it to either upper case
+// or lower case.
+func (is *CaseChangingStream) LA(offset int) int {
+	in := is.CharStream.LA(offset)
+	if in < 0 {
+		// Such as antlr.TokenEOF which is -1
+		return in
+	}
+	if is.upper {
+		return int(unicode.ToUpper(rune(in)))
```
KvanTTT: Maybe it's better to convert case at the beginning one time and use upper/lower values from a prepared array for the performance reason? `LA` function can be called multiple times.

bramp: I could cache the result, but I didn't want this to work by reading the full input up front, because this data could be streamed, or extremely large.

However, I don't think caching will save much, as this is quite a cheap operation for the CPU. But I'm happy to benchmark it.

bramp:
Sure, I'd be happy to contribute, Java, C++ and Python[23] (I don't know / have access to C# or Swift). Let's resolve this one first, then I'll work on the other languages.

runtime/Go/antlr/case_changing_stream.go
KvanTTT: Ok, got it. Maybe it's right.

bramp:
I just added a similar version for Java https://github.com/antlr/antlr4/pull/2048

parrt:
hi gang. How is this different than simply running tolower() or whatever on the input string?

bramp:
In my example, I want to use a NewFileStream, which gives me no opportunity to tolower the input. Basically this just wraps an existing stream and does that for me.

parrt:
It seems to me that what one really wants is insensitivity during the lexing process but to leave the actual input as is, right?

bramp:
Sure, if the lexer could treat the input in a case insensitive way, that would be ideal. I don't like sticking an extra stream in front, I'd rather the author of the lexer hide this detail from me. But currently there are multiple grammars that require lower or upper case input, and this helps me use them.

Alternatively as a step in that direction, this stream could be kept inside the lexer, and put there by antlr during the code generation.

parrt:
It seems like we should simply tweak `LA()` so `input.LA(1)` returns toupper or change `getExistingTargetState()` and `computeTargetState()`  and any other places that test characters in the LexerATNSimulator.  of course that is a lot scarier to change

bramp:
And there would be a change to the grammar file to indicate that input should always be uppercased? Because presumably some grammars would want to be case sensitive.

parrt:
That's a good question. Is it a function of the grammar purely or could the same grammar be used in both case-sensitive and insensitive situations?

bramp:
The samples I've seen on https://github.com/antlr/grammars-v4 have all been written assuming the input would all be one case, and not something you'd want to selectively turn on and off. They do this because it makes the grammar easier to write. So in those cases it would be a function of the grammar. However, I can't speak for all exotic grammars.

parrt:
Hmm...yeah,I can imagine it being a function of the grammar.  That said, given my limited time I think we have a simple solution which is to manually alter the string going into the upper or lower case. The programmer can do this easily with their own wrapper class or a simple function call if they have the input as a string. In that case, I'm going to close. Sorry but thanks for the effort.

bramp:
Sure, the programmer can write their own wrapper, but providing one in the core library makes that so much easier.  When I first started using antlr, I took a pre-written grammar (the mysql one in fact), and it took me a while to realise I had to uppercase everything, then it took me another while to understand the Stream interface, and figure out how to adapt the LA function to do what I need... Lets avoid that burden on a first time user, and make this easy for them with this pre-written and tested wrapper class.

parrt:
Well,I have a lot of targets that we need to be updated and documentation and tests and so on. Probably not worth it when I personally just call toupper after reading in a string and before passing to ANTLR.

bramp:
Ok sounds reasonable. I am willing to provide implementations for some of the other targets, but not all of them. However, I'll maintain this code outside of the core library, until there is an appetite for it. Thanks.

KvanTTT:
@parrt, this stream save chars case. It only influences on lexer, but not on the parser or parse tree bypassing. It's roughly just to call "toupper" for an input string.

On my opinion, it's very demanded feature. See the list of issues in grammars and main repository:
https://github.com/antlr/grammars-v4/issues/854
https://github.com/antlr/grammars-v4/issues/878
https://github.com/antlr/grammars-v4/issues/919

parrt:
hiya. I can see how it's useful occasionally but it seems like we would need to do this for all targets not just Go before adding it.  Maybe I can come up with a Java version to see how it feels.

bramp:
I sent a PR for a Java implementation here: https://github.com/antlr/antlr4/pull/2048

KvanTTT:
I've implemented such streams for C#, C# optimized, Go, Java, JavaScript, Python2, Python3 runtimes. See `AntlrCaseInsensitiveInputStream` in respective runtime directories: https://github.com/KvanTTT/DAGE/tree/master/AntlrGrammarEditor/Runtimes. С++ and Swift are lacking for now.

KvanTTT:
Yet another case insensitive issue: https://github.com/antlr/grammars-v4/issues/921

parrt:
Ok, so this wrapper implementation that @bramp  has provided seems flexible actually.  It works with existing streams but we just wrap that stream in an object that upper or lower case is things as needed without changing the actual stream. If we could bring other targets in line with #2048 and bring it all together as a PR, I think that would be acceptable.

parrt:
@KvanTTT @bramp can you guys summarize the status of this insensitive stream stuff? I'm willing to merge a PR that has all target impl you have created but let's get into one PR with tests?

bramp:
I had time for this ~2 months ago when I sent the first PR, but in the past few weeks been busy... However, if you are willing to accept just the languages we've already covered, then I can make time one evening this week pull it all together into one PR with tests.

parrt:
Ok, great, let's do just the languages you've got. it'd be great to pull into a single PR. I might pull in and add a few more tests and tweak API.

bramp:
I've pulled in a Java version into this (with updates based on #2048).

I've looked at @KvanTTT's implementations, and they seem to take a string, and just upper or lowercase the full thing. So not exactly the same as this implementation.

However, I've added a javascript version, but I don't see where the existing Javascript tests are. I see runtime-testsuite/test/org/antlr/v4/test/runtime/javascript/... but that seems full of Java code. The same is true for the other languages as well. Other than Go and Java there doesn't seem to be unit tests for any of the languages.  The tests that are there look like integration tests, and it is unclear where to start. Please advise.

KvanTTT:
I think your implementation is more proper than mine.

KvanTTT:
Is it also possible to include a test with another culture (Turkish)?

parrt:
looking good! @bramp the [target unit tests are all done generically](https://github.com/antlr/antlr4/blob/master/doc/antlr-project-testing.md) except for specific API function call testing.  So all targets are tested for functionality using the descriptors but testing actual function calls needs to be done in that target language, hence, the specific tests like [TestCharStreams.java](https://github.com/antlr/antlr4/blob/master/runtime-testsuite/test/org/antlr/v4/test/runtime/java/TestCharStreams.java), which is under `runtime-testsuite/test/org/antlr/v4/test/runtime/java`.

does that help? The problem we will have here is that we don't currently have an option to turn on case insensitivity in the descriptors. I think it's okay to simply add some tests in the target specific area as you have done.

@bramp @KvanTTT  anything else before I merge this?

KvanTTT:
LGTM, but ideally it would be great to have a test with culture sensitive setup, see [CultureSensitiveLowerCase](https://www.tutorialspoint.com/compile_java_online.php) sample.

>The problem we will have here is that we don't currently have an option to turn on case insensitivity in the descriptors.

It would be great to implement a `caseInsensitive` option in grammar up to 4.7.2 version. Something like this:

```antlr
```

I've already [done](https://github.com/antlr/antlr4/pull/1092) it in a long time ago :) But case insensitive streams much better approach in my opinion.

parrt:
Well that caseInsensitive option would need to work in conjunction with the streams. we can also add caseInsensitive option to the test descriptors eventually maybe.

ericvergnaud:
Agreed a test flag would make more sense than a grammar option.

KvanTTT:
Without such option case insensitive issues in grammars-v4 repository will appear again.

mike-lischke:
This is a long discussion here and it could be I'm missing something, but I don't agree with the solution(s) provided here. It seems you stepped into the same old trap like so many people before: simple upper/lower casing works only for ASCII input reliably. It already fails with ANSI input (see below). We left this habour a long time ago - the world speaks Unicode.

You don't want **all** your input to be case insensitive (and casing is a pretty sensitive issue in Unicode, sometimes case conversion depends on specific contexts). Instead what you want is that your **keywords** are matched regardless of the written case, correct? So, it makes sense to add handling of different letter cases to keyword rules only.

A good way is what grammar authors have used for years, which is mostly about defining all letter-cases in single letter lexer rules, like:

```
fragment A: 'A' | 'a';
fragment B: 'B' | 'b';
...

KEYWORD1: K E Y W O R D '1';
```

This way you can also define exactly what you want to match, e.g. when a char cannot directly be derived by a case transformation, but should still match. Keep also in mind that case conversion can change the number of letters (e.g. the german `ß` is usually converted to `SS` when upper-cased). There can be other issues you don't even thought about in your solution (like normalization forms where characters come in either composed or separated, which can lead to totally wrong results by using plain upper/lower case functions).

However, if this all about ingoring case for the ASCII charset then just use the single lexer rules as outlined above and you are done with it, without have to add new charstream classes to each target.

KvanTTT:
>You don't want all your input to be case insensitive (and casing is a pretty sensitive issue in Unicode, sometimes case conversion depends on specific contexts). Instead what you want is that your keywords are matched regardless of the written case, correct? So, it makes sense to add handling of different letter cases to keyword rules only.

The text of the input stream is unaffected. Example: input `BeGiN` would match lexer rule `BEGIN` but `getText()` would return `BeGiN` string.

>A good way is what grammar authors have used for years, which is mostly about defining all letter-cases in single letter lexer rules, like:

This approach is slower and more ugly unlike the approach with case insensitive input stream.

>Keep also in mind that case conversion can change the number of letters (e.g. the german ß is usually converted to SS when upper-cased).

We are using simple culture invariant case transformation. You german `ß` won't be converted, see [CultureCaseConversionIssues ](https://ideone.com/IvtDOd) sample. Moreover, such samples very rare and this is not important in most cases.

I'm an author/editor of several case insensitive SQL grammars and I think such feature is required and simplifies grammar development process. It's not necessary to use such stream if you don't want.

parrt:
Should we use `toUpperCase(Locale locale)`, at least for the java target, that supports a locale? It would mean an extra constructor arg and CharStreams.toUpper() arg.

KvanTTT:
I think we have to use the default Local in all runtimes for excluding cases when one character converter to several (`ß` -> `SS`). Also, lexer should not depend on current user Local.

parrt:
I guess that makes sense. The lexer assumes specific characters and can't have them split into multiple in the input stream.

mike-lischke:
> The text of the input stream is unaffected. Example: input BeGiN would match lexer rule BEGIN but getText() would return BeGiN string.

You are converting all text whenever the lexer needs to match a letter, no? So, you are also converting all the things that aren't even case sensitive (like numbers, strings, comments). You didn't answer the core question: this is all about keyword matching, yes or no? If yes, limit case conversion to keywords only.

>>fragment A: 'A' | 'a';
>>fragment B: 'B' | 'b';
>>...
>>KEYWORD1: K E Y W O R D '1';
>
>This approach is slower and more ugly unlike the approach with case insensitive input stream.

This is all but ugly and I doubt that it has more performance impact than a case conversion stream. It does exactly what's actually needed and it does it explicitly, so people looking at the grammar **know** this is supposed to handle all letter cases. Having to use a case conversion stream is a hidden incredient, required to make the grammar work. Which is an extra stumbling stone for the unexperienced. That's what I'd call ugly.

>>    Keep also in mind that case conversion can change the number of letters (e.g. the german ß is usually converted to SS when upper-cased).
>
>We are using simple culture invariant case transformation. You german ß won't be converted, see CultureCaseConversionIssues sample. Moreover, such samples very rare and this is not important in most cases.

Oh, so the "solution" is it make it work somehow, ignoring all the things that are more difficult to get them to work? Oh my 🙄 

>I'm an author/editor of several case insensitive SQL grammars and I think such feature is required and simplifies grammar development process.

Hmm, I added the letter case rules once and never had to touch them again. Super easy to do and they can even be copied to every other grammar, if needed. I don't see any complication with them.

parrt:
Interesting. Sounds like keywords are the only issue and we could trivially make a keyword rule generator script or better yet an IDE command like I have for autogenerating other rules in intellij for antlr.

mike-lischke:
@parrt I got another idea. Could we introduce a new keyword in the ANTLR4 language to tell ANTLR to implicitly handle a lexer rule case-insensitive? Let's say in addition of `fragment` we could add `ignorecase` and use that when defining a keyword:

```
fragment caseignore KEYWORD1: 'KEYWORD1';
```

That would probably fit best for both, easy use and explicit notation. And it wouldn't require any target changes. Internally ANTLR4 would include alle letter cases in the interval sets for that rule in the lexer. Still only works for ASCII, but sounds like a good compromise.

parrt:
Could work. It's only euro languages that have cap/lower right? How can we support all upper/lower?

parrt:
maybe `options {...}` for rule instead of new keyword.

davesisson:
Here are the case folding rules from unicode.org:

ftp://ftp.unicode.org/Public/UNIDATA/CaseFolding.txt

mike-lischke:
`options` is more expressive, but yes that would work too.

davesisson:
And if you have troubles reading that, here's are some charts:

http://unicode.org/charts/case/

KvanTTT:
> You are converting all text whenever the lexer needs to match a letter, no? So, you are also converting all the things that aren't even case sensitive (like numbers, strings, comments).

It does not matter in real scenarios. Semantics should be checked after the parsing stage.

> You didn't answer the core question: this is all about keyword matching, yes or no? If yes, limit case conversion to keywords only.

Yes. I've already answered that fragment rules worsen the grammar.

>Oh, so the "solution" is it make it work somehow, ignoring all the things that are more difficult to get them to work? Oh my 🙄

Nobody worried about it with my case insensitive approach with SQL grammars.

>fragment caseignore KEYWORD1: 'KEYWORD1';

This does not make sense because in general entire grammar or at least modes are case insensitive, but not certain tokens.

>maybe options {...} for rule instead of new keyword.

I've already implemented it even for modes: https://github.com/antlr/antlr4/pull/1092 But it has not been accepted. See also related issue: https://github.com/antlr/antlr4/issues/1002 At the beginning I also adhered to the fragment rules instead of case-insensitive stream. But later I changed my opinion.
Moreover, I would like to implement such an option in the next ANTLR version (4.7.2).

>Super easy to do and they can even be copied to every other grammar if needed. I don't see any complication with them.

This is a duplication pattern that should be eliminated.

All IDE's and text editors have a 'case insensitive' option to find the certain text and this suits everyone. Why can ANTLR not have such option?

KvanTTT:
>Still only works for ASCII, but sounds like a good compromise.

At least Cyrillic is also working fine, see https://ideone.com/IvtDOd

parrt:
Given a unicode string, shouldn't toUpperCase() (without locale) work correctly?

parrt:
Altering the grammar with an option has the advantage of not requiring new streams in every target.

parrt:
btw, @sharwell used a stream approach which makes me think it's also not a hack. His approach does toUpperCase() once on copy of stream which is faster and also supports beta becoming two char 'SS'. Disadvantage: must replicate across streams and also that grammar doesn't indicate case-insensitivity.

parrt:
yep, streams leave text alone (in functionality) but make the lexer think everything is upper or lower.

KvanTTT:
> Given a unicode string, shouldn't toUpperCase() (without locale) work correctly?

The `Character.toUpperCase('а'));` is not depend on the users locale. 

>Disadvantage: must replicate across streams and also that grammar doesn't indicate case-insensitivity.

Other runtimes also have a culture-invariant character case changing methods. I checked it at least for Java, C#, JavaScript, Go (I can attach confirming links later). We also can introduce case insensitive option in grammar and use such streams if it's set up.

Maybe it makes sense to add not only case sensitivity option, but also `culture` (`invariant` by default, `ru`, `tr`, `de` etc.) or `locale` option. But I'm afraid it may not work across all runtimes.

parrt:
So `Character.toUpperCase(x))` will work for any unicode x? Why do they support a locale I wonder?

davesisson:
For an example where uppercasing will fail, try the Turkish dotted i:

http://www.i18nguy.com/unicode/turkish-i18n.html

parrt:
Yeah, but the *unicode* dotted i should upper case properly as it's not the english i right? toUpperCase should handle it

parrt:
Actually, doesn't work. Turkish reuses english 'i' I think. i get same upper case char here when it should be two diff ones.

```
class T {
     public static void main(String []args) {
        System.out.printf("%d\n", (int)Character.toUpperCase('\u0069'));
        System.out.printf("%d\n", (int)Character.toUpperCase('\u0131'));
     }
}
beast:/tmp $ javac T.java; java T
73
73
```

KvanTTT:
> So Character.toUpperCase(x)) will work for any unicode x? Why do they support a locale I wonder?

The Locale influence only on string `toUpperCase` method, but not on the character `toUpperCase` method (see my code above). I can prepare a table for all Unicode characters.

>For an example where uppercasing will fail, try the Turkish dotted i:

It's failed because of its culture depended, see https://ideone.com/IvtDOd

```
Invariant: 
UPPER_CHAR(i) == I
UPPER_STR(i) == I

Turkish: 
LOWER_CHAR(I) == i
LOWER_STR(I) == ı
UPPER_CHAR(i) == I
UPPER_STR(i) == İ

Deutsch: 
UPPER_CHAR(ß) == ß
UPPER_STR(ß) == SS

Russian: 
UPPER_CHAR(а) == А
UPPER_STR(а) == А
```

parrt:
thanks! So it would seem `option { caseInsensitive=TR; }` or whatever would be needed, not just case insensitive. that makes sense as grammar should indicate the exact characters/sets it is matching. A stream approach hides all of that and makes it easy to mismatch stream / grammar I'd say.

parrt:
I wonder if the simpler solution is just have antlr's IDE alter a rule inplace or have a trivial script that does it.

parrt:
Referring to Mike's:

```
fragment A: 'A' | 'a';
fragment B: 'B' | 'b';
...
KEYWORD1: K E Y W O R D '1';
```

All that is really needed is the fragment rules because:

```
KEYWORD1: K E Y W O R D '1';
```

is no more burden than

```
KEYWORD1: 'keyword1' ;
```

I would say.  I can easily add the fragment generator to the IDE and/or we can simply add that list to the doc for cut/paste.  And we can refer people to these PR for streams and include them in the doc if we want. I think that satisfies everything.  I just can't see beta to **two** chars right + Turkish etc...

KvanTTT:
```ANTLR
KEYWORD1: K E Y W O R D '1';
```

A new unreachable token warning from https://github.com/antlr/antlr4/pull/2066 won't work with such fragment rules.

Ok, you have the freedom to do what you think right. But I won't transform tokens in existing T-SQL, PL/SQL, and other case insensitive grammars.

parrt:
Ok, I'll sleep on it. I think maybe doc is the right way to go on this rather than enforcing a specific vision in the tool.

mike-lischke:
A word about `locale`: don't mix that with `charset`. A `charset` describes a mapping of numbers to charactes. Unicode itself is a charset (however, usually ones speaks of "encoding" instead). A locale is a collection of rules that describe various aspects about a culture/language and has nothing to do how characters are map. A locale determines time + number formats, sorting rules (collation)  and others ([here's the C++11 locale](http://en.cppreference.com/w/cpp/locale/locale) description with more details). This is what makes this so complicated with Unicode. And my question (and suggestion) about focusing on keywords tried to reflect that.

By focusing on keywords we don't need to worry about locales at all. We don't touch text that doesn't need to be case folded. We usually don't have trouble with special cases like the turkish I or the german `ß` (which is called: `ess-zet` or `sharp s` btw.), since they rarely are part of keywords. In general we can simplify things a lot by just focusing on keywords. We don't need comments or strings to be case insensitive, as we never have to match them in a lexer. Converting them in the input stream is just a big waste of CPU cycles and memory slots and therefore also Sam's solution is not good either.

@KvanTTT: by using a new `caseInsensitive` option you wouldn't need to have letter fragments. The lexer internally would handle that. As I wrote before it would probably boil down to implicitly extending the interval sets used to match a character. So, from an implementation standpoint it should be relatively easy to do (and can be done fully in the Java code, no target involved then).

@parrt One thing remains to be clarified then: what is a keyword? That's a weak point in my approach. It's tricky to say *this* is a keyword and *that* is not, in a lexer grammar. You would need to limit where the new option can be applied to. In any case, I would leave out culture/locale processing, as ANTLR is not an NLP tool, but made for programming languages, where some common rules exist (like: keywords use the ASCII charset) that simplify processing. So, one way could be to say: case insensitivity can only be applied to lexer rules that only match ASCII input. That should still cover most usage scenarios, but keeps implementation work small.

parrt:
Thanks everybody! I think I'm going to try out the documentation cut/paste angle ala @mike-lischke's approach and send in a pull request that incorporates streams provided by the other folks as options. then we can comment on that before I pull it in.

parrt:
For your review gentlemen, https://github.com/antlr/antlr4/pull/2146   I have tried to incorporate with cherry picking the resources created by @bramp so that he can get credit for the contribution.

balahsen:
Hi,
I agree with mike-lischke: 
It's about lexer characters for reserved (or key) word for some grammar that is case insensitive like Oracle (PL/SQL).
The lexer should accept upper or lower key word without syntax error for case insensitive grammar.

parrt:
Closing As I'm going the documentation route https://github.com/antlr/antlr4/pull/2146

KvanTTT:
> Converting them in the input stream is just a big waste of CPU cycles and memory slots and therefore also Sam's solution is not good either.

I've made [a benchmark](https://github.com/KvanTTT/AntlrBenchmarks/tree/master/AntlrLetterCaseBenchmark) (C# Runtime) for comparison of grammar and runtime case comparison approaches. It shows that grammar based approach a bit faster:

```
BenchmarkDotNet=v0.12.1, OS=Windows 10.0.17763.2114 (1809/October2018Update/Redstone5)
AMD Ryzen 7 3700X, 1 CPU, 16 logical and 8 physical cores
.NET Core SDK=5.0.101
  [Host]     : .NET Core 3.1.11 (CoreCLR 4.700.20.56602, CoreFX 4.700.20.56604), X64 RyuJIT
  DefaultJob : .NET Core 3.1.11 (CoreCLR 4.700.20.56602, CoreFX 4.700.20.56604), X64 RyuJIT


|                     Method |     Mean |     Error |    StdDev | Ratio |
|--------------------------- |---------:|----------:|----------:|------:|
|   CaseInsensitiveLexerTest | 6.769 ms | 0.0282 ms | 0.0236 ms |  1.00 |
| CaseInsensitiveGrammarTest | 6.258 ms | 0.1018 ms | 0.0903 ms |  0.92 |
```

But, if grammar approach is used, the generated parser is a bit bigger.

> So, one way could be to say: case insensitivity can only be applied to lexer rules that only match ASCII input.

It's not completely true. I know well-known language in Russia (1С) that have [case insensitive cyrillic lexemes](https://github.com/1c-syntax/bsl-parser/blob/c0f9c91b119a1d4341e90a64a7aa8a4d9dc5d585/src/main/antlr/SDBLLexer.g4#L229-L261) (А а Б б В в ...). But I've found a simple solution for not dealing with complex locale case comparison: just do not change case if length of resulting char (actually string) more than 1 or use culture inveriant case changing if possible. It keeps equal length of input and normalized string. For instance, the following string:

```
A a ß Д д
```

is being transformed to

```
A A ß Д Д
```

You can take a look at [my implementation](https://github.com/KvanTTT/DAGE/commit/495b510a10278a2c7e7b9dcc4b43bc44d03fe163#diff-fc0f22c4c027949fac68cc3f5319f651bb94b976878a2175eab5e23dc6ed22c4R6-R9).

Python, JavaScript, PHP require one-length character check, but C#, Java, Go, Dart do correct transfomation using system functions.

PR #2048, opened 2017-10-07
[Java] Add a new CharStream that converts the symbols to upper or lower case.

bramp:
This is useful for many of the case insensitive grammars found at https://github.com/antlr/grammars-v4/ which assume the input would be all upper or lower case. Related discussion can be found at antlr/antlr4test-maven-plugin#1

This is similar to the version for Go https://github.com/antlr/antlr4/pull/2046

### DIFF
--- a/runtime/Java/src/org/antlr/v4/runtime/CaseChangingCharStream.java
+++ b/runtime/Java/src/org/antlr/v4/runtime/CaseChangingCharStream.java
@@ -3,9 +3,13 @@ package org.antlr.v4.runtime;
 import org.antlr.v4.runtime.misc.Interval;
 
 /**
- * This class wraps an existing {@link CharStream} and converts all returned
- * symbols to upper or lower case. This is useful for case-insensitive grammars
- * where all input is assumed to be either upper or lower case.
+ * This class supports case-insensitive lexing by wrapping an existing
+ * {@link CharStream} and forcing the lexer to see either upper or
+ * lowercase characters. Grammar literals should then be either upper or
+ * lower case such as 'BEGIN' or 'begin'. The text of the character
+ * stream is unaffected. Example: input 'BeGiN' would match lexer rule
+ * 'BEGIN' if constructor parameter upper=true but getText() would return
+ * 'BeGiN'.
  */
 public class CaseChangingCharStream implements CharStream {
 
--- a/runtime/Java/src/org/antlr/v4/runtime/CaseChangingCharStream.java
+++ b/runtime/Java/src/org/antlr/v4/runtime/CaseChangingCharStream.java
@@ -1,0 +1,77 @@
+package org.antlr.v4.runtime;
+
+import org.antlr.v4.runtime.misc.Interval;
+
+/**
+ * This class wraps an existing {@link CharStream} and converts all returned
+ * symbols to upper or lower case. This is useful for case-insensitive grammars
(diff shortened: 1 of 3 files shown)


bramp:
Per discussion on #2046 this can be implemented on a case-by-case bases by the users of the grammar.

scadgek:
Why is this closed? Looks like this one is already widely used as a kind of reference implementation.

scadgek:
Probably these changes should be moved to #2046 according to [the last Mr. Parr's comment](https://github.com/antlr/antlr4/pull/2046#issuecomment-341175722).

runtime/Java/src/org/antlr/v4/runtime/CharStreams.java
```diff
@@ -303,4 +303,24 @@ public static CodePointCharStream fromChannel(
 			channel.close();
 		}
 	}
+
+	/**
+	 * Takes the stream and forces all symbols to uppercase.
```
parrt: perhaps say "Takes the stream and forces all symbols to uppercase for lexing purposes but leaves the original text as-is."?

runtime/Java/src/org/antlr/v4/runtime/CaseChangingCharStream.java
```diff
@@ -0,0 +1,77 @@
+package org.antlr.v4.runtime;
+
+import org.antlr.v4.runtime.misc.Interval;
+
+/**
```
parrt: Perhaps we could change the comment as follows?

This class supports case-insensitive lexing by wrapping an existing
{@link CharStream} and forcing the lexer to see either upper or
lowercase characters. Grammar literals should then be either upper or
lower case such as 'BEGIN' or 'begin'. The text of the character
stream is unaffected. Example: input BeGiN would match lexer rule
'BEGIN' if constructor parameter upper=true but getText() would return
BeGiN.

runtime/Java/src/org/antlr/v4/runtime/CharStreams.java
```diff
@@ -303,4 +303,24 @@ public static CodePointCharStream fromChannel(
...
+	 *
+	 * @param in
+	 * @return
+	 */
+	public static CharStream forceUpper(CharStream in) {
```
parrt: Could we use toUpper perhaps? asUpper?

runtime/Java/src/org/antlr/v4/runtime/CharStreams.java
bramp: Done.

runtime/Java/src/org/antlr/v4/runtime/CaseChangingCharStream.java
bramp: SGTM.

runtime/Java/src/org/antlr/v4/runtime/CharStreams.java
bramp: toUpper it is., and changed forceLower to toLower.

KvanTTT:
@bramp could you also add a test with Turkish culture and `I` letter? See this [java sample](http://tpcg.io/inOWDS).

parrt:
I'm closing as I'm going the documentation route per https://github.com/antlr/antlr4/pull/2146

PR #2057, opened 2017-10-14
Split Travis tests into three phases, and add caching and timeout management

ewanmellor:
Split Travis tests into three phases.  This gives a "smoke-test" phase which just runs the Java tests, to get a quick check that compilation is working.  It is followed by the "main-test" phase that tests each target on their preferred platform (e.g. Swift on macOS, Python on Linux) followed by "test-extended" which runs all the remaining tests (all the other Java variants, and any other platforms supported by each target).

This means that those tests in the later phases won't run unless the earlier phases have passed.  This should vastly improve our Travis turnaround.

This PR also changes the matrix so that we get some coverage across all of openjdk7,8 and oraclejdk7,8 and moves the Go and Javascript tests to Trusty just so that they are more up-to-date.  This adds one additional Java test run, so now we compare oraclejdk8 and openjdk8 in case of problems in that regard.

Note that the test matrix has not been extended to cover oraclejdk9.  This currently fails for us with a build issue, so it is not included here.

The C++ tests have been broken into three groups, for reliability.

Also:
* add some caching to improve turnaround speeds;
* add a travis_wait call when running the tests so that Travis does not time out the long-running-but-silent TestLeftRecursion tests.
* remove the meaningless custom apt configuration in the section for macOS .NET tests.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -7,6 +7,11 @@ cache:
   - $HOME/.m2
   - $HOME/Library/Caches/Homebrew
 
+stages:
+  - smoke-test
+  - main-test
+  - extended-test
+
 matrix:
   include:
     - os: linux
@@ -16,6 +21,7 @@ matrix:
         - TARGET=cpp
         - CXX=g++-5
         - GROUP=ALL
+      stage: main-test
       addons:
         apt:
           sources:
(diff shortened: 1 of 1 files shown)


parrt:
@ericvergnaud Can you give this a quick sniff? It seems like an extension of your earlier work to break things into parsing test phases etc.

PR #2063, opened 2017-10-21
Update boot.py and swift-target.md to reflect Swift 4 migration changes.

hanjoes:
Follow up PR for #2032.


hanjoes:
@parrt this is followup updates for @ewanmellor's PR #2032

PR #2081, opened 2017-10-29
Place quotes around the arguments in the error messages.

bramp:
This makes it easier for a human to understand errors such as `symbol type conflicts with generated code in target language or runtime`. It's not the type of the symbol that's at fault, it is that there is a symbol named 'type'.

### DIFF
--- a/tool/src/org/antlr/v4/tool/ErrorType.java
+++ b/tool/src/org/antlr/v4/tool/ErrorType.java
@@ -28,56 +28,56 @@ public enum ErrorType {
 	 *
 	 * <p>cannot write file <em>filename</em>: <em>reason</em></p>
 	 */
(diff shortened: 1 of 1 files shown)


bramp:
I realise this is going to break a lot of tests, that test for the error string. If there is no objection to this change, I'll go through and fix them.

parrt:
Hi. This is a very reasonable idea but I think it's probably too late to change as many tools outside of antlr would rely on specific error format.

KvanTTT:
I think it's easy to fix these tools. There are too many issues with symbol conflicts: https://github.com/antlr/grammars-v4/issues/950, https://github.com/antlr/grammars-v4/issues/857

sharwell:
:memo: The form currently used by ANTLR was an intentional design decision in #653.

KvanTTT:
Ok then.

PR #2085, opened 2017-10-29
Fix the -message-format command line option.

ewanmellor:
ErrorManager.setFormat was being called before the command line options
were parsed in handleArgs.  This meant that setFormat was always called
with the default, and so the command line option never took effect.

This option apparently only worked for 2.5 hours on Sep 6 2012 ;-)

Closes #992.

PR #2086, opened 2017-10-29
Improve the "gnu" message format

ewanmellor:
Improve the "gnu" message format so that the messages are parsed by IDEs (I'm testing with Xcode but other IDEs should appreciate these changes too).

1. Output the full path to the file, when using any message format other than the default "antlr".  This means that IDEs will be able to find the file correctly.
2. Add the column number to the error message.
3. Add a note to the documentation.

Depends upon #2085 , because the non-default message formats don't work without that.


parrt:
I guess since nobody could use that format it's okay to change!

PR #2095, opened 2017-11-01
Support utf8 in win demo

gendalph:
not full support, but more than nothig


parrt:
@mike-lischke Are you okay with this change?

parrt:
@gendalph Can you address Mike's comment?

gendalph:
Use antlr4/runtime/Cpp/demo/Linux/main.cpp linux file for solve win32 console unicode troubles?

And #ifdefs for win part and includes?

gendalph:
The demo is needed for a quick decision whether to use antlr4 further or to find alternatives, and if the limitations and shortcomings are immediately visible, then it is much more difficult to make a decision in favor of antlr4.

A simple fully working cross platform demo is the best solution, but if half the code is platform dependent and the result also depends slightly on the system, then it is probably worth dividing the code into 2 parts?

mike-lischke:
@gendalph Sorry, I don't understand your answers. There's an own file for each platform already. However, with your patch you are adding yet another main.cpp (so we have 4 then) and I don't see what this is good for. The one I created for Windows works out of the box and many people have tested the demo already.

gendalph:
But file present only for example but not part of cmake build.
VC project broken and print errors.

runtime/Cpp/demo/Windows/antlr4-cpp-demo/main.cpp

gendalph:
I agree that you can include the file in the cmake and discard the new file
 set (main_src ${PROJECT_SOURCE_DIR}/demo/Windows/antlr4-cpp-demo/main.cpp)

And use OutputDebugString only for demo and discard patch for console output.

mike-lischke:
Ah I see what you are after. Yes, ok, change the current main.cpp file to use the console if you can get it to print UTF-8 properly. That OutputDebugString call is just a hack as I couldn't get the console to print Unicode at that time. You can remove that once the console output is ok.

parrt:
@gendalph let me know when this is ready. i'd like to push 4.7.1 soon

PR #2132, opened 2017-11-22
Fix Xpath in Cpp runtime 

### LINKED ISSUE #2037: [Cpp] Xpath doesn't work
Xpath does not work in cpp runtime. No matter what path I use, the `evaluate` function always return the base context tree.

As an example, the following code always return the same tree (even if I change the `xpath_enum`) : 

```
std::string xpath_enum = "fakepath";

antlr4::tree::xpath::XPath path(_parser, xpath_enum);

std::vector<antlr4::tree::ParseTree *> res = path.evaluate(context);
```

malickf:
close #2037

The split function does not behave as in other runtime. It should accept a string instead of a file as argument in order to behave as in other runtime. I just did it.

### DIFF
--- a/contributors.txt
+++ b/contributors.txt
@@ -171,3 +171,4 @@ YYYY/MM/DD, github id, Full name, email
 2017/05/29, rlfnb, Ralf Neeb, rlfnb@rlfnb.de
 2017/10/29, gendalph, Максим Прохоренко, Maxim\dotProhorenko@gm@il.com
 2017/11/02, jasonmoo, Jason Mooberry, jason.mooberry@gmail.com
+2017/11/22  Malick F.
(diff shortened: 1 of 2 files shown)


malickf:
As an example, with this fix we can use directly : 

```
  const std::string xpath_str ="*//*/enumerator/Identifier";  											
  antlr4::tree::xpath::XPath path(parser, xpath_str);
```

malickf:
it seems that https://github.com/antlr/antlr4/blob/46b3aa98cc8d8b6908c2cabb64a9587b6b973e6c/runtime/Cpp/runtime/src/tree/xpath/XPath.cpp#L125 doesn't work. 

Is there any unit test with this function ?

malickf:
There is other problems, the

malickf:
The main problem is that this function :  https://github.com/antlr/antlr4/blob/990d4848b265b28af4243ce37734d1bd0d424393/runtime/Cpp/runtime/src/tree/xpath/XPath.cpp#L91
suffers of object slicing. I believe it is required to make a large commit to fix it ...

PR #2133, opened 2017-11-24
[JavaScript] Fix Interval usage in ErrorStrategy.js

zqlu:
I was using the Antlr4 JavaScript version in my project, but get an error related to ErrorStrategy.js.
ErrorStrategy::reportNoViableAlternative method try to new a Interval object, but pass two Token as constructor argument, but Interval class assume the arguments to be numbers. So I fixed by this PR.


ericvergnaud:
@parrt blessed

PR #2066, opened 2017-10-22
New unreachable token value warning

KvanTTT:
Partial fix for #1072 (see unit tests).

I tested my version on [grammars-v4]() repository and found a couple of such warnings:

* warning(184): [LessLexer.g4:224:0](https://github.com/antlr/grammars-v4/blob/master/less/LessLexer.g4#L224): One of the token FORMAT values unreachable. % is always overlapped by token PERC
* warning(184): [ZLexer.g4:478:0](https://github.com/antlr/grammars-v4/blob/master/z/ZLexer.g4#L478): One of the token ID0 values unreachable. _ is always overlapped by token ARGUMENT

Also I discovered and fixed such warnings in ANTLR unit tests itself (testInvalidLexerCommand, testLexerCommandArgumentValidation).

Issue #970, opened 2015-08-04
No error/incorrect code generation when importing lexer grammar with modes into a combined grammar


WalkerCodeRanger:
Using ANTLR 4.5.1, if you import a lexer with modes into a combined grammar it does not report the error that modes are not supported in combined grammars.  Furthermore, it does not output the modes in the `modeNames` field of the lexer.  I assume that means the generated lexer won't handle the modes correctly, though I haven't tested because I have moved on to separate lexer and parser grammars due to all the limitations & issues of combined grammars.

I reported this as a distinct issue from #160 and #256, because the first seems to be about making it actually work and the second is about importing into a lexer grammar.  This issue is just to get a proper error message for importing into combined grammars.

To reproduce, use the following files:

**Mode.g4**

	lexer grammar Mode;
	Char : . -> mode(MODE1);
	mode MODE1;
	WS : ' ' -> mode(DEFAULT_MODE);

**Parse.g4**

	grammar Parse;
	import Mode;
	file : Char* EOF;

Then generate lexer and parser for Parse.g4

niccroad:
At the present time it appears importing a modal lexer grammar into a combined grammar produces an error at latest in, 4.7.0. I think this issue has been addressed elsewhere in the project already.

There is a test case in the merge above which demonstrates this during grammar combination.

Issue #256, opened 2013-05-19
Problems with lexical modes inside an imported grammar


giorgiofran:
Hi All,

 I'm dealing with a big lexer grammar, and I wanted to divide it into subGrammars through the import action.

But, it seems that you cannot split grammars into chunks if you use lexical modes (i.e. you cannot use modes inside the imported grammar).

The question is: is it a bug or is it like that by design?

Regards,
  Giorgio

WalkerCodeRanger:
I would also like the ability to define tokens that exist in multiple modes.  Yes I know about using [-> type()](http://stackoverflow.com/q/15807322/268898).  Still it would be nice.  But rather than the import that @giorgiofran suggests,  I think a better syntax would be to be able to list multiple modes in a section.  For example:

     mode MODE1;
     // mode 1 tokens
     mode MODE2;
     // mode 2 tokens
     mode MODE1, MODE2;
     // tokens that exist in both mode 1 and 2

lawley:
I'm looking at a simple(?) use case where the mode is only referenced by the imported lexer

may-bee:
I would like to see this feature as well. Nevertheless, I tried the workaround with -> type() syntax, which blows up my grammar quite a lot. But the there seems to be an additional problem with this. -> type() semantics does not include the execution of token action of the origin token. Thus, you have to specify your token actions redundantly for each reused token.
Am I right or have I overseen anything related?

sharwell:
> But there seems to be a follow-up issue with this solution as `-> type()` semantics does not include the execution of token action of the origin token. Thus, you have to specify your token actions redundantly for each reused token.

This is correct. This feature was actually added in #408, but we restored the original behavior for referenced rules in #428 because many grammars relied heavily on the original behavior and changing it would make upgrading difficult or impossible.

may-bee:
Is there any chance to see such a feature in the next months? Can you estimate the development effort of a solution like @WalkerCodeRanger has shown some posts before?

parrt:
It's a worthy feature to consider but unfortunately there could be some compatibility issues and also on focusing primarily on both fixing, updating the documentation, building plug-ins at the moment.

niccroad:
I think the original comment by @giorgiofran is addressed by the change to resolve #160 above. Also the grammars posted during 2013 should work with this change (though I didn't try them).

Maybe this issue is now discussing something else however where tokens can now be shared between modes?

Lorthiz:
[...]

Issue #160, opened 2013-02-19
Support importing multi-mode lexer grammars


sharwell:
Note that specific semantics for this feature have not yet been established.

mar9000:
Not, with this addition I get this exception (I'm extending ANTLRv4Lexer).

> Exception in thread "main" java.util.EmptyStackException
	at org.antlr.v4.runtime.Lexer.popMode(Lexer.java:216)
	at org.antlr.v4.runtime.atn.LexerPopModeAction.execute(LexerPopModeAction.java:82)
	at org.antlr.v4.runtime.atn.LexerActionExecutor.execute(LexerActionExecutor.java:193)
	at org.antlr.v4.runtime.atn.LexerATNSimulator.accept(LexerATNSimulator.java:395)
	at org.antlr.v4.runtime.atn.LexerATNSimulator.failOrAccept(LexerATNSimulator.java:328)
	at org.antlr.v4.runtime.atn.LexerATNSimulator.execATN(LexerATNSimulator.java:259)
	at org.antlr.v4.runtime.atn.LexerATNSimulator.match(LexerATNSimulator.java:143)
	at org.antlr.v4.runtime.Lexer.nextToken(Lexer.java:165)

Another difference is that my lexer gets a method NESTED_ARG_ACTION_action() that ANTLR lexer does not have.

zhaohuiyang:
Could anyone please share an update on this issue?

parrt:
Hmm..Unfortunately I'm not going to be able to get to this for the next 4.5.1 release

TheEricWilson:
If the modes are not in the thing being imported, but only in the part doing the import would that work?

parrt:
Actually, you don't need to import a lexer. Just use it as a separate lexer g4 file, right?

TheEricWilson:
Yes the wording of several things seems to imply to me that I can not do includes if I have modes in my lexer.

parrt:
Well, it's true, no includes of modes but it's easier to not import the lexer into the parser. just use as separate file. I'll close for now.

TheEricWilson:
I am trying to import modes into a lexer and not the parser

parrt:
ah. ok, reopening.

may-bee:
Thinking out loud about the semantics of this feature:

* imported tokens/rules/named actions of the default state will be appended to the root grammars default state at the bottom
 * tokens/rules of modes of imported grammars will _not_ be appended to the root grammars default state as it is currently the case.
[...]

PR #2135, opened 2017-11-28
This is one implementation addressing issue 160.

niccroad:
Change to support import of lexer grammars containing modes into other
lexer grammars. The semantics for this are,
   * sets of channels from all grammars are merged
   * rules of modes found in an imported grammar which are in the root
     grammar are merged into the root grammar mode.
   * modes which are not in the root grammar are added to the root
     grammar, excluding modes which become empty due to a re-definition of
     rules in the root grammar.


niccroad:
Hi Terrence Parr/Sam Harwell,
I have been using antlr4 for some grammars I am working on and found a need for importing multi-mode lexer grammars into other lexer grammars. This didn't seem too difficult so I implemented this and some test cases. Would be happy to discuss other behaviour for the import but the implementation here suited my requirements and seemed broadly to implement Terrence Parr's description of import as compositing two grammar files and replacing slave rules with parent rules.

parrt:
@sharwell What do you think of importing lexers with modes? Semantics described seem ok but I'm wondering about edge cases that could screw us. Maybe rule/tokentype ordering?

KvanTTT:
I like this improvement, seems like in-demand: #256, #160.

Also, another issue should be considered: https://github.com/antlr/antlr4/issues/970

@niccroad did you test grammars import with combined grammars?

parrt:
@KvanTTT Could you do me a favor and look at the code to see if it seems right to you? thanks!

KvanTTT:
For sure, but later, until the end of week.

niccroad:
I didn't look at importing modal lexers into combined grammars. Will add a test case for that I think.

parrt:
thanks

niccroad:
The latest change demonstrates importing a modal lexer grammar into a composite grammar. Of course this just generates an error to state that composite grammars should not contain modes. As you can't create a composite grammar containing modes I think this error should be the correct behavior of an import creating such a composite grammar. The test simply checks for the error when the composite grammar is parsed.

niccroad:
"Maybe rule/tokentype ordering?"
Using the nomenclature of master for the importing and slave for imported grammars.

I didn't think that importing the rules and modes creates any additional issues that importing the lexer rules does not already introduce here. Hopefully using the same strategy for ordering the modal rules as for importing other lexer rules accounts for this anyway. I did think that intuitively importing the slave rules before the master rules (not the existing strategy of after) was the better strategy, but presently the import puts the slave rules after the master ones. This means that lexer disambiguation between rules which are a prefix for other lexer rules must happen inside one lexer grammar at present and the slave grammar should not declare tokens which are a suffix of master rules.

Probably once you start breaking up the lexer rules you will want to declare and use a common set of lexer tokens between the master and slave grammars (as I do). The solution for this is to declare your tokens in the slave lexer grammar which results in them having the same token id numbers when they are generated in the master grammar. This means a parser grammar which builds against the slave lexer grammar can also use the master lexer at run-time. Note to this the token ordering is slave before master tokens which is why this works. But if you do want a fixed set of token ids for two related lexer grammars then declaring tokens seems reasonable so I didn't see a need to worry about this if you import without declaring tokens.

Reordering the lexer rules of import seems likely to break a wide variety of existing grammars. 

Additional consideration, my understanding of the potential ordering issues here is also likely to be quite superficial. I managed to start my latest grammars off by declaring my IDENTIFIER rule before all my KEYWORD rules. Then it took me a while to work out why it failed to parse in simple test cases, with the error to the effect of expecting 'class' at 'class'.

KvanTTT:
Overall LGTM. But ideally it would be great to resolve these issues:

* Write at least one test with conflicted channel names.
* Add info about modes and channels merging to [doc](https://github.com/antlr/antlr4/blob/master/doc/grammars.md#grammar-imports).

parrt:
@niccroad That ID before keyword thing is what concerns me but it's already a problem as you say. If you define ID in imported slave, is it perforce "before" any keywords in grammar?

niccroad:
@parrt, No any rules imported into a existing mode are suffixed onto the modes rules list, and the rule import loop didn't change and does this already. Actually this explains why its slave after master at import time.

I think you can run afoul of this strategy if you define, or redefine ID in master which may cause all the slave keywords to match ID however.

parrt:
ok, so any more objections?

niccroad:
Latest change updates the documentation and adds a test case for a channel rule which conflicts. I would be more than happy if my documentation changes get rephrased as necessary (my writing style tends to be more verbose than necessary).

parrt:
ok, guys. thanks! looks good.

niccroad:
@parrt @sharwell 

Have been thinking about the keyword / ID lexer issue a bit further. It might be possible to add a warning when the lexer will completely consume one rule inside another.

The process would happen during the algorithm which converts a lexer NFA into a DFA (I am assuming this happens somewhere in the tool, but it may not be, it may be using Thompson's algorithm on the NFA for several reasons or even the exponential version NFA version).

Anyway where this happens when you identify all the terminal states out of the constructed DFA, you check if another rule is always in the same state set as the terminal rules DFA state set, and has a lower rule number, e.g if ID matches every terminal state of 'keyword' and has a lower rule number. If that is so then the lower rule number state will always match the input of the higher rule number and the higher rule is redundant. My thought is that this should produce a warning.

If you think there is already a NFA to DFA process in the lexer tool to build this into I might give it a go. If not its probably beyond my scope to introduce such a process primarily to produce a warning (unfortunately).

parrt:
I just pulled in a warning for ID vs keyword from @KvanTTT :)

niccroad:
Might have a look at that change and give it a try. A cursory look indicates it doesn't find keywords shadowed by ID's however. I think that needs to be done by finding the unreachable tokens in the DFA created after NFA to DFA conversion (and in Antlr4 per mode obviously). I think Kvan's change will pick up the direct keyword clashes however, which is a start.

Also note, my last parser theory course was more than 10 years ago, and its not something I do day to day. I am probably over simplifying what problems can be solved by that conversion algorithm.

KvanTTT:
>A cursory look indicates it doesn't find keywords shadowed by ID's however.

Yes, my solution from #2066 finds warnings for only simple cases with duplicated tokens. Deeper analysis requires at least DFA's difference algorithm implementation.

The general satisfactory solution will check a difference between token's DFA and throw a warning if it's not an empty, i.e. `A \ B != Ø` where `A` is a token declaration written before the `B` token declaration.

Let check this term in practice.

This sample will throw a warning because of difference between `ID` rule and `A` rule is `[B-Z][A-Z]*`, i.e. not an empty.

```antlr
ID : [A-Z]+;
A : 'A';
```

This sample won't throw a warning because of difference is an empty language.

```antlr
A : 'A';
ID : [A-Z]+;
```

The following sample also will throw a warning:

```antlr
A : 'A';
B : 'A';
```

Despite the fact that ANTLR lexer covers not only regular languages but [context-free](https://github.com/antlr/antlr4/blob/master/doc/lexer-rules.md#recursive-lexer-rules) too I suggest checking only not recursive tokens. Those exclude tokens like this:

```
ACTION : '{' ( ACTION | ~[{}] )* '}' ;
```

Such tokens are rare and do not significantly affect warnings checking.

PR #2137, opened 2017-11-29
Fix circular dependencies

asmarques1990:


parrt:
@ericvergnaud could you check this one? thanks!

ewanmellor:
@asmarques1990 There are 96 new test failures with this change.  At first glance, it looks like everything that used to print with a name now just prints as integer lists.  For example:

    TestSemPredEvalParser>BaseRuntimeTest.testOne:118->BaseRuntimeTest.testParser:170 expected:<([file_ (para (paraContent s) \n \n) (para (paraContent] \n x) \n \n) <EOF>)...> but was:<([[] ([8] ([12 8] s) \n \n) ([9] ([12 9]] \n x) \n \n) <EOF>)...>

ericvergnaud:
Hi, sorry can't bless this:
 - it breaks the tests
 - it relies on hard coded strings
Please use the deferred require pattern, see example with:
require('./ConcreteCategoryDeclaration').resolve();
in top level index.js

PR #2140, opened 2017-12-01
[CSharp] Add a new CharStream that converts the symbols to upper or lower case.

DavidMoraisFerreira:
Ported @bramp's java implementation (#2048 and #2046) for case insensitive grammars to the CSharp runtime.

### DIFF
--- a/runtime/CSharp/runtime/CSharp/Antlr4.Runtime/CaseChangingCharStream.cs
+++ b/runtime/CSharp/runtime/CSharp/Antlr4.Runtime/CaseChangingCharStream.cs
@@ -1,0 +1,47 @@
+﻿using System;
+using Antlr4.Runtime.Misc;
+
+namespace Antlr4.Runtime
+{
+    public class CaseChangingCharStream : ICharStream
+    {
+        private ICharStream stream;
+        private bool upper;
+
+        public CaseChangingCharStream(ICharStream stream, bool upper)
+        {
+            this.stream = stream;
+            this.upper = upper;
+        }
+
+        public int Index => stream.Index;
+
+        public int Size => stream.Size;
+
+        public string SourceName => stream.SourceName;
+
+        public void Consume() => stream.Consume();
+
+        [return: NotNull]
+        public string GetText(Interval interval) => stream.GetText(interval);
+
+        public int LA(int i)
+        {
+            int c = stream.LA(i);
+
+            if (c <= 0) return c;
+
+            var o = Convert.ToChar(c);
+
+            if (upper) return Convert.ToInt32(char.ToUpper(o));
+
+            return Convert.ToInt32(char.ToLower(o));
+        }
+
+        public int Mark() => stream.Mark();
+
+        public void Release(int marker) => stream.Release(marker);
+
+        public void Seek(int index) => stream.Seek(index);
+    }
+}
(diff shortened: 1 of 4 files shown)


runtime/CSharp/runtime/CSharp/Antlr4.Runtime/CaseChangingCharStream.cs
```diff
@@ -0,0 +1,99 @@
...
+            return stream.GetText(interval);
+        }
+
+        public int LA(int i)
+        {
+            int c = stream.LA(i);
+
+            if (c <= 0) return c;
+
+            var o = Convert.ToChar(c);
```
KvanTTT: Please use casing instead of `Convert` methods. It's faster.

```csharp
char o = (char)c;
```

runtime/CSharp/runtime/CSharp/Antlr4.Runtime/CaseChangingCharStream.cs
```diff
@@ -0,0 +1,99 @@
...
+
+            if (upper) return Convert.ToInt32(char.ToUpper(o));
```
KvanTTT: Use an invariant method here: `char.ToUpperInvariant(o)`. Lexer should not depend on users culture.

runtime/CSharp/runtime/CSharp/Antlr4.Runtime/CaseChangingCharStream.cs
```diff
@@ -0,0 +1,99 @@
...
+
+            return Convert.ToInt32(char.ToLower(o));
```
KvanTTT: Casting: `return (int)char.ToLower(o)`.

runtime/CSharp/runtime/CSharp/Antlr4.Runtime/CaseChangingCharStream.cs
```diff
@@ -0,0 +1,99 @@
...
+        public string GetText(Interval interval)
+        {
+            return stream.GetText(interval);
+        }
+
+        public int LA(int i)
+        {
+            int c = stream.LA(i);
+
+            if (c <= 0) return c;
```
KvanTTT: The one-line statement is not used in C# runtime. Please use the following style:

```csharp
if (c <= 0)
{
    return c;
}
```

DavidMoraisFerreira:
Thanks for the review @KvanTTT.

KvanTTT:
Could you also add runtime tests with chars case changing? See related [PR](https://github.com/antlr/antlr4/pull/2046/files) as a sample.

Also, include a test with another culture (Turkish), see [this code fragment](https://www.jdoodle.com/a/fLz) for example

DavidMoraisFerreira:
@KvanTTT I am not sure about which csproj the tests should be added to, as there don't seem to be any CSharp tests. I'd greatly appreciate some pointers :)

ericvergnaud:
Since this PR  addresses a rather generic requirement I would definitely support having target agnostic tests. 
These tests are all in Java, since any end to end test involves Java for the tool. 
From there the Java tests launch the target runtime as a child process.

Envoyé de mon iPhone

DavidMoraisFerreira:
I agree @ericvergnaud. I noticed that this PR didn't make it into #2146. Do I need to make further changes?

parrt:
Hi @DavidMoraisFerreira Just added it now so I am closing this one.

PR #2141, opened 2017-12-03
[JavaScript] enable auto-suggest engines by exposing the lexer ATN as a property, similar to the parser ATN

oranoran:
In order to allow JavaScript clients to auto-suggest completions, they need access to the lexer's ATN, not just the parser's.

See usage in:
https://github.com/oranoran/antlr4-autosuggest-js


ericvergnaud:
blessed

PR #2165, opened 2017-12-15
Add Visual Studio 2017 project to Cpp runtime. Also change define for static library setting to be ANTLR4CPP_STATIC, in all projects.

niccroad:


mike-lischke:
The addition of the VS 2017 project is fine, but why changing the other projects? They are fine as they are.

niccroad:
All the visual studio projects define ANTLR4CPP_EXPORT for static library configurations. They should define ANTLR4CPP_STATIC for static library configurations, I think. Otherwise you can get unresoved external symbols when other projects are linked against the static libraries built by these projects.

mike-lischke:
No, `ANTLR4CPP_EXPORT` is defined for a dynamic lib,  see `antlr4-common.h`. I'd like to keep the dynamic lib approach for the demo.

niccroad:
I only changed the define for the "Debug Static" and "Release Static" configurations of the projects. Are you using one of the static library configurations but linking against the dynamic library?

I think the static define is preferable for the static library configurations as in this case the antlr4cpp projects can be included in some other solution and the appropriate library configurations are built with correct exports. I don't really understand what the issue is with building the static library configurations with ANTLR4CPP_STATIC defined (rather than ANTLR4CPP_EXPORT).

The "Debug DLL" and "Release DLL" configurations are still built with ANTLR4CPP_EXPORT defined.

mike-lischke:
Ah, I see now. I didn't think of looking in which configuration you actually changed the flag.

PR #2166, opened 2017-12-15
Performance improvements (and looks like a bug fix) for C++ runtime.

niccroad:
Performance improvements (and looks like a bug fix) for C++ runtime.
On same grammar and inputs times now 50% less, memory about 25% less.
Probably needs to be tested against more grammars.

Most of the changes are C++ specific however merging a two level indirection hash map lookup to a one level of indirection hash map lookup could probably be applied to all run-times.

I have merged another pull request for a 2017 visual studio project in here already.


runtime/Cpp/runtime/src/atn/PredictionContext.cpp
```diff
@@ -152,16 +152,14 @@ Ref<PredictionContext> PredictionContext::mergeSingletons(const Ref<SingletonPre
...
   }
 
-  Ref<PredictionContext> parentA = a->parent;
-  Ref<PredictionContext> parentB = b->parent;
   if (a->returnState == b->returnState) { // a == b
-    Ref<PredictionContext> parent = merge(parentA, parentB, rootIsWildcard, mergeCache);
+    Ref<PredictionContext> parent = merge(a->parent, b->parent, rootIsWildcard, mergeCache);
 
     // If parent is same as existing a or b parent or reduced to a parent, return it.
-    if (parent == parentA) { // ax + bx = ax, if a=b
```
niccroad: This an the following if statement appears to be a bug with the initial C++ runtime. Its doing pointer comparison, not object equality comparison.

niccroad:
Note, performance improvements are in warm-up phases. For continuous running of a grammar without significant ATNState warm-up this may only a small difference.

runtime/Cpp/runtime/src/atn/PredictionContext.cpp
```diff
@@ -233,15 +204,13 @@ Ref<PredictionContext> PredictionContext::mergeRoot(const Ref<SingletonPredictio
...
     }
     if (a == EMPTY) { // $ + x = [$,x]
-      std::vector<size_t> payloads = { b->returnState, EMPTY_RETURN_STATE };
-      std::vector<Ref<PredictionContext>> parents = { b->parent, nullptr };
-      Ref<PredictionContext> joined = std::make_shared<ArrayPredictionContext>(parents, payloads);
+      std::vector<PredictionContextItem> contexts = {PredictionContextItem(b->parent, b->returnState), PredictionContextItem(nullptr, EMPTY_RETURN_STATE)};
+      Ref<PredictionContext> joined = std::make_shared<ArrayPredictionContext>(contexts);
       return joined;
     }
     if (b == EMPTY) { // x + $ = [$,x] ($ is always first if present)
```
niccroad: Can anybody explain the history of this comment? It appears to contradict the code (in that $ is always added second if present). Its also in the Java run-time. The contradiction should be removed.

runtime/Cpp/runtime/src/atn/PredictionContext.cpp
mike-lischke: This is not a bug. Look at the original Java code:

			// if parent is same as existing a or b parent or reduced to a parent, return it
			if ( parent == a.parent ) return a; // ax + bx = ax, if a=b
			if ( parent == b.parent ) return b; // ax + bx = bx, if a=b

That's a pointer comparison, not object equality. Object equality in Java is done via `parent.equals(a.parent)`. Here the real parent is required. Object equality is used where a set of properties is compared to other objects with that set of properties.

runtime/Cpp/runtime/src/atn/ArrayPredictionContext.cpp
```diff
@@ -10,49 +10,61 @@
...
 }
 
-ArrayPredictionContext::ArrayPredictionContext(std::vector<Ref<PredictionContext>> const& parents_,
-                                               std::vector<size_t> const& returnStates)
-  : PredictionContext(calculateHashCode(parents_, returnStates)), parents(parents_), returnStates(returnStates) {
-    assert(parents.size() > 0);
-    assert(returnStates.size() > 0);
+ArrayPredictionContext::ArrayPredictionContext(std::vector<PredictionContextItem> contexts_)
+  : PredictionContext(calculateHashCode(contexts_)),
+    contexts(std::move(contexts_)) {
```
mike-lischke: A bit unusual to pass in a copy and move that, instead of using a reference and copy that. Any good reason why you didn't stay with the `const&` approach used throught the runtime?

runtime/Cpp/runtime/src/atn/ParserATNSimulator.cpp
```diff
@@ -451,7 +451,7 @@ std::unique_ptr<ATNConfigSet> ParserATNSimulator::computeReachSet(ATNConfigSet *
 
   // First figure out where we can reach on input t
   for (auto &c : closure_->configs) {
-    if (is<RuleStopState *>(c->state)) {
+    if (c->state->getStateType() == ATNState::RULE_STOP) {
```
mike-lischke: This is something I wanted to change for quite some time already! Now make it a `switch`...

runtime/Cpp/runtime/src/atn/PredictionContext.cpp
```diff
@@ -101,48 +100,46 @@ Ref<PredictionContext> PredictionContext::merge(const Ref<PredictionContext> &a,
...
+  if (a->isSingleton() && b->isSingleton()) {
+    return mergeSingletons(std::static_pointer_cast<SingletonPredictionContext>(a),
+                           std::static_pointer_cast<SingletonPredictionContext>(b), rootIsWildcard, mergeCache);
   }
 
   // At least one of a or b is array.
   // If one is $ and rootIsWildcard, return $ as * wildcard.
   if (rootIsWildcard) {
-    if (is<EmptyPredictionContext>(a)) {
+    if (a->isEmpty()) {
```
mike-lischke: Do you have any evidence that using a virtual function is actually faster than a dynamic cast? Would be interesting to see some numbers which support this switch.

Btw, the name `isEmpty()` is not wisely choosen as it seems to indicate the context might be empty (regardless of it's type). It is probably clearer to call that `isEmptyContext` or similar, instead. Sounds like a tiny problem, but I'd like to keep the original semantic (the test is about a type check, not it's content). Not a strong required, however.

runtime/Cpp/runtime/src/atn/PredictionContext.cpp
```diff
@@ -180,38 +175,14 @@ Ref<PredictionContext> PredictionContext::mergeSingletons(const Ref<SingletonPre
...
-      Ref<PredictionContext> a_ = std::make_shared<ArrayPredictionContext>(parents, payloads);
-      if (mergeCache != nullptr) {
-        mergeCache->put(a, b, a_);
-      }
-      return a_;
-    }
-
-    // parents differ and can't merge them. Just pack together
-    // into array; can't merge.
-    // ax + by = [ax,by]
```
mike-lischke: Tbh, I don't understand this removal. In what sense does it change (improve) handling, compared to the original code?

runtime/Cpp/runtime/src/atn/PredictionContext.cpp
mike-lischke: This is something Ter or Sam must answer.

runtime/Cpp/runtime/src/atn/PredictionContext.cpp
```diff
@@ -252,130 +221,100 @@ Ref<PredictionContext> PredictionContext::mergeArrays(const Ref<ArrayPredictionC
   const Ref<ArrayPredictionContext> &b, bool rootIsWildcard, PredictionContextMergeCache *mergeCache) {
 
   if (mergeCache != nullptr) {
-    auto existing = mergeCache->get(a, b);
-    if (existing) {
-      return existing;
-    }
-    existing = mergeCache->get(b, a);
-    if (existing) {
+    if (const Ref<PredictionContext>& existing = mergeCache->get(a, b)) {
```
mike-lischke: You are ignoring the (b, a) combination here. Doesn't that significantly change the merge algorithm?

runtime/Cpp/runtime/src/atn/PredictionContext.cpp
```diff
@@ -252,130 +221,100 @@ Ref<PredictionContext> PredictionContext::mergeArrays(const Ref<ArrayPredictionC
...
-      }
-      else { // ax+ay -> a'[x,y]
-        Ref<PredictionContext> mergedParent = merge(a_parent, b_parent, rootIsWildcard, mergeCache);
-        mergedParents[k] = mergedParent;
-        mergedReturnStates[k] = payload;
+  std::vector<PredictionContextItem> contexts;
+  contexts.reserve(a->contexts.size() + b->contexts.size());
+  std::vector<PredictionContextItem>::const_iterator i = a->contexts.begin(), i_end = a->contexts.end();
+  std::vector<PredictionContextItem>::const_iterator j = b->contexts.begin(), j_end = b->contexts.end();
+  while (i != i_end && j != j_end) {
```
mike-lischke: Looks good, however, can you save the original comments, at least those that describe the operation (e.g. `// ax+ay -> a'[x,y]`)?

runtime/Cpp/runtime/src/atn/PredictionContext.cpp
```diff
@@ -252,130 +221,100 @@ Ref<PredictionContext> PredictionContext::mergeArrays(const Ref<ArrayPredictionC
...
-
-  // ml: this part differs from Java code. We have to recreate the context as the parents array is copied on creation.
-  if (combineCommonParents(mergedParents)) {
-    mergedReturnStates.resize(mergedParents.size());
-    M = std::make_shared<ArrayPredictionContext>(mergedParents, mergedReturnStates);
+  if (antlrcpp::Arrays::equals(contexts, b->contexts)) {
+      if (mergeCache != nullptr) {
+          mergeCache->put(a, b, b);
+      }
+      return b;
```
mike-lischke: Here I am again at a point where I cannot understand what you changed and can only hope you know what you are doing. The patch creates bigger differences (also in naming) compared to the original Java code, but maybe this is the price we have to pay to make this more "C++-ish"...

runtime/Cpp/runtime/src/atn/PredictionContext.h
```diff
@@ -10,243 +10,263 @@
...
-    std::vector<std::string> toStrings(Recognizer *recognizer, int currentState);
-    std::vector<std::string> toStrings(Recognizer *recognizer, const Ref<PredictionContext> &stop, int currentState);
-  };
-
-  struct PredictionContextHasher {
-    size_t operator () (const Ref<PredictionContext> &k) const {
-      return k->hashCode();
+    namespace atn {
+
+        struct PredictionContextHasher;
```
mike-lischke: Looks like you indented the atn namespace block. This was not so before by intention. Please revert that.

runtime/Cpp/runtime/src/atn/PredictionContext.cpp
niccroad: Sorry my bad, I should have checked that against the original run-time.

runtime/Cpp/runtime/src/atn/ArrayPredictionContext.cpp
niccroad: The intention here is to avoid a call to the vector copy constructor (which will copy the contents). There is a related move in PredictionContext.cpp,

Ref<ArrayPredictionContext> M = std::make_shared<ArrayPredictionContext>(std::move(contexts));

runtime/Cpp/runtime/src/atn/PredictionContext.cpp
niccroad: I have added some comments to make this more clear. Lookup of the merge cache is now symmetric, 

`mergeCache->get(a, b) == mergeCache->get(b, a)`

runtime/Cpp/runtime/src/atn/PredictionContext.cpp
niccroad: Have put these comments back in and added the same to mergeSingletonIntoArray as well. I also modified mergeSingletonIntoArray so it should be doing exactly the same for a single item context as mergeArrays does not a many item context. In particular for empty the parents of a merged item don't need to be the same.

runtime/Cpp/runtime/src/atn/PredictionContext.cpp
niccroad: I renamed a few of the iterators and put back the original comments in this regard. Hopefully that makes it more clear what the lineage of the code is. In particular there is a new merge method specifically for merging a singleton into an array. The merge method for this is based on mergeArrays but by doing this there is no need to allocate a temporary ArrayPredictionContext in order to call mergeArrays.

runtime/Cpp/runtime/src/atn/PredictionContext.cpp
niccroad: This block seemed un-necessary to me. I have updated it slightly so the parent object of a,b is used in case that happens, but I can't see how two copies of the same parent can exist given the context caching.

runtime/Cpp/runtime/src/atn/PredictionContext.cpp
niccroad: isEmpty was a pre-existing method, but will change that as its ripe for miss-interpretation. I already at first missed the subtle point that an EmptyPredictionContext has a SingletonPredictionContext as a base class.
My evidence for the virtual function calls was that they were showing up in my VC++ profile runs. But if you know how dynamic_cast is implemented then the virtual function calls should be at least as fast as any dynamic_cast. Its possible that gcc/clang will figure out that the result of dynamic_cast is only being type checked against at high optimization levels but not on my release build profiles. I was going to use typeid() but thought this might cause subtle issues after any further changes to this class hierarchy.
Will try with/without and produce some numbers to support this change however.

niccroad: I compiled versions of the code with is<> and with isEmptyContext()/isSingletonContext() just for the code in PredictionContext.cpp. While I really noticed the is calls in ParserATNSimulator.cpp the timings indicate that the virtual function calls are an improvement here as well.
`
                             Virtual               dynamic_cast
Example 1(SLL)     806ms                840ms
                 (LL)      980ms                1019ms
Example 2(SLL)    2755ms               2826ms
                 (LL)     4951ms               4993ms
Example 3(SLL)    959ms                 1028ms
                 (LL)     20369ms             21239ms
Example 4 (SLL)   145ms                 152ms
                 (LL)    161919ms           166516ms
`
This is what I would expect, virtual function calls should be slightly faster than dynamic_cast in general.

niccroad:
I think all the comments have been addressed in some form by the latest push. At the same time I checked out the virtual function calls I also compared C++ runtime 4.7.2 with and without these changes.
                                  Optimized          Original
Example1 (SLL)             807ms              1487ms
                 (LL)              980ms              1821ms            
Example2 (SLL)             2755ms            5030ms
                 (LL)               4951ms           9111ms
Example3 (SLL)             959ms               1746ms            
                 (LL)              20369ms           38550ms          
Example4 (SLL)            145ms               299ms               
                 (LL)              160919ms         371028ms        

Again averages of 5 runs. So the version here runs in around 50% of the original time on the same grammar and examples.

runtime/Cpp/runtime/src/atn/ArrayPredictionContext.cpp
mike-lischke: You are making a copy, when you call the function. I don't understand what your goal is here. And I don't see any need to change the established approach. By using `const&` in the function signature you avoid a copy, but later on you have to copy anyway (to keep a local duplicate). I don't see a reason why we should change that.

mike-lischke:
Yes, looks pretty good now. Only the VS project changes should be removed and that little `const&` issue be solved. After that we are good to go, I think.

Regarding the numbers you posted: that's the improvements we get by your patch? That's pretty impressive! I'm eager to try this out once we got this patch merged.

niccroad:
I am going to check the performance difference using move construction of the vector vs copy construction to try to resolve that discussion (as I think its significant).

I have seen around 50% less time performance improvements on the same grammar (derived from Cpp14 grammar example) and files. May main caveat being my machine is probably not so good execution environment for a parser, its not a large lap top with limited cores and memory. But this should probably benefit a better equipped machine also.
It may be targeting a poorly optimized grammar as well, especially regarding the warm-up. The Cpp14 grammar example still very much looks like the Cpp14 appendix.

runtime/Cpp/runtime/src/atn/PredictionContext.cpp
```diff
@@ -159,10 +159,10 @@ Ref<PredictionContext> PredictionContext::mergeSingletons(const Ref<SingletonPre
     Ref<PredictionContext> parent = merge(a->parent, b->parent, rootIsWildcard, mergeCache);
 
     // If parent is same object as existing a or b parent or reduced to a parent, return it.
-    if (parent == a->parent) { // ax + bx = ax, if a=b
+    if (*parent == *a->parent) { // ax + bx = ax, if a=b
       return a;
     }
-    if (parent == b->parent) { // ax + bx = bx, if a=b
+    if (*parent == *b->parent) { // ax + bx = bx, if a=b
       return b;
```
mike-lischke: Now the wrong comparison is back?

niccroad: Sorry, was experimenting with that and left it in.

runtime/Cpp/runtime/src/atn/ArrayPredictionContext.cpp
niccroad: After adding the proper rvalue reference semantics this doesn't hit the vector copy constructor here, which was the initial intention. Unfortunately, I didn't quite understand all the move semantics stuff in my first attempt.

niccroad:
Somebody with CI build permissions will need to re-run this job at some stage. I don't think I have permissions to do this. The failure appears to be just a temporary unavailable site for some of the install packages.

mike-lischke:
You don't see the restart button in travis, when you click on `Details`?
restart

WalterCouto:
In the original PredictionContext::mergeArrays there was a bug on line 281 where the logic differed from the java runtime:

``

WalterCouto:
In the orignal PredictionContext::mergeArrays there was a bug on Line 281 where the logic differs from java:  It currently is:

`bool both$ = payload == EMPTY_RETURN_STATE && a_parent && b_parent;`

and should instead match java as: 

`bool both$ = payload == EMPTY_RETURN_STATE && !a_parent && !b_parent;`


The new code in here replicates the same logic error.  This bug will cause the `assert(a & B)` in the merge method to fire when the case happens.

parrt:
@mike-lischke so this is approved and I should merge?

mike-lischke:
@parrt @niccroad We have to fix that regression yet, which @WalterCouto mentioned (good catch btw. Walter).

WalterCouto:
I have fixed the bug reported above in the original code from master in PR #2445. Haven't had the time ti see if can see where a corresponding fix would be in the modified code in the PR.

mike-lischke:
Thanks @WalterCouto, however @niccroad should include the fix in this PR as we are merging *his* patch at the end. Maybe I should have been more clear in my previous comment.

runtime/Cpp/runtime/src/atn/PredictionContext.cpp
```diff
@@ -252,130 +230,123 @@ Ref<PredictionContext> PredictionContext::mergeArrays(const Ref<ArrayPredictionC
...
+    std::vector<PredictionContextItem>::const_iterator i = a->contexts.begin(), // walks a
+                                                       iEnd = a->contexts.end();
+    std::vector<PredictionContextItem>::const_iterator j = b->contexts.begin(), // walks b
+                                                       jEnd = b->contexts.end();
+    while (i != iEnd && j != jEnd) {
+      if (i->returnState == j->returnState) {
+        // same payload (stack tops are equal), must yield merged singleton
+        size_t payload = i->returnState;
+        // $+$ = $
+        if (i->parent && j->parent && (payload == EMPTY_RETURN_STATE || *i->parent == *j->parent)) {
```
WalterCouto: @mike-lischke and @niccroad, This line replicates the bug in line 281 which would have stated `bool both$ = payload == EMPTY_RETURN_STATE && !a_parent && !b_parent;`  So one suggestion for fixing this bug in this version would be to have the following line 252 be modified to:
```
bool both$ = payload == EMPTY_RETURN_STATE && !(i->parent) && !(j->parent);
bool ax_ax = (i->parent && j->parent) && *i->parent == *j->parent; // ax+ax -> ax
if (both$ || ax_ax) {
```

runtime/Cpp/runtime/src/atn/PredictionContext.h
```diff
@@ -217,36 +224,62 @@ namespace atn {
...
     std::string toString() const;
     size_t count() const;
 
   private:
-    std::unordered_map<Ref<PredictionContext>,
-      std::unordered_map<Ref<PredictionContext>, Ref<PredictionContext>, PredictionContextHasher, PredictionContextComparer>,
-      PredictionContextHasher, PredictionContextComparer> _data;
-
+    Ref<PredictionContext> _missing;
+    std::unordered_map<PredictionContextPair, Ref<PredictionContext>> _data;
```
garzon: It seems that I have to add copy constructor to PredictionContextPair `PredictionContextPair::PredictionContextPair(const PredictionContextPair& o): lhs(o.lhs), rhs(o.rhs) { }` to make `std::unordered_map<PredictionContextPair, Ref<PredictionContext>>` works for MSVC2017?

runtime/Cpp/runtime/src/atn/PredictionContext.cpp
jwakely: > Do you have any evidence that using a virtual function is actually faster than a dynamic cast? Would be interesting to see some numbers which support this switch.

It should always be no slower, and probably faster. A virtual call just looks up an entry in the vtable and calls it. A `dynamic_cast` has to do a lot more work to handle upcasts, downcasts, multiple inheritance hierarchies, etc. Even for a simple inheritance heirarchy, the C++ runtime still tends to implement dynamic cast using a complex algorithm that can handle the general case of complex hierarchies.
https://wyfeizj.wordpress.com/2011/03/25/the-rtti-implementation-of-vs2010/
https://tinodidriksen.com/2010/04/cpp-dynamic-cast-performance/ (I have no idea why this benchmark uses `reinterpret_cast` instead of `static_cast`, that is a smell, but doesn't affect the result in these specific tests).

A virtual function call is a lot easier to optimize too, as it's eligible for inlining via full devirtualization, or speculative devirt. http://hubicka.blogspot.com/2014/02/devirtualization-in-c-part-4-analyzing.html

PR #2168, opened 2017-12-17
Add missing -std=c++11 flag when using non-appleclang compiler on macosx

ajaypanyala:
When using non-appleclang compiler on macosx, i.e when using upstream (llvm.org) Clang,   
```
if("${CMAKE_VERSION}" VERSION_GREATER 3.1.0)
  set(CMAKE_CXX_STANDARD 11)
  set(CMAKE_CXX_STANDARD_REQUIRED ON)
 ....
```
 does not add -std=c++11 to the list of compile flags when building the Cpp runtime sources causing the build to fail. This happens when I am using any cmake version >= 3.1.

An alternative fix to the one suggested in this PR is to cmake_minimum_required (VERSION 3.0) instead of 2.8. Not sure why exactly, but the setting minimum version to 3.0 fixes the missing c++11 flag problem (i.e. I would not need to explicitly add the flag as done in this PR).


ajaypanyala:
Removed the comment. Thanks!

ajaypanyala:
@mike-lischke Could you please restart the one testcase that failed. From the build log, it's not clear why it fails.

PR #2176, opened 2017-12-19
This resolves the recently opened issue 2175 (a memory leak).

niccroad:
Function should still make the same calculation but without leaking memory in the process.

### DIFF
--- a/runtime/Cpp/runtime/src/atn/ParserATNSimulator.cpp
+++ b/runtime/Cpp/runtime/src/atn/ParserATNSimulator.cpp
@@ -697,6 +697,7 @@ std::vector<Ref<SemanticContext>> ParserATNSimulator::getPredsForAmbigAlts(const
 
 std::vector<dfa::DFAState::PredPrediction *> ParserATNSimulator::getPredicatePredictions(const antlrcpp::BitSet &ambigAlts,
   std::vector<Ref<SemanticContext>> altToPred) {
+  bool containsPredicate = false;
   for (size_t i = 1; i < altToPred.size(); i++) {
     const Ref<SemanticContext>& pred = altToPred[i];
 
@@ -704,11 +705,14 @@ std::vector<dfa::DFAState::PredPrediction *> ParserATNSimulator::getPredicatePre
     assert(pred != nullptr);
 
     if (pred != SemanticContext::NONE) {
-      return std::vector<dfa::DFAState::PredPrediction*>();
+      containsPredicate = true;
+      break;
     }
   }
 
   std::vector<dfa::DFAState::PredPrediction*> pairs;
+  if (!containsPredicate) return pairs;
+
   for (size_t i = 1; i < altToPred.size(); i++) {
     if (ambigAlts.test(i)) {
       pairs.push_back(new dfa::DFAState::PredPrediction(altToPred[i], (int)i)); /* mem-check: managed by the DFAState it will be assigned to after return */
--- a/runtime/Cpp/runtime/src/atn/ParserATNSimulator.cpp
+++ b/runtime/Cpp/runtime/src/atn/ParserATNSimulator.cpp
@@ -1,4 +1,4 @@
-﻿/* Copyright (c) 2012-2017 The ANTLR Project. All rights reserved.
+/* Copyright (c) 2012-2017 The ANTLR Project. All rights reserved.
  * Use of this file is governed by the BSD 3-clause license that
  * can be found in the LICENSE.txt file in the project root.
  */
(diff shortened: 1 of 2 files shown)


runtime/Cpp/runtime/src/atn/ParserATNSimulator.cpp
```diff
@@ -697,24 +697,22 @@ std::vector<Ref<SemanticContext>> ParserATNSimulator::getPredsForAmbigAlts(const
...
 
     // unpredicted is indicated by SemanticContext.NONE
     assert(pred != nullptr);
 
-    if (ambigAlts.test(i)) {
-      pairs.push_back(new dfa::DFAState::PredPrediction(pred, (int)i)); /* mem-check: managed by the DFAState it will be assigned to after return */
-    }
     if (pred != SemanticContext::NONE) {
-      containsPredicate = true;
+      return std::vector<dfa::DFAState::PredPrediction*>();
```
garzon: Please note that this line of code is wrong. The whole thing should be something like
```cpp
std::vector<dfa::DFAState::PredPrediction*> pairs;
    bool containsPredicate = false;
    for (size_t i = 1; i < altToPred.size(); i++) {
      Ref<SemanticContext> pred = altToPred[i];
  
      // unpredicted is indicated by SemanticContext.NONE
      assert(pred != nullptr);
  
      if (pred != SemanticContext::NONE) {
        containsPredicate = true;
      }
    }
    
    if (!containsPredicate) return pairs;
    
    for (size_t i = 1; i < altToPred.size(); i++) {
      if (ambigAlts.test(i)) {
        pairs.push_back(new dfa::DFAState::PredPrediction(altToPred[i], (int)i)); /* mem-check: managed by the DFAState it will be assigned to after return */
      }
    }
  
    return pairs;
```

niccroad: Doh, Thanks!

runtime/Cpp/runtime/src/atn/ParserATNSimulator.cpp
```diff
@@ -697,24 +697,26 @@ std::vector<Ref<SemanticContext>> ParserATNSimulator::getPredsForAmbigAlts(const
...
 
     // unpredicted is indicated by SemanticContext.NONE
     assert(pred != nullptr);
 
-    if (ambigAlts.test(i)) {
-      pairs.push_back(new dfa::DFAState::PredPrediction(pred, (int)i)); /* mem-check: managed by the DFAState it will be assigned to after return */
-    }
     if (pred != SemanticContext::NONE) {
       containsPredicate = true;
+      break;
```
garzon: I don't think using `break` here is a good idea, since it is supposed to check all `pred` is not `nullptr`.

mike-lischke: It just needs to find one predicate that is non NONE.

garzon: @mike-lischke I mean there is still an assertion statement `assert(pred != nullptr);` above in the loop and the `break` makes the check not to cover all `pred`, in this commit. Whatever, the "c++-ish" solution you mentioned is good with it.

mike-lischke: The assertion is probably not needed anymore. But it doesn't hurt either.

The first loop is fine because all it needs to check is that at least one non-NONE context is in the list. So it can break on the first one it finds.

garzon: @mike-lischke I know that, but I think you didn't get my point. I mean, if a `pred` is not NONE and `break` here in this case, the assertion will not be executed for the other `pred`s afterwards.

niccroad: I have stuck the assert in the second loop. But in general relying on asserts for correctness is very bad style, so the function flow should not be altered to trigger an assert (I think).

niccroad:
Yes. Looks good to me.

garzon:
@niccroad looks good now, though the assertion is duplicated. But it seems it is not supposed to be stuck at the assertion and everything seems okay for me.

niccroad:
@mike-lischke Wasn't sure what the intention was with your comment to the commit (You should also be allowed to commit changes here directly as a reviewer) but I have merged it into this pull request now.

mike-lischke:
OK, thanks. Good to go now.

niccroad:
Does anybody know a way to restart a failed  job on Circle-CI for a temporary problem?

mike-lischke:
When you open the `Details` link you see the main build page. On the far right there's a button to re-trigger the build. I just did that.

mike-lischke:
This PR can be closed. I added it in my PR #2383 (more by accident as I forgot I had the patch merged already to my repo).

PR #2184, opened 2017-12-27
[Python3] return result instead of None in visitor

jkmar:
Previously ParseTreeVisitor in the visitChildren method returned None when it shouldn't visit the next child, so if you overrode the shouldVisitNextChild you would also have to override the visitChildren method only to change that return statement.


ericvergnaud:
Hi, thanks for this.
@parrt blessed

parrt:
Does the same  Bug occur in Python 2 as well?

ericvergnaud:
Indeed the same fix needs to be applied.
@jkmar may I ask that you make it part of this PR?

jkmar:
@ericvergnaud sure, I've also applied the same fix to Python 2.

PR #2187, opened 2017-12-30
Force usage of 'previousContext'.

kosak:
Some generated grammars define but don't use the variable 'previousContext'.

This is a problem for customers who are compiling with warnings, e.g.
-Wall -Werror.

Those customers get this message from gcc:
  error: variable ‘previousContext’ set but not used [-Werror=unused-but-set-variable]

An easy fix is to just force the compiler to consider this variable always
used.


kosak:
Hi, friendly ping? Would you be willing to accept this change?

parrt:
Done manually since no contributors.txt signature.

PR #2190, opened 2018-01-01
fixes #2188

### LINKED ISSUE #2188: Python parser generation: rule can't have same name as python identifier
If a grammar contains a rule with the same name as a python keyword, then the generated parser will contain a function with the name of a python keyword - this creates a python syntax error.

For example, if the grammar `Test` has the rule `if: IDENTIFIER;` then in TestLangParser.py the following code is generated

``` python
   class IfContext(ParserRuleContext):
            #...
            def if(self): # Syntax error: identifier expected
                    localctx = TestLangParser.IfContext(self, self._ctx, self.state)
                   #...
```

Stack trace when running antlr4-python3-runtime:

```
Traceback (most recent call last):
  File "C:/Users/Ben/Desktop/benlang/grammar/parser.py", line 3, in <module>
    from BenLangVisitor import BenLangVisitor
  File "C:\Users\Ben\Desktop\benlang\grammar\BenLangVisitor.py", line 6, in <module>
    from BenLangParser import BenLangParser
  File "C:\Users\Ben\Desktop\benlang\grammar\BenLangParser.py", line 338
    def if(self):
         ^
SyntaxError: invalid syntax
```

Exact grammar extract:
```
grammar BenLang;		
prog:		statement progRest;
progRest:	SEMICOLON statement progRest | ;
statement: 	expr | bool_expr;

if:		IF LBRACKET bool_expr RBRACKET;
```

Python version: 3.6
Antlr version: 4.7.1

ericvergnaud:
List of keywords was incomplete


ericvergnaud:
@parrt please also merge this

parrt:
thanks, @ericvergnaud !!

PR #2191, opened 2018-01-01
move python 2 back to main test stage since it’s a standalone target

ericvergnaud:
no code change


ericvergnaud:
@parrt please also merge this

PR #2192, opened 2018-01-01
Use csharp as main test target

ericvergnaud:
since it is faster


ericvergnaud:
@parrt please also merge this

PR #2198, opened 2018-01-06
fix indentation in javascript target examples

kasbah:
I could maybe add the needed `require` statements at the top as well.


ericvergnaud:
@parrt blessed

PR #2200, opened 2018-01-11
enable Precompiled headers for faster compiles

chund:
enable PrecompiledHeader and MultiProcessorCompilation for faster builds in Visual Studio.
maybe someone could also add this to the CMake build or other platforms if needed.


niccroad:
You can use the /FI flag to always include a particular header at the start of every source file. This means no change to the code is necessary for PCH builds.

PR #2202, opened 2018-01-15
Support for CMake Packages in the Cpp Runtime

xgcssch:
Hello!
I added support for the use of the CMake Package feature in the Cpp Runtime.
The documentation is included in the commits.

Further included is a small fix for the FreeBSD platform. The default compiler ther is Clang, but it was only detected on Linux based systems.
I was able to test the Packages with Windows/VS2017 and FreeBSD/Clang.

Best regards
Sönke


xgcssch:
It turned out that the CMake function didn't work for combined lexer/parser files.
I will open another pull Request. 
Sorry

Issue #2211, opened 2018-01-25
VS2017 link error when using antlrcpp::Any


EternalPhane:
Hello. I'm trying to use `antlrcpp::Any` and got the following errors when building with VS2017:
```
error LNK2019: unresolved external symbol "public: virtual __thiscall antlrcpp::Any::Base::~Base(void)" (??1Base@Any@antlrcpp@@UAE@XZ) referenced in function "public: virtual __thiscall antlrcpp::Any::Derived<int>::~Derived<int>(void)" (??1?$Derived@H@Any@antlrcpp@@UAE@XZ)
fatal error LNK1120: 1 unresolved externals
```
Test.cpp:
```cpp
#include <antlr4-runtime.h>
#include <iostream>

int main()
{
	int i = 0;
	antlrcpp::Any test(i);
	std::cout << test.as<int>() << std::endl;
	return 0;
}
```
However, if I move the definition of `antlrcpp::Any::Base::~Base()` inline, it works as expected. I don't know whether this is a bug of VS2017 or my usage of `antlrcpp::Any` is wrong in the above code. Could you please address this issue?

Chengifei:
I have the same problem, a wild guess might be the symbol is being not exported.

kfwalther:
I also had the same problem when extending the TParserBaseVisitor class with my own visitor class. The antlrcpp::Any::Base::~Base() in the .cpp file isn't seen by the linker for some reason. As @EternalPhane suggested, copying the destructor definition inline into my custom Visitor class header file fixed the issue. I've only tested on VS2017, so not sure if the other VS versions are seeing the same problem. Maybe @mike-lischke can provide insight on the issue.

mike-lischke:
Unfortunately, I missed the fact that the Any class is not exported, when I approved the patch which changed the d-tor. Previously it was a header only class. In fact I think now that move of the d-tor to a cpp file was unnecessary, mea culpa. So, the best solution is to move the d-tor back to the header.

EternalPhane:
@mike-lischke I've made a pr for this issue. Please check #2213. Thanks.

siliconvoodoo:
it still doesn't work but now with the move constructor for some reason. and a bunch of other functions.

`error LNK2001: unresolved external symbol "__declspec(dllimport) public: __cdecl antlrcpp::Any::Any(struct antlrcpp::Any &&)" (__imp_??0Any@antlrcpp@@QEAA@$$QEAU01@@Z)
[...]

PR #2213, opened 2018-01-29
Fix #1855, #2211

### LINKED ISSUE #1855: Support passing std::unique_ptr through antlrcpp::Any

EternalPhane:
Make `clone()` function of `antlrcpp::Any::Derive` return either new `Derive*` or `nullptr` depending on whether template type is copy-constructible.

Related: #1855


daniele-niero:
Hello,

Is this already merged to the master branch? I have compiled the cpp runtime and I still have the problem.

mike-lischke:
No it's not merged yet.

EternalPhane:
Ping...

mike-lischke:
@parrt This is a pure C++ patch and all checks passed. Can you please merge?

PR #2223, opened 2018-02-11
Add an Automatic-Module-Name

### LINKED ISSUE #2163: Please publish JDK 9 Automatic-Module-Name entries
The jar files (particularly for the runtime) should contain `Automatic-Module-Name` entries so that modular JDK 9 projects have a stable name to use in module `requires` clauses. I'd suggest `org.antlr.antlr4.runtime` as this is close to what you're already using as the OSGi symbolic name, but with the hyphen replaced with a dot (module names at the language level can't contain hyphens).

I can submit a PR to make the changes if you like. My current interest in this is getting [jgrapht modularized](https://github.com/jgrapht/jgrapht/issues/448).

io7m:
This adds an Automatic-Module-Name entry to the runtime jar in order to
provide a stable name upon which other modules can depend. The module
name chosen was "org.antlr.antlr4.runtime". This closely matches the
Maven artifact name with the obvious change that the module name
doesn't contain a hyphen (hyphens can't be used in module names at
the language level in Java).

I have signed the contributor agreement in a separate commit below.

Fix #2163

### DIFF
--- a/contributors.txt
+++ b/contributors.txt
@@ -180,3 +180,4 @@ YYYY/MM/DD, github id, Full name, email
 2017/12/03, oranoran, Oran Epelbaum, oran / epelbaum me
 2017/12/20, kbsletten, Kyle Sletten, kbsletten@gmail.com
 2017/12/27, jkmar, Jakub Marciniszyn, marciniszyn.jk@gmail.com
+2018/02/11, io7m, Mark Raynsford, code@io7m.com
--- a/runtime/Java/pom.xml
+++ b/runtime/Java/pom.xml
@@ -96,6 +96,7 @@
 						<phase>process-classes</phase>
 						<configuration>
 							<instructions>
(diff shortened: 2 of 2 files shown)


parrt:
hiya. as we don't support even java 8 yet, is there any chance this will mess up the jar for < java 9?

io7m:
'Ello.

No chance of compatibility problems. It's just a manifest attribute and it's ignored on anything prior to Java 9.

parrt:
ok, sounds good to me. thanks.

io7m:
Thanks!

io7m:
@parrt Any chance I could nudge you into releasing 4.7.2? I can only see one thing left on the milestones and that already seems to be resolved based on the conversation in the ticket.

parrt:
It's an involved process and it's not my focus at the moment. Sorry!

PR #2227, opened 2018-02-18
fix git checkout to tag usage

songtianyi:


parrt:
@songtianyi I just tried it and it works for me.  A lot of tutorials look like they have `tags/` on there.

PR #2231, opened 2018-02-25
[CPP] Fix Xpath - problem of object slicing 

### LINKED ISSUE #2037: [Cpp] Xpath doesn't work

malickf:
This PR close #2037 , there is a problem of [object slicing](https://en.wikipedia.org/wiki/Object_slicing) in function  `XPath::split(const std::string &path)`  . It returns different derived class and loose information. I changed the return value to be references in order to keep the whole information.  Also the parameter `path` is incorrectly treated as a file instead of a string (to behave as in other runtime)

I have tested and now Xpath is working. 

See also https://github.com/antlr/antlr4/pull/2132


parrt:
Looks like we have some build errors:

```
Building ANTLR4 C++ runtime (if necessary) at /home/travis/build/antlr/antlr4/runtime-testsuite/target/classes/Cpp
In file included from /home/travis/build/antlr/antlr4/runtime-testsuite/target/classes/Cpp/runtime/src/atn/ParserATNSimulator.cpp:35:0:
/home/travis/build/antlr/antlr4/runtime-testsuite/target/classes/Cpp/runtime/src/atn/ParserATNSimulator.h: In constructor ‘antlr4::atn::ParserATNSimulator::ParserATNSimulator(antlr4::Parser*, const antlr4::atn::ATN&, std::vector<antlr4::dfa::DFA>&, antlr4::atn::PredictionContextCache&)’:
/home/travis/build/antlr/antlr4/runtime-testsuite/target/classes/Cpp/runtime/src/atn/ParserATNSimulator.h:371:19: warning: ‘antlr4::atn::ParserATNSimulator::parser’ will be initialized after [-Wreorder]
     Parser *const parser;
                   ^
...
```

mike-lischke:
I revoke my approval. Seems I had a bad day when I approved this patch. It's odd in several ways.

malickf:
I'm using this patch and didn't face any build errors. By revoking it, you broke Xpath. I will post here a detailed example when I'll have time.

mike-lischke:
@malickf Strange you didn't see an error `for each (...)` is not valid C++ (`std::for_each` is), which is what one build error is about.

Also, you used pointers in the `Ref` instances you added. Usually (and throughout the rest of the C++ code) a shared pointer is defined as e.g. `Ref<XPathElement>` instead of `Ref<XPathElement *>`.

malickf:
@mike-lischke , thank you for the information and sorry I wasn't aware of the portability problem of `for each` , I build with VS 2015 and so I did not get this issue ( https://stackoverflow.com/questions/197375/visual-c-for-each-portability). Ok for shared pointer. I will come up with a proper fix later on.

PR #2232, opened 2018-02-26
 Fixed warning from MingGW and error from Clang Windows

ChaseOxide:
Reasons for changes are in commit messages.

In general, it fixes:
- In MingGW
   ```
   warning: declaration 'class std::exception' does not declare anything
   class ANTLR4CPP_PUBLIC std::exception; // Needed for VS 2015.
   ```
- In Clang (I tested with clang-tidy on Windows specifically)
   ```
   error: forward declaration of class cannot have a nested name specifier [clang-diagnostic-error]
     class ANTLR4CPP_PUBLIC std::exception; // Needed for VS 2015.
   ```

Side note: This sort of reverts the commit [19f584d](https://github.com/antlr/antlr4/commit/19f584da05bcc277f9804a596765acfc3c8172c5), where the `#ifdef` flag for the forward declaration was removed.
Another note: Seems like there's more than just `std::exception` that requires forward declaration.

### DIFF
--- a/contributors.txt
+++ b/contributors.txt
@@ -180,3 +180,4 @@ YYYY/MM/DD, github id, Full name, email
 2017/12/03, oranoran, Oran Epelbaum, oran / epelbaum me
 2017/12/20, kbsletten, Kyle Sletten, kbsletten@gmail.com
 2017/12/27, jkmar, Jakub Marciniszyn, marciniszyn.jk@gmail.com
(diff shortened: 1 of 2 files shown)


parrt:
We are getting an error here:

```
E: Unable to locate package g++-5
E: Couldn't find any package by regex 'g++-5'
apt-get.diagnostics
apt-get install failed
```

Is that a spurious error or a real problem? @mike-lischke

parrt:
It looks to be a typical spurious inability to download a file from CI.

PR #2233, opened 2018-02-27
Tidy build doc by removing git clone from Compile section

jcttrll:
The `git clone` instructions in the "Compile" section duplicate instructions already given in the previous section, "Get the source".


parrt:
Thanks. I tweaked it myself in a slightly different way so I will close this one.

PR #2235, opened 2018-02-28
Improved ExternalAntlr4Cpp and added CMake option for static CRT (Fixes #1872)

### LINKED ISSUE #1872: ExternalAntlr4Cpp.cmake can't find generated files with single run

ExternalAntlr4Cpp.cmake determines the list of `generated_files` via `GLOB` after ANTLR runs. However, these generated files haven't been _generated_ yet! That means `GLOB` won't see them and therefore the dependencies will be empty.

A simple work-around seems to be to run the build (`cmake` and `make`) until it fails with a linking error, then run the build again. The second invocation of `cmake` will find the generated files from the first run.

Was this behavior intended? (I've tried adding `GENERATED TRUE` to the properties for the generated file, but that didn't change anything.)

ChaseOxide:
CMake minimum required 3.7 for `SOURCE_SUBDIR` options in ExternalProject, otherwise need to manually pass all generator-related args to configure command.

I have removed the old macro antlr4cpp_process_grammar as it requires a lot of parameters that may not be needed. Most changes are specified in commit messages.

On side note please let me know if `ANTLR4_` or `ANTLR4CPP_` prefix would be a better choice for ExternalAntlr4Cpp, as I noticed the definition `ANTLR4CPP_STATIC` has `ANTLR4CPP_` prefix.

I have tested this [demo](https://github.com/ChaseOxide/antlr-demo) under the following environment:
* Ninja + MSVC 19.12 + Windows 10
* Visual Studio 2017 + MSVC 19.12 + Windows 10
* Visual Studio 2015 + MSVC 19.0 + Windows 10
* NMake Makefiles + MSVC 19.12 + Windows 10
* Unix Makefiles + AppleClang 8.1.0 + macOS 10.12.6 (with -j14)
* Xcode + AppleClang 8.1 + macOS 10.12.6
* Unix Makefiles + GNU 5.4 + Ubuntu 16.04.3 LTS (with -j14)
* Ninja + GNU 5.4 + Ubuntu 16.04.3 LTS

Note: On the demo `ANTLR4_GIT_REPOSITORY` is set to `https://github.com/ChaseOxide/antlr4.git`

Known issues:
* Building both antlr4_static and antlr4_shared will force project to rebuild everything after building `clean` target in Ninja (Windows)


runtime/Cpp/cmake/ExternalAntlr4Cpp.cmake
```diff
@@ -1,223 +1,131 @@
...
+  if(MINGW)
+    set(ANTLR4_SHARED_LIBRARIES
+        ${ANTLR4_OUTPUT_DIR}/libantlr4-runtime.dll.a)
+    set(ANTLR4_RUNTIME_LIBRARIES
+        ${ANTLR4_OUTPUT_DIR}/libantlr4-runtime.dll)
+  elseif(CYGWIN)
+    set(ANTLR4_SHARED_LIBRARIES
+        ${ANTLR4_OUTPUT_DIR}/libantlr4-runtime.dll.a)
+    set(ANTLR4_RUNTIME_LIBRARIES
+        ${ANTLR4_OUTPUT_DIR}/cygantlr4-runtime-4.7.1.dll)
```
mike-lischke: Why's here a version number used for the binary? Should be named like the other libraries.

runtime/Cpp/cmake/README.md
```diff
@@ -0,0 +1,139 @@
...
+add_executable(demo main.cpp
+               ${ANTLR_SampleGrammarLexer_CXX_OUTPUTS}
+               ${ANTLR_SampleGrammarParser_CXX_OUTPUTS})
+target_link_libraries(demo antlr4_static)
+```
+
+## Documentation for FindANTLR
+
+The module defines the following variables:
+```
```
KvanTTT: Please use line breaks before and after code snippets. Here and further.

runtime/Cpp/cmake/README.md
```diff
@@ -0,0 +1,139 @@
...
+* `OUTPUT_DIRECTORY` - the output directory for the generated files. By default it uses `${CMAKE_CURRENT_BINARY_DIR}`
+* `DEPENDS_ANTLR` - the dependent target generated from antlr_target for the current call
+* `COMPILE_FLAGS` - additional compile flags for ANTLR tool
+* `DEPENDS` - specify the files on which the command depends. It works the same way `DEPENDS` in [add_custom_command()](https://cmake.org/cmake/help/v3.11/command/add_custom_command.html)
+* `LEXER` - specify the input file is a lexer grammar
+* `PARSER` - specify the input file is a parser grammar
+* `LISTENER` - tell ANTLR tool to generate a parse tree listener
+* `VISITOR` - tell ANTLR tool to generate a parse tree visitor
+
+### Examples
```
KvanTTT: Line break after the header.

runtime/Cpp/cmake/ExternalAntlr4Cpp.cmake
ChaseOxide: CMake for Antlr4 under Cygwin generates cygantlr4-runtime-4.7.1.dll instead of cygantlr4-runtime.dll due to the declaration of SOVERSION. I just tested that building still works even if version isn't included, or IMPORTED_LOCATION is not set at all. However, it will mean the documentation in README.md has to change as I claim that ANTLR4_RUNTIME_LIBRARIES is the path to shared runtime library.

mike-lischke:
The PR must be updated so we can approve it. Terence won't merge it until then (and the release window for new stuff is pretty small).

runtime/Cpp/cmake/ExternalAntlr4Cpp.cmake
mike-lischke: Will you do that change yet?

chrisaycock:
@mike-lischke Can I do anything to help? I see a few cosmetic notes to the README, updates needed for the merge conflict, and your note on the hard-coded version number. Anything else?

How would I "take over" a PR anyway? The OP deserves full credit rather than me.

chrisaycock:
@mike-lischke I have opened PR #2388 to tweak some things from this patch.

PR #2267, opened 2018-03-29
Universal Windows Support (UWP/UAP)

thegoldenmule:
We are using ANTLR4 on HoloLens, which runs on UWP. This platform is very close to `netstandard`, but we needed to make a handful of changes. The only real difference here is that we needed to abstract out `Windows.Console.Out` and `Err` and replace them with `TextWriter` implementations that simply forward to `Diagnostics.Debug`. This is the MS recommended way of writing to stdout and stderr.

### DIFF
--- a/contributors.txt
+++ b/contributors.txt
@@ -1,4 +1,4 @@
-ANTLR Project Contributors Certification of Origin and Rights
+﻿ANTLR Project Contributors Certification of Origin and Rights
 
 All contributors to ANTLR v4 must formally agree to abide by this
 certificate of origin by signing on the bottom with their github
@@ -181,3 +181,4 @@ YYYY/MM/DD, github id, Full name, email
 2017/12/20, kbsletten, Kyle Sletten, kbsletten@gmail.com
 2017/12/27, jkmar, Jakub Marciniszyn, marciniszyn.jk@gmail.com
 2018/02/11, io7m, Mark Raynsford, code@io7m.com
+2018/03/29, thegoldenmule, Benjamin Jordan, bmj465@gmail.com
--- a/runtime/CSharp/runtime/CSharp/Antlr4.Runtime/Antlr4.Runtime.dotnet.csproj
+++ b/runtime/CSharp/runtime/CSharp/Antlr4.Runtime/Antlr4.Runtime.dotnet.csproj
@@ -3,7 +3,7 @@
     <Company>The ANTLR Organization</Company>
     <Version>4.7.1.1</Version>
 	<NeutralLanguage>en-US</NeutralLanguage>
-    <TargetFrameworks>netstandard1.3;net35</TargetFrameworks>
+    <TargetFrameworks>netstandard1.3;net35;uap10.0</TargetFrameworks>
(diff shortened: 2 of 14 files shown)


thegoldenmule:
In addition-- I don't believe I can fix one of the errors as the UWP SDK has to be installed on the build machine.

thegoldenmule:
VS2013 build is fixed-- now it appears to be the lack of UAP SDK on the build machine.

ericvergnaud:
You would need to install it as part of the Travis build. Check the script in .travis folder

Envoyé de mon iPhone

thegoldenmule:
Merci, I will check that out next.

xied75:
Sure, most are not needed to get it work in UWP. Current nuget package works perfectly, even with "system.console" there, it won't affect anything. Just run the antlr java tool, get those .cs files into your solution, compile, and release in the store, everything works fine.

thegoldenmule:
You're doing this on HoloLens? That has not been my experience.

xied75:
No I'm not fortune enough to afford a HoloLens, my UWP app is ever tested/released on Windows 10 Desktop and Phone. Antlr4 netstandard1.3 work was done by me long time ago, purely to "support" UWP. If HoloLens behave differently than Desktop and Phone (which also use an ARM processor, if my memory is correct), then apparently Microsoft is to blame for telling us it's "Universal".

I understand having "system.console" hard coded is wrong, but in the Desktop and Phone, it seems not getting any troubles, that's why I was asking in the beginning. So what exactly did you observe when you compile or deploy? Are you using Unity as well?

Besides, antlr by natural is a Java project, the ultimate **upstream** is the Java version, i.e. the Runtime in Java, from what I saw there, e.g. the ATNSimulator is already deprecated, thus ideally we in the C# world better follow suite rather than patching on something legacy.

thegoldenmule:
I am using Unity, yes. Also, the HoloLens is actually x86, not ARM-- though in theory that shouldn't matter. I have not had much luck in using netstandard on HoloLens in general, to be honest. Even other libraries that support netstandard I've had to tweak.

When I compile, I see:

```
The type or namespace 'Console' does not exist in the namespace "System".
```

xied75:
> I am using Unity

That might explains your not-so-smooth experience with netstandard, while mine is smooth so far. Unity is using its own toolchain (Mono and IL2CPP stuff); while Visual Studio would call CoreRT to aot C# source code, anything is not used would not be in the final release package.

Back to your problem, surely we should try to get your code compile, but getting a new UAP target is not necessary, I would suggest that you make a new PR, just replace all those relic of "console.write" to "debug.write", if @ericvergnaud agree?

One more question to Eric, do we make nuget release whenever we merge? How would the versioning go? 4.7.1.2?

thegoldenmule:
I am not using Mono or IL2CPP, I am using a .NET backend (this is an option in Unity for this platform). I am open to `Debug.Write`, though this changes the internals of ANTLR more significantly.

thegoldenmule:
To clarify: you can tell Unity essentially not to mess with your C#. It will spit out a VS solution with multiple C# projects that include your C#. VS then compiles these with the .NET compiler or with the .NET Native toolchain. Mono and IL2CPP are not involved.

Issue #2234, opened 2018-02-28
[CPP] Names conflict


pollnossa:
Hi Mike,
I'm writing my project on Linux using antlr 4.7. When I started to port it on Windows I've encountered with the typical issue - conflict in variable's names.  ```static const Ref<dfa::DFAState> ERROR;``` from https://github.com/antlr/antlr4/blob/7a2b95ab53f488b9e26d4af7409be6d5ca1f9eb9/runtime/Cpp/runtime/src/atn/ATNSimulator.h#L19 overlaps with ```#define ERROR               0``` from *wingdi.h* on my system. 

Here is the simple code which will not compile due to name's conflict:
```c
#include <Windows.h>
#include "antlr4-runtime.h"

int main()
{
    return 0;
}
```

Bad solution for such case is to change the order of includes for which the project will be compiled successfully:
```
#include "antlr4-runtime.h"
#include <Windows.h>
```
Good solution is to rename  ```static const Ref<dfa::DFAState> ERROR;``` and not use ALL_CAPS all names except macro names - https://goo.gl/UDpvdA
Thank you.

PR #2279, opened 2018-04-21
[CPP] Fixing name conflict #2234

SuperFola:
Hi everyone !

This is my first pull request here, if anything is missing please let me know.

I am just fixing the issue #2234 (replacing `ERROR` by `DFAStateError` to avoid name conflict when including `windows.h`) because it is a bit annoying to review a whole project and fix include order because of this (even if `NOGDI` is defined before including `windows.h`, I have this problem).

Have a good day, SuperFola

### DIFF
--- a/runtime/Cpp/runtime/src/atn/ErrorInfo.h
+++ b/runtime/Cpp/runtime/src/atn/ErrorInfo.h
@@ -28,7 +28,7 @@ namespace atn {
     /// </summary>
     /// <param name="decision"> The decision number </param>
     /// <param name="configs"> The final configuration set reached during prediction
-    /// prior to reaching the <seealso cref="ATNSimulator#ERROR"/> state </param>
+    /// prior to reaching the <seealso cref="ATNSimulator#DFAStateError"/> state </param>
     /// <param name="input"> The input token stream </param>
     /// <param name="startIndex"> The start index for the current prediction </param>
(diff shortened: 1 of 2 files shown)


SuperFola:
You mean you used this workaround instead of renaming `ERROR` to something else ?

If my patch is generating more errors to you, could I see your build log to fix them please ? Thanks

mike-lischke:
In what sense does this name conflict with anything? It's not a global type but a static variable of the ATNSimulator class. You cannot always name class variables etc. in a way to avoid trouble with every possible 3rd and 4th party software.

SuperFola:
Well, I shouldn't have named it a "name conflict" tht's right. It's more than a problem with naming conventions, variables shouldn't be named with only uppercase letters.
A side effect is that it's conflicting with a widely used "library": the WinAPI, since it defined `ERROR`

mike-lischke:
You are right. If that was written in C++ right from the start, the naming would certainly be different. But this is a C++ port of a Java library, which tries to stay close to the original code, in order to ease code reviews, correctness checks and later changes. I have used it a number of projects on Windows and never had trouble with it, wrt. naming (except for global names, which I could easily solve).

On Windows, when you use the windows.h header file, include that as the first think in your PCH and the ANTLR4 runtime after it. That should avoid trouble with duplicate names in this situation.

I'm really not inclined to change variable names, unless they are also changed in the original Java source.

PR #2280, opened 2018-04-24
File path handling fixes

solussd:
Ensure file paths passed to STGroup[File] are absolute
- Fixes AWS Lambda deployments
- gracefully handle grammars without source fileNames

See: https://github.com/walmartlabs/lacinia/issues/175

### DIFF
--- a/tool/src/org/antlr/v4/Tool.java
+++ b/tool/src/org/antlr/v4/Tool.java
@@ -838,7 +838,7 @@ public class Tool {
 		// or just or the relative path recorded for the parent grammar. This means
 		// that when we write the tokens files, or the .java files for imported grammars
 		// taht we will write them in the correct place.
-		if (fileNameWithPath.lastIndexOf(File.separatorChar) == -1) {
+		if ((fileNameWithPath == null) || (fileNameWithPath.lastIndexOf(File.separatorChar) == -1)) {
 			// No path is included in the file name, so make the file
 			// directory the same as the parent grammar (which might sitll be just ""
 			// but when it is not, we will write the file in the correct place.
--- a/tool/src/org/antlr/v4/codegen/Target.java
+++ b/tool/src/org/antlr/v4/codegen/Target.java
@@ -26,6 +26,8 @@ import org.stringtemplate.v4.STGroupFile;
 import org.stringtemplate.v4.StringRenderer;
 import org.stringtemplate.v4.misc.STMessage;
 
+import java.net.URL;
+
 /** */
 public abstract class Target {
 	/** For pure strings of Java 16-bit Unicode char, how can we display
@@ -504,9 +506,10 @@ public abstract class Target {
 
 	public boolean templatesExist() {
 		String groupFileName = CodeGenerator.TEMPLATE_ROOT + "/" + getLanguage() + "/" + getLanguage() + STGroup.GROUP_FILE_EXTENSION;
+		URL url = Thread.currentThread().getContextClassLoader().getResource(groupFileName);
 		STGroup result = null;
 		try {
-			result = new STGroupFile(groupFileName);
+			result = new STGroupFile(url.getPath());
 		}
 		catch (IllegalArgumentException iae) {
 			result = null;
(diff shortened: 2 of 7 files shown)


solussd:
I don't understand the CI failure, any insight would be appreciated.

parrt:
@solussd looks like we have some build errors

parrt:
@solussd looks like build errors in the Target.java file you changed:

```
[ERROR] /C:/projects/antlr4/tool/src/org/antlr/v4/codegen/Target.java:[512,50] incompatible types: java.net.URL cannot be converted to java.lang.String
[ERROR] /C:/projects/antlr4/tool/src/org/antlr/v4/codegen/Target.java:[527,50] incompatible types: java.net.URL cannot be converted to java.lang.String
```

solussd:
@parrt This PR depends on changes here: https://github.com/antlr/stringtemplate4/pull/199

That PR was closed as "fixed" in another PR, but the other PR didn't implement a constructor for STGroupFile that takes a URL.

parrt:
Ah. Can i just add `public STGroupFile(URL url) { this(url, "UTF-8",'<', '>'); }`?

solussd:
That could be sufficient. I can test tomorrow morning w/ my AWS Lambda use case if you add that constructor.

solussd:
I had rewritten the constructors so they all bottom out in loadGroupURL instead of loadGroupFile.

parrt:
While moving the constructors would mean it uses a different style so we should move those back.  I'm trying to build but am being unsuccessful now with later versions of Java. tools.jar  apparently no longer exists. damn them!  any idea what happened to the Java compiler objects? are they in a public API somewhere now?

solussd:
@parrt I'm not sure, sorry. W.r.t. the constructors, I didn't change the interface to the existing ones, I just consolidated the implementation under the URL one.

ewanmellor:
@parrt I've got a change to fix the tools.jar issue, I'll post that in a little while.

parrt:
Actually, it's no problem. I forgot we require java 1.7 so it's gotta be there.

parrt:
Hi @solussd i have added the ctor to ST4: https://github.com/antlr/stringtemplate4/commit/b5a82e04e45be2b4d88e2498ebda4cdc1fedb1b0

I hope to release ST4 this weekend so please let me know :) thanks!

solussd:
@parrt works with updated ST4. Thanks!

solussd:
@parrt yes, this PR. I’ve updated it. Yes, it looks like 4.1 has what I need. Will update in a moment.

parrt:
Ok. also i'm not sure you can add others to contrib list. they need to add commits for their own name.

solussd:
@parrt I did that to fix a merge conflict, technically, they were already there (though git might show it as being added in the diff of this PR).

parrt:
ah. thanks

tool/src/org/antlr/v4/codegen/Target.java
```diff
@@ -504,9 +506,10 @@ public boolean grammarSymbolCausesIssueInGeneratedCode(GrammarAST idNode) {
 
 	public boolean templatesExist() {
 		String groupFileName = CodeGenerator.TEMPLATE_ROOT + "/" + getLanguage() + "/" + getLanguage() + STGroup.GROUP_FILE_EXTENSION;
+		URL url = Thread.currentThread().getContextClassLoader().getResource(groupFileName);
 		STGroup result = null;
 		try {
-			result = new STGroupFile(groupFileName);
+			result = new STGroupFile(url);
```
parrt: It looks to me as though STGroupFile's ctor calls `getURL()` which does exactly that getResource thing. What additional functionality are you providing? i.e., where does ST4 go wrong? Maybe we shoulda fixed that in ST4 :)

solussd: I believe is related to when it is called. I’d have to dig into it again.

solussd: It's also possible the PR you merged into ST4 instead of mine (199) makes this unnecessary.

parrt: ah. cool. if you have time, could you find the minimal changeset that works? I'm terrified of changing this stuff in the tool!

solussd: Tested without the classloader in ANTLR (instead letting ST4 + getURL do the context class loader lookup) and it works. PR updated to what I believe is a minimal changeset.

tool/src/org/antlr/v4/Tool.java
```diff
@@ -397,7 +397,7 @@ public void processNonCombinedGrammar(Grammar g, boolean gencode) {
 
 		if ( generate_ATN_dot ) generateATNs(g);
 
-		if ( g.tool.getNumErrors()==0 ) generateInterpreterData(g);
+		if (gencode && g.tool.getNumErrors()==0 ) generateInterpreterData(g);
```
parrt: ok, looking good but do we need this tweak for gencode? I'm worried someone might rely on it.

solussd: This seems like a safe change to me– code is still generated when executed form the cli, but won't generate code when executed programmatically. I think there was an issue with it writing files in a readonly environment (e.g., deployed as a Lambda).

solussd: Thanks for looking at it!

PR #2295, opened 2018-05-23
Update icu4j to resolve CVEs

srvance:
There are 4 CVEs against the current version of icu4j used by ANTLR. This PR updates to the latest version, which resolves all of the CVEs.

These were identified with the [OWASP Dependency Check](https://jeremylong.github.io/DependencyCheck/dependency-check-gradle/configuration.html).

To the extent my environment is configured, all tests pass. For configurations I don't have (e.g., Python 3.5), the tests fail identically between versions.

Details:
=====
Published Vulnerabilities
CVE-2017-14952  suppress

Severity:High
CVSS Score: 7.5 (AV:N/AC:L/Au:N/C:P/I:P/A:P)
CWE: CWE-415 Double Free

Double free in i18n/zonemeta.cpp in International Components for Unicode (ICU) for C/C++ through 59.1 allows remote attackers to execute arbitrary code via a crafted string, aka a "redundant UVector entry clean up function call" issue.
CONFIRM - http://bugs.icu-project.org/trac/changeset/40324/trunk/icu4c/source/i18n/zonemeta.cpp
MISC - http://www.sourcebrella.com/blog/double-free-vulnerability-international-components-unicode-icu/
Vulnerable Software & Versions:

cpe:/a:icu-project:international_components_for_unicode:59.1::~~~c%2fc%2b%2b~~ and all previous versions
CVE-2017-17484  suppress

Severity:High
CVSS Score: 7.5 (AV:N/AC:L/Au:N/C:P/I:P/A:P)
CWE: CWE-119 Improper Restriction of Operations within the Bounds of a Memory Buffer

The ucnv_UTF8FromUTF8 function in ucnv_u8.cpp in International Components for Unicode (ICU) for C/C++ through 60.1 mishandles ucnv_convertEx calls for UTF-8 to UTF-8 conversion, which allows remote attackers to cause a denial of service (stack-based buffer overflow and application crash) or possibly have unspecified other impact via a crafted string, as demonstrated by ZNC.
MISC - https://github.com/znc/znc/issues/1459
MISC - https://ssl.icu-project.org/trac/attachment/ticket/13490/poc.cpp
MISC - https://ssl.icu-project.org/trac/changeset/40714
MISC - https://ssl.icu-project.org/trac/changeset/40715
MISC - https://ssl.icu-project.org/trac/ticket/13490
MISC - https://ssl.icu-project.org/trac/ticket/13510
Vulnerable Software & Versions:

[...]

PR #2299, opened 2018-05-31
Java.stg: use static functions instead of inline arrays

dhalperi:
We hit a Java limit on the amount of code that's allowed to execute in the static initializer for a class.

```
[ERROR] /home/travis/build/batfish/batfish/projects/batfish/target/generated-sources/antlr4/org/batfish/grammar/cisco/CiscoLexer.java:[14,1] error: code too large
[INFO] 1 error
```

See [this PR](https://github.com/batfish/batfish/pull/1533) for more information.

A really simple fix is in the attached template: instead of defining the large public `String[]` constants directly, move them into `private static` factory functions.

### DIFF
--- a/tool/resources/org/antlr/v4/tool/templates/codegen/Java/Java.stg
+++ b/tool/resources/org/antlr/v4/tool/templates/codegen/Java/Java.stg
@@ -236,9 +236,12 @@ public class <parser.name> extends <superClass; null="Parser"> {
 	<endif>
 	public static final int
 		<parser.rules:{r | RULE_<r.name> = <r.index>}; separator=", ", wrap, anchor>;
-	public static final String[] ruleNames = {
-		<parser.ruleNames:{r | "<r>"}; separator=", ", wrap, anchor>
-	};
+	private static String[] makeRuleNames() {
+		return new String[] {
+			<parser.ruleNames:{r | "<r>"}; separator=", ", wrap, anchor>
+		};
+	}
+	public static final String[] ruleNames = makeRuleNames();
 
 	<vocabulary(parser.literalNames, parser.symbolicNames)>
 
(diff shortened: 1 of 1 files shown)


dhalperi:
Hi @sharwell @parrt , any availability to take a look at this change? It's simple and improves scalability of ANTLR. Since it looks like 4.7.2 release may be imminent with the various CVE fixes in https://github.com/antlr/antlr4/pull/2295 , would be nice to get this in before that if possible.

parrt:
Hmm...well, I'd hate to break stuff that relied on current impl. This is more of a Java issue anyway, right?

dhalperi:
I am not a Java expert, but I thought pretty hard and couldn't figure out how this change could break anything. The guarantee before was that after the static initializer ran, there was a `static String[] array`. That is maintained. (Two of the three changes are to `private` arrays, btw, which aren't even exposed to users/extenders parser implementations.)

Yes, this is working around a Java limitation... but it's not one that's going anywhere.

parrt:
hahah. yeah, we're stuck with it. I'd have to think about this for awhile. are you sure you need such big actions in your lexer/parser?  why not move the bulk of code into functions elsewhere?  Create superclass and put them there?

dhalperi:
Sorry if I wasn't clear -- these arrays are only the arrays that contain the list of tokens, their string names, or their rule names. Lexer file is [here](https://github.com/batfish/batfish/blob/master/projects/batfish/src/main/antlr4/org/batfish/grammar/cisco/CiscoLexer.g4).

I think this is just what ANTLR does when you have lots of tokens in your lexer (in our case, several thousand). Is there a way we can break this up?

parrt:
OH! It's our arrays. @sharwell it looks to be a correct fix that doesn't affect external code. it moves static init code to a static function. What ya tink?

sharwell:
Looks reasonable to me. Doesn't guarantee the problem cannot occur, but seems to at least double the supported size of these arrays by placing them in separate methods.

parrt:
@dhalperi could you sign the contributors.txt file too? i'll approve it then.

dhalperi:
https://github.com/antlr/antlr4/blob/master/contributors.txt#L156 ?

dhalperi:
Thanks both!

PR #2302, opened 2018-06-03
Fixes #2301

### LINKED ISSUE #2301: TypeError: '<' not supported for PrecedencePredicate (Python 3 runtime)
I get the exception (in ANTLR 4.7.1):

 `TypeError: '<' not supported between instances of 'PrecedencePredicate' and 'PrecedencePredicate'`

since the ` __lt__()` dunder is not implemented in PrecedencePredicate - it is almost certainly a porting issue from Python 2 since I see a `__cmp__()` method which is, of course, no longer honored in Python 3.

The issue is easily "fixed" by implementing all the equivalent Rich Methods: 
```python
    #def __cmp__(self, other):
    #    return self.precedence - other.precedence

    def __lt__(self, other):
        return self.precedence < other.precedence

    def __le__(self, other):
        return self.precedence <= other.precedence

    def __gt__(self, other):
        return self.precedence > other.precedence

    def __ge__(self, other):
        return self.precedence >= other.precedence 
```
Note: I have not implemented `__ne__()` since it is automatically implemented ad "not `__eq__()`", and that looks desirable in this case (to me).

It is "fixed" since I no longer get the exception, but I'm not sure whether this change is correct - I cannot even know how the change impacts the resulting AST - an ANTLR expert's review is certainly needed.

Additional side note, not strictly related to this issue: I see that `__eq__()` checks for the instance type:
```python
    def __eq__(self, other):
        if self is other:
            return True
        elif not isinstance(other, PrecedencePredicate):
            return False
        else:
            return self.precedence == other.precedence
```
shouldn't the original  `__cmp__()` , and hence the ported Rich Methods, check for the instance type as well?

ericvergnaud:


ericvergnaud:
@parrt any chance you can merge this. it's an important fix

PR #2311, opened 2018-06-15
Updated csharp-target doc

alecont:
Added a 'using' statement for namespace using Antlr4.Runtime.Tree, so that IParseTree can be found without fully qualified name.
Fixed typo in property Parser.BuildParseTree.


ericvergnaud:
@parrt blessed

PR #2312, opened 2018-06-15
BaseTestJava.java: Allow for custom template

HSorensen:
To support complex test case I needed to use a custom StringText template.

With the proposed change the template can be defined together with the unit test that needs this.

### DIFF
--- a/runtime-testsuite/test/org/antlr/v4/test/runtime/java/BaseJavaTest.java
+++ b/runtime-testsuite/test/org/antlr/v4/test/runtime/java/BaseJavaTest.java
@@ -452,6 +452,25 @@ public class BaseJavaTest implements RuntimeTestSupport {
 	}
 
 	protected String execLexer(String grammarFileName,
+            String grammarStr,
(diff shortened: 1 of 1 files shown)


ericvergnaud:
Why not write a plain unit test? The template is only useful for multi-target unit tests, to make the production grammar fragments portable.

HSorensen:
@ericvergnaud Good question. Some background: For some time I have been maintaining an ANTLR fork that enable to include source code into the current scanning stream. Up to now this has been for the Java target only, but I have had a request for C/C++ and I will be needing a Python port as well. In order to enable this I will propose a new portable lexer action and refactor the current implementation to minimize the changes needed in the Lexer core nextToken() loop. As I was testing all of this I needed to create a unit test that require a different test template. This pull request is more in preparation for larger pull request.

ericvergnaud:
Why are you not just enhancing the existing template in your fork?
Sorry to sound like I'm pushing back, but not sure we can support user defined templates.

PR #2322, opened 2018-07-04
Add support for cpack.

jgoppert:
This adds support for cpack packaging. I find this useful for linux binary distribution of the C++ runtime since there are no official binaries yet.

I'm currently hosting a binary built on ubuntu 18.04 here if any one finds it useful: https://github.com/jgoppert/antlr4/releases/download/4.7.1-cpack/LIBANTLR4-4.7.1-Linux.deb

Example of how to make a deb package with cpack:

```bash
mkdir build
cd build
cmake .. 
cpack -G DEB
```


mike-lischke:
@jgoppert pls fix the conflict yet

@parrt This is a pure C++ issue and can be merged without side effects.

PR #2351, opened 2018-08-22
Fix #2350

### LINKED ISSUE #2350: [cpp] Setting `contextSuperClass='antlr4::RuleContextWithAltNum'` doesn't work
Because [this](https://github.com/antlr/antlr4/blob/master/tool/resources/org/antlr/v4/tool/templates/codegen/Cpp/Cpp.stg#L545) should really be like `<struct.name>() = default`.

AmatanHead:


parrt:
Did the change myself as I didn't see a signature for contributors file.

PR #2358, opened 2018-09-09
Assign token names to string literals in parse rule contexts

### LINKED ISSUE #2347: Assign token name to generated ParserRuleContext for automatically resolved tokens
Consider the following grammar:

```antlr
grammar test;
root: Token;
Token: 'token';
```

After the parser generation, I have the following `RootContext`:

```java
public static class RootContext extends ParserRuleContext {
	public TerminalNode Token() { return getToken(testParser.Token, 0); }
	public RootContext(ParserRuleContext parent, int invokingState) {
		super(parent, invokingState);
	}
	@Override public int getRuleIndex() { return RULE_root; }
}
```

But for such grammar:

```antlr
grammar test;
root: 'token';
Token: 'token';
```

The terminal node `Token` is missing in `RootContext`:

```java
public static class RootContext extends ParserRuleContext {
	public RootContext(ParserRuleContext parent, int invokingState) {
		super(parent, invokingState);
	}
	@Override public int getRuleIndex() { return RULE_root; }
}
```

It would be convenient to generate the same contexts for the abovementioned grammars.

Full codes and diff are available on https://www.diffchecker.com/eg2AIeJN

KvanTTT:
fixes #2347

### DIFF
--- a/tool-testsuite/test/org/antlr/v4/test/tool/TestCodeGeneration.java
+++ b/tool-testsuite/test/org/antlr/v4/test/tool/TestCodeGeneration.java
@@ -31,6 +31,7 @@ import java.util.ArrayList;
 import java.util.List;
 
 import static org.junit.Assert.assertFalse;
+import static org.junit.Assert.assertNotEquals;
 
 public class TestCodeGeneration extends BaseJavaToolTest {
 	@Before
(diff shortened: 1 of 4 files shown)


parrt:
Hmm...well, this would have to be implemented in all other targets optimally. I am not sure diverging even more with Java target is worth the increased value.

KvanTTT:
It should work for all targets. But I'll test it on our big grammars set ASAP.

parrt:
Oh, i get it. it doesn't change code gen, it just makes it look like the token ref had a label. ok. please do test on a few big grammars. thanks.

KvanTTT:
Eventually, I've checked the new version on our grammars-v4 repository on Java runtime. Now I can say that everything is fine :)

PR #2363, opened 2018-09-21
Fix CaseInsensitiveInputStream.js

jcnguyen:
Assign and use the correct case function. Define properties `index` and `size`. Return a code point instead of a char in the `LA` function. Call the base seek function instead of getText in the `seek` function.


doc/resources/CaseInsensitiveInputStream.js
```diff
@@ -39,8 +51,8 @@ CaseInsensitiveInputStream.prototype.release = function(marker) {
     return this._stream.release(marker);
 };
 
-CaseInsensitiveInputStream.prototype.seek = function(_index) {
-    return this._stream.getText(start, stop);
+CaseInsensitiveInputStream.prototype.seek = function(index) {
+    return this._stream.seek(index);
```
jamesgpearce: There are a few functions (`reset`, `consumer`, `seek`) that don't need to be `return`ed from.

jcnguyen: Updated

jcnguyen:
Closing because of this fix: https://github.com/antlr/antlr4/commit/91f4b89ee666d6499a9974e2c92b8bf3ea9ec2a9

PR #2364, opened 2018-09-24
sign contributors, add Richard Neil Pittman

rnpittman:


rnpittman:
I only added my name to the contributor text file at the top.  I don't how that could result in a build break.

parrt:
Hi. Shouldn't this go in the grammars-v4 repo?

PR #2374, opened 2018-10-08
[Python] Add missing lexer super class imports.

renatahodovan:
Until now, the generated Python3 code imported the custom parser
superclasses relatively. However, this only worked in Python3 if
the module was inside a package, since relative imports rely on
__name__ to determine the current module's position in the package
hierarchy. In case of a standalone script, this was always __main__
and hence these relative imports failed.
The patch handles this issue them same way as it is handled by
listener imports.


ericvergnaud:
Thanks!
@parrt blessed

Issue #2304, opened 2018-06-07
NullReferenceException in CSharp target


slt3:
I have following grammar:
```
grammar FormulaCalc;
expr: formula EOF;
formula 
	 : formula '^' formula 
	 | formula '*' formula
	 | formula '/' formula
	 | formula '%' formula
	 | formula '+' formula
	 | variable
	 | number
	 | formula '-' formula
     | '('formula')'
	 | function
     ;
function : zero_param_function
		 |one_param_function
		 |two_param_function;
zero_param_function : FUNNAME0 '('')';
one_param_function : FUNNAME1 '('parameter')';
two_param_function : FUNNAME2 '('parameter ',' parameter')';
parameter: number|variable;
FUNNAME0 : 'rand';
FUNNAME1 : 'sin'|'cos'|'tan'|'cot'|'asin'
		 |'acos'|'atan'|'acot'|'log'|'ln'|'exp';
FUNNAME2 :  'sqrt'|'pow'|'mod';
number : '-'? unsigned_number | '+'? unsigned_number;
unsigned_number : 'pi'| EULER |FPOINTNUMBER;
EULER : 'e'|'E';
FPOINTNUMBER : DIGIT+ ('.'DIGIT+)?;
DIGIT : [0-9];
variable : '{'word'}';
LETTER : [a-z]|[A-Z];
SPECIALLETTER : SIGN
	|'^'|'"'|'$'|'%'|'&'|'/'|'('|')'
	|'['|']'|'='|'?'|'#'|'_'|'.'|'~'
	|':'|','|';'|'<'|'>'
	;
word : (EULER|'^'|'*'|'/'|'%'|'+'|'-'|','
     |'('|')'
     |LETTER|FPOINTNUMBER|SPECIALLETTER
	 |FUNNAME0|FUNNAME1|FUNNAME2|'pi')+;
SIGN : '+'|'-';
WS : [ \t\r\n]+ -> skip;
```
When i run the parser:
```
String input = "5+1";
ICharStream stream = CharStreams.fromstring(input);
ITokenSource lexer = new FormulaCalcLexer(stream);
 ITokenStream tokens = new CommonTokenStream(lexer);
FormulaCalcParser parser = new FormulaCalcParser(tokens);
parser.BuildParseTree = true;
IParseTree tree = parser.expr();
```
the NullReferenceException is thrown:

Object reference not set to an instance of an object.
    Boolean Equals(System.Object)

[...]

PR #2375, opened 2018-10-08
NullReferenceException in CSharp target #2304   

xsIceman:
updated SingletonPredictionContext in the C# runtime to match that of Java as described in issue #2304. I face the same issue with my IPS parser


runtime/CSharp/runtime/CSharp/Antlr4.Runtime/Atn/SingletonPredictionContext.cs
```diff
@@ -79,7 +79,7 @@ public override bool Equals(object o)
                 return false;
             }
 			Antlr4.Runtime.Atn.SingletonPredictionContext other = (Antlr4.Runtime.Atn.SingletonPredictionContext)o;
-            return returnState == other.returnState && parent.Equals(other.parent);
+            return returnState == other.returnState && (parent != null && parent.Equals(other.parent));
```
ericvergnaud: How about when parent is null?

xsIceman: That is the change i want to add, when parent == null i want to return false. before we did a compare on null and got a nullrefexception. This is the same check that is done in java code SingletonPredictionContext.java.

ericvergnaud: Then maybe the Java code also needs to be fixed. Please get this right - maybe add a reusable utility to compare objects.

parrt: It looks like what the Java target does and I'm terrified of changing what that target does so maybe we should allow this within the C# to match the functionality at least?

zhech2:
I am also running into this issue.  For starters can the behavior of the c# code be changed to match the java code. Which this pull request already does?

ericvergnaud:
@parrt blessed

parrt:
@xsIceman can you fix the contributors.txt file? It has weird old date: 2018/10/08,

parrt:
@xsIceman can you fix the contributors.txt file? It has weird old date: 2018/10/08?

yonigibbs:
Hi @parrt, is this PR planned to be merged at some point? Is there anything we can do to help here? We're running into the same issue. Thanks!

parrt:
I guess date is ok. just old.

parrt:
Hmm...seems this [fails to pass a test](https://travis-ci.org/github/antlr/antlr4/jobs/734045778#L2955) @ericvergnaud can you take a peek?

ericvergnaud:
@parrt the failure seems to be with a swift test so not sure how it can relate to this PR. Did you also merge other stuff?

parrt:
I don't think I've included any swift stuff in a while.  maybe I need to try to run the tests locally.  It looks like the npm thing was missing  on an AppVeyor fail.  did you see this one? https://ci.appveyor.com/project/parrt/antlr4/builds/35672295

 maybe it got fixed later

ericvergnaud:
That is unrelated I think.
The npm thing was from me, I disabled JS on AppVeyor for now due to that issue, and asked help from AppVeyor experts but it’s not blocking (the tests pass on Travis)

Envoyé de mon iPhone

PR #2379, opened 2018-10-13
Unify generated file names (for Go runtime)

### LINKED ISSUE #2378: [Go] Generated lexer and parser file names are always in lower case with underscore
The following command

```
Test.g4 -Dlanguage=Go
```

generates `test_lexer.g4` and `test_parser.g4` instead of `TestLexer.g4` and `TestParser.g4` like other runtimes.

KvanTTT:
fixes #2378


grundic:
I think this is intentional, so that generated files would comply to standard Goland naming convention.
Please have a look here: https://stackoverflow.com/questions/35694137/file-name-convention-for-compound-words

parrt:
I'm going to close this as it looks intentional.

Issue #2097, opened 2017-11-02
C# ParseTreePatternMatcher.Compile fails


appel1:
ParserInterpreter creates an instance of ParserATNSimulator with null for the decisionToDFA parameter. ParserATNSimulator.AdaptivePredict throws a NullReferenceException if it is null. 

```
grammar Test;

options {
    language=CSharp;
}

text: TEXT* EOF;

TEXT: [a-z]+;

WS: [ \t\r\n]+ -> skip;
```

```csharp
var lexer = new TestLexer(new AntlrInputStream("text"));
var parser = new TestParser(new CommonTokenStream(lexer));
var context = parser.text();

var pattern = parser.CompileParseTreePattern("<TEXT>", TestParser.RULE_text);
```

```
Antlr4.Runtime.Tree.Pattern.ParseTreePatternMatcher+CannotInvokeStartRule: Object reference not set to an instance of an object. ---> System.NullReferenceException: Object reference not set to an instance of an object.
   at Antlr4.Runtime.Atn.ParserATNSimulator.AdaptivePredict(ITokenStream input, Int32 decision, ParserRuleContext outerContext)
   at Antlr4.Runtime.ParserInterpreter.VisitState(ATNState p)
   at Antlr4.Runtime.ParserInterpreter.Parse(Int32 startRuleIndex)
   at Antlr4.Runtime.Tree.Pattern.ParseTreePatternMatcher.Compile(String pattern, Int32 patternRuleIndex)
   --- End of inner exception stack trace ---
   at Antlr4.Runtime.Tree.Pattern.ParseTreePatternMatcher.Compile(String pattern, Int32 patternRuleIndex)
   at Antlr4.Runtime.Parser.CompileParseTreePattern(String pattern, Int32 patternRuleIndex, Lexer lexer)
   at Antlr4.Runtime.Parser.CompileParseTreePattern(String pattern, Int32 patternRuleIndex)
   at ConsoleApp3.Program.Main(String[] args) in ConsoleApp3\ConsoleApp3\Program.cs:line 19
```

ericvergnaud:
Is this with the official tool/runtime or the tunnel vision labs one?

Envoyé de mon iPhone

appel1:
This is with the Antlr4.Runtime.Standard v4.7.1 NuGet package. 

Relevant parts of the code:
[...]

PR #2382, opened 2018-10-18
Properly build decisionToDFA

edirgarcia:
this small contribution solves this issue:
https://github.com/antlr/antlr4/issues/2097


kaby76:
Parse tree matching does not work in the C# runtime v4.7.2. Reproduced this problem and fix using equivalent C# of the tests for parse tree pattern matching at https://github.com/antlr/antlr4/blob/master/tool-testsuite/test/org/antlr/v4/test/tool/TestParseTreeMatcher.java.  Crashes on "DFA dfa = decisionToDFA[decision];"  in ParserATNSimulator.cs because decisionToDFA is null, set to null in  "Interpreter = new ParserATNSimulator(this, atn, **null**, null);" in ParserInterpreter.cs.  After reading the Java code, I notice that like decisionToDFA not being set, "PredictionContextCache sharedContextCache = new PredictionContextCache();" field is not defined, and null always passed in "Interpreter = new ParserATNSimulator(this, atn, null, **null**);". I would think this runtime would be similar to that in Java. Is this change blocked because "contributors.txt" is in conflict?

ericvergnaud:
@kaby76 No it just fell off the radar.

ericvergnaud:
@parrt blessed, not sure why these lines you added on 06-06-2015 in the Java runtime were not ported to the C# runtime...

PR #2388, opened 2018-10-29
Improved ExternalAntlr4Cpp and added CMake option for static CRT (Fixes #1872)

### LINKED ISSUE #1872: ExternalAntlr4Cpp.cmake can't find generated files with single run

chrisaycock:
This is a refresh of PR #2235. All credit goes to the OP, @ChaseOxide.


chrisaycock:
Hold off on this PR for the moment. I'm still testing some things.

chrisaycock:
CMake style guides [ [1](https://community.kde.org/Policies/CMake_Coding_Style), [2](https://llvm.org/docs/CMakePrimer.html) ] don't have any spaces between a directive and the parenthesis, so it should be `if()` everywhere. I notice that the older `CMakeLists.txt` files mix the styles; I will fix them all to be `if()`.

mike-lischke:
@parrt This is the improved PR for the one you just closed. C++ build stuff only, so it can be merged with no harm. Thank you.

PR #2398, opened 2018-11-11
Fix for antlr#2164.

mike-lischke:
A different approach to the problem, now really fixing the leak.


chrisaycock:
I hate to reopen this debate, but @chund is right; that is definitely a double free. In C++, `throw` **copies** the exception object and then invokes the destructor. So the `catch` object (in the generated `parse()` function from ...Parser.cpp) now contains a copy of the pointer that has just been freed. So once `catch` ends, the destructor is run a second time on that pointer.

The solution is to simply switch to a `shared_ptr`.  This [PR](https://github.com/antlr/antlr4/pull/2551) almost has it; there shouldn't be a destructor at all for `NoViableAltException` if `_deadEndConfigs` is a `shared_ptr`.

mike-lischke:
@chund @chrisaycock since this PR is already closed let's discuss this issue in the reference bug report.

PR #2409, opened 2018-11-16
Fix install dependency for 'typing' library in python3.4

amykyta3:
Use of the "typing" library introduced in Python 3.5 is the only thing preventing the Antlr4 runtime to be used with the older Python 3.4.

Fortunately there is a backport available that resolves this dependency: https://pypi.org/project/typing/
Added a version-specific dependency so that the "typing" backport will get installed automatically when using versions of python older than 3.5


parrt:
Hmm...I'm not sure we need to support 3.4.  @ericvergnaud ?

amykyta3:
I'm not particularly hellbent on 3.4 either. I expect user share to start dropping significantly especially since Ubuntu 14.04 LTS (only remaining active version that ships with 3.4) is due to lose support this coming April.

That said, I still encounter 3.4 pretty often and it appears that the Antlr runtime is compatible except for this one thing so it might be an easy fix.

ericvergnaud:
Looks like typing.io is a prerequisite i.e. it's in the .stg.
@parrt blessed

PR #2418, opened 2018-11-18
Fix #1955 javascript

ericvergnaud:


parrt:
Thanks @ericvergnaud !!

PR #2423, opened 2018-11-25
fix NPE

ericvergnaud:
fix NPE encountered with a left-recursive grammar such as:
stuff:
   stuff1       # Stuff1
   | stuff2    # Stuff2
   | (stuff1 | stuff2) stuff3
 ;

Does not happen in Java because null keys are allowed


sharwell:
@ericvergnaud can you add a test for this so other runtimes can be validated against the desired behavior?

PR #2445, opened 2018-12-20
C++: fix bug in PredictionContext::mergeArrays

WalterCouto:
In the orignal PredictionContext::mergeArrays there was a bug on Line 281 where the logic differs from java: It currently is:
bool both$ = payload == EMPTY_RETURN_STATE && a_parent && b_parent;
and should instead match java as:
bool both$ = payload == EMPTY_RETURN_STATE && !a_parent && !b_parent;


WalterCouto:
The build failure doesn't look like it has anything to do with the change but with signing failure

mike-lischke:
@parrt In the case @niccroad does not add this fix to his patch you could merge it first and then this one, instead.

PR #2447, opened 2018-12-20
Change bitset upperbound to another higher arbitrary value to allow f…

WalterCouto:
…or larger number of rules

We had a complex grammar file hit past this upperbound.  One solution is to put this up higher, another is to refactor into a container that can grow.

### DIFF
--- a/runtime/Cpp/runtime/src/support/BitSet.h
+++ b/runtime/Cpp/runtime/src/support/BitSet.h
@@ -9,7 +9,7 @@
 
 namespace antlrcpp {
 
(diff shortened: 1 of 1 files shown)


WalterCouto:
signing of contributor list is in PR #2445

WalterCouto:
The build failure doesn't look like it has anything to do with the change but with signing failure

parrt:
@WalterCouto to give you credit, can you sign contributors.txt?

WalterCouto:
@parrt I signed the contributes list in a different PR that was already merged in (PR #2445)

PR #2449, opened 2018-12-23
add Interval for python3 and correct type annotation

youkaichao:
I have noticed that in python2-runtime, there's this ``Interval`` class inside ``IntervalSet`` module. However, the ``Interval`` class is missing in python3-runtime.

The only direct reference to ``Interval`` class is in ``TokenStreamRewriter.getText`` with out-of-date annotation(it refers to ``Interval.Interval`` which doesn't exist). So I update the annotation for python2 & python3.


youkaichao:
see #2451 , this pr is out-of-date.

PR #2451, opened 2018-12-23
bugfix for python-antlr4-runtime

youkaichao:
In this pr, I want to : 

- fix one bug
- add the missing ``Interval`` class
- add default arguments for ``TokenStreamRewriter.getText``

details are given in reviews below.

### DIFF
--- a/runtime/Python2/src/antlr4/TokenStreamRewriter.py
+++ b/runtime/Python2/src/antlr4/TokenStreamRewriter.py
@@ -135,7 +135,7 @@ class TokenStreamRewriter(object):
 
         if stop == len(self.tokens.tokens) - 1:
             for op in indexToOp.values():
-                if op.index >= len(self.tokens.tokens) - 1: buf.write(
+                if op.index > len(self.tokens.tokens) - 1: buf.write(
                     op.text)  # TODO: this check is probably not needed
 
         return buf.getvalue()
--- a/runtime/Python2/src/antlr4/TokenStreamRewriter.py
+++ b/runtime/Python2/src/antlr4/TokenStreamRewriter.py
@@ -100,15 +100,17 @@ class TokenStreamRewriter(object):
         return self.programs.setdefault(program_name, [])
 
     def getDefaultText(self):
-        return self.getText(self.DEFAULT_PROGRAM_NAME, Interval(0, len(self.tokens.tokens)))
+        return self.getText(self.DEFAULT_PROGRAM_NAME, Interval(0, len(self.tokens.tokens) - 1))
 
-    def getText(self, program_name, interval):
+    def getText(self, program_name=None, interval=None):
         """
         :type interval: IntervalSet.Interval
         :param program_name:
         :param interval:
         :return:
         """
+        program_name = program_name or self.DEFAULT_PROGRAM_NAME
+        interval = interval or Interval(0, len(self.tokens.tokens) - 1)
         rewrites = self.programs.get(program_name)
         start = interval.start
         stop = interval.stop
--- a/runtime/Python2/src/antlr4/TokenStreamRewriter.py
+++ b/runtime/Python2/src/antlr4/TokenStreamRewriter.py
@@ -104,7 +104,7 @@ class TokenStreamRewriter(object):
 
     def getText(self, program_name, interval):
         """
-        :type interval: Interval.Interval
+        :type interval: IntervalSet.Interval
         :param program_name:
         :param interval:
         :return:
--- a/runtime/Python3/src/antlr4/IntervalSet.py
+++ b/runtime/Python3/src/antlr4/IntervalSet.py
@@ -13,7 +13,7 @@ class Interval(object):
     def __init__(self, start, stop):
         self.start = start
         self.stop = stop
-        self.range = xrange(start, stop)
+        self.range = list(range(start, stop))
 
     def __contains__(self, item):
         return item in self.range
--- a/runtime/Python3/src/antlr4/IntervalSet.py
+++ b/runtime/Python3/src/antlr4/IntervalSet.py
@@ -8,6 +8,29 @@ from io import StringIO
 import unittest
 from antlr4.Token import Token
 
+class Interval(object):
+
+    def __init__(self, start, stop):
+        self.start = start
+        self.stop = stop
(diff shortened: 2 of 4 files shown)


runtime/Python2/src/antlr4/TokenStreamRewriter.py
```diff
@@ -100,15 +100,17 @@ def getProgram(self, program_name):
         return self.programs.setdefault(program_name, [])
 
     def getDefaultText(self):
-        return self.getText(self.DEFAULT_PROGRAM_NAME, Interval(0, len(self.tokens.tokens)))
+        return self.getText(self.DEFAULT_PROGRAM_NAME, Interval(0, len(self.tokens.tokens) - 1))
```
youkaichao: although range check is performed in ``TokenStreamRewriter.getText``, this line is misleading.

runtime/Python2/src/antlr4/TokenStreamRewriter.py
```diff
@@ -100,15 +100,17 @@ def getProgram(self, program_name):
...
 
-    def getText(self, program_name, interval):
+    def getText(self, program_name=None, interval=None):
```
youkaichao: Default arguments are backward-compatible and make apis more friendly ^_^

runtime/Python2/src/antlr4/TokenStreamRewriter.py
```diff
@@ -100,15 +100,17 @@ def getProgram(self, program_name):
...
         """
-        :type interval: Interval.Interval
+        :type interval: IntervalSet.Interval
```
youkaichao: Correct out-of-date doc

runtime/Python2/src/antlr4/TokenStreamRewriter.py
```diff
@@ -133,7 +135,7 @@ def getText(self, program_name, interval):
 
         if stop == len(self.tokens.tokens) - 1:
             for op in indexToOp.values():
-                if op.index >= len(self.tokens.tokens) - 1: buf.write(
+                if op.index > len(self.tokens.tokens) - 1: buf.write(
```
youkaichao: This is a bug. If I insert something after the last token, that ``insertation`` will be executed twice!

A minimal working example is:

test.g4
```
grammar test;
ruleABC : 'aa' EOF ;
```

test.txt
```
aa
```

test.py
```
import sys
from antlr4 import *
from testLexer import testLexer
from testParser import testParser
from antlr4.TokenStreamRewriter import TokenStreamRewriter
from collections import namedtuple
from testListener import testListener
Interval = namedtuple("Interval", ["start", "stop"])

class MyListener(testListener):

    def __init__(self, tokens:BufferedTokenStream) -> None:
        super().__init__()
        self.writer = TokenStreamRewriter(tokens)

    def getText(self):
        return self.writer.getText(TokenStreamRewriter.DEFAULT_PROGRAM_NAME, Interval(0, len(self.writer.tokens.tokens)))

    def enterRuleABC(self, ctx:testParser.RuleABCContext):
        self.writer.insertAfterToken(ctx.stop, 'it appears first!\n')


def main(argv):
    input = FileStream('test.txt' if len(argv) <= 1 else argv[1])
    lexer = testLexer(input)
    parser = testParser(stream)
    tree = parser.ruleABC()
    myListener = MyListener(stream)
    walker = ParseTreeWalker()
    walker.walk(myListener, tree)

    print(myListener.getText())


if __name__ == '__main__':
    main(sys.argv)
```

the output is : 
```
aait appears first!
it appears first!
```

The purpose here is to insert **one** sentence. However, it ends up that **two** sentences are inserted!

runtime/Python3/src/antlr4/IntervalSet.py
```diff
@@ -8,6 +8,29 @@
...
+    def __iter__(self):
+        return iter(self.range)
+
+    def __getitem__(self, idx):
+        if idx == 0:
+            return self.start
+        elif idx == 1:
+            return self.stop
+        raise IndexError('Interval index out or range [{}]'.format(idx))
+
```
youkaichao: I have noticed that in python2-runtime, there's this ``Interval`` class inside ``IntervalSet`` module. However, the ``Interval`` class is missing in python3-runtime.
It is sort of strange.

youkaichao:
#2449  is also mine. but it is useless now.

ericvergnaud:
Hi,
thanks for this, but it seems you are trying to achieve too many objectives in just 1 PR
Can you please split in:
 - bug fix
 - API improvement
Re Interval, range does the exact same job so doesn't add any value. Only reason to have it in Python 2 is because Python 2 range object does not support r.start and r.stop.

youkaichao:
@ericvergnaud  I will split this pr.

Re Interval: 
`Python 3 range object does the exact same job`
basically it's the same but there is one difference:
python 3 range object:
```
>>>x = range(1, 4)
>>>x[3]
4
```
python 2 Interval:
```
class Interval(object):

    def __init__(self, start, stop):
        self.start = start
        self.stop = stop
        self.range = xrange(start, stop)

    def __contains__(self, item):
        return item in self.range

    def __len__(self):
        return self.stop - self.start

    def __iter__(self):
        return iter(self.range)

    def __getitem__(self, idx):
        if idx == 0:
            return self.start
        elif idx == 1:
            return self.stop
        raise IndexError('Interval index out or range [{}]'.format(idx))
```
```
>>>a=Interval(1,4)
>>>a[3]
index out of range!
```

one more thing to note:
``Interval`` object is referenced in ``TokenStreamRewriter.getText``. I think it is better to keep consistent.

youkaichao:
@ericvergnaud  one more thing about splitting this pr: I have to sign the ``contributors.txt`` in each pr cause I don't know which pr will be merged first. Then I have to resolve conflict if my prs are all accepted. I feel bad =_=

youkaichao:
the bugfix part is split into #2452

ericvergnaud:
Thanks for this, but not comfortable to include this Interval class which adds no significant functionality.

youkaichao:
Then the doc in ``getText`` should be updated and it will be better to provide default arguments ^_^

youkaichao:
see #2452

PR #2452, opened 2018-12-24
bugfix for python-runtime/TokenStreamRewriter.getText

youkaichao:
This is a bug. If I insert something after the last token, that ``insertation`` will be executed twice!

A minimal working example is:

test.g4
```
grammar test;
ruleABC : 'aa' EOF ;
```

test.txt
```
aa
```

test.py
```
import sys
from antlr4 import *
from testLexer import testLexer
from testParser import testParser
from antlr4.TokenStreamRewriter import TokenStreamRewriter
from collections import namedtuple
from testListener import testListener
Interval = namedtuple("Interval", ["start", "stop"])

class MyListener(testListener):

    def __init__(self, tokens:BufferedTokenStream) -> None:
        super().__init__()
        self.writer = TokenStreamRewriter(tokens)

    def getText(self):
        return self.writer.getText(TokenStreamRewriter.DEFAULT_PROGRAM_NAME, Interval(0, len(self.writer.tokens.tokens)))

    def enterRuleABC(self, ctx:testParser.RuleABCContext):
        self.writer.insertAfterToken(ctx.stop, 'it appears first!\n')


def main(argv):
    input = FileStream('test.txt' if len(argv) <= 1 else argv[1])
    lexer = testLexer(input)
    parser = testParser(stream)
    tree = parser.ruleABC()
    myListener = MyListener(stream)
    walker = ParseTreeWalker()
    walker.walk(myListener, tree)

    print(myListener.getText())


if __name__ == '__main__':
    main(sys.argv)
```

the output is : 
```
aait appears first!
it appears first!
```

The purpose here is to insert **one** sentence. However, it ends up that **two** sentences are inserted!

### DIFF
--- a/runtime/Python2/src/antlr4/TokenStreamRewriter.py
+++ b/runtime/Python2/src/antlr4/TokenStreamRewriter.py
@@ -88,7 +88,8 @@ class TokenStreamRewriter(object):
     def delete(self, program_name, from_idx, to_idx):
         if isinstance(from_idx, Token):
             self.replace(program_name, from_idx.tokenIndex, to_idx.tokenIndex, "")
-        self.replace(program_name, from_idx, to_idx, "")
+        else:
+            self.replace(program_name, from_idx, to_idx, "")
 
     def lastRewriteTokenIndex(self, program_name=DEFAULT_PROGRAM_NAME):
         return self.lastRewriteTokenIndexes.get(program_name, -1)
@@ -104,7 +105,7 @@ class TokenStreamRewriter(object):
 
     def getText(self, program_name, start, stop):
         """
-        :type interval: Interval.Interval
+        :type interval: IntervalSet.Interval
         :param program_name:
         :param interval:
         :return:
@@ -173,13 +174,13 @@ class TokenStreamRewriter(object):
             if any((iop is None, not isinstance(iop, TokenStreamRewriter.InsertBeforeOp))):
                 continue
             prevInserts = [op for op in rewrites[:i] if isinstance(op, TokenStreamRewriter.InsertBeforeOp)]
-            for i, prevIop in enumerate(prevInserts):
+            for prev_index, prevIop in enumerate(prevInserts):
                 if prevIop.index == iop.index and type(prevIop) is TokenStreamRewriter.InsertBeforeOp:
                     iop.text += prevIop.text
-                    rewrites[i] = None
+                    rewrites[prev_index] = None
                 elif prevIop.index == iop.index and type(prevIop) is TokenStreamRewriter.InsertAfterOp:
                     iop.text = prevIop.text + iop.text
-                    rewrites[i] = None
+                    rewrites[prev_index] = None
             # look for replaces where iop.index is in range; error
             prevReplaces = [op for op in rewrites[:i] if type(op) is TokenStreamRewriter.ReplaceOp]
             for rop in prevReplaces:
--- a/runtime/Python3/src/antlr4/TokenStreamRewriter.py
+++ b/runtime/Python3/src/antlr4/TokenStreamRewriter.py
@@ -41,7 +41,10 @@ class TokenStreamRewriter(object):
         self.insertAfter(token.tokenIndex, text, program_name)
 
     def insertAfter(self, index, text, program_name=DEFAULT_PROGRAM_NAME):
-        self.insertBefore(program_name, index + 1, text)
+        op = self.InsertAfterOp(self.tokens, index + 1, text)
+        rewrites = self.getProgram(program_name)
+        op.instructionIndex = len(rewrites)
+        rewrites.append(op)
 
     def insertBeforeIndex(self, index, text):
         self.insertBefore(self.DEFAULT_PROGRAM_NAME, index, text)
@@ -84,8 +87,9 @@ class TokenStreamRewriter(object):
 
     def delete(self, program_name, from_idx, to_idx):
         if isinstance(from_idx, Token):
-            self.replace(program_name, from_idx.tokenIndex, to_idx.tokenIndex, None)
-        self.replace(program_name, from_idx, to_idx, None)
+            self.replace(program_name, from_idx.tokenIndex, to_idx.tokenIndex, "")
+        else:
+            self.replace(program_name, from_idx, to_idx, "")
 
     def lastRewriteTokenIndex(self, program_name=DEFAULT_PROGRAM_NAME):
         return self.lastRewriteTokenIndexes.get(program_name, -1)
@@ -95,6 +99,9 @@ class TokenStreamRewriter(object):
 
     def getProgram(self, program_name):
         return self.programs.setdefault(program_name, [])
+        
+    def getDefaultText(self):
+        return self.getText(self.DEFAULT_PROGRAM_NAME, 0, len(self.tokens.tokens))
 
     def getText(self, program_name, start:int, stop:int):
         """
@@ -114,7 +121,7 @@ class TokenStreamRewriter(object):
         indexToOp = self._reduceToSingleOperationPerIndex(rewrites)
         i = start
         while all((i <= stop, i < len(self.tokens.tokens))):
-            op = indexToOp.get(i)
+            op = indexToOp.pop(i, None)
             token = self.tokens.get(i)
             if op is None:
                 if token.type != Token.EOF: buf.write(token.text)
@@ -134,7 +141,7 @@ class TokenStreamRewriter(object):
             if any((rop is None, not isinstance(rop, TokenStreamRewriter.ReplaceOp))):
                 continue
             # Wipe prior inserts within range
-            inserts = [op for op in rewrites[:i] if isinstance(rop, TokenStreamRewriter.InsertBeforeOp)]
+            inserts = [op for op in rewrites[:i] if isinstance(op, TokenStreamRewriter.InsertBeforeOp)]
             for iop in inserts:
                 if iop.index == rop.index:
                     rewrites[iop.instructionIndex] = None
@@ -148,7 +155,7 @@ class TokenStreamRewriter(object):
                 if all((prevRop.index >= rop.index, prevRop.last_index <= rop.last_index)):
                     rewrites[prevRop.instructionIndex] = None
                     continue
-                isDisjoint = any((prevRop.last_index<rop.index, prevRop.index>rop))
+                isDisjoint = any((prevRop.last_index<rop.index, prevRop.index>rop.last_index))
                 isSame = all((prevRop.index == rop.index, prevRop.last_index == rop.last_index))
                 if all((prevRop.text is None, rop.text is None, not isDisjoint)):
                     rewrites[prevRop.instructionIndex] = None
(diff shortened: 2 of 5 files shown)


ericvergnaud:
Seems to make sense, but the current code equals the Java code, so maybe the bug is not in getText ?
Are you able to compare with the Java behaviour

youkaichao:
Sorry, I'm not familiar with Java. But by looking at the code in ``getText`` in java, I think this bug still exists. Can you test it? 

The bug is in ``getText`` though, in the following code in python:
```
        while all((i <= stop, i < len(self.tokens.tokens))):
            op = indexToOp.get(i)
            token = self.tokens.get(i)
            if op is None:
                if token.type != Token.EOF: buf.write(token.text)
                i += 1
            else:
                i = op.execute(buf)

        if stop == len(self.tokens.tokens)-1:
            for op in indexToOp.values():
                if op.index > len(self.tokens.tokens)-1: buf.write(op.text)
```

if ``stop == len(self.tokens.tokens)-1`` and there's an operation there, then this operation will be executed in the while loop and in the following code, resulting the operation to be executed twice.

parrt:
Hmm...surely i have a unit test for adding after last token...I see `testInsertAfterLastIndex` which seems to show the java one working.

```java
}    @Test public void testInsertAfterLastIndex() throws Exception {
        LexerGrammar g = new LexerGrammar(
                                             "lexer grammar T;\n"+
                                             "A : 'a';\n" +
                                             "B : 'b';\n" +
                                             "C : 'c';\n");
        String input = "abc";
        LexerInterpreter lexEngine = g.createLexerInterpreter(new ANTLRInputStream(input));
        CommonTokenStream stream = new CommonTokenStream(lexEngine);
        stream.fill();
        TokenStreamRewriter tokens = new TokenStreamRewriter(stream);
        tokens.insertAfter(2, "x");
        String result = tokens.getText();
        String expecting = "abcx";
        assertEquals(expecting, result);

    }
```

youkaichao:
Uh, I know it. in python3, it is ``op = indexToOp.get(i)``. But the Java version is ``indexToOp.remove(i);`` and the python2 version is ``op = indexToOp.pop(i, None)``.

@ericvergnaud  @parrt

parrt:
@ericvergnaud blessed?

ericvergnaud:
There seems to be something rotten in the kingdom of Denmark... I’m not comfortable with the code base diverting


Envoyé de mon iPhone

parrt:
Not sure I follow, @ericvergnaud as it looks like something different about the py3 version that is already diverged from java, py2.

ericvergnaud:
I mean it seems the proposed fixes only address side effects, not the root cause.
I need to look into this closer - not my code originally 

Envoyé de mon iPhone

parrt:
Ah. Ok, well, the Java target seems to do right thing here.

ericvergnaud:
@youkaichao thanks for this, makes sense
there is a test suite for this in the python 2 runtime, not sure why it is not present in the python 3 one
would you mind copying it so we get 100% confident that this change actually works for all use cases?

youkaichao:
@ericvergnaud  yeah, I did it just now. 

I'm astonished that such a buggy code exists for such a long time and the python3-runtime is totally a mess.

To make it pass the test, some bugfix are included as well. so #2455  is closed.

runtime/Python2/src/antlr4/TokenStreamRewriter.py
```diff
@@ -88,7 +88,8 @@ def deleteIndex(self, index):
     def delete(self, program_name, from_idx, to_idx):
         if isinstance(from_idx, Token):
             self.replace(program_name, from_idx.tokenIndex, to_idx.tokenIndex, "")
-        self.replace(program_name, from_idx, to_idx, "")
+        else:
+            self.replace(program_name, from_idx, to_idx, "")
```
youkaichao: see #2455

runtime/Python3/src/antlr4/TokenStreamRewriter.py
```diff
@@ -165,28 +172,31 @@ def _reduceToSingleOperationPerIndex(self, rewrites):
...
-                if reduced.get(op.index): raise ValueError('should be only one op per index')
-                reduced[op.index] = op
+        reduced = {}
+        for i, op in enumerate(rewrites):
+            if op is None: continue
+            if reduced.get(op.index): raise ValueError('should be only one op per index')
+            reduced[op.index] = op
 
-            return reduced
+        return reduced
```
youkaichao: Just to find a mis-indented block ...... I don't know what to say

runtime/Python3/src/antlr4/TokenStreamRewriter.py
```diff
@@ -234,4 +247,9 @@ def __init__(self, from_idx, to_idx, tokens, text):
         def execute(self, buf):
             if self.text:
                 buf.write(self.text)
-            return self.last_index + 1
\ No newline at end of file
+            return self.last_index + 1
+            
+        def __str__(self):
```
youkaichao: ``__str__`` is needed when this object is used for formatting string.

runtime/Python3/src/antlr4/TokenStreamRewriter.py
ericvergnaud: @youkaichao welcome to open source, I did not write this code, which strictly speaking is not part of the runtime. I guess @vlad needed it in Python 2 and lazily copied it to Python 3 without testing it. Thanks for your help!

ericvergnaud:
@youkaichao let me know when you think this is ready

youkaichao:
@ericvergnaud  I think it's ready now. You can wait until the checks to be passed, but I don't think it ever runs the python tests...

runtime/Python3/src/antlr4/TokenStreamRewriter.py
```diff
@@ -165,28 +175,31 @@ def _reduceToSingleOperationPerIndex(self, rewrites):
         for i, iop in enumerate(rewrites):
             if any((iop is None, not isinstance(iop, TokenStreamRewriter.InsertBeforeOp))):
                 continue
-            prevInserts = [op for op in rewrites[:i] if isinstance(iop, TokenStreamRewriter.InsertBeforeOp)]
-            for prevIop in prevInserts:
-                if prevIop.index == iop.index:
+            prevInserts = [op for op in rewrites[:i] if isinstance(op, TokenStreamRewriter.InsertBeforeOp)]
+            for i, prevIop in enumerate(prevInserts):
```
ericvergnaud: are you sure about reusing i in an inner loop?

runtime/Python3/src/antlr4/TokenStreamRewriter.py
```diff
@@ -112,12 +122,12 @@ def getText(self, program_name, interval):
         if start < 0: start = 0
 
         # if no instructions to execute
-        if not rewrites: return self.tokens.getText(interval)
+        if not rewrites: return self.tokens.getText(Interval(start, stop))
```
ericvergnaud: same here

youkaichao: No, it's not the same. The code above does some range check and make sure ``start`` and ``stop`` are  in proper range. If we just pass the original ``interval`` with an out-of-range ``stop``, error will be raised in ``BufferedTokenStream.getText``

runtime/Python3/src/antlr4/TokenStreamRewriter.py
youkaichao: Yeah, I have noticed this tricky code. The same code appears in python-2-runtime so I just copied it.

In the inner loop, only ``i`` of ``enumerate(prevInserts)`` is used. And when the inner loop ends, ``i`` becomes ``len(prevInserts) - 1`` which depends on the operation sequence.

The core problem is that ``for`` loop is not a scope in python, so the ``i`` of the inner-loop just covers the outter ``i``.

Since you point this out, I can fix the bug in py2 & py3.

runtime/Python3/src/antlr4/TokenStreamRewriter.py
ericvergnaud: I mean pass range(start, stop)
I already mentioned I was not keen to include Interval in python 3

runtime/Python3/src/antlr4/TokenStreamRewriter.py
```diff
@@ -95,14 +100,19 @@ def setLastRewriteTokenIndex(self, program_name, i):
 
     def getProgram(self, program_name):
         return self.programs.setdefault(program_name, [])
+        
+    def getDefaultText(self):
+        return self.getText(self.DEFAULT_PROGRAM_NAME, Interval(0, len(self.tokens.tokens)))
```
ericvergnaud: use range please, not Interval

runtime/Python3/src/antlr4/TokenStreamRewriter.py
```diff
@@ -95,14 +100,19 @@ def setLastRewriteTokenIndex(self, program_name, i):
...
+    def getText(self, program_name=None, interval=None):
         """
-        :type interval: Interval.Interval
+        :type interval: IntervalSet.Interval
         :param program_name:
         :param interval:
         :return:
         """
+        program_name = program_name or self.DEFAULT_PROGRAM_NAME
+        interval = interval or Interval(0, len(self.tokens.tokens))
```
ericvergnaud: use range please, not Interval

runtime/Python3/src/antlr4/IntervalSet.py
```diff
@@ -8,6 +8,29 @@
 import unittest
 from antlr4.Token import Token
```
ericvergnaud: as mentioned, no need for this class, when range does the job

runtime/Python3/src/antlr4/TokenStreamRewriter.py
```diff
@@ -112,12 +119,12 @@ def getText(self, program_name, interval):
         if start < 0: start = 0
 
         # if no instructions to execute
-        if not rewrites: return self.tokens.getText(interval)
+        if not rewrites: return self.tokens.getText((start, stop))
```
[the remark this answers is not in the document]
youkaichao: Using ``range`` will cause error(cannot pass some tests) .  Because ``range(start, stop)[1]`` returns ``start + 1`` rather than ``stop``.

youkaichao:
I feel it's necessary to comment here about the ``Interval`` class. It's indeed because python3 ``range`` object is not the same as ``Interval`` especially that ``Interval.__getitem__`` behaves somewhat odd and this characteristic is used by ``TokenStream.getText``.  Removing ``Interval`` will cause some tests to fail.

ericvergnaud:
Python 2 Interval does not have a __getitem__ and TokenStream.getText expects a tuple so not sure I follow the reasoning...

Envoyé de mon iPhone

youkaichao:
check https://github.com/antlr/antlr4/blob/master/runtime/Python2/src/antlr4/IntervalSet.py#L27 ?

ericvergnaud:
Not sure how that piece of code made it to master, I will clean things tomorrow 

Envoyé de mon iPhone

parrt:
@ericvergnaud ok, so is this one ready for merging?

ericvergnaud:
@parrt not yet

youkaichao:
@ericvergnaud  so I should modify the code for the new api?

ericvergnaud:
Yes, the new API is merged on master already, so you can just rebase

youkaichao:
@ericvergnaud  ok, I have rebased.

runtime/Python2/src/antlr4/TokenStreamRewriter.py
```diff
@@ -104,7 +105,7 @@ def getDefaultText(self):
 
     def getText(self, program_name, start, stop):
         """
-        :type interval: Interval.Interval
+        :type interval: IntervalSet.Interval
```
ericvergnaud: No longer valid

ericvergnaud:
Good stuff! Thanks for making the effort :-)
@parrt blessed

runtime/Python2/src/antlr4/TokenStreamRewriter.py
youkaichao: OK, I'll fix it in another PR with some other fixes.

PR #2455, opened 2018-12-27
Bugfix/pythonruntime

youkaichao:
@ericvergnaud  Here is an obvious bug.

If we call ``TokenStreamRewriter.delete`` with 2 tokens as argument, then ``replace`` will be called twice, and an exception will occur at the second time, cause ``replace`` requires two ``int`` arguments.


ericvergnaud:
@parrt blessed

youkaichao:
included in #2452

PR #2456, opened 2018-12-28
[CPP] Fix Xpath. Derived class of XPathElement are now transmitted by pointers.

malickf:
Derived class of XPathElement are now transmitted by pointers and the evaluate() function of XPathElement is now a pure virtual function to prevent future object slicing problems.

This is a cleaner version of an old pull request, recently rejected (see https://github.com/antlr/antlr4/pull/2231 ) 

The main problem was that `XPath::getXPathElement`  (https://github.com/antlr/antlr4/blob/98dc2c0f0249a67b797b151da3adf4ffbc1fd6a1/runtime/Cpp/runtime/src/tree/xpath/XPath.cpp#L91) was returning a derived class of XpathElement without a pointer. As the result some informations was lost and the Xpath framework was broken. 

I also changed the `XPath::split` (https://github.com/antlr/antlr4/blob/98dc2c0f0249a67b797b151da3adf4ffbc1fd6a1/runtime/Cpp/runtime/src/tree/xpath/XPath.cpp#L31) to properly treat the argument as a string and not as a filepath. In this way the runtime behave as other runtime (for instance see https://github.com/antlr/antlr4/blob/98dc2c0f0249a67b797b151da3adf4ffbc1fd6a1/runtime/CSharp/runtime/CSharp/Antlr4.Runtime/Tree/Xpath/XPath.cs#L91)

### DIFF
--- a/contributors.txt
+++ b/contributors.txt
@@ -211,3 +211,4 @@ YYYY/MM/DD, github id, Full name, email
 2018/11/15, amykyta3, Alex Mykyta, amykyta3@users.noreply.github.com
 2018/11/29, hannemann-tamas, Ralf Hannemann-Tamas, ralf.ht@gmail.com
 2018/12/20, WalterCouto, Walter Couto, WalterCouto@users.noreply.github.com
+2018/12/28, Malick F.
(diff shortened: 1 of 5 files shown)


malickf:
I'm closing this merge as https://github.com/antlr/antlr4/pull/2667 solves the issue.

mike-lischke:
Oh, sorry, I even approved this patch. Seems I totally forgot about it. Anyway, I'm glad it's fixed now.

PR #2458, opened 2019-01-01
Fix python get text api

ericvergnaud:
Not sure how some stuff made it to master but incorrect semantics with Interval and getText have led to lots of confusion.
Purpose of this PR is to restore a unambiguous API such that contributors can rely on a clear contract.


ericvergnaud:
@parrt please merge this.
This is a breaking change in the python API so will need to describe it in next release notes

parrt:
Thanks so much for cleaning that up, @ericvergnaud !!

Issue #2317, opened 2018-06-27
Specifying an exception after a rule definition doesn't generate python code for python3 target


wkhemir:
Antlr4 doesn't generate correct python code when specifying an exception after a rule definition.

Here is an example using antlr 4.7.1 with python3 target:

```
grammar SomeParser;

options {
    language=Python3;
}

@header {
from antlr4.error.Errors import ParseCancellationException
}

model
    : EOF
    ;
    catch[ParseCancellationException] {pass}
```

The generated `model` fonction is the following:

```
    def model(self):

        localctx = BonsaiParser.ModelContext(self, self._ctx, self.state)
        self.enterRule(localctx, 0, self.RULE_model)
        try:
            self.enterOuterAlt(localctx, 1)
            self.state = 28
            self.match(BonsaiParser.EOF)
        catch (ParseCancellationException) {
        	pass
        }
        finally:
            self.exitRule()
        return localctx
```

`catch (ParseCancellationException) { pass }` is not a valid python syntax.
It should be `except ParseCancellationException: pass`

PR #2462, opened 2019-01-03
Fix code generation for catch statements in python

wkhemir:
Related issue: #2317


ericvergnaud:
Hi,
thanks for this, looks good!
Do you think you would be able to enrich our test suite with a test for this?

ericvergnaud:
@parrt blessed

PR #2486, opened 2019-02-06
Fix Index Error in Python3 GetChild

### LINKED ISSUE #2485: Python3 ParserRuleContext getChild casues Index Error
This line should check that we don't try to access positions in the list that do not exist. Since lists in python3 are indexed starting with position 0, the check between the position we try to access and the list length should be a strict `> i`, not the current `>= i`, which leads to IndexErrors when checking a position equal to the number of elements in the list.

https://github.com/antlr/antlr4/blob/de316ff5ce3fa7c05ad706335ae274e382c92709/runtime/Python3/src/antlr4/ParserRuleContext.py#L116

ralucado:
Fixes #2485


ericvergnaud:
Hi,

thanks for this. Could you also update Python 2 if required?

ralucado:
> Hi,
> 
> thanks for this. Could you also update Python 2 if required?

I checked, Python2 is already correct.

ericvergnaud:
@parrt blessed

PR #2487, opened 2019-02-06
[WIP] Generate a valid tree structure when errors occur

### LINKED ISSUE #30: add rule context object factory?
``` java
public static class LExprBaseContextFactory {
    public SContext createS(ParserRuleContext<Token> parent, int state) {
        return new SContext(parent, state);
    }
    public EContext createE(ParserRuleContext<Token> parent, int state, int _p) {
        return new EContext(parent, state, _p);
    }
    public MultContext createMult(EContext parent) { return new MultContext(parent); }
    public AddContext createAdd(EContext parent) { return new AddContext(parent); }
    public IntContext createInt(EContext parent) { return new IntContext(parent); }
}
```

### LINKED ISSUE #1972: Error recovery should create missing tokens for first path to recovery point
Currently, error recovery does not create tree nodes or tokens for grammar elements that appear between the point where recovery efforts start and where they finally succeed. The skipped tokens are simply skipped. This causes problems for listeners and visitors which expect the parse tree to return non-null values for element accessors for required references.

For example, consider the following:

```antlr
someElement
  : first second third
  ;
```

In this code, it is clear that `someElement` can never parse in a way that `first`, `second`, or `third` is missing. Therefore, it is reasonable to expect that the corresponding accessors in code, `first()`, `second()`, and `third()`, will all return non-null trees. We can ensure these conditions are met by modifying the error recovery to follow the first path to the end of a rule, manufacturing rule and terminal nodes as necessary to ensure the resulting parse tree contains a minimally-correct structure.

* Take "First path to end"
    * Do not loop back for `+` or `*` blocks
    * Do not enter `*` or `?` blocks
    * Otherwise take the first alternative for decisions
* Create a context node when following a `RuleTransition`
* Create a zero-width `TerminalNode` when traversing a match transition

### Background

It's straightforward to see how this could improve the ability of users to reason about the shape of a parse tree without constantly worrying about what happens if/when an error occurs. The *shape* will be valid even if the *content* is not, which improves over the current situation where visitors and listeners that operate on trees produced by erroneous code must worry about *both* potentially being invalid.

However, this became a particularly obvious usability problem during the implementation of the TypeScript target. TypeScript's type system has distinct representations of null (we used `undefined` uniformly), which impacted the generated accessors:

```antlr
someElement
  : element // This accessor returns a non-null type
  ;

someOptionalElement
  : element? // This accessor returns a nullable type
  ;
```

In order to ensure the accessor methods behaved accurately with respect to the type system, the accessor generated for `SomeElementContext.element()` *throws* in the TypeScript target instead of returning a null value. However, if the error recovery effort traversing through `someElement` actually ensured that it contained an `ElementContext` child, the type system would be accurate and users would not need to worry about exceptions except in cases where they truly failed to write code matching a form appearing in the original grammar.

sharwell:
Fixes #30
Fixes #1972

🚧 Needs tests, including tests involving left-recursive rules


runtime/Java/src/org/antlr/v4/runtime/Parser.java
```diff
@@ -249,6 +249,15 @@ public Token matchWildcard() throws RecognitionException {
 		return t;
 	}
 
+	protected ParserRuleContext createContext(int ruleIndex, int altIndex, ParserRuleContext parent, int invokingState, Object... arguments) {
+		InterpreterRuleContext context = new InterpreterRuleContext(parent, invokingState, ruleIndex);
+		if (altIndex > 0) {
+			context.setAltNumber(altIndex);
```
sharwell: 💭 Maybe we should make this the caller's responsibility? Right now it's not consistently called.

runtime/Java/src/org/antlr/v4/runtime/DefaultErrorStrategy.java
```diff
@@ -766,4 +776,98 @@ protected void consumeUntil(Parser recognizer, IntervalSet set) {
...
+	 */
+	protected void buildErrorTree(Parser recognizer) {
+		final ATN atn = recognizer.getATN();
+		int state = recognizer.getState();
+		ParserRuleContext context = recognizer.getContext();
+		final IntegerStack returnStateStack = new IntegerStack();
+		while (true) {
+			ATNState p = atn.states.get(state);
+			int edge;
+			switch (p.getStateType()) {
```
sharwell: 💭 This should probably handle the decision for a rule:

* If an error occurs in a rule before passing the rule entry decision and the rule has labeled alternatives, we need to make sure the correct context is constructed
* We need to call `context.setAltNumber`

runtime/Java/src/org/antlr/v4/runtime/Parser.java
```diff
@@ -249,6 +249,15 @@ public Token matchWildcard() throws RecognitionException {
 		return t;
 	}
 
+	protected ParserRuleContext createContext(int ruleIndex, int altIndex, ParserRuleContext parent, int invokingState, Object... arguments) {
```
sharwell: 💭 Right now `arguments` is available but not used. Options include:

1. Omit `arguments` and rely on the context constructor that doesn't take any
2. Require `arguments` match the form of the context constructor that includes all arguments
3. Allow `arguments` to be omitted by the called (use the context constructor that doesn't include them), or specified (use the context constructor that passes arguments)

If we do allow arguments, we will likely need a helper function like this:

```
private static Object getArgumentOrDefault(Object[] arguments, int index, Object defaultValue) {
  if (arguments == null || index >= arguments.length || arguments[index] == null) {
    return defaultValue;
  }

  return arguments[index];
}
```

runtime/Java/src/org/antlr/v4/runtime/Parser.java
sharwell: ➡️ Now the callers responsibility

runtime/Java/src/org/antlr/v4/runtime/DefaultErrorStrategy.java
sharwell: ➡️ Now fixed

runtime/Java/src/org/antlr/v4/runtime/Parser.java
sharwell: ➡️ Implemented the third option (arguments are optional)

KvanTTT:
Thank you for such very useful improvement! It moves ANTLR parse tree one step closer to full fidelity tree and removes potential NRE during bypass. Do you have plans of implementation leading-trivia hidden tokens binding that Roslyn analyzer has? See the linked PR https://github.com/antlr/antlr4/pull/1667

Issue #2481, opened 2019-01-30
Lots of compiler warnings when compiling c++ library with gcc


majojoe:
When compiling antlr4 c++ library with a the gcc compiler on Ubuntu 18.04 using ExternalAntlr4Cpp I get many of the following type of error:
======================================================
In file included from /home/jo/localgit/ .... /antlr4_runtime/src/antlr4_runtime/runtime/Cpp/runtime/src/atn/PredictionContext.h:10:0,
                 from /home/jo/localgit/..../antlr4_runtime/src/antlr4_runtime/runtime/Cpp/runtime/src/atn/ATNSimulator.h:11,
                 from /home/jo/localgit/.../antlr4_runtime/src/antlr4_runtime/runtime/Cpp/runtime/src/Recognizer.cpp:12:
/home/jo/localgit/..../antlr4_runtime/src/antlr4_runtime/runtime/Cpp/runtime/src/atn/ATNState.h:73:26: warning: type attributes ignored after type is already defined [-Wattributes]
   class ANTLR4CPP_PUBLIC ATN;
                          ^~~
======================================================
I tried that also on mingw and VS 2017. Everywhere the same. While I like antlr4 and everything works as expected, the compiler warnings are very annoying and make an unprofessional impression.
The bug seems to be related to #2450 .

PR #2488, opened 2019-02-06
Feature/fix gcc compiler warnings

majojoe:
This fixes the compiler warnings with gcc (mingw) and fixes issue #2481 .
I'm sorry, but I couldn't compile the cpp tests since there is a linker error (tests broken or maybe a dependency is missing? - Have not investigated further). The java tests work. 
However the runtime compiles with clang and gcc without warnings and the runtime seems to work flawlessly for me.


ritchiecarroll:
FYI, the change from `class ANTLR4CPP_PUBLIC ATN;` to `class ATN;` fixed my build warnings using gcc on Linux...

PR #2490, opened 2019-02-07
Feature added: pretty print tree output

huperniketes:


huperniketes:
Instead of a very long line of code that looks like it was written by a masochistic LISP programmer, the output should be nicely formatted over several lines with indentation.

mike-lischke:
@parrt This PR can be closed. We have an alternative PR merged already.

Issue #2366, opened 2018-09-29
[C++] possible optimization of ParserATNSimulator


garzon:
I have found that `antlr4::atn::ATNConfigSet::add` takes 25.76% of whole CPU time in my parsing-intensive program. And I am wondering if I can modify `ParserATNSimulator.cpp`(line 726) to the following, which 
significantly reduces the time of `antlr4::atn::ATNConfigSet::add` to 8.76%. In this way the time-consuming memory allocation and hash calculating and tracking of `config` in `ATNConfigSet::add` could be prevented.

If there is no problem with it, I could make a PR later.

```c++
size_t ParserATNSimulator::getSynValidOrSemInvalidAltThatFinishedDecisionEntryRule(ATNConfigSet *configs,
  ParserRuleContext *outerContext)
{
  std::pair<size_t, size_t> sets = splitAccordingToSemanticValidity(configs, outerContext);
  //std::unique_ptr<ATNConfigSet> semValidConfigs(sets.first);
  //std::unique_ptr<ATNConfigSet> semInvalidConfigs(sets.second);
  size_t alt = sets.first;//getAltThatFinishedDecisionEntryRule(semValidConfigs.get());
  if (alt != ATN::INVALID_ALT_NUMBER) { // semantically/syntactically viable path exists
    return alt;
  }
  // Is there a syntactically valid path with a failed pred?
  /*if (!semInvalidConfigs->configs.empty()) {
    alt = getAltThatFinishedDecisionEntryRule(semInvalidConfigs.get());
    if (alt != ATN::INVALID_ALT_NUMBER) { // syntactically viable path exists
      return alt;
    }
  }
  return ATN::INVALID_ALT_NUMBER;*/
  return sets.second;
}

/*size_t ParserATNSimulator::getAltThatFinishedDecisionEntryRule(ATNConfigSet *configs) {
  misc::IntervalSet alts;
  for (auto &c : configs->configs) {
    if (c->getOuterContextDepth() > 0 || (is<RuleStopState *>(c->state) && c->context->hasEmptyPath())) {
      alts.add(c->alt);
    }
  }
  if (alts.size() == 0) {
    return ATN::INVALID_ALT_NUMBER;
  }
  return alts.getMinElement();
}*/

inline static void add_alts(misc::IntervalSet &alts, const Ref<ATNConfig> &c) {
	if (c->getOuterContextDepth() > 0 || (is<RuleStopState *>(c->state) && c->context->hasEmptyPath())) {
[...]

PR #2493, opened 2019-02-11
Optimize getSynValidOrSemInvalidAltThatFinishedDecisionEntryRule

sharwell:
Implements the optimization identified by @garzon.

See antlr/antlr4#2366

### DIFF
--- a/runtime/Java/src/org/antlr/v4/runtime/atn/ParserATNSimulator.java
+++ b/runtime/Java/src/org/antlr/v4/runtime/atn/ParserATNSimulator.java
@@ -1297,65 +1297,51 @@ public class ParserATNSimulator extends ATNSimulator {
 	protected int getSynValidOrSemInvalidAltThatFinishedDecisionEntryRule(ATNConfigSet configs,
 																		  ParserRuleContext outerContext)
 	{
-		Pair<ATNConfigSet,ATNConfigSet> sets =
-			splitAccordingToSemanticValidity(configs, outerContext);
-		ATNConfigSet semValidConfigs = sets.a;
-		ATNConfigSet semInvalidConfigs = sets.b;
-		int alt = getAltThatFinishedDecisionEntryRule(semValidConfigs);
-		if ( alt!=ATN.INVALID_ALT_NUMBER ) { // semantically/syntactically viable path exists
-			return alt;
-		}
-		// Is there a syntactically valid path with a failed pred?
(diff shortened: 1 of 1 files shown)


runtime/Java/src/org/antlr/v4/runtime/atn/ParserATNSimulator.java
```diff
@@ -1297,65 +1297,51 @@ else if (altToPred[i] != SemanticContext.NONE) {
...
-		}
-		return ATN.INVALID_ALT_NUMBER;
-	}
 
-	protected int getAltThatFinishedDecisionEntryRule(ATNConfigSet configs) {
-		IntervalSet alts = new IntervalSet();
-		for (ATNConfig c : configs) {
-			if ( c.getOuterContextDepth()>0 || (c.state instanceof RuleStopState && c.context.hasEmptyPath()) ) {
-				alts.add(c.alt);
+			if (c.getOuterContextDepth() == 0 && !(c.state instanceof RuleStopState && c.context.hasEmptyPath())) {
```
sharwell: 📝 This condition should be verified in detail next to the original, which was part of `getAltThatFinishedDecisionEntryRule`

runtime/Java/src/org/antlr/v4/runtime/atn/ParserATNSimulator.java
```diff
@@ -1297,65 +1297,51 @@ else if (altToPred[i] != SemanticContext.NONE) {
...
-			if ( alt!=ATN.INVALID_ALT_NUMBER ) { // syntactically viable path exists
-				return alt;
+		// The minimum alternative that finished the decision entry rule and is
+		// valid according to semantic predicates.
+		int minAltSucceeded = Integer.MAX_VALUE;
+		// The minimum alternative that finished the decision entry rule but is
+		// not valid according to semantic predicates. Only considered if
+		// minAltSucceeded does not produce an alternative.
+		int minAltFailed = Integer.MAX_VALUE;
+		for (ATNConfig c : configs) {
```
sharwell: 📝 These are not ordered by alternative, so it's not safe to break out early unless we reach a case where `minAltSucceeded` is equal to the minimum alternative in `configs`, but that value isn't cached and I'm not sure it would be worthwhile to handle it specifically. The current implementation does not make any assumptions about order so it's a safe first step here.

PR #2500, opened 2019-02-20
I don't understand how I can override existing grammar datasets to make my code fully async/await.

ameyinvent:


ericvergnaud:
Hey @ameyinvent 
not sure anyone is willing to review so many changes in just one PR
there seems to be a lot of historical commits here
suggest you rebase from master, and submit 1 PR on a per atomic change basis

KvanTTT:
Please close this MR since it generates notification noise on every git event.

PR #2502, opened 2019-03-01
Better word choice

atlx:


parrt:
thank you for the submission. I changed it locally to be "special cases for..."

PR #2505, opened 2019-03-07
Feature: Add pretty print AST to CPP runtime

foxever:
Add pretty print option, default is non pretty print, the second argument is pretty print flag.
Ref #2490 

```cpp
tree->toStringTree(&parser, true);
```
Here's the demo output
```
(main 
        (stat 
            (expr 
                (id 🍴)) = 
            (expr 
                (expr 
                    (id 🍐)) + 
                (expr "😎")) ;) 
        (stat 
            (expr 
                (expr 
                    (expr ( 
                        (expr ( 
                            (expr ( 
                                (expr 
                                    (expr 
                                        (id x)) * 
                                    (expr 
                                        (id π))) )) )) )) * 
                    (expr 
                        (id µ))) + 
                (expr 
                    (id ∰))) ;) 
        (stat 
            (expr 
                (expr 
                    (id a)) + 
                (expr ( 
                    (expr 
                        (expr 
                            (expr 
                                (id x)) * 
                            (expr ( 
                                (expr 
                                    (expr 
                                        (id y)) ? 
                                    (expr 0) : 
                                    (expr 1)) ))) + 
                        (expr 
                            (id z))) ))) ;) <EOF>)
```

### DIFF
--- a/runtime/Cpp/runtime/src/RuleContext.h
+++ b/runtime/Cpp/runtime/src/RuleContext.h
@@ -110,15 +110,15 @@ namespace antlr4 {
     ///  (root child1 .. childN). Print just a node if this is a leaf.
     ///  We have to know the recognizer so we can get rule names.
     /// </summary>
-    virtual std::string toStringTree(Parser *recog) override;
+    virtual std::string toStringTree(Parser *recog, bool pretty = false) override;
 
     /// <summary>
     /// Print out a whole tree, not just a node, in LISP format
     ///  (root child1 .. childN). Print just a node if this is a leaf.
     /// </summary>
-    virtual std::string toStringTree(std::vector<std::string> &ruleNames);
+    virtual std::string toStringTree(std::vector<std::string> &ruleNames, bool pretty = false);
 
-    virtual std::string toStringTree() override;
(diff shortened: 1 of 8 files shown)


runtime/Cpp/runtime/src/RuleContext.h
```diff
@@ -110,15 +110,15 @@ namespace antlr4 {
...
     /// </summary>
-    virtual std::string toStringTree(Parser *recog) override;
+    virtual std::string toStringTree(Parser *recog, bool pretty=false) override;
 
     /// <summary>
     /// Print out a whole tree, not just a node, in LISP format
     ///  (root child1 .. childN). Print just a node if this is a leaf.
     /// </summary>
-    virtual std::string toStringTree(std::vector<std::string> &ruleNames);
+    virtual std::string toStringTree(std::vector<std::string> &ruleNames, bool pretty=false);
```
mike-lischke: Please use a space on each side of the equal sign (do this for all occurences in this patch).

foxever: fixed

PR #2514, opened 2019-03-18
DOCS: do not use python built-in as variable name

carlodri:
do not use `input` as  a variable name since it is a [Python built-in function](https://docs.python.org/3/library/functions.html#input)


ericvergnaud:
@parrt blessed

PR #2517, opened 2019-03-21
Fix imports in the Python2 sources.

renatahodovan:
Contrary to Python3, the lexers and parser generated for Python2 target
support only relative imports which causes a failure in case of
standalone scripts. The patch adapts these imports to work in both
cases.


ericvergnaud:
Thanks Renata!
@parrt blessed

Issue #1098, opened 2016-01-26
Incorrect tokens line number for files with '\r' newline.


KvanTTT:
I have the following grammar:
```ANTLR
lexer grammar T;
ID: [a-z]+;
WS: [ \n] -> skip;
```
And the following input: ```"a" + newLine + "b" + newLine + "c"```

* If newLine is **\r\n**, I got this:
```
[@0,0:0='a',<1>,1:0]
[@1,3:3='b',<1>,2:0]
[@2,6:6='c',<1>,3:0]
[@3,7:6='<EOF>',<-1>,3:1]
```
newLine == **\n**:
```
[@0,0:0='a',<1>,1:0]
[@1,2:2='b',<1>,2:0]
[@2,4:4='c',<1>,3:0]
[@3,5:4='<EOF>',<-1>,3:1]
```
newLine == **\r**:
```
[@0,0:0='a',<1>,1:0]
[@1,2:2='b',<1>,1:2]
[@2,4:4='c',<1>,1:4]
[@3,5:4='<EOF>',<-1>,1:5]
```

In last case line numer for all tokens is 1 and it's wrong. I suggest to support '\r' newLine character. Another variants described at [Wikipedia](https://en.wikipedia.org/wiki/Newline#Representations), but they are too rarely used. Moreover '\r' newline supported by most part of modern editors (IDEA, Visual Studio, Notepad++ etc.).

jimidle:
It's probably easier to just convert the \r to \n before you parse it.

KvanTTT:
I thing it's not a very good idea. In this case ```\r\n``` will be treated as two lines, whereas it's only one line. Moreover it's better to leave input stream as is.

jimidle:
There is no universally applicable solution for this - you will always find
an exception case and you start to add little bits of overhead trying to
cater for everyone. In the C runtime I just looked for a newline and
ignored CR, but made it so that you could change the triggering character
at runtime. That's about the best you can do generically I think. Checking
for a following \n after a \r is all adding overhead, and where do you stop.

You could add your own code to supply tokens and change the line numbers
based on your own tracking, but I would just change the \r to \n on the
fly. You could do that with your own input stream for instance.

Jim

ericvergnaud:
Hi,

historically, ‘\r’ is not a new line, but rather a carriage return, hence the ‘\r\n’ sequence.
so not sure where you take your input from, but you might be better fixing the input.

Eric

parrt:
[...]

PR #2519, opened 2019-03-24
Fix #1890 Standalone CR should be recognized as line separator

### LINKED ISSUE #1890: Carriage return is not treated as newline
When files having '\r' as newlines are parsed, the line numbers and the character positions in line do not correctly reflect the situation. For instance:

Grammar:

    grammar Hello;            // Define a grammar called Hello
    r  : 'hello' ID ;         // match keyword hello followed by an identifier
    ID : [a-z]+ ;             // match lower-case identifiers
    WS : [ \t\r\n]+ -> skip ; // skip spaces, tabs, newlines, \r (Windows) 

Input file: [input.txt](https://github.com/antlr/antlr4/files/1036864/input.txt)

Output:

     grun Hello r -tokens input.txt
     [@0,0:4='hello',<'hello'>,1:0]
     [@1,7:11='parrt',<ID>,1:7]
     [@2,12:11='<EOF>',<EOF>,1:12]

_2nd and 3rd token don't lie on the 1st line. The parser is failing to recognize CR as newline._

Seems to relate to [this](https://github.com/antlr/antlr4/blob/master/runtime/Java/src/org/antlr/v4/runtime/atn/LexerATNSimulator.java#L738) part of the code.

nixel2007:
Hello!

I've made changes to treat stand-alone CR as line separator.
This changes were tested on millions of LOC with-out any noticable perfomance regress. Also I use my fork to parse files everyday as a part of static analysis process.

If antlr4 authors want I can make the same changes for all other supported languages.

### DIFF
--- a/runtime/Java/src/org/antlr/v4/runtime/atn/LexerATNSimulator.java
+++ b/runtime/Java/src/org/antlr/v4/runtime/atn/LexerATNSimulator.java
@@ -70,7 +70,6 @@ public class LexerATNSimulator extends ATNSimulator {
 	/** The index of the character relative to the beginning of the line 0..n-1 */
 	protected int charPositionInLine = 0;
 
-
 	public final DFA[] decisionToDFA;
 	protected int mode = Lexer.DEFAULT_MODE;
 
@@ -735,8 +734,13 @@ public class LexerATNSimulator extends ATNSimulator {
 		if ( curChar=='\n' ) {
 			line++;
 			charPositionInLine=0;
-		}
-		else {
+		} else if ( curChar == '\r' ) {
+			int nextChar = input.LA(2);
+			if ( nextChar != '\n') {
+				line++;
+				charPositionInLine=0;
+			}
+		} else {
 			charPositionInLine++;
 		}
 		input.consume();
--- a/contributors.txt
+++ b/contributors.txt
@@ -230,3 +230,4 @@ YYYY/MM/DD, github id, Full name, email
 2019/09/10, amorimjuliana, Juliana Amorim, juu.amorim@gmail.com
 2019/09/17, kaz, Kazuki Sawada, kazuki@6715.jp
 2019/09/28, lmy269, Mingyang Liu, lmy040758@gmail.com
+2019/03/24, nixel2007, Nikita Gryzlov, nixel2007@gmail.com


KvanTTT:
Link to an issue: #1098

nixel2007:
Ping?

KvanTTT:
@parrt maybe it makes sense to accept the `\r` line separator because a lot of users encounter the same problem and report them. On the other hand, the fix looks simple.

parrt:
If there is an isolated `\r` it doesn't reset line nor bump char position. doesn't nothing. is it an error?  i think only old macs used `\r` in isolation...

ericvergnaud:
Given the very nature of Antlr parsers i.e. softwares that parse input, I think we should keep options open. If for some reason someone needs to treat isolated \r in a special manner, they should be able to do so. So maybe we should

KvanTTT:
As I've already written in [the issue comment](https://github.com/antlr/antlr4/issues/1098#issuecomment-1007425775):

> From Wiki: Commodore 8-bit machines (C64, C128), Acorn BBC, ZX Spectrum, TRS-80, Apple II series, Oberon, the classic Mac OS, MIT Lisp Machine and OS-9 uses \r as the new line separator. But there are also more rare line separators as \036, \025 that I have never seen.

> But it looks like many users encounter the problem with \r separator. Also all text editors and IDE that I've tested support such line separator (Notepad++, VS Code, IntelliJ, even Windows Notepad). Thus, it's not a very rare line separator.

> I suggest supporting it especially since the fix is quite simple.

But this pull request is not okay because it does not have a unit test and fixes only Java runtime (other runtimes should accept `\r` as well).

nixel2007:
@KvanTTT 

See OP:

> If antlr4 authors want I can made the same changes for all other supported languages.

KvanTTT:
Also, [BaseRuntimeTest](https://github.com/antlr/antlr4/blob/master/runtime-testsuite/test/org/antlr/v4/test/runtime/BaseRuntimeTest.java#L436-L453) should be improved to accept `\r\n` and `\r` line separators (now it converts `\r\n` to `\n` that's bad).

runtime/Java/src/org/antlr/v4/runtime/atn/LexerATNSimulator.java
```diff
@@ -735,8 +734,13 @@ public void consume(CharStream input) {
...
 			charPositionInLine=0;
-		}
-		else {
+		} else if ( curChar == '\r' ) {
+			int nextChar = input.LA(2);
+			if ( nextChar != '\n') {
+				line++;
+				charPositionInLine=0;
+			}
+		} else {
```
KvanTTT, line 743: Maybe it's better to change to the following?

```suggestion
		} else if ( curChar == '\r' ) {
			if (input.LA(2) == '\n') {
			    input.consume();
			}
			line++;
		    charPositionInLine=0;
		} else {
```

It looks more optimal since it does not require reading `\n` twice.

KvanTTT (commented):
* A unit test is required (`LexerExec/CarriageReturnAsLineSeparator.txt`). Also, [readDescriptor](https://github.com/antlr/antlr4/blob/master/runtime-testsuite/test/org/antlr/v4/test/runtime/BaseRuntimeTest.java#L436-L453) should be improved to accept `\r` and don't change to `\n` all line separators from the input stream.
* Implementation for other runtime is required as we've discussed.

ericvergnaud:
Hi,
given the very nature of ANTLR, not sure any of this should be hard-coded.
If the input is not text, the CR might NOT mean what it means in a text editor.
It's a bit of a chicken-and-egg situation because the lexer uses CR/LF do detect line breaks before it actually parses them...
Would it make sens to have this be a configuration of the input stream ?
Something like:
inputStream = inputStream.withLineBreaksPattern("\r?\n")
such that users can adapt to whatever situation they need, although the default would be the above ?

KvanTTT:
> If the input is not text, a CR might NOT mean what it means in a text editor.

I think it's a very rare case when ANTLR is used for not text parsing (In my opinion it's not very suitable for binary parsing) and when `\r` is not treated as a new line. If so, `\r\n` and `\n` also should be defined by the user but they are defined by default.

Single `\n` is also line break without carriage return but actually, it's always used as `\r\n`.

I'm basing on the most widespread situations case and ANTLR practical usage. In most cases, `\r` is a line separator. If it's not, behavior can be changed as you suggest.

> Would it make sense to have this be a configuration of the lexer ?

I have no objection but it complicates runtimes. Anyway by default line separator should be `(\r?\n|\r)` because almost all text tools (if not all) support it and it surprises the user if a tool doesn't support such a separator.

ericvergnaud:
> > If the input is not text, a CR might NOT mean what it means in a text editor.
> 
> I think it's a very rare case when ANTLR is used for not text parsing (In my opinion it's not very suitable for binary parsing) and when `\r` is not treated as a new line. If so, `\r\n` and `\n` also should be defined by the user but they are defined by default.
> 
> Single `\n` is also line break without carriage return but actually, it's always used as `\r\n`.
> 
> I'm basing on the most widespread situations case and ANTLR practical usage. In most cases, `\r` is a line separator. If it's not, behavior can be changed as you suggest.
> 
> > Would it make sense to have this be a configuration of the lexer ?
> 
> I have no objection but it complicates runtimes. Anyway by default line separator should be `(\r?\n|\r)` because almost all text tools (if not all) support it and it surprises the user if a tool doesn't support such a separator.
> 
> > lexer.setLineBreaksPattern("\r?\n")
> 
> Also, actually binary files don't have "lines" at all.

You'd be surprised how many people use antlr for use cases it wasn't designed for.
It's also a very rare case when text has CR without LF, so changing the current behavior for a rare case, when it can be made optional doesn't sound like the best approach.

parrt:
Hi guys, Thanks very much for helping me think through this problem. Excellent analyses. My thoughts:

* `\r` in isolation is a rare case these days
* People parsing `\r` only text right now must be doing something; the obvious solution is simply to either replace with `\n` prior to processing or using a custom character stream or file loader
* Existing code might rely on current behavior, across multiple targets
* Making this change in Java means changing all targets to suit, possibly updating documentation
* It's unclear what the effect on run-time speed would be because the Java JIT could be inlining consume() at the moment; adding more code including a conditional could prevent the inlining and cause a virtual method call.  Given how much we tuned the java target, I would be very careful about changing something that affects every character consumption

In the end, given the simple solution and potential issues arising from a big change across targets for a rare situation is probably not worth it. Consequently, I'm going to close this one with the recommendation that people use a custom character stream or file loader.

parrt:
> Also, [BaseRuntimeTest](https://github.com/antlr/antlr4/blob/master/runtime-testsuite/test/org/antlr/v4/test/runtime/BaseRuntimeTest.java#L436-L453) should be improved to accept `\r\n` and `\r` line separators (now it converts `\r\n` to `\n` that's bad).

a unit test that checked `\r\n` would be good.

nixel2007:
Thanks for your opinion, @parrt 

My 2 cents: instead of changing character stream (which could lead to position errors) user may override LexerATNSimulator:
https://github.com/1c-syntax/bsl-parser/blob/b32b5c83b7afe84831764f6e88a794bce613cda9/src/main/java/com/github/_1c_syntax/bsl/parser/CRAwareLexerATNSimulator.java

and pass it to lexer as `member`:

https://github.com/1c-syntax/bsl-parser/blob/b32b5c83b7afe84831764f6e88a794bce613cda9/src/main/antlr/BSLLexer.g4#L29-L35

parrt:
Ah. I wondered about subclassing to change consume() directly. thanks!

KvanTTT:
> a unit test that checked \r\n would be good.

I started working on it in another pull request: [Preserve line separators for input runtime tests dat](https://github.com/antlr/antlr4/pull/3483). But there are some problems because git changes EOL depending on OS and other options. On the other hand, we don't have another input format that allows chars escaping: `1\r\n2\r\n3`. I'm thinking about how to implement such a test in the best way.

@nixel2007 yes, `CRAwareLexerATNSimulator` is a much more preferable option since it does not require changing all input string.

nixel2007:
> But there are some problems because git changes EOL depending on OS and other options. 

we fixed EOLs on test files with .gitattributes files

KvanTTT:
Yes, I've tried to ignore EOL changing for the entire directory:

```
/runtime-testsuite/resources/org/antlr/v4/test/runtime/descriptors/*.txt -text
```

But maybe it makes sense to set up `\r\n` just to the certain file:

```
LineSeparator_CRLF.txt eol=crlf
LineSeparator_LF.txt eol=lf
```

KvanTTT:
BTW, it looks like ANTLR 3 also does not treat single `\r` as line separator because line numbers are wrong with such grammars (ANTLR 3 is used for parsing ANTLR 4 grammars).

cxambs:
> Thanks for your opinion, @parrt
> 
> My 2 cents: instead of changing character stream (which could lead to position errors) user may override LexerATNSimulator: https://github.com/1c-syntax/bsl-parser/blob/b32b5c83b7afe84831764f6e88a794bce613cda9/src/main/java/com/github/_1c_syntax/bsl/parser/CRAwareLexerATNSimulator.java
> 
> and pass it to lexer as `member`:
> 
> https://github.com/1c-syntax/bsl-parser/blob/b32b5c83b7afe84831764f6e88a794bce613cda9/src/main/antlr/BSLLexer.g4#L29-L35

I am trying this approach in CSharp, but Consume is not a virtual member.
Suggestions?

cxambs:
And, of course, shadowing it won't work as well.
Probably methods from the generated Lexer could be all virtual.
Without that option for now, what are my options?

PR #2520, opened 2019-03-24
Clean up namespace of generated Python3 listeners

akosthekiss:
Added `del <file.parserName>` to the end of the `ListenerFile`
template the same way it is used in `VisitorFile`.


ericvergnaud:
@parrt blessed

PR #2527, opened 2019-04-07
Add incremental parsing support

dberlin:
This commit adds incremental parsing support to ANTLR4
. 
I have only updated the Java target, and the out-of-tree typescript target (see https://github.com/tunnelvisionlabs/antlr4ts/pull/414), but it should be *very* easy to update the other targets for someone who understands that language.  The changes are deliberately minimal.

The Java version here is actually a backport of the typescript version, and took O(2 hours).
(as an aside, i have not written Java in a few years, so i totally expect there are things that could be done better). The comments were originally written for the typescript version, I will go through and clean them up.

A detailed description of how it works is [here](https://github.com/dberlin/antlr4ts/blob/incremental/IncrementalParsing.md) (which also lists the outstanding issues), but it's a very straightforward implementation of detection of rules that could be affected by token changes.  Rule contexts that can't have been affected by a set of token changes are reused and the rules are not re-run.  To account for possibly infinite lookahead/lookbehind, we keep track of how far ahead/behind the parser looked last time on each rule, and use that as the bounds to detect changes in.

The tests currently test on a simple grammar and the JavaLR grammar (which exercises the left recursion removal support).

The only class i've added that requires anything even mildly interesting of the runtime is the IncrementalParserData class. 
Most of the work there is related to changing the start/end tokens of rule contexts to realign them with the token stream changes.  If you only care about the text of the parse tree, and not the position/etc info, this is obviously unnecessary. I have not made this an option.

To track changed tokens and stream adjustments,  the Java version of IncrementalParseData uses TreeMap/TreeSet.  The Typescript versions uses arrays of ranges and binary search (see https://github.com/dberlin/antlr4ts/blob/incremental/src/IncrementalParserData.ts)

I am happy to encapsulate this into a data structure in the runtime if anyone thinks it is worth it. 

As for why do this at all: Yes, ANTLR is actually pretty fast.
My use case is a bit weird - large [GCode](https://en.wikipedia.org/wiki/G-code) files, which are often 20+ megabytes. As such, a single parse takes 6-10 seconds (for a 20 meg file).
Users often make small edits to various pieces.
(It's part of a vscode extension).

Lexing GCode is also completely trivial to do in a contextless fashion. 
The incremental parser brings the reparse time down to <50ms.

I may get around to adding incremental lexing. As i'm sure Terrence knows, this is " trickier".

I have the beginnings of support (elsewhere) based on some papers, but it is incestuous (the parser tells the lexer what tokens could be valid at a given change point and the lexer tries those rules).  There are ways that don't do this, but some require being able to store/rewind/replay the transition state at each token, etc.

### DIFF
--- a/runtime/Java/src/org/antlr/v4/runtime/IncrementalParser.java
+++ b/runtime/Java/src/org/antlr/v4/runtime/IncrementalParser.java
@@ -1,0 +1,175 @@
+/*
+ * Copyright 2019 The ANTLR Project. All rights reserved.
+ * Licensed under the BSD-3-Clause license. See LICENSE file in the project root for license information.
+ */
+package org.antlr.v4.runtime;
+
+import org.antlr.v4.runtime.misc.Interval;
+import org.antlr.v4.runtime.tree.ErrorNode;
+import org.antlr.v4.runtime.tree.ParseTreeListener;
+import org.antlr.v4.runtime.tree.TerminalNode;
+
+/**
+ * Incremental parser implementation
+ * <p>
+ * There are only two differences between this parser and the underlying regular
+ * Parser - guard rules and min/max tracking
+ * <p>
+ * The guard rule API is used in incremental mode to know when a rule context
+ * can be reused. It looks for token changes in the bounds of the rule.
+ * <p>
+ * The min/max tracking is used to track how far ahead/behind the parser looked
+ * to correctly detect whether a token change can affect a parser rule in the future (IE when
+ * handed to the guard rule of the next parse)
+ *
+ * @note See IncrementalParsing.md for more details on the theory behind this.
+ * In order to make this easier in code generation, we use the parse
+ * listener interface to do most of our work.
+ */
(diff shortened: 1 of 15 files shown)


runtime/Java/src/org/antlr/v4/runtime/IncrementalParser.java
```diff
@@ -0,0 +1,175 @@
...
+ * handed to the guard rule of the next parse)
+ *
+ * @note See IncrementalParsing.md for more details on the theory behind this.
+ * In order to make this easier in code generation, we use the parse
+ * listener interface to do most of our work.
+ */
+public abstract class IncrementalParser extends Parser implements ParseTreeListener {
+	// Current parser epoch. Incremented every time a new incremental parser is
+	// created.
+	private static int _PARSER_EPOCH = 0;
```
Clashsoft: This should be an `AtomicInteger` to avoid race conditions when multiple threads instantiate this class.

PR #2535, opened 2019-04-25
MSVC build improvements

vvs31415:
Improve MSVC++ 2017 build


vvs31415:
If this PR is good, why is it just sitting here for more than five months?

PR #2551, opened 2019-05-09
use shared_ptr to avoid multi delete

### LINKED ISSUE #2550: [c++ runtime] reportNoViableAlternative()  causing BAD_ACCESS
CrashGrammar.g4:
```antlr
grammar CrashGrammar;

// combined from https://github.com/antlr/grammars-v4/blob/master/sqlite/SQLite.g4  parse + sql_stmt_list
parse: (';'* stmt ( ';'+ stmt )* ';'*)* EOF 
;

stmt: 'A' 'B'?
;

WS: [\u000B\t\r\n ]+  -> skip
;

ANY: .
;
```
main.cpp:
```cpp
#include <iostream>

#include "antlr4-runtime.h"
#include "CrashGrammarLexer.h"
#include "CrashGrammarParser.h"


using namespace antlr_sdl;

int main(int argc, char **argv) {
    std::stringstream ss(R"__(
        A B ;
        A _ ;
    )__");
    antlr4::ANTLRInputStream antlr_input(ss);
    CrashGrammarLexer lexer(&antlr_input);
    antlr4::CommonTokenStream token_stream(&lexer);
    CrashGrammarParser parser(&token_stream);
    antlr4::tree::ParseTree *ast = parser.parse();
    std::cout << ast->toStringTree(&parser) << std::endl;
    return 0;
}
```
Crashes on parser doing `parse()`.
Backtrace  shows the  `NoViableAltException` destructing on `reportNoViableAlternative` return, which is attempting to delete `_deadEndConfigs` that has been deleted

pnck:
fix  #2550 : use shared_ptr instead of raw pointer to avoid multiple delete

### DIFF
--- a/runtime/Cpp/runtime/src/NoViableAltException.cpp
+++ b/runtime/Cpp/runtime/src/NoViableAltException.cpp
@@ -22,7 +22,7 @@ NoViableAltException::NoViableAltException(Parser *recognizer, TokenStream *inpu
 
 NoViableAltException::~NoViableAltException() {
   if (_deleteConfigs)
-    delete _deadEndConfigs;
+    _deadEndConfigs = nullptr;
 }
 
 Token* NoViableAltException::getStartToken() const {
@@ -30,5 +30,5 @@ Token* NoViableAltException::getStartToken() const {
 }
 
 atn::ATNConfigSet* NoViableAltException::getDeadEndConfigs() const {
-  return _deadEndConfigs;
+  return _deadEndConfigs.get();
 }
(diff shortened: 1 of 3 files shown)


pnck:
not fixing.  still work in progress

runtime/Cpp/runtime/src/NoViableAltException.cpp
```diff
@@ -22,13 +22,13 @@ NoViableAltException::NoViableAltException(Parser *recognizer, TokenStream *inpu
 
 NoViableAltException::~NoViableAltException() {
   if (_deleteConfigs)
-    delete _deadEndConfigs;
+    _deadEndConfigs = nullptr;
```
chrisaycock: I think the assignment to `nullptr` triggers a `delete` regardless of the `use_count()`. If I just gut the destructor (make it an empty function), then your code works for me.

PR #2562, opened 2019-05-30
Fix #2550 

### LINKED ISSUE #2550: [c++ runtime] reportNoViableAlternative()  causing BAD_ACCESS

pnck:
fix #2550

I found that the `deadEndConfigs` is from a `unique_ptr` whose lifetime was handled by the `DFAState`.

when the `NoViableAltException` was thrown, the current dfa state makes no more sense, so i think it's ok to transfer the ownership of `ATNConfigSet` into `NoViableAltException` and let it deleted

### DIFF
--- a/contributors.txt
+++ b/contributors.txt
@@ -216,3 +216,4 @@ YYYY/MM/DD, github id, Full name, email
 2019/03/13, base698, Justin Thomas, justin.thomas1@gmail.com
 2019/03/18, carlodri, Carlo Dri, carlo.dri@gmail.com
 2019/05/02, askingalot, Andy Collins, askingalot@gmail.com
+2019/05/30, pnck, hio131@gmail.com
--- a/runtime/Cpp/runtime/src/NoViableAltException.cpp
+++ b/runtime/Cpp/runtime/src/NoViableAltException.cpp
@@ -11,20 +11,13 @@ using namespace antlr4;
 
 NoViableAltException::NoViableAltException(Parser *recognizer)
   : NoViableAltException(recognizer, recognizer->getTokenStream(), recognizer->getCurrentToken(),
-                         recognizer->getCurrentToken(), nullptr, recognizer->getContext(), false) {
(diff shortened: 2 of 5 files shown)


pnck:
😷 it's quite hard to deal with ATNConfigSet*

so many parameter is passed as the raw pointer

chrisaycock:
Unfortunately this failed on my [original problem](https://github.com/empirical-soft/empirical-lang/issues/2). After lots of trial and error, the [solution I came-up with](https://github.com/empirical-soft/empirical-lang/pull/19) has been to quite literally disable that `delete` statement. Ie., I just allow the memory to leak.

The leak won't be too bad for my cases, though I imagine other people will have trouble with it. I don't see an easy answer.

PR #2585, opened 2019-07-01
[C++]: Rethink about Any class

gfphoenix78:
Hi All,
Any is used to be a container for the parser to return a value to users. There are some things I concern:
* Is it necessary to have a copy constructor or copy assignment?
* Move semantics is more meaningful for this class.
* There should be a simple way to represent parent-child class relationship.
  * If I pass a derived class into `Any`, I should be able to get a pointer of the parent class
* Simple design, it need not be generic.


mike-lischke:
@parrt This PR can be closed. In the next iteration of the C++ runtime (sometimes next year)  I'll switch to C++17 and use the `std::any` class instead.

PR #2588, opened 2019-07-07
Make channelNames and modeNames final

Fokko:
Spotbugs is complaining that the `channelNames` and `modeNames` should be final.

### DIFF
--- a/tool/resources/org/antlr/v4/tool/templates/codegen/Java/Java.stg
+++ b/tool/resources/org/antlr/v4/tool/templates/codegen/Java/Java.stg
@@ -915,11 +915,11 @@ public class <lexer.name> extends <superClass; null="Lexer"> {
 	public static final int
 		<rest(lexer.modes):{m | <m>=<i>}; separator=", ", wrap, anchor>;
 	<endif>
-	public static String[] channelNames = {
+	public static final String[] channelNames = {
 		"DEFAULT_TOKEN_CHANNEL", "HIDDEN"<if (lexer.channels)>, <lexer.channels:{c| "<c>"}; separator=", ", wrap, anchor><endif>
 	};
 
-	public static String[] modeNames = {
+	public static final String[] modeNames = {
 		<lexer.modes:{m| "<m>"}; separator=", ", wrap, anchor>
 	};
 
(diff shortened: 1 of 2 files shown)


Fokko:
My personal Travis does run: https://travis-ci.org/Fokko/antlr4/builds/555506198

tool/resources/org/antlr/v4/tool/templates/codegen/Java/Java.stg
```diff
@@ -915,11 +915,11 @@ public class <lexer.name> extends <superClass; null="Lexer"> {
 	public static final int
 		<rest(lexer.modes):{m | <m>=<i>}; separator=", ", wrap, anchor>;
 	<endif>
-	public static String[] channelNames = {
+	public static final String[] channelNames = {
```
ibrahimAlii: Hi, What is the advantage of make `channelNames` to `final`  while I can still modify internal data i.e : `channelNames[0] = "ma me mo"`.
Should be there a better solution ?

Fokko: Thanks for the comment.

The thing with the `final` modifier is that you cannot change the memory address of the variable. So `Class.channelNames = new String[] {}` will give you an error on compile time. If we want to make the content immutable.

If we want to make the collection itself immutable, we need to change to `List<String>`: https://docs.oracle.com/en/java/javase/12/core/creating-immutable-lists-sets-and-maps.html
I'm fine with that as well, but then we're changing the public API. In other words, if some people are using the `channelNames` and `modeNames` in their application, then they need to update their application code as well. We can do this, but then we should wait for a major release.

Any opinion on this @parrt ?

parrt:
Hi. Given that we don't NEED to do this, I'm going to close but thanks for the PR!

Fokko:
This is sad. The code generated by Antlr triggers code quality rules in other projects, it is easy to ignore these, but it doesn't solve the actual problem. This PR addresses the root cause. I would say if you take software engineering serious, then we actually **need** this.

Fokko:
@parrt Anything I can do to change your mind on this?

parrt:
nope. sorry.

PR #2597, opened 2019-07-13
Various Travis configuration updates

ewanmellor:
Replace all the JDK versions used on Travis, since Travis and Oracle have EOL'd the previous versions.

    openjdk7 -> openjdk10
    openjdk8 -> openjdk11
    oraclejdk8 -> oraclejdk11

Change the Linux C# tests on Travis to use the current version of Mono (5.20.1.19 at the moment) instead of an ancient snapshot (3.12.1) that no longer exists.  This also means that the Mono build will match the Travis worker (running xenial) instead of being a version built for wheezy.

Change the Linux C++ tests on Travis to use Trusty.  This is what we've been using historically, but Travis switched the default Linux to Xenial in May 2019.  Put it back to Trusty, because the build currently doesn't work on Xenial at the moment :-(

Remove .travis/before-install-linux-{cpp,java,python2,python3}.sh, and remove some lines from before-install-linux-{javascript,go}.sh.  These contained apt commands for configuring the Mono project's GPG key, which is not used in any of these builds, and pointless apt-get update calls.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -20,7 +20,7 @@ matrix:
   include:
     - os: linux
       compiler: clang
-      jdk: openjdk7
+      jdk: openjdk10
       env:
         - TARGET=cpp
         - CXX=g++-5
@@ -37,7 +37,7 @@ matrix:
             - clang-3.7
     - os: linux
       compiler: clang
-      jdk: openjdk7
+      jdk: openjdk10
       env:
         - TARGET=cpp
         - CXX=g++-5
@@ -54,7 +54,7 @@ matrix:
             - clang-3.7
     - os: linux
       compiler: clang
-      jdk: openjdk7
+      jdk: openjdk10
       env:
         - TARGET=cpp
         - CXX=g++-5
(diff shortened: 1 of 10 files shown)


ericvergnaud:
Hi,
as we speak, we guarantee java7 compatibility.
Just because it's EOL'd by Oracle only means you have to pay for support, not that people have stopped using it.
I don't have a problem using jdk11 to build, but not comfortable dropping Java7 validation altogether.

ewanmellor:
@ericvergnaud Right, but support for our current matrix has been withdrawn by Travis, so it simply doesn't work any more (look at all the recent errored builds).  We have to change something.

We should be reasonable about what we do on a public free service like Travis, so we can't expand the matrix indefinitely.  What we're doing at the moment is testing JDK 7 but not testing JDK 10 or JDK 11, and that is not sensible.

Don't forget, this test matrix is only what applies to the Antlr bleeding-edge development.  If someone is willing to run on an EOL JDK on an EOL operating system, then they can use a prior release of Antlr too.  It's not reasonable to expect a volunteer project like this to have a wider support window than Oracle.

ericvergnaud:
Mmmm… the test matrix is what applies to the release process, which includes bug fixes.
We release on green builds, and there is no other build.
We cannot claim that we support JDK7 if we no longer test it.
We cannot drop it without checking with our user base.
Suggest we drop oracle JDK altogether, keep openJDK7 and openJDK11, seems reasonable to expect that if it works for these it also works for intermediate versions.

ewanmellor:
Or we drop official JDK 7 support, and tell anyone who needs it to use Antlr 4.7.2.  It is 2011-vintage software.  It is unreasonable for us to support it when no-one else is doing so.

ericvergnaud:
Not sure what 'us' means here. The Java target is supported by Terence only.
It would have to be his decision. Don't underestimate the inertia in certain sectors.
At my place (a global bank), nobody is using Java 9 yet.
@parrt what do you think?

ewanmellor:
This “global bank” is welcome to come and dedicate some resources to this project. As it stands, no tests have passed for two months and no one has done anything about it.

Ewan.

On Jul 13, 2019, at 19:36, ericvergnaud <notifications@github.com<mailto:notifications@github.com>> wrote:


Not sure what 'us' means here. The Java target is supported by Terence only.
It would have to be his decision. Don't underestimate the inertia in certain sectors.
At my place (a global bank), nobody is using Java 9 yet.
@parrt<https://github.com/parrt> what do you think?

ericvergnaud:
'global personal time' has submitted PR #2598 which adresses your concern.
This buys us time to discuss jdk7 support.
I'm all for adding jdk11.
I am vehemently opposed to dropping jdk8.

parrt:
I agree that tests should pass! :)  No point in forcing latest version of java unless we change the software in a way that needs > 7. I'm still using 8 I think :)  I'll jump over to @ericvergnaud 's PR now.

ewanmellor:
Closing.  Superceded by #2598 and #2599.

PR #2598, opened 2019-07-14
Use openjdk instead of oraclejdk

ericvergnaud:


ericvergnaud:
@parrt this fix addresses 2 urgent Travis CI issues:
 - when running on Travis, oracle won't let you use their jdk7 or jdk8 anymore
   (you can see: "Expected feature release number in range of 9 to 14, but got: 8" in the logs)
 - there was an 'open' dependency on Jetty, which ended up recently switching to Jetty 10 which does not run on jdk8
It also generalises use of jdk8 for all builds except the backwards compatible Java for jdk7 one

ericvergnaud:
@mike-lischke can you check why the cpp build fails? It seems to release to code generation, not to the environment

ericvergnaud:
> @mike-lischke can you check why the cpp build fails? It seems to relate to code generation, not to the environment

@mike-lischke it works when using the trusty travis-ci environment.
I find it a bit scary though that the generated code would compile on trusty but not on the latest travis env...

mike-lischke:
Hard to say what's wrong, since I don't see the previous compiler errors.

ericvergnaud:
> Hard to say what's wrong, since I don't see the previous compiler errors.

check this one: https://travis-ci.org/antlr/antlr4/jobs/558417739

mike-lischke:
Hmm, I see, a linker error. I cannot fix right now, however, as I'm on vacation and don't have my dev environment with me (just the spare laptop). I'll look after it when I return.

parrt:
This looks good to me. should we wait until @mike-lischke  his back from vacation or push this one and fix C++ later? @ericvergnaud @ewanmellor

ericvergnaud:
No it's all good when using trusty, so suggest we move forward.
Many thanks to @ewanmellor for his great work which I leveraged.
And let's engage a jdk support survey?

parrt:
Do you mean something like https://github.com/surveyproject/surveyproject @ericvergnaud ?

Ok, merging.  thanks to you and @ewanmellor !

ericvergnaud:
@parrt yes something like that
I tried to post a survey on the discussion group, but it got blocked, not sure why.

PR #2604, opened 2019-07-21
[Swift] Check process exit status during tests.

ewanmellor:
Check the exit status of the processes run during the runtime-testsuite
for Swift.  This means that we'll catch low-level issues earlier.

This changeset also changes the exception handling so that exceptions
throw as far up as possible so that we don't proceed with a test when
something has already broken.


ewanmellor:
Rebased to current master.

Issue #2274, opened 2018-04-11
Cancellation support in C# runtime


KvanTTT:
See duplicated issue in C# optimized runtime: [https://github.com/tunnelvisionlabs/antlr4cs/issues/277](https://github.com/tunnelvisionlabs/antlr4cs/issues/277).

ericvergnaud:
Can you elaborate?

KvanTTT:
Maybe later. I'm waiting for @sharwell response.

PR #2610, opened 2019-07-27
[Java] Check for thread interrupt status in potentially exponential codepaths

ryanpbrewster:
## Interrupting runaway parsers
Some grammars take an exponential amount of time to parse (relative to the input size).

One way to safeguard a program from these kinds of inputs is to run the parser in a separate thread and interrupt it if it's not finished after a reasonable amount of time. In Java, computations are expected to occasionally check if the thread they're running on has been interrupted. This PR adds such a check in one particularly important codepath.

Here is an example grammar that can trigger exponential runtime on inputs like `((()))`:
```
grammar Foo;
foo
    : '()'
    | '(' foo ')'
    | '(' foo ')' { notifyErrorListeners("Extra closing ')'"); } ')'
    ;
```

Here is an example of a program that defensively runs the parser in a separate thread and interrupts it after 5 seconds. I have checked that the current behavior does not terminate, and with the changes in this PR it does.

```
  public static void main(String[] args) throws InterruptedException {
    String input = args[0];
    Thread t = new Thread(() -> {
      FooLexer lexer = new FooLexer(CharStreams.fromString(input));
      FooParser parser = new FooParser(new CommonTokenStream(lexer));
      System.out.println(parser.foo());
    });
    t.start();
    Thread.sleep(5000);
    t.interrupt();
    Thread.sleep(60000);
  }
```

## What is the parser doing that takes so long
```
	at org.antlr.v4.runtime.atn.ArrayPredictionContext.equals(ArrayPredictionContext.java:82)
	at java.base/java.util.Objects.equals(Objects.java:77)
	at java.base/java.util.Arrays.equals(Arrays.java:3186)
	at org.antlr.v4.runtime.atn.ArrayPredictionContext.equals(ArrayPredictionContext.java:82)
	at java.base/java.util.Objects.equals(Objects.java:77)
	at java.base/java.util.Arrays.equals(Arrays.java:3186)
	at org.antlr.v4.runtime.atn.ArrayPredictionContext.equals(ArrayPredictionContext.java:82)
	at java.base/java.util.Objects.equals(Objects.java:77)
	at java.base/java.util.Arrays.equals(Arrays.java:3186)
	at org.antlr.v4.runtime.atn.ArrayPredictionContext.equals(ArrayPredictionContext.java:82)
	at java.base/java.util.HashMap.getNode(HashMap.java:572)
	at java.base/java.util.HashMap.get(HashMap.java:557)
	at org.antlr.v4.runtime.atn.PredictionContextCache.get(PredictionContextCache.java:36)
	at org.antlr.v4.runtime.atn.PredictionContext.getCachedContext(PredictionContext.java:556)
	at org.antlr.v4.runtime.atn.ATNSimulator.getCachedContext(ATNSimulator.java:103)
	at org.antlr.v4.runtime.atn.ATNConfigSet.optimizeConfigs(ATNConfigSet.java:217)
	at org.antlr.v4.runtime.atn.ParserATNSimulator.addDFAState(ParserATNSimulator.java:2124)
	at org.antlr.v4.runtime.atn.ParserATNSimulator.addDFAEdge(ParserATNSimulator.java:2078)
	at org.antlr.v4.runtime.atn.ParserATNSimulator.computeTargetState(ParserATNSimulator.java:619)
	at org.antlr.v4.runtime.atn.ParserATNSimulator.execATN(ParserATNSimulator.java:454)
	at org.antlr.v4.runtime.atn.ParserATNSimulator.adaptivePredict(ParserATNSimulator.java:393)
```

### DIFF
--- a/contributors.txt
+++ b/contributors.txt
@@ -221,3 +221,4 @@ YYYY/MM/DD, github id, Full name, email
 2019/07/11, olowo726, Olof Wolgast, olof@baah.se
 2019/07/16, abhijithneilabraham, Abhijith Neil Abraham, abhijithneilabrahampk@gmail.com
 2019/07/26, Braavos96, Eric Hettiaratchi, erichettiaratchi@gmail.com
+2019/07/26, ryanpbrewster, Ryan Brewster, ryanpbrewster@google.com
--- a/runtime/Java/src/org/antlr/v4/runtime/atn/ArrayPredictionContext.java
+++ b/runtime/Java/src/org/antlr/v4/runtime/atn/ArrayPredictionContext.java
@@ -7,6 +7,7 @@
 package org.antlr.v4.runtime.atn;
 
 import java.util.Arrays;
+import org.antlr.v4.runtime.misc.ParseCancellationException;
 
 public class ArrayPredictionContext extends PredictionContext {
 	/** Parent can be null only if full ctx mode and we make an array
@@ -62,6 +63,9 @@ public class ArrayPredictionContext extends PredictionContext {
 
 	@Override
 	public boolean equals(Object o) {
+		if (Thread.interrupted()) {
+			throw new ParseCancellationException(new InterruptedException());
+		}
 		if (this == o) {
 			return true;
 		}


sharwell:
> Some grammars take an exponential amount of time to parse (relative to the input size).

This should not be possible. The algorithm is O(n<sup>4</sup>) in the worst case.

sharwell:
@ryanpbrewster Can you try the same test using my fork (groupId **com.tunnelvisionlabs** in Maven Central instead of **org.antlr**). It more aggressively ensures that two equal `PredictionContext` objects are also reference-equal, which avoids the recursive `equals` evaluation seen in the stack trace above.

ryanpbrewster:
Thanks for the pointers. I'll test with the fork and report back.

ryanpbrewster:
Experimental runtimes:
```
1 --- 1
2 --- 4
3 --- 1
4 --- 2
5 --- 4
6 --- 6
7 --- 8
8 --- 10
9 --- 15
10 --- 21
11 --- 26
12 --- 32
13 --- 28
14 --- 30
15 --- 47
16 --- 58
17 --- 41
18 --- 38
19 --- 43
20 --- 58
21 --- 69
22 --- 129
23 --- 232
24 --- 458
25 --- 948
26 --- 1847
27 --- 3648
28 --- 7190
29 --- 14416
30 --- 28750
```

ryanpbrewster:
I'm running into errors using the `com.tunnelvisionlabs` artifact for the Antlr runtime
```
java.lang.NoClassDefFoundError: org/antlr/v4/runtime/misc/Pair
```
Trying to simplify the build process.

ryanpbrewster:
Got it working, it's much faster.
`com.tunnelvisionlabs` results:
```
10 --- 19
20 --- 49
30 --- 73
40 --- 57
50 --- 45
60 --- 65
70 --- 77
80 --- 94
90 --- 121
100 --- 157
110 --- 193
120 --- 263
130 --- 306
140 --- 356
150 --- 439
160 --- 524
170 --- 598
180 --- 709
190 --- 819
200 --- 934
210 --- 1096
220 --- 1244
230 --- 1422
240 --- 1569
250 --- 1755
260 --- 1979
270 --- 2176
280 --- 2441
290 --- 2769
300 --- 2991
```

ryanpbrewster:
Unfortunately your suggestion in https://github.com/antlr/antlr4/pull/2610#pullrequestreview-269671655 is not quite deep enough in the codepath to address the pathological case described in my original comment. The runtime is spending all of its time churning inside of `ArrayPredictionContext`, so it does not encounter the `computeTargetState` method during the time-consuming part of the computation.

KvanTTT:
Is this issue related to https://github.com/antlr/antlr4/issues/2274?

ryanpbrewster:
It seems like it. I imagine that the C# runtime could use a similar approach.

ryanpbrewster:
@sharwell any idea how much work it would be to port the reference-based equality from `com.tunnelvisionlabs` to `org.antlr`? The performance improvement seems potentially huge.

danielsun1106:
FYI. After comparison, the highly optimized fork of @sharwell is much more efficient than the official reference implementation, so the new Parrot parser of Apache Groovy(https://github.com/apache/groovy) is based on https://github.com/tunnelvisionlabs/antlr4

ericvergnaud:
One strong requirement is to keep the same algo for all language targets, such that it’s easy to spread bug fixes and minor enhancements.

ryanpbrewster:
Agreed, the fork seems significantly more performant. After looking through the README, I still believe that the mainline branch of Antlr should respect thread interrupts, regardless of its performance.

ryanpbrewster:
@sharwell, thoughts on merging this PR as it stands? Are there benchmarks or tests you'd like to see?

ryanpbrewster:
ping

sharwell:
I'm hesitant to add anything to this intensely-hot path. I'll need to think about alternatives that may work.

ryanpbrewster:
It seems unlikely that this approach will be merged. We are working around this issue in a more brute-force way.

Issue #2606, opened 2019-07-26
[C++][CMake] find_package() fails when the C++ runtime is make-install'ed


delins:
I followed [README.md](https://github.com/antlr/antlr4/blob/master/runtime/Cpp/README.md) to compile and install antlr4 on Linux. I turned `ANTLR4_INSTALL` on in the `cmake` step and I left out the `DESTDIR` variable in the `make install` step. This results in the following directory with CMake files:
```
$ ls -l /usr/local/lib/cmake/antlr4
total 24
-rw-r--r-- 1 root root 1607 jul 25 20:27 antlr4-config-version.cmake
-rw-r--r-- 1 root root 7632 jul 25 20:27 antlr4-generator-config.cmake
-rw-r--r-- 1 root root 1138 jul 25 20:27 antlr4-runtime-config.cmake
-rw-r--r-- 1 root root 3235 jul 25 20:27 antlr4-targets.cmake
-rw-r--r-- 1 root root 1427 jul 25 20:27 antlr4-targets-release.cmake
```

I then followed [Antlr4Package.md](https://github.com/antlr/antlr4/blob/master/runtime/Cpp/cmake/Antlr4Package.md) to add antlr4 to my project. However, `find_package(antlr4-runtime REQUIRED)` fails: CMake can't find `antlr4-runtime`. This makes sense, given CMake's [Search procedure](https://cmake.org/cmake/help/latest/command/find_package.html#search-procedure)

One solution is to follow CMake's advice in the error message:
```
Add the installation prefix of "antlr4-runtime" to CMAKE_PREFIX_PATH or set
  "antlr4-runtime_DIR" to a directory containing one of the above files.
```

But I'd say it's cleaner to give the `runtime` and `generator` packages each their own directory (or perhaps there's another way that I'm unaware of). Otherwise, could a warning about this issue be added to  [Antlr4Package.md](https://github.com/antlr/antlr4/blob/master/runtime/Cpp/cmake/Antlr4Package.md)? It's not much of a problem if you're a seasoned CMake user, but if you barely use it it's quite a hassle to debug (i.e. actually finding the root cause and not simply following CMake's advice).

mike-lischke:
Yes, the given suggestion is probably the best solution here and I will add a note to the Antlr4Package.md file.

Nic30:
I am suffering from similar issue.

[...]

PR #2631, opened 2019-08-21
C++ code cleanup

mike-lischke:
- Converted back all TO_DO items to TODO (they were renamed ANTLR4 TODO items, but I had some for C++ which I all wanted to fix).
- Removed some default copy assignment operators which clang warned about.
- Added cmake build informations as reported in #2606


mike-lischke:
@parrt This is just a clean up for the C++ target. No behavior change has been introduced, so I think it's pretty safe to merge this PR, thanks.

PR #2635, opened 2019-08-27
Fix escaping of angle brackets

wurzelpeter:
Only minor fix to the markdown documentation


wurzelpeter:
I don't have a clue why those changes break the csharp CI build. Any comments on this?

PR #2638, opened 2019-09-03
General XPath fixes for the Python3 runtime

JohnnyonFlame:
This commit:
- Fixes some code that does not reflect the current API
- Fixes wrong XPath library behavior, such as failing to apply the "invert" flag on certain operations
- Does general style touch-ups

Further work is necessary, such as implementing tests for this functionality, so that further changes to the API or Runtime won't break functionality, and maybe separate the interpreter data from the library code. For now, a code review would be stellar.

### DIFF
--- a/runtime/Python3/src/antlr4/xpath/XPath.py
+++ b/runtime/Python3/src/antlr4/xpath/XPath.py
@@ -47,7 +47,7 @@
 # <p>
 # Whitespace is not allowed.</p>
 #
-from antlr4 import CommonTokenStream, DFA, PredictionContextCache, Lexer, LexerATNSimulator
+from antlr4 import CommonTokenStream, DFA, PredictionContextCache, Lexer, LexerATNSimulator, ParserRuleContext, TerminalNode
 from antlr4.InputStream import InputStream
 from antlr4.Parser import Parser
 from antlr4.RuleContext import RuleContext
@@ -134,7 +134,7 @@ class XPathLexer(Lexer):
         if _action is not None:
             _action(localctx, actionIndex)
         else:
-            raise Exception("No registered action for:" + str(ruleIndex))
+            raise Exception("No registered action for: %d" % ruleIndex)
 
     def ID_action(self, localctx:RuleContext , actionIndex:int):
         if actionIndex == 0:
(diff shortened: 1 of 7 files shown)


runtime/Python3/src/antlr4/xpath/XPath.py
```diff
@@ -210,24 +210,32 @@ def recover(self, e):
...
+                ttype = (tsource.ruleNames.index(word) + 1) if word in tsource.ruleNames    else None
+            else:
+                litType = tsource.literalNames.index(word)  if word in tsource.literalNames else None
+
+            # Decide which one is it
+            ttype = ttype or litType
 
-            if ttype==Token.INVALID_TYPE:
-                raise Exception( word + " at index " + str(wordToken.startIndex) + " isn't a valid token name")
+            if not ttype:
```
ericvergnaud: mmm... this is javascript style not python, please keep the intent explicit

runtime/Python3/src/antlr4/xpath/XPath.py
```diff
@@ -246,18 +254,16 @@ def evaluate(self, t:ParseTree):
...
 
         work = [dummyRoot]
-
-        for i in range(0, len(self.elements)):
-            next = set()
+        for element in self.elements:
+            work_next = list()
             for node in work:
-                if len( node.children) > 0 :
+                if not isinstance(node, TerminalNode) and node.children:
```
ericvergnaud: That is probably wrong,g, not sure there is guarantee that node.children is never empty

runtime/Python3/src/antlr4/xpath/XPath.py
```diff
@@ -210,24 +210,32 @@ def recover(self, e):
...
+            tsource = self.parser.getTokenStream().tokenSource
+
+            ttype = litType = None
+            if wordToken.type == XPathLexer.TOKEN_REF:
+                ttype = (tsource.ruleNames.index(word) + 1) if word in tsource.ruleNames    else None
+            else:
+                litType = tsource.literalNames.index(word)  if word in tsource.literalNames else None
+
+            # Decide which one is it
+            ttype = ttype or litType
```
ericvergnaud: mmm... this is javascript style not python, please keep the intent explicit
also in Python 1 variable is probably sufficient

runtime/Python3/src/antlr4/xpath/XPath.py
```diff
@@ -210,24 +210,32 @@ def recover(self, e):
...
+            if not ttype:
+                raise Exception("%s at index %d isn't a valid token name" % (word, wordToken.tokenIndex))
             return XPathTokenAnywhereElement(word, ttype) if anywhere else XPathTokenElement(word, ttype)
 
         else:
+            ruleIndex = self.parser.ruleNames.index(word) if word in self.parser.ruleNames else None
 
-            if ruleIndex==-1:
-                raise Exception( word + " at index " + str(wordToken.getStartIndex()) + " isn't a valid rule name")
+            if not ruleIndex:
```
ericvergnaud: same here

ericvergnaud:
Hi,
thanks for this
Any chance you could provide some tests that can help validate this PR?

runtime/Python3/src/antlr4/xpath/XPath.py
JohnnyonFlame: https://www.python.org/dev/peps/pep-0008/
>For sequences, (strings, lists, tuples), use the fact that empty sequences are false.
>
>```Python3
>Yes: if not seq:
>     if seq:
>
>No:  if len(seq):
>     if not len(seq):
>```

JohnnyonFlame:
@ericvergnaud 
First, thank you for the review.
I am going to port some of the test of the Java runtime when possible. I was first waiting for a code review while taking a look at some other problems.

And yeah, the implicit intent is a bad habit, _oops_.

JohnnyonFlame:
@ericvergnaud Implemented some of the recommended changes, corrected some bugs and included a port of _testValidPaths_ from "_tool-testsuite/test/org/antlr/v4/test/tool/TestXPath.java_"

runtime/Python3/src/antlr4/xpath/XPath.py
```diff
@@ -210,24 +210,31 @@ def recover(self, e):
...
 
+        word = wordToken.text
         if wordToken.type==XPathLexer.WILDCARD :
-
             return XPathWildcardAnywhereElement() if anywhere else XPathWildcardElement()
 
         elif wordToken.type in [XPathLexer.TOKEN_REF, XPathLexer.STRING]:
+            tsource = self.parser.getTokenStream().tokenSource
+
+            ttype = -1
```
ericvergnaud: I would set type to Token.INVALID_TYPE

runtime/Python3/src/antlr4/xpath/XPath.py
```diff
@@ -210,24 +210,31 @@ def recover(self, e):
...
+            if wordToken.type == XPathLexer.TOKEN_REF:
+                if word in tsource.ruleNames:
+                    ttype = tsource.ruleNames.index(word) + 1
+            else:
+                if word in tsource.literalNames:
+                    ttype = tsource.literalNames.index(word)
 
-            if ttype==Token.INVALID_TYPE:
-                raise Exception( word + " at index " + str(wordToken.startIndex) + " isn't a valid token name")
+            if ttype == -1:
```
ericvergnaud: such that you can test against it
not sure tsource... call may or not return it

runtime/Python3/src/antlr4/xpath/XPath.py
ericvergnaud: Not sure I support that pep...

ericvergnaud:
Thanks for this
Apart from 2 minor style comments it looks good!

runtime/Python3/src/antlr4/xpath/XPath.py
JohnnyonFlame: Passes the tests it seems, not too sure what would be a better choice.

JohnnyonFlame:
Thank you. Included the proposed style changes.

ericvergnaud:
Cool!
@parrt blessed

PR #2642, opened 2019-09-07
lexer declarations and constructor extension

sirkon:
Having `@lexer::members` support for Go target is not enough. Just like C++, Go needs `@lexer::declarations` to add needed fields into a lexer and I also added `@lexer::construction` block (I don't know how you call it in ANTLR4) to extend `New<Gramman Name>Lexer` constructor code.

This is how I used this:

```
...

@lexer::declarations {
    comments map[int]map[int]string // This adds a new field "comments"
}

@lexer::construction {
    l.comments = map[int]map[int]string{} // This inits "comments" field in the constructor
}

...
```

### DIFF
--- a/tool/resources/org/antlr/v4/tool/templates/codegen/Go/Go.stg
+++ b/tool/resources/org/antlr/v4/tool/templates/codegen/Go/Go.stg
@@ -1564,6 +1564,10 @@ func New<lexer.name>(input antlr.CharStream) *<lexer.name> {
 	l.LiteralNames = staticData.LiteralNames
 	l.SymbolicNames = staticData.SymbolicNames
 	l.GrammarFileName = "<lexer.grammarFileName>"
+	<if(namedActions.structconstruction)>
+	// Grammar author supplied struct members construction code.
+	<namedActions.structconstruction>
+	<endif>
 	// TODO: l.EOF = antlr.TokenEOF
 
 	return l


historical-contributors-agreement.txt
```diff
@@ -234,6 +234,7 @@ YYYY/MM/DD, github id, Full name, email
 2019/08/27, wurzelpeter, Markus Franke, markus[hyphen]franke[at]web[dot]de
 2019/09/10, ImanHosseini, Iman Hosseini, hosseini.iman@yahoo.com
 2019/09/03, João Henrique, johnnyonflame@hotmail.com
+<<<<<<< HEAD
```
KvanTTT, line 237: This line looks like leftover of conflict resolving. Please remove.

historical-contributors-agreement.txt
```diff
@@ -328,3 +329,4 @@ YYYY/MM/DD, github id, Full name, email
 2021/12/31, Biswa96, Biswapriyo Nath, nathbappai@gmail.com
 2022/03/07, chenquan, chenquan, chenquan.dev@gmail.com
 2022/03/15, hzeller, Henner Zeller, h.zeller@acm.org
+2023/04/29, sirkon, Denis Cheremisov, denis.cheremisov@gmail.com
```
KvanTTT, line 332: Now there is no need to change this file. Please revert.

KvanTTT (changes requested):

historical-contributors-agreement.txt
sirkon: Oops, removed. Removed the last line too.

KvanTTT (changes requested):
Please rebase on the latest master and fix conflicts if it's still actual.

sirkon:
`structmembers` thing was added doing everything my `decalarations` did, so I've removed mine.

But `construction` equivalent is missing in the upstream. So, this PR can still be useful.

Also renamed it into `structconstruction` to be inline with the existing code. Added if check is not NULL as well.

PR #2644, opened 2019-09-10
Fix data race in LexerATNSimulator

neko1235:
There is a potential memory consistency problem.
https://docs.oracle.com/javase/tutorial/essential/concurrency/memconsist.html


ericvergnaud:
Hi
Not sure volatile is more than a hint for the jvm.
Have you tried ThreadLocal?

neko1235:
> Hi
> Not sure volatile is more than a hint for the jvm.
> Have you tried ThreadLocal?

"volatile" is more than just a hint for the jvm. It introduces the happens-before relationship, meaning that a write to a volatile field happens before every subsequent read of that same field.

ThreadLocal would fix the data race, however the counter will be no longer global.

neko1235:
The `match_calls` counter is not used anywhere outside of the class. However, it's a public field in a public class. If you don't need to maintain backward compatibility then it's probably better to remove it.

If the counter is still needed then an AtomicInteger is probably better.

ericvergnaud:
Yes an AtomicInteger would be much better than the proposed synchronized block
My guess is that this very unsafe var is only used for algorithm tuning

neko1235:
> Yes an AtomicInteger would be much better than the proposed synchronized block
> My guess is that this very unsafe var is only used for algorithm tuning

Making it an AtomicInteger will break backward compatibility anyway.
If it's only used for algorithm tuning, then it's better to remove it or comment it out.

runtime/Java/src/org/antlr/v4/runtime/atn/LexerATNSimulator.java
```diff
@@ -78,8 +78,6 @@ protected void reset() {
 
 	protected final SimState prevAccept = new SimState();
 
-	public static int match_calls = 0;
```
sharwell: 📝 If there is a concern about binary compatibility, we can keep this field (mark it `@Deprecated`) and just stop writing to it. However, I don't think binary compatibility for this field is a concern. I'll let @parrt make the call on it.

parrt: Always scary but I think it's cool.

Issue #2523, opened 2019-03-26
JavaScript Target: Poor Performance when parsing large expressions


jackandjohn:
I'm using ANTLR-4.7.2 with a grammar file which support nesting expressions recursively, and find large nested expressions cost too much time to parse. It seems like the same problem https://github.com/antlr/antlr4/issues/2468 .

Stack Trace:
![image](https://user-images.githubusercontent.com/10008741/55000053-9cae2980-500c-11e9-8de7-0e4fe8dcaa44.png)

ericvergnaud:
As stated in #2468, the only known possible improvement is with your grammar. Please move this to the google discussion group.

KvanTTT:
@ericvergnaud also it's can be closed.

Issue #2468, opened 2019-01-08
Poor performance of Javascript parser with recursive patterns


corvana-nbaker:
We have a pretty simple grammar similar to javascript which supports nesting expressions recursively. In general the javascript parser is performing adequately. However, when large nested expressions are encountered our parse time can balloon to 15 seconds or more. e.g. FOO(BAR(BAZ("something"))). 

I have a trivial example demonstrating this problem here: https://github.com/corvana-nbaker/antlr4-example

I'm hoping someone will identify an issue with our grammar. If not maybe this can help optimize the javascript runtime. Profiling does show that much of the processing time is spent in .toString() calls as reported in https://github.com/antlr/antlr4/issues/2360

KvanTTT:
Have you tried to replace the fragment 

```antlr
| CASE '(' expr ')'          #FunctionExpr
| CASE '(' expr ',' expr ')'          #FunctionExpr
| CASE '(' expr ',' expr ',' expr ')'          #FunctionExpr
| CASE '(' expr ',' expr ',' expr ',' expr ')'          #FunctionExpr
| CASE '(' expr ',' expr ',' expr ',' expr ',' expr ')'          #FunctionExpr
| CASE '(' expr ',' expr ',' expr ',' expr ',' expr ',' expr ')'          #FunctionExpr
| CASE '(' expr ',' expr ',' expr ',' expr ',' expr ',' expr ',' expr ')'          #FunctionExpr
| CASE '(' expr ',' expr ',' expr ',' expr ',' expr ',' expr ',' expr ',' expr ')'          #FunctionExpr
| CASE '(' expr ',' expr ',' expr ',' expr ',' expr ',' expr ',' expr ',' expr ',' expr ')'          #FunctionExpr
| CASE '(' expr ',' expr ',' expr ',' expr ',' expr ',' expr ',' expr ',' expr ',' expr ',' expr ')'          #FunctionExpr
```

with the following one:

```antlr
| CASE '(' expr (',' expr (',' expr (',' expr (',' expr (',' expr (',' expr (',' expr (',' expr (',' expr)?)?)?)?)?)?)?)?)? ')'          #FunctionExpr
```

Moreover, you should not escape `[` char inside `[]` interval here: https://github.com/corvana-nbaker/antlr4-example/blob/96626e8bb7c4f8b152caa25785c69292f3fb165d/antlr/Query.g4#L86

corvana-nbaker:
[...]

PR #2657, opened 2019-09-28
Fix javascript parser cache performance issue

lmy269:
Javascript parser is leveraging {} to cache the contexts. Each time when calling Get,Set, the key is converted to string first. In some cases (nested functions), the performance is very bad due to this. When it deals with simple string like "abs(abs(abs(abs(abs(abs(abs(abs(abs(1)))))))))", it takes minutes to finish. The time can easily reach to hours, as time complexity looks like to be O(2^N). 
This change is to use Map to store the real object in the key to avoid that. After this change, the time is decreased to less than 1 second to handle that string.
I believe #2468 and #2523 will also be addressed by this PR.

This's a call stack when parsing nested functions. And I saw there were huge minor GCs caused by the toString().
callstack

### DIFF
--- a/contributors.txt
+++ b/contributors.txt
@@ -228,3 +228,4 @@ YYYY/MM/DD, github id, Full name, email
 2019/09/10, marcospassos, Marcos Passos, marcospassos.com@gmail.com
 2019/09/10, amorimjuliana, Juliana Amorim, juu.amorim@gmail.com
 2019/09/17, kaz, Kazuki Sawada, kazuki@6715.jp
+2019/09/28, lmy269, Mingyang Liu, lmy040758@gmail.com
--- a/runtime/JavaScript/src/antlr4/PredictionContext.js
+++ b/runtime/JavaScript/src/antlr4/PredictionContext.js
@@ -660,13 +660,13 @@ function getCachedPredictionContext(context, contextCache, visited) {
 	if (context.isEmpty()) {
 		return context;
 	}
-	var existing = visited[context.hashCode()] || null;
+	var existing = visited.get(context) || null;
 	if (existing !== null) {
 		return existing;
 	}
 	existing = contextCache.get(context);
 	if (existing !== null) {
-		visited[context.hashCode()] = existing;
+		visited.put(context, existing);
 		return existing;
 	}
 	var changed = false;
@@ -686,7 +686,7 @@ function getCachedPredictionContext(context, contextCache, visited) {
 	}
 	if (!changed) {
 		contextCache.add(context);
-		visited[context.hashCode()] = context;
+		visited.put(context, context);
 		return context;
 	}
 	var updated = null;
@@ -699,8 +699,8 @@ function getCachedPredictionContext(context, contextCache, visited) {
 		updated = new ArrayPredictionContext(parents, context.returnStates);
 	}
 	contextCache.add(updated);
-	visited[updated.hashCode()] = updated;
-	visited[context.hashCode()] = updated;
+	visited.put(updated, updated);
+	visited.put(context, updated);
 
 	return updated;
 }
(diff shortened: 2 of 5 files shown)


ericvergnaud:
Hi,

thanks for this, notably the stack trace.

this PR aggregates changes which seem related but actually address different concerns.

1) Using the standard Map instead of the in-house one only introduces browser compatibility risk since they both follow the same semantics. I'll give you that this would much less a problem today than it was 6 years ago when this was implemented. But I'd be interested in numbers that demonstrate a performance improvement from that particular change. Only then can we decide if it's worth sacrificing old browsers.

2) Using the object itself as a key, rather than an explicit key. The purpose of the cache is NOT just to locate a given object by key, rather re-use the existing one. 

Unfortunately, in JavaScript the following expression:
{ stuff: "a" } === { stuff: "a" } 
returns false. 

As a result, your proposed change is guaranteed to end up in an OOM when reusing the parser multiple times.

Thanks for identifying a remaining toString in map key computation.

Would you be willing to try replace these by equivalent calls to Murmur3 in a dedicated PR?

lmy269:
Hi Eric,
Thanks for your comments!
The performance is significant for our user case. To parse a query like below, the time come down to 400ms from 220s. I'll share the perf screen at the bottom.

----------------Query Start-----------------------------
**if**(Column_1=='test0','result0',
**if**(in(['test1','test2','test3'],Column_1),'result1',
**if**(in(['test4','test5'],Column_1),'result2',
**if**(in(['test6','test7','test8'],Column_1),'result3',
**if**(in(['test9','test10','test11'],Column_1),'result4',
**if**(Column_1 =='test12','result5',
**if**(Column_1 =='test13','result6',
**if**(in(['test14','test15'],Column_2),'result7',
**if**(Column_2 == 'test16','result8',
**if**(Column_1=='test17','result9',
'result10'
))))))))))
----------------Query End-----------------------------

For your second concern, I agree that cache is used to re-use the existing values. And it depends on how the key generated. It's true that { stuff: "a" } === { stuff: "a" } returns false in JS. But if the objects are pointing to the same reference, then key1 === key2 can return true. The problem is not just in JS, but also the other languages. (key1 can be not equal to key2 even if they have the same value inside)

In JS, we use "new SingletonPredictionContext()" to create the contexts, which can be used in context key later on. And so are other languages like C#, C++. They all use reference to do the lookup, not the serialized value. Why JS need to do things special?

----------------Perf comparison--------------------------
From the screenshots, we can see the calls are quite balanced in the second one. It's even impossible for me to find which time range is used to process that particular query, as it's too short.
1. Before:
perf1
2. After:
perf2

lmy269:
@ericvergnaud , updated with hash code, the performance looks even better than the old way. Let me know if it's right way to use Murmur3. Thanks!

ericvergnaud:
I'm afraid the way you do it there is a risk of collision.
The hash code is not the key. 
Suggest you leverage the provided Set for caching in PredictionContextCache.
Maybe also for other changes, but would need time to look into the details.

lmy269:
@ericvergnaud Just realized I only need to use the Map you created to do the cache, which handles the hashcode and collision already.

ericvergnaud:
Does that address your performance issue?

lmy269:
Yes, I tested, it works well, no toString() calls found.

ericvergnaud:
@parrt blessed

parrt:
Thanks @lmy269 and @ericvergnaud !!

lmy269:
@parrt , Hi Terence, thanks for merging the code, may I know when do you plan to release next version? I do see some similar discussions around this in the internet. And this change improves our production a lot. Thanks again.

parrt:
Hi @lmy269 I have it scheduled for Dec 16th week. I'll work with @ericvergnaud and crew to figure out proper version number and what all we can sneak in.  I'm in middle of intense teaching until about that time.

PR #2663, opened 2019-10-03
Java.stg: only cast if necessary

dhalperi:
Also a bit of code reuse.

### DIFF
--- a/tool-testsuite/test/org/antlr/v4/test/tool/TestActionTranslation.java
+++ b/tool-testsuite/test/org/antlr/v4/test/tool/TestActionTranslation.java
@@ -98,8 +98,8 @@ public class TestActionTranslation extends BaseJavaToolTest {
 		String action = "x, $ID.text+\"3242\", (*$ID).foo(21,33), 3.2+1, '\\n', "+
 						"\"a,oo\\nick\", {bl, \"fdkj\"eck}";
 		String expected =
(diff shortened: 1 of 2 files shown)


dhalperi:
Already in contributors list.

dhalperi:
FWIW, this PR mostly improves readability of generated code. For a way-too-large grammar, the total generated Parser.java decreases by 55K in 11M (0.4%), and the class files by a bit more (closer to 1%-2%). This is likely not much actual performance change.

sharwell:
If this was much earlier in the product development, I probably would have been very much in favor of it. However, considering this is only a code style change (the compiler removes the unnecessary casting so it's not a performance issue), my instinct is to play things safe and keep them the way they are.

dhalperi:
The compiled class files are different, so it's not completely removed by
the compiler.

sharwell:
@dhalperi Can you show an example bytecode difference? I'm curious where this would be different.

dhalperi:
Attaching the dev (this PR) and master (with 4.7.2 release) versions of the generated Java and compiled .class files. The classes are different, but I'm not sure how to check the compiled bytecode atm.

[Archive.zip](https://github.com/antlr/antlr4/files/3722459/Archive.zip)

sharwell:
@dhalperi You can compare the bytecode like this:

```
javap -p -c dev.class > dev.bc
javap -p -c master.class > master.bc
```

Then diff the **master.bc** and **dev.bc** to reveal that the compiled bytecode is identical for the two cases.

PR #2665, opened 2019-10-05
Fix codestyle and bugs in PHP.stg

amorimjuliana:


marcospassos:
Thank you, @amorimjuliana!

@parrt, all green!

PR #2666, opened 2019-10-11
Added @foo::footer to go with @foo::header in Java target

PuppyPi:
These two lines in this one file are all that's needed to support footers as well as headers in "impure" grammars!

This is important because with footers you can include a supertype in grammar file.  You could already add a class as a static nested class in ::members already, but not a superclass since classes can't extend any of their nested classes in Java (not for any technical reason I think, just philosophical java reasons I guess)

But including (non-public) superclasses inside grammars via footers allows them to use the same package selection mechanism the generated lexer/parser classes do, instead of hardcoding the package name into a separate file :>

### DIFF
--- a/contributors.txt
+++ b/contributors.txt
@@ -230,3 +230,4 @@ YYYY/MM/DD, github id, Full name, email
 2019/09/10, amorimjuliana, Juliana Amorim, juu.amorim@gmail.com
 2019/09/17, kaz, Kazuki Sawada, kazuki@6715.jp
 2019/09/28, lmy269, Mingyang Liu, lmy040758@gmail.com
+2019/10/11, PuppyPi, Sean [privacy], puppyofcodez@gmail.com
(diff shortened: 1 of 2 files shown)


PuppyPi:
@sharwell I don't understand, exactly which files are you talking about?  I searched for ParserFile for example and found no template with a name like that.

And <namedActions.header> does appear in the same file I added <namedActions.footer> to, so I figured that would be the best way to add it.

sharwell:
The syntax is correct, but it's in the wrong place. Both of the changes need to move up higher in the Java.stg file, to be in the same location as the header lines.

PuppyPi:
@sharwell Ahhh I understand now.  I didn't notice the template was structured like that.  Yes, this is a much better way, thanks for the suggestion :)

sharwell:
@PuppyPi Sorry my replies were a bit confusing. I was working on my phone at the time and couldn't get you the direct links to the correct locations. I see you did figure out what I was mentioning though. 👍

PuppyPi:
@sharwell  No worries!  Thanks for approving :)

PR #2667, opened 2019-10-12
Another attempt at solving this issue

mike-lischke:
There was already a patch to fix this issue #2037, but it had some problems and was closed by the author after a while.


mike-lischke:
@parrt Will we see a 4.7.3 this autumn? Would be good if this patch could go into it.

parrt:
@mike-lischke yep, i have it scheduled for mid December. will that work?

mike-lischke:
Yes, perfect, thanks!

PR #2672, opened 2019-10-21
Made some variable names more consistent with the other parts.

euske:
Hello, we're developing an automated system that detects inconsistent variable names in a large software project. Our system checks if each variable name is consistent with other variables in the project in its usage pattern, and proposes correct candidates if inconsistency is detected. This is a part of academic research that we hope to publish soon, but as a part of the evaluation, we applied our systems to your projects and got a few interesting results. We carefully reviewed our system output and manually created a patch to correct a few variable names. We would be delighted if this patch is found to be useful. If you have a question or suggestion regarding this patch, we'd happily answer. Thank you.

### DIFF
--- a/runtime/Java/src/org/antlr/v4/runtime/DiagnosticErrorListener.java
+++ b/runtime/Java/src/org/antlr/v4/runtime/DiagnosticErrorListener.java
@@ -76,8 +76,8 @@ public class DiagnosticErrorListener extends BaseErrorListener {
 		String decision = getDecisionDescription(recognizer, dfa);
 		BitSet conflictingAlts = getConflictingAlts(ambigAlts, configs);
 		String text = recognizer.getTokenStream().getText(Interval.of(startIndex, stopIndex));
-		String message = String.format(format, decision, conflictingAlts, text);
-		recognizer.notifyErrorListeners(message);
+		String msg = String.format(format, decision, conflictingAlts, text);
+		recognizer.notifyErrorListeners(msg);
 	}
 
 	@Override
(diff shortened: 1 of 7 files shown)


runtime/Java/src/org/antlr/v4/runtime/DiagnosticErrorListener.java
```diff
@@ -76,8 +76,8 @@ public void reportAmbiguity(Parser recognizer,
 		String decision = getDecisionDescription(recognizer, dfa);
 		BitSet conflictingAlts = getConflictingAlts(ambigAlts, configs);
 		String text = recognizer.getTokenStream().getText(Interval.of(startIndex, stopIndex));
-		String message = String.format(format, decision, conflictingAlts, text);
-		recognizer.notifyErrorListeners(message);
+		String msg = String.format(format, decision, conflictingAlts, text);
+		recognizer.notifyErrorListeners(msg);
```
sharwell: 💭 I prefer `message` to `msg`

runtime/Java/src/org/antlr/v4/runtime/misc/FlexibleHashMap.java
```diff
@@ -64,9 +64,9 @@ public FlexibleHashMap(AbstractEqualityComparator<? super K> comparator, int ini
 		this.initialBucketCapacity = initialBucketCapacity;
 	}
 
-	private static <K, V> LinkedList<Entry<K, V>>[] createEntryListArray(int length) {
+	private static <K, V> LinkedList<Entry<K, V>>[] createEntryListArray(int capacity) {
```
sharwell: 💭 `length` appears to be the more correct term here. Input particular, note that the result has a `length` field which corresponds to this parameter, and does not have a `capacity` field at all.

parrt:
Well, it's not a high priority for me and, while likely a good thing, it's not high enough priority for me to worry about. sorry!

PR #2673, opened 2019-10-22
Fix main entry point for JavaScript runtime npm package

zqlu:
The main entry point for JavaScript runtime package is not right: after publish to npm, the main field should point to index.js.

[https://unpkg.com/browse/antlr4@4.7.2/](https://unpkg.com/browse/antlr4@4.7.2/)


ericvergnaud:
@parrt blessed

PR #2674, opened 2019-10-23
JavaScript runtime bugfixes

ethall:
While adding TypeScript declaration files to the JavaScript runtime module (#2678), I've found and resolved various issues. More detail on each commit can be found it its description.

### Inconsistencies
- bcfeea6
- 5738df1

### Bugs
- d7a82a2
- 87788e7
- 6712106
- Everything after "Updated contributors.txt"

### Checklist
- [x] Signed contributors.txt.
- [x] Done creating type declaration files. (More bugs may be squashed in the meantime.)
- [x] Updated this description with the PR number. (See above.)
- [ ] Added test coverage for all changes.

### DIFF
--- a/contributors.txt
+++ b/contributors.txt
@@ -230,3 +230,4 @@ YYYY/MM/DD, github id, Full name, email
 2019/09/10, amorimjuliana, Juliana Amorim, juu.amorim@gmail.com
 2019/09/17, kaz, Kazuki Sawada, kazuki@6715.jp
 2019/09/28, lmy269, Mingyang Liu, lmy040758@gmail.com
+2019/10/20, ethall, Evan Hall, evan.tom.hall@gmail.com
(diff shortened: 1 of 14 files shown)


ericvergnaud:
Hi,
Thanks for this.
Re so called 'bugs', I suspect this could simply be dead code (inherited from the Java->JS migration).
The only way to integrate these changes would be if you provide tests that provide evidence of the bug and the fix.

ethall:
Sure, not a problem.

ethall:
@ericvergnaud 
If I've understood the documentation correctly, these tests should live in
`runtime-testsuite/test/org/antlr/v4/test/runtime/javascript/*/api`?

ericvergnaud:
suggest you only provide them for node, in 1 file named for example ApiTests under the node folder

ethall:
I've found an alternative solution ([antlr4ts](https://github.com/tunnelvisionlabs/antlr4ts)) that provides a more fully-featured ANTLR4 runtime than the pure JavaScript runtime maintained by the ANTLR project.  Although `antlr4ts` is in alpha at the time of writing, I have yet to run into any significant problems using it.

I've archived my fork in case anyone from the future wants to pick up where I left off.

PR #2678, opened 2019-10-26
JavaScript runtime typings

ethall:
Added TypeScript declaration files to the existing ANTLR4 JavaScript runtime package. The following rules were adhered to during creation:
* Types were inferred based on usage and cross-referenced against the Java runtime implementation.
* All object properties are public unless accessor methods exist.
* Private members are elevated to protected visibility.
I figured that if someone is using one of these members, then having the visibility be protected allows them to create a simple subclass that exposes said member.  If the visibility were changed to private, then they would be forced to perform a more significant rewrite.
* Method visibility mirrors the Java implementation unless a method is used outside of its class.
* Methods unique to the JavaScript implementation are kept public.

**NOTE:** the year in not included in the copyright header, since I wasn't sure what the range should be.


ethall:
I'm also not sure if the existing Maven project will have problems capturing the declaration files or if the project files ought to be updated in general.

KvanTTT:
What about ANTLR TypeScript runtime [antlr4ts](https://github.com/tunnelvisionlabs/antlr4ts)?

ethall:
@KvanTTT 
It's marked as an alpha, so I was hesitant to use it.

ethall:
I don't have the bandwidth to see this through to merge, but I've archived my fork in case anyone from the future wants to pick up where I left off.

PR #2679, opened 2019-10-29
Fix placement of @members in Javascript target lexers

tehbone:
When @members is used in the lexer grammar, the content gets placed in an arbitrary location that is out of scope of the generated lexer. The template was updated so the content gets placed in the constructor, allowing for the creation of class members. This is consistent with the placement of members for parser grammars in the Javascript target.

### DIFF
--- a/contributors.txt
+++ b/contributors.txt
@@ -230,3 +230,4 @@ YYYY/MM/DD, github id, Full name, email
 2019/09/10, amorimjuliana, Juliana Amorim, juu.amorim@gmail.com
 2019/09/17, kaz, Kazuki Sawada, kazuki@6715.jp
 2019/09/28, lmy269, Mingyang Liu, lmy040758@gmail.com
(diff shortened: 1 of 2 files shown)


ericvergnaud:
mmm... this is a breaking change...

tehbone:
Can you elaborate? I do realize that it could impact existing grammars that may use members for lexers in its current form, but to that end I don't believe it sets out to do what it should be doing.

ericvergnaud:
Well one can create a member like this:
MyLexer.prototype.myMethod = function() {};

tehbone:
I'm fine with this being rejected then. I only point out that it is an oddity that parsers have members generated in the constructor while the lexers do not.

ericvergnaud:
True.
May I suggest you run a survey on the google discussion group to find out who uses lexer members in javascript?
If nobody in 2 weeks, we can consider blessing.

ericvergnaud:
@parrt blessed

PR #2684, opened 2019-11-17
Fixed installation instruction of the PHP runtime

adamwojs:
Fixed the following issues in PHP ANTLR Runtime installation instructions:

* `composer require`  should be used o add dependency to your project instead of  `composer install`  (without manual edit of `composer.json`)
* Base on the `composer.json` corrent name of the package is `antlr/antlr4-php-runtime`  (ref. https://github.com/antlr/antlr-php-runtime/blob/master/composer.json#L2)


parrt:
Hi.  I merged something that has changed this to the following:

```
composer require antlr4-php-runtime
```

 is that correct as well? if not, let me know and I can update.

adamwojs:
Hi @parrt!

Correct command to add php runtime to your project dependency is 

```
composer require antlr/antlr4-php-runtime
```

as described on package page https://packagist.org/packages/antlr/antlr4-php-runtime and [README.md file in repository](https://github.com/antlr/antlr-php-runtime/blob/master/README.md#2-install-the-php-antlr-runtime). Executing command without vendor name (`antlr/` prefix)

```
composer require antlr4-php-runtime
```

will fail with `"antlr4-php-runtime" is not a valid alias.` error:

parrt:
fixed and pushed

Issue #2693, opened 2019-12-01
[C# runtime] Parser.Profile = true causes crash, v4.7.2.


kaby76:
I've been looking at the three Java grammars, and tested each of them on 40k+ Java files (it took weeks to run). In order to figure out where some of the performance problems are, I decided to turn on the Parser.Profile option to get some diagnostics. Unfortunately, the parse crashed on a null ptr deref. I traced it down to a version skew between the Java runtime and the C# runtime. It looks like the last time the C# code was generated from the Java code by Sharpen was prior to 2014, and it seemed to have included this problem. I updated the C# runtime to correspond to what the Java runtime code is currently, and Profile=true works great now.

Here are the diffs. The file ProfilingATNSimulator.cs had evil tabs in it, so I converted them to spaces, which git diff doesn't diff very well. I'm going to make a pull request for this change.

    diff --git a/runtime/CSharp/runtime/CSharp/Antlr4.Runtime/Atn/AmbiguityInfo.cs b/runtime/CSharp/runtime/CSharp/Antlr4.Runtime/Atn/AmbiguityInfo.cs
    index fbe0bd7e8..7433c6d52 100644
    --- a/runtime/CSharp/runtime/CSharp/Antlr4.Runtime/Atn/AmbiguityInfo.cs
    +++ b/runtime/CSharp/runtime/CSharp/Antlr4.Runtime/Atn/AmbiguityInfo.cs
    @@ -1,4 +1,4 @@
    -/* Copyright (c) 2012-2017 The ANTLR Project. All rights reserved.
    +/* Copyright (c) 2012-2019 The ANTLR Project. All rights reserved.
      * Use of this file is governed by the BSD 3-clause license that
      * can be found in the LICENSE.txt file in the project root.
      */
    @@ -41,6 +41,9 @@ namespace Antlr4.Runtime.Atn
         /// <since>4.3</since>
         public class AmbiguityInfo : DecisionEventInfo
         {
    +        /** The set of alternative numbers for this decision event that lead to a valid parse. */
    +        public BitSet ambigAlts;
    +
         /// <summary>
         /// Constructs a new instance of the
         /// <see cref="AmbiguityInfo"/>
    @@ -48,19 +51,31 @@ namespace Antlr4.Runtime.Atn
         /// specified detailed ambiguity information.
         /// </summary>
         /// <param name="decision">The decision number</param>
    -        /// <param name="state">
    -        /// The final simulator state identifying the ambiguous
[...]

PR #2694, opened 2019-12-01
Fix https://github.com/antlr/antlr4/issues/2693, [C# runtime] Parser.…

kaby76:
…Profile = true causes crash, v4.7.2.

I'm not in https://github.com/antlr/antlr4/blob/master/contributors.txt, so here's my info, and hereby agree to it.

2019/12/01, Ken Domino, kenneth.domino@domemtech.com


ericvergnaud:
@kaby76 can you fix the AppVeyor build?

kaby76:
This PR is a year old; I forgot all the details of the change. Basically, someone forgot to propagate the changes in the Java target runtime to the C# runtime. That's why the bug is there, for 5+ years. And, it's likely there are other bugs like this. For the Appveyor build, the log indicates failures in the PHP target, which I haven't touched.

Since I haven't looked at this for a year, and hopefully the PHP target now stabilized, I'm going to close this PR, and redo the PR from scratch with the current sources on the main branch.

PR #2699, opened 2019-12-11
Java: don't unnecessary fetch all tokens when grabbing an interval of text (match C# logic)

WalterCouto:
When an exception is thrown, it calls getText to get the text of the tokens involved but the current Java implementation first fetches the rest of the tokens instead of just using the tokens it already has. C#'s version is correct (fill() is called only when asking for the entire stream's text). C++ runtime had the same bug but was fixed by https://github.com/antlr/antlr4/pull/2532


WalterCouto:
@sharwell this was a bug discovered in both C++ and Java api's that was fixed as part of https://github.com/antlr/antlr4/pull/2532 but the Java part was pulled out to a seperate PR.  Please review at your convenience.

dhalperi:
This patch does look correct, but I'm curious how it really matters? Since the buffered token stream never clears the front of its buffer, eventually the whole thing will be kept in memory anyway.

I guess amortizing the overhead of seeking across the entire file doesn't hurt, but does it help?

WalterCouto:
> This patch does look correct, but I'm curious how it really matters? Since the buffered token stream never clears the front of its buffer, eventually the whole thing will be kept in memory anyway.
> 
> I guess amortizing the overhead of seeking across the entire file doesn't hurt, but does it help?

There are a couple of ways we saw this problem, but if you end up parsing small amounts of text or your errors are near the end of your large input file, you may not notice it as badly.

1. If you encounter an error and want to understand where the predictor stopped looking for a valid path, i.e. where the problem token is probably located as apposed to the start location of rule that threw the exception, can be gotten by looking how far into the buffer you read.  We use this property to guide users closer to the problem spot and overall it gives a good indication in many cases especially when compared to the default error position.  Because of this bug, upon first error your entire buffer is filled in and thus you have no glue were the predictor stopped it's look ahead for the rest of the parsing of the file.
2. if you parse large files and to be get to processing the statements as early as possible (i.e. don't wait for the entire file to parse all statements before processing the first statement) then if there was an syntax error in that file (especially close to the beginning) you loose the ability to parse statements without waiting for the entire file to be parsed and thus users see this lag that seems random as it only appears significant if you are processing a large file and there is an error near the beginning.

WalterCouto:
@parrt is this PR being considered for 4.9?

runtime/Java/src/org/antlr/v4/runtime/BufferedTokenStream.java
```diff
@@ -441,15 +441,17 @@ protected int previousTokenOnChannel(int i, int channel) {
 
 	@Override
 	public String getText() {
+		fill();
 		return getText(Interval.of(0,size()-1));
 	}
 
 	@Override
 	public String getText(Interval interval) {
+		lazyInit();
```
sharwell, line 450: 📝 The old behavior was correct here. `getText` has a [documented](https://github.com/antlr/antlr4/blob/9e64dfc6e99f7dcf7cae492927d37290daf57734/runtime/Java/src/org/antlr/v4/runtime/TokenStream.java#L54-L56) [precondition](https://github.com/antlr/antlr4/blob/9e64dfc6e99f7dcf7cae492927d37290daf57734/runtime/Java/src/org/antlr/v4/runtime/TokenStream.java#L30-L33) ([link](https://github.com/antlr/antlr4/blob/9e64dfc6e99f7dcf7cae492927d37290daf57734/runtime/Java/src/org/antlr/v4/runtime/IntStream.java#L115-L117)) that an [initializing method](https://github.com/antlr/antlr4/blob/9e64dfc6e99f7dcf7cae492927d37290daf57734/runtime/Java/src/org/antlr/v4/runtime/IntStream.java#L15-L23) must be called before `getText` can be called.

runtime/Java/src/org/antlr/v4/runtime/BufferedTokenStream.java
```diff
@@ -441,15 +441,17 @@ protected int previousTokenOnChannel(int i, int channel) {
 
 	@Override
 	public String getText() {
+		fill();
```
sharwell, line 444: The old behavior was correct here. See https://github.com/antlr/antlr4/pull/2699#discussion_r504692562.

runtime/Java/src/org/antlr/v4/runtime/BufferedTokenStream.java
```diff
@@ -441,15 +441,17 @@ protected int previousTokenOnChannel(int i, int channel) {
...
 	}
 
 	@Override
 	public String getText(Interval interval) {
+		lazyInit();
 		int start = interval.a;
 		int stop = interval.b;
 		if ( start<0 || stop<0 ) return "";
-		fill();
+		sync(stop);
```
sharwell, line 454: 📝 This is the only part of the change that really works. I'd keep this line and remove the others.

sharwell (changes requested):
See comments in code

runtime/Java/src/org/antlr/v4/runtime/BufferedTokenStream.java
WalterCouto: Agreed. This matches up with how getTokens() method works which is just use the current list of tokens. We should also fix up C# and C++ to do the same?

runtime/Java/src/org/antlr/v4/runtime/BufferedTokenStream.java
WalterCouto: Agreed, however I noticed that getTokens which is logically doing a similar action, does call lazyInit() so should not getText() and getTokens() be consistent?

parrt:
@sharwell  if you approve of this PR in its entirety, I will merge; I don't have time to check through all of the issues.

PR #2711, opened 2020-01-03
C++ VS project updates for next ANTLR4 release

mike-lischke:
This patch contains a few changes to build now VS 2017/19 binaries instead of VS 2013/15 (with the deployment script). Also a warning was fixed (suppressed) and the docs updated wrt. the C++ target.

### DIFF
--- a/doc/releasing-antlr.md
+++ b/doc/releasing-antlr.md
@@ -379,7 +379,7 @@ There are links to the artifacts in [download.html](http://www.antlr.org/downloa
 
 The C++ target is the most complex one, because it addresses multiple platforms, which require individual handling. We have 4 scenarios to cover:
 
-* **Windows**: static and dynamic libraries for the VC++ runtime 2013 or 2015 (corresponding to Visual Studio 2013 or 2015) + header files. All that in 32 and 64bit, debug + release.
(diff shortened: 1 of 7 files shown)


mike-lischke:
@parrt Here's the promised patch. Depending on which prebuilt binaries you want to provide for download you will need Visual Studio 2017 and/or Visual Studio 2019. In the last ANTLR4 release we only had libs for VS 2015 IIRC, so it should be fine to only provide VS 2019 libs now. The script will check and compile only with those versions that really exist on your box. It will even succeed if you have no VS at all installed :-D (you just don't get any binaries to publish then). The release doc has been updated too.

parrt:
I only have 2019 so we're cool. thanks!

parrt:
Getting some errors:
Screen Shot 2020-01-03 at 8 40 31 AM

parrt:
Screen Shot 2020-01-03 at 8 58 11 AM

mike-lischke:
> Getting some errors:
> Screen Shot 2020-01-03 at 8 40 31 AM

I got similar errors, seems something specific to VS 2019. However I could fix them. Can you check if that path given in the error message exists? Maybe it's something that is different for the community version of VS (I have professional)?

PR #2733, opened 2020-01-22
Fix #2727: improve setTokenFactory in Cpp target

### LINKED ISSUE #2727: cpp target: Lexer::setTokenFactory template function unusable
The `Lexer` class provides the [`setTokenFactory` template function](https://github.com/antlr/antlr4/blob/master/runtime/Cpp/runtime/src/Lexer.h#L98-L101), but its function signature means that its body can't compile. The function is defined as follows:

    template<typename T1>
    void setTokenFactory(TokenFactory<T1> * factory) {
        this->_factory = factory;
    }

where [`this->_factory`](https://github.com/antlr/antlr4/blob/master/runtime/Cpp/runtime/src/Lexer.h#L34) is declared as a [`std::shared_ptr`](https://en.cppreference.com/w/cpp/memory/shared_ptr):

    Ref<TokenFactory<CommonToken>> _factory;

The function can't work because `std::shared_ptr<T>` provides no assignment operator from `T*`, nor does it provide an implicitly converting constructor from `T*` (i.e., its `T*` constructor is marked `explicit`). `Lexer` works with the default `CommonTokenFactory` only because it assigns its `_factory` using `CommonTokenFactory::DEFAULT`, which is a `shared_ptr` of the same type as `_factory`.

I believe `setTokenFactory` on `Lexer`, `Parser`, and `TokenSource` should be changed to the following signature:

    template<typename T1>
    void setTokenFactory(Ref<TokenFactory<T1>> const& factory);

and the implementation of `Lexer::setTokenFactory` should be:

    template<typename T1>
    void setTokenFactory(Ref<TokenFactory<T1>> const& factory) {
        this->_factory = std::static_pointer_cast<TokenFactory<CommonToken>>(factory);
    }

Because of the declared type of `Lexer::_factory`, it's necessary for the `factory` argument here to be convertible to a `std::shared_ptr<TokenFactory<CommonToken>>`.

vinoski:
Change the setTokenFactory template function in the Lexer, Parser, and
TokenSource classes to be parameterized by Factory type rather than by
token type. Change the setTokenFactory signature to require the
function argument to be a Ref of the Factory template parameter type
so it can be correctly converted and assigned to the token factory
member. The type of that member is Ref<TokenFactory<CommonToken>>,
which implies that the Factory template parameter must be of a type
convertible to TokenFactory<CommonToken>.

### DIFF
--- a/runtime/Cpp/runtime/src/Lexer.h
+++ b/runtime/Cpp/runtime/src/Lexer.h
@@ -31,7 +31,7 @@ namespace antlr4 {
 
   protected:
     /// How to create token objects.
-    Ref<TokenFactory<CommonToken>> _factory;
+    TokenFactory<CommonToken> *_factory;
 
   public:
     /// The goal of all lexer rules/methods is to create a token object.
@@ -100,7 +100,7 @@ namespace antlr4 {
       this->_factory = factory;
     }
 
-    virtual Ref<TokenFactory<CommonToken>> getTokenFactory() override;
+    virtual TokenFactory<CommonToken>* getTokenFactory() override;
 
     /// Set the char stream and reset the lexer
     virtual void setInputStream(IntStream *input) override;
--- a/runtime/Cpp/runtime/src/CommonTokenFactory.cpp
+++ b/runtime/Cpp/runtime/src/CommonTokenFactory.cpp
@@ -11,7 +11,7 @@
 
 using namespace antlr4;
 
-const Ref<TokenFactory<CommonToken>> CommonTokenFactory::DEFAULT = std::make_shared<CommonTokenFactory>();
+const std::unique_ptr<TokenFactory<CommonToken>> CommonTokenFactory::DEFAULT(new CommonTokenFactory);
 
 CommonTokenFactory::CommonTokenFactory(bool copyText_) : copyText(copyText_) {
 }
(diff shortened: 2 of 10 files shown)


runtime/Cpp/runtime/src/Lexer.h
```diff
@@ -95,9 +95,9 @@ namespace antlr4 {
     virtual void pushMode(size_t m);
     virtual size_t popMode();
 
-    template<typename T1>
-    void setTokenFactory(TokenFactory<T1> *factory)  {
-      this->_factory = factory;
+    template<typename Factory>
+    void setTokenFactory(Ref<Factory> const& factory)  {
+      this->_factory = std::static_pointer_cast<TokenFactory<CommonToken>>(factory);
```
mike-lischke: I'm afraid this is the wrong approach. Look at the Java source code, which uses `public void setTokenFactory(TokenFactory<?> factory)`, so it's clear the token class must be templated, not the token source class. And your cast wouldn't work if something else but a CommonTokenFactory is passed in.

This part of the C++ runtime is a bit tricky, as there's no simple equivalent for the Java generic used for it (C++ cannot directly define templated member variables). The correct approach to use here is a simple pointer. This is also what is used in the other classes (e.g. see `setInterpreter` and `addErrorListener` in Recognizer.h, which all just use a plain object pointer). The caller of `setTokenFactory` is reponsible to manage the memory used by the factory instance. This works also well with the factory function for token factories (see `TokenFactory::create`, which creates a unique pointer).

vinoski: Great, thanks for this guidance. There were a number of ways to go here; I went with one that preserved the current type of the `Lexer::_factory` member. Your feedback clarifies that changing the type of that member is ok, and I can see the rest will flow from that. I'll post a revision here shortly.

PR #2739, opened 2020-01-30
Unicode properties split

jamesjer:
Building ANTLR 4.8 on a Fedora Rawhide machine with icu4j 65.1, maven says:
```
[ERROR] Failed to execute goal org.apache.maven.plugins:maven-compiler-plugin:3.8.1:compile (default-compile) on project antlr4: Compilation failure
[ERROR] /builddir/build/BUILD/antlr4-4.8/tool/target/generated-sources/tool/src/org/antlr/v4/unicode/UnicodeData.java:[60122,28] code too large
```
This is a proposed split to keep the code within the Java limits.  I have tested this split and it works for me.

On a personal note, it has been nearly 20 years since I last pestered Terence about java.g, which no longer seems to be in the ANTLR distribution.  I think that was just before JDK 1.5 was released.  Those were the days....


jamesjer:
I have updated this pull request to resolve merge conflicts.  This is still needed for Fedora.

parrt:
@ericvergnaud I'm not wise to the ways of Unicode. Can you check this for sanity?

ericvergnaud:
@parrt noted, will do

ericvergnaud:
@jamesjer I struggle to understand how this change can possibly address a code size issue, when both current and proposed generated UnicodeData. java contain all the literals... both classes should have almost the same size

jamesjer:
The issue isn't the amount of code in the class, but rather the amount of code in method `addPropertyAliases`, which starts at line 60122, as mentioned in the error message.  Without this patch, and with icu4j 65.1, `addPropertyAliases` contains 2440 lines of this form:
```
propertyAliases.put("Name".toLowerCase(Locale.US), "name".toLowerCase(Locale.US));
```
Compiling those 2440 lines with OpenJDK 8 (which we were using in January) or OpenJDK 11 (which we are using now) apparently produces more than 64 Kb of bytecode.  This patch breaks that method up into 2 methods: `addPropertyAliases1`, with 871 such lines, and `addPropertyAliases2`, with 1569 such lines.  Each of those can be compiled into less than 64 Kb of bytecode.

ericvergnaud:
According to the jvm spec the max size of a 'Code' attribute is a u4 i.e. unsigned int. 
See https://docs.oracle.com/javase/specs/jvms/se7/html/jvms-4.html#jvms-4.7.3
So not sure what is causing this observed 64k limit.
This smells like a Fedora-specific bug... In theory, the same java source code should produce the same .class on all platforms so that code should also break on other platforms.  
Have you tried:
 - running on Fedora the tool built on a non problematic platform
 - using a different jdk on Fedora
(I don't thing your PR would break anything but it might not be addressing the root cause)

jamesjer:
That link appears to be for Java 7, right?  If you look at the equivalent link for JDK 8:

https://docs.oracle.com/javase/specs/jvms/se8/html/jvms-4.html#jvms-4.7.3

you will see that the `code_length` description has gained some text.  Now it reads:
```
The value of code_length must be greater than zero (as the code array must not be empty) and less than 65536. 
```

That text is unchanged in the JDK 11 and 14 versions of the document.  I don't think the JDK is the deciding factor here.  I think it is the version of icu4j that matters.  The number of property aliases has simply grown too large.  Compiling the patched code with OpenJDK 11 yields a 23516 byte `addPropertyAliases1` method, and a 42363 byte `addPropertyAliases2` method.  The sum of those is just barely too large for a single method.

ericvergnaud:
Well noted re code_length.
But then how is it that icu4j, a pure java library, returns a different set of unicode entities across platforms, when they claim the exact opposite? 
_"ICU is widely portable and gives applications the same results on all platforms and between C/C++ and Java software"_

ericvergnaud:
Looks like the icu4j version you are using is icu4j 65.1, when the one referred to by the tool pom.xml is 61.1.
Is that intentional?

ebourg:
We are hitting the same issue in Debian when building antlr4 with icu4j 68.2. The issue appears if you move the icu4j version in `tool/pom.xml` past 64.1.

PR #2742, opened 2020-02-03
Javascript: migrate prototypical error object to es6 classes

carocad:
This PR refactors the error object declarations to:
- minimum node version to 10.x to support es6 classes

EDIT 07.02.2020:
This PR now also:
- adds a webpack configuration for publishing the Javascript runtime as a single `antlr4.js` file
- adapts the documentation for publishing with antlr4 js accordingly
- removes the conditional node.js require in favour of the webpack bundle
- removes the custom async browser require script in favour of the webpack bundle


ericvergnaud:
It's a good start!
Since this code will no longer run in old browsers, may I require that the runtime be packaged (and babelized) ?

carocad:
> Since this code will no longer run in old browsers, may I require that the runtime be packaged (and babelized) ?

I am a bit confused on this. Afaik the javascript runtime is not provided as a minified file but the user is expected to use webpack to get it in the browser. Do you mean for me to update the documentation of [javascript target](https://github.com/antlr/antlr4/blob/master/doc/javascript-target.md) ?

ericvergnaud:
You’re correct that the runtime is currently source only. This was acceptable because the source could be used ‘as is’ which would no longer be the case. 
It would be great if, as part of this modernization, you could provide a runtime directly usable in a browser. We can then update the docs accordingly.

Envoyé de mon iPhone

ericvergnaud:
Am I correct in guessing you are using lebab?

carocad:
> Am I correct in guessing you are using lebab?

No, I use Webstorm auto convertion system. First time that I have used it though. Seems to work pretty well.

> If so, suggest you also convert require into import

Unfortunately this is not possible. I already tried it and I couldnt get it working even with the latest version of Node.js (see [docs](https://nodejs.org/api/esm.html#esm_ecmascript_modules)). On my local machine node wouldnt even parse the refactored files. Therefore I dropped that idea and went with classes directly.

> We could further evolve to 1 file per class and leverage export default ...

This is already possible. Right now the code is exposed as `module.exports = { MyClass }` which creates an unnecessary wrapping object. The equivalent of `export default` is `module.exports = MyClass`, which is perfectly valid js.

Btw, the tests on the PR are failing but due to a [connection error on Travis](https://travis-ci.org/antlr/antlr4/jobs/647321480#L1165) which has nothing to do with the actual changed code. Are you able to request a re-run of that specific job ? I am not able to do that on Travis CI.

On another topic, how should we proceed with the refactoring ? do we wait until this PR is approved and merged by the author of Antlr4 or we keep working on this branch until the Js runtime is done ?

ericvergnaud:
@parrt blessed
n.b. more to come but I'd rather watch this progress gradually

ericvergnaud:
Hi,

> Le 8 févr. 2020 à 22:09, Camilo Roca <notifications@github.com> a écrit :
> 
> Am I correct in guessing you are using lebab?
> 
> No, I use Webstorm auto convertion system. First time that I have used it though. Seems to work pretty well.
> 
> If so, suggest you also convert require into import
> 
> Unfortunately this is not possible. I already tried it and I couldnt get it working even with the latest version of Node.js (see docs <https://nodejs.org/api/esm.html#esm_ecmascript_modules>). On my local machine node wouldnt even parse the refactored files. Therefore I dropped that idea and went with classes directly.
> 
> 


For require conversion, maybe lelab works better ? (I used it on another project)

> We could further evolve to 1 file per class and leverage export default ...
> 
> This is already possible. Right now the code is exposed as module.exports = { MyClass } which creates an unnecessary wrapping object. The equivalent of export default is module.exports = MyClass, which is perfectly valid js.
> 
> 

Let’s do that then :-)

> Btw, the tests on the PR are failing but due to a connection error on Travis <https://travis-ci.org/antlr/antlr4/jobs/647321480#L1165> which has nothing to do with the actual changed code. Are you able to request a re-run of that specific job ? I am not able to do that on Travis CI.
> 
> 

It’s passing now
> On another topic, how should we proceed with the refactoring ? do we wait until this PR is approved and merged by the author of Antlr4 or we keep working on this branch until the Js runtime is done ?
> 
> —
> 
> 
Small PRs are preferable, I’ve blessed this one.

Cheers, and thanks so much for helping out :-)

PR #2744, opened 2020-02-10
Fix CMake packaging

### LINKED ISSUE #2606: [C++][CMake] find_package() fails when the C++ runtime is make-install'ed

julibert:
This pull request fixes #2606. It installs `antrl4-generator` and `antlr4-runtime` in separate files following the CMake's [search procedure](https://cmake.org/cmake/help/latest/command/find_package.html#search-procedure).

### DIFF
--- a/contributors.txt
+++ b/contributors.txt
@@ -238,3 +238,4 @@ YYYY/MM/DD, github id, Full name, email
 2019/11/17, felixn, Felix Nieuwenhuizhen, felix@tdlrali.com
 2019/11/18, mlilback, Mark Lilback, mark@lilback.com
 2020/02/02, carocad, Camilo Roca, carocad@unal.edu.co
+2020/02/10, julibert, Julián Bermúdez Ortega, julibert.dev@gmail.com
--- a/runtime/Cpp/CMakeLists.txt
+++ b/runtime/Cpp/CMakeLists.txt
@@ -141,11 +141,12 @@ if (ANTLR4_INSTALL)
   include(CMakePackageConfigHelpers)
 
   if(NOT ANTLR4_CMAKE_DIR)
-    set(ANTLR4_CMAKE_DIR ${CMAKE_INSTALL_LIBDIR}/cmake/antlr4 CACHE STRING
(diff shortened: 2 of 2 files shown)


mike-lischke:
@parrt This is a C++ only patch and ready to merge.

PR #2746, opened 2020-02-11
Javascript: migrate prototypical DFA objects to es6 classes

carocad:
This PR is a follow up on #2742 to refactor the DFA object declarations to:


ericvergnaud:
@carocad no comment, thanks for this
@parrt approved

PR #2496, opened 2019-02-15
Fix errors in JavaScript run-time.

hemvignesh:
Mostly they are references to bogus properties / forgotten imports / forgotten exports.

Found them while I was attempting to bundle ANTLR4's JavaScript with [Closure Compiler](https://developers.google.com/closure/compiler/) which does some static analysis.

PR #2749, opened 2020-02-17
 Javascript: migrate prototypical Tree objects to es6 classes

carocad:
This PR is a follow up on #2746.

The (original) scope was to refactor the Tree object declarations to:

However, during the refactoring a couple of problems arised which increased the scope to also include:
- change the js test suite string templates to avoid using `antlr4.tree.ParseTreeListener.call(this)`. This is a consecuence of this PR refactor. Classes unlike functions cannot be "called" only instantiated. A possible substitution for this would be `new antlr....`. However a closer inspection showed that this is an interface with no inner members therefore it makes no sense to instantiate an objec and throw it away inmediately.
- expose some "classes" as `module.export = MyClass`. This is a follow up on [this comment](https://github.com/antlr/antlr4/pull/2742#issuecomment-583798501) from @ericvergnaud to make the JS target closer to the Java one. As I was already touching these files it made sense to do the change inmediately.

Note to @ericvergnaud :
- During this refactoring I found out that the antlr4 implementation has a bug [here](https://github.com/antlr/antlr4/compare/master...carocad:tree-classes?expand=1#diff-d14b4c5bc8a076a45a5d88cf81fcea0eR54). There is a cyclic dependency which is not being resolved correctly. This causes the `const ATN = require ....` to return an empty object which means that `ATN.INVALID_ALT_NUMBER` returns `undefined` as opposed to `0`. Since fixing that bug would require quite some refactoring I left it there. This bug has been there even before I started the refactoring. See https://github.com/antlr/antlr4/pull/2496#discussion_r259058850


runtime/JavaScript/src/antlr4/RuleContext.js
```diff
@@ -30,7 +30,6 @@ var INVALID_INTERVAL = require('./tree/Tree').INVALID_INTERVAL;
 var INVALID_ALT_NUMBER = require('./atn/ATN').INVALID_ALT_NUMBER;
```
ericvergnaud: I think this should be: ...).ATN.INVALID_ALT_NUMBER;

runtime-testsuite/test/org/antlr/v4/test/runtime/javascript/browser/BaseBrowserTest.java
```diff
@@ -656,7 +656,6 @@ protected void writeParserTestFile(String parserName,
 			"				};\r\n" +
 			"\r\n" +
 			"				TreeShapeListener = function() {\r\n" +
-			"					antlr4.tree.ParseTreeListener.call(this);\r\n" +
```
ericvergnaud: how about converting this test class to ES6?

runtime-testsuite/test/org/antlr/v4/test/runtime/javascript/node/BaseNodeTest.java
```diff
@@ -627,7 +627,6 @@ protected void writeParserTestFile(String parserName, String lexerName,
 						+ "var <visitorName> = require('./<visitorName>').<visitorName>;\n"
 						+ "\n"
 						+ "function TreeShapeListener() {\n"
-						+ "	antlr4.tree.ParseTreeListener.call(this);\n"
```
ericvergnaud: how about converting this test class to ES6?

runtime/JavaScript/src/antlr4/RuleContext.js
carocad: this is fine as it is because I changed the export of ATN.js to be `module.exports = ATN;` which means that the object itself is exported. This is precisely the equivalent of export default.

runtime-testsuite/test/org/antlr/v4/test/runtime/javascript/browser/BaseBrowserTest.java
carocad: I thought the main point of the browser tests was that they should work everywhere ? if I convert this to es6 classes then we wouldnt be able to do that right ?

runtime/JavaScript/src/antlr4/RuleContext.js
ericvergnaud: so it's fine now? I thought you said it was undefined?

runtime-testsuite/test/org/antlr/v4/test/runtime/javascript/node/BaseNodeTest.java
ericvergnaud: node should support this one

runtime/JavaScript/src/antlr4/RuleContext.js
carocad: sorry for the misunderstanding. The situation is as follows:
- the ATN module is exported directly so its members are directly available as shown above.
- the cyclic dependency prevents the ATN module from being correctly imported, therefore node just returns an empty object `{}` which results in `INVALID_ALT_NUMBER` being `undefined`. However, that has nothing to do with the way the object is exposed but rather with the cyclic dependency itself.

runtime-testsuite/test/org/antlr/v4/test/runtime/javascript/browser/BaseBrowserTest.java
ericvergnaud: The browser tests are not run as part of the CI, but I suspect they will fail...

carocad: I tried running them and indeed they fail. Not only because of this refactoring but it seems that the previous ones also added some errors 😞 

The browser tests are still using `lib/require` even though we removed it. I rolled that back locally but still ran into errors due to the missing module `fs`. Since the whole library is expected to be webpacked.

I see two options right now:
- modify the browser test suite to use webpack
- rollback some of the changes for lib/require and conditional `fs` module import

Let me know what you think.

ericvergnaud: Preferred option is "modify the browser test suite to use webpack"
I did that in another project (run webpack from Java) and it's quite a piece of work...

carocad: > (run webpack from Java) and it's quite a piece of work...

The "piece of work" was meant ironically ? My first view of the setup for those tests is that it is definitely not going to be easy :/

I was already inclining towards a rollback and leave this test suite to the end 🙈

ericvergnaud: It's not my IP unfortunately 

Another approach could be to pre-webpack the text class making the parent class external?

carocad:
hey @ericvergnaud I think it is more appropiate to continue the discussion over here from: https://github.com/antlr/antlr4/pull/2749#discussion_r381216543

So I have good news and bad news:

good: I think I have a fix for the failing tests. I basically did a rollback of the `require.js` polyfill but on a `mock` directory together with the `fs` module. This allows a browser to "require" the complete antlr library and the node.js ones as well. I testes it manually and it "worked". I havent pushed the changes upstream yet due to the problem below.

bad: I tried to run the tests on different browsers but in all cases they fail with empty output 😞 . Eventually I started doubting the tests themselves and rolled back to commit 2d7f727a00b70554df172c9048b91712f5dceeb0 and tried the tests; and they also fail 😨 .

Therefore right now I am stuck. There are no documentation on the browser test suite (so I dont know if I need a special setup) and even the tests before the refactoring started are failing for me. Any hints/suggestion are welcomed.

ericvergnaud:
@carocad I will try and look into this by end of next week

ericvergnaud:
Ok, I've been looking into this, and from what I can see:
 - the browser testing framework stopped working a loooong time ago (probably end 2016)
 - nobody has complained since
 - nobody's using the runtime js files as is anymore i.e. they're always packaged
 - the new package does the es2015 conversion (correct?)
 - the runtime does not rely on any browser runtime behavior
From there, I think it's ok to simply drop these tests suites.
I will do that in a separate PR.

carocad:
> the new package does the es2015 conversion (correct?)

I just double checked this and for full disclosure [here](https://gist.github.com/carocad/2da39f4d918fd25a704b7a3c3da92724) is the webpack processed antlr4 runtime. It seems that it is still using classes and other "newer" stuff. Therefore I think I would need to modify the webpack configuration (separate PR?).

Btw, how do we proceed with this PR then ? Is it ok as it is ?

ericvergnaud:
do we still need a fix for the INVALID_ALT_NUMBER?

carocad:
> do we still need a fix for the INVALID_ALT_NUMBER?

definitely, but as a mention before that would require quite some refactoring on the way that several modules are imported and exported. Therefore I wanted to tackle that separatedly as it is not a problem that arised due to this PR but rather it has been there from the very beginning.

ericvergnaud:
how about the following:
const INVALID_ALT_NUMBER = require(...). INVALID_ALT_NUMBER || 0;
that would ensure the correct value for now and I can look into it later...

carocad:
For transparency here is the cyclic dependency cycle. I got it using [madge](https://github.com/pahen/madge)
```bash
$ madge --circular src/antlr4/index.js 
Processed 47 files (1.1s) 

✖ Found 3 circular dependencies!

1) LL1Analyzer.js > PredictionContext.js > RuleContext.js > atn/ATN.js
2) RuleContext.js > tree/Trees.js > ParserRuleContext.js
3) RuleContext.js > tree/Trees.js
```

and the complete dependency graph starting at `antlr4/index.js`. The `red` parts are the cyclic dependencies.


deps

ericvergnaud:
> For transparency here is the cyclic dependency cycle. I got it using [madge](https://github.com/pahen/madge)
> 
> ```shell
> $ madge --circular src/antlr4/index.js 
> Processed 47 files (1.1s) 
> 
> ✖ Found 3 circular dependencies!
> 
> 1) LL1Analyzer.js > PredictionContext.js > RuleContext.js > atn/ATN.js
> 2) RuleContext.js > tree/Trees.js > ParserRuleContext.js
> 3) RuleContext.js > tree/Trees.js
> ```
> 
> and the complete dependency graph starting at `antlr4/index.js`. The `red` parts are the cyclic dependencies.
> 
> deps

My usual workaround in ES2015 is as follows:

in _imported_ script
```
var imported = null;

exports.resolve = function() {
    imported = require(...);
} 
```

in _importing_ script
```
var stuffScript = require("some-script");
var stuff = stuffScript.stuff;
stuffScript.resolve() ;
```

not sure how this can be achieved in ES6 though...

ericvergnaud:
@parrt blessed

PR #2753, opened 2020-02-21
Add contextImplements rule option

StochasticTinkr:
Allows a new "contextImplements" rule option. The value of this option will be used to add an `implements` clause on the Java class declaration of the rule's Context class.

I wasn't quite sure the most appropriate place to add this code, so if it belongs in another part of the processing, let me know and I can move it.  If there is any additional changes you would like, also let me know.


StochasticTinkr:
As stated in the related issue, I don't think this solves what I wanted.

PR #2754, opened 2020-02-21
replace gramamr with grammar

StochasticTinkr:


ericvergnaud:
Thanks for this. Have you signed the contributors agreement? (appreciate it might seem overkilling for fixing typos)

StochasticTinkr:
I have another (more substantial) pull request, where I did. If you insist, I can do it in this one, but it definitely seems like overkill.

ericvergnaud:
I saw the other one, which might take longer to merge, hence why I'd rather you add it to this one to start with.

StochasticTinkr:
Fair enough. I’ll add it when I’m next in front of my computer

ericvergnaud:
@parrt blessed

PR #2755, opened 2020-02-22
Drop browser tests

ericvergnaud:


ericvergnaud:
@parrt I might need your help with this one... 
this is only about javascript testing
mvn test script works fine locally, but fails on travis, complaining it's not finding any tests
not sure it that has to do with some test generation config?

ericvergnaud:
@parrt forget about the help, pb fixed
Not sure why git decided to ignore those test files when I renamed the package...
Good to merge

PR #2761, opened 2020-02-23
Javascript: browser support increased with babel transformation

carocad:
This PR is a follow up on:
- https://github.com/antlr/antlr4/pull/2749#issuecomment-589905171
- https://github.com/antlr/antlr4/pull/2755

This PR adds a babel transformation to the webpack configuration of antlr4.js. This "enables transforms for ES2015+" as stated on [babel website](https://babeljs.io/setup#installation).

According to [this website](https://browserl.ist/) this would default to a global coverage of `90.77%`.

For a comparison of the output code you can check [this gist](https://gist.github.com/carocad/2da39f4d918fd25a704b7a3c3da92724)

### DIFF
--- a/runtime/JavaScript/.babelrc
+++ b/runtime/JavaScript/.babelrc
@@ -1,0 +1,3 @@
+{
+  "presets": ["@babel/preset-env"]
+}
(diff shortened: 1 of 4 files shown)


ericvergnaud:
@parrt blessed

PR #2768, opened 2020-03-04
Drop maven2 artifact maven-project in favor of maven-core

jamesjer:
The other maven artifacts are maven3 already.  The necessary classes are in maven-core, so give it compile scope instead of test scope.

My addition to contributors.txt is part of the first pull request I submitted: https://github.com/antlr/antlr4/pull/2739


parrt:
I don't really understand maven very well.  and I'm worried this would break the current build...

PR #2771, opened 2020-03-08
Javascript: migrate prototypical ATN objects to ES6

carocad:
This PR is a follow up on #2749.

The scope of this PR is to refactor the ATN object declarations to:
- refactor exports to directly expose a single class whenever possible

NOTE: I also renamed [this](https://github.com/antlr/antlr4/compare/master...carocad:atn-es6?expand=1#diff-f9342208b7997e9e39d3e59ae7f0c90cL338) "constructor" to be the `equals` method as it made no sense to have two constructors and for consistency with the `SemanticContext.java` implementation


ericvergnaud:
Hi,

thanks for this

re "NOTE: I also renamed this "constructor" to be the equals method as it made no sense to have two constructors and for consistency with the SemanticContext.java implementation"
I couldn't find the corresponding change (the link doesn't lead to a specific chunk)

Cheers

carocad:
@ericvergnaud it seems that github doesnt show the file due to the file being "collapsed" i.e diff too big.

Anyway, [here is the actual diff](https://github.com/antlr/antlr4/commit/6ebe0d13170d2a09abda0765372b45ebb00530cb#diff-f9342208b7997e9e39d3e59ae7f0c90cL338).

Specifically this below:

```js
OR.prototype = Object.create(SemanticContext.prototype);
OR.prototype.constructor = OR;

OR.prototype.constructor = function(other) {
	if (this === other) {
		return true;
	} else if (!(other instanceof OR)) {
		return false;
	} else {
		return this.opnds === other.opnds;
```
As you can see the code declares the constructor of `OR` as being the function `OR` but then it overwrites it with another "constructor" which checks if the two objects are `equal`.

ericvergnaud:
it does look like a bug, but then I wonder how the whole thing works...

ericvergnaud:
maybe we're lacking some tests for semantic predicates

carocad:
> it does look like a bug, but then I wonder how the whole thing works...

a simple 
```js
const foo = new OR("a", "b")
console.log(foo)
```
at the end of the `SemanticContext.js` file shows that the `constructor` isnt called when calling `new` bur rather the "real" constructor is being called. It simply sets the values on the prototype chain. I guess that this was working due to that "misconception" 🤷‍♂

ericvergnaud:
@parrt blessed

ericvergnaud:
@carocad if not already done, would you be able to provide a doc for using this ES6 version locally?
I've tried npm link, which ends up badly (ES2015 functions inheriting from ES6 classes...)
I've tried webpack, but that creates a browser only minimized file

carocad:
> would you be able to provide a doc for using this ES6 version locally?

Not sure what you mean. Locally it is simply using node.js so a standard `const antlr4 = require('./src/antlr4')` would do (assuming that you are in the js runtime dir).

That is also what the test suite does. Generates the code and copies the js runtime locally so that it is available. 

> I've tried npm link, which ends up badly (ES2015 functions inheriting from ES6 classes...)

mmmm 🤔 . I tried `npm link` as per the [official documentation](https://docs.npmjs.com/cli/link) and it worked as expected. I was able to import the library with `const antlr4 = require('antlr4')`. Nodejs displays the `antlr4` object as containing functions however I think that is normal; as classes are just sugared functions.

I could do an extra guide on how to use it but right now I don't know exactly what to put there as it is simply "require it as you would any other node.js library" 🤷‍♂️

ericvergnaud:
I suspect the issue I encountered related to using pre ES6 generated parser and lexer with latest ES6 runtime in node
I’ll double check, and if the case, we'll need to document that to limit support

Issue #2450, opened 2018-12-23
Compiling warning


blob84:
I have compiled the c++ target following the instructions and I've got a lot of warning:

`/home/user/antlr4/runtime/src/atn/ATNState.h:73:26: warning: type attributes ignored after type is already defined [-Wattributes]
   class ANTLR4CPP_PUBLIC ATN;
                          ^~~
/home/user/antlr4/demo/generated/TParser.h:122:38: warning: extra ‘;’ [-Wpedantic]
     antlr4::Token *idToken = nullptr;;
                                      ^`
I'm using `gcc version 6.3.0 20170516 (Debian 6.3.0-18+deb9u1) 
`

mike-lischke:
I'm afraid I can't do anything about either issue. The attribute warning seems to be a peculiarity of certain compilers. The forward declarations are perfectly ok and accepted without any trouble by VS, clang and also gcc (as far as I have tried, on various platforms). This has also been discussed in the [ANTLR Google group](https://groups.google.com/forum/#!topic/antlr-discussion/ULujLaottNs) last September.

The additional semicolon is caused by the way string templates are used by the code generator in ANTLR4. There are nested ST rules, which sometimes are used in places where they must not add a semicolon and in other places where a semicolon is required. I tried for hours to solve that and gave up (I believe the only true solution would be to restructure the ST rules, which is not a good idea for compatibilitiy reasons). I decided it's better to carry on with 2 semicolons (which are no syntax error, just useless) instead of getting code errors for missing ones.

This issue can be closed @parrt.

PR #2773, opened 2020-03-09
Fix compiler warning cpp

jmcken8:
GCC generates a warning when forward-declaring a class if that class has
already declared with `__attributes__`:

    warning: type attributes ignored after type is already defined [-Wattributes]

This is a bug in GCC: https://gcc.gnu.org/bugzilla/show_bug.cgi?id=39159

This works around that bug by only forward-declaring the class if it hasn't
already been declared.

Addresses (parts of): #2450, #2481


mike-lischke (commented):
If this patch is still valid then please fix the contributors.txt file conflict and we can merge it.

mike-lischke:
The original author can just add more patches to his PR to fix the issue. If the OP is not interested in fixing it then we should close it and you may pick it up and post a new PR.

jmcken8:
```
java.io.IOException: Process exited with status 143:
```

It looks like the swift process was killed with SIGTERM. Any idea why that would happen?

mike-lischke (approved):

mike-lischke:
That's frequently happens and you can ignore that.

@parrt This patch is ready to merge.

Issue #2759, opened 2020-02-22
Typo in Java runtime documentation


XsongyangX:
The ParseTreeWalker class has a typo the runtime documentation of the enterRule method. Here is the link to the docs: https://www.antlr.org/api/Java/index.html
![image](https://user-images.githubusercontent.com/10444460/75095456-de1fbf80-5562-11ea-8a4f-aee0d41cef8d.png)

ericvergnaud:
Would you mind providing a PR?

XsongyangX:
I don't really understand what the right formulation should be.

ericvergnaud:
This was copied from the Java target, maybe you can check there?

XsongyangX:
I found the doc entry in the [repository](https://github.com/antlr/antlr4/blob/225249fdaef5a1d6bdc3694ab0e1fc936cf7f527/runtime/Java/src/org/antlr/v4/runtime/tree/ParseTreeWalker.java#L37) but it displays the same thing. I don't think there is a correct formulation that is already written down, since the docs are automatically generated from the source code.

ericvergnaud:
Fair enough.

What this is supposed to explain is the following sequence:
1° walker calls 'enterEveryRule(ctx)
2° walker calls rule specific 'enterRuleXXX'
3° this entire sequence (steps 1 to 5) is triggered for each child rule if any
4° walker calls rule specific 'exitRuleXXX'
5° walker calls 'exitEveryRule'

Original java doc is here:
https://github.com/antlr/antlr4/blob/master/runtime/Java/src/org/antlr/v4/runtime/tree/ParseTreeWalker.java
(you might want to also update other language targets)

XsongyangX:
Thanks, I'll look into this. If I bump into problems, I'll post more here.

XsongyangX:
Alright I made a pull request

PR #2782, opened 2020-03-17
Clarify and expand documentation for runtime ParseTreeWalker

XsongyangX:
My pull request based on the issue #2759 and the answer given by @ericvergnaud . I applied the changes to all target languages. Hopefully, the documentation style conforms to all the target languages and is accepted by each respective compiler.

### DIFF
--- a/runtime/CSharp/runtime/CSharp/Antlr4.Runtime/Tree/ParseTreeWalker.cs
+++ b/runtime/CSharp/runtime/CSharp/Antlr4.Runtime/Tree/ParseTreeWalker.cs
@@ -9,6 +9,16 @@ namespace Antlr4.Runtime.Tree
     {
         public static readonly ParseTreeWalker Default = new ParseTreeWalker();
 
+        /// <summary>
+        /// Performs a walk on the given parse tree starting at the root and going down recursively
+	    /// with depth-first search. On each node, 
+        /// <see cref="ParseTreeWalker.EnterRule(IParseTreeListener, IRuleNode)"/> is called before
+	    /// recursively walking down into child nodes, then
+	    /// <see cref="ParseTreeWalker.ExitRule(IParseTreeListener, IRuleNode)"/>
+        /// is called after the recursive call to wind up.
+        /// </summary>
+        /// <param name="listener">The listener used by the walker to process grammar rules</param>
+        /// <param name="t">The parse tree to be walked on</param>
         public virtual void Walk(IParseTreeListener listener, IParseTree t)
         {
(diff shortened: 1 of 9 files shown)


runtime/Cpp/runtime/src/tree/ParseTreeWalker.h
```diff
@@ -15,15 +15,32 @@ namespace tree {
     static ParseTreeWalker &DEFAULT;
 
     virtual ~ParseTreeWalker();
-
+    
+    /// <summary>
+    /// Performs a walk on the given parse tree starting at the root and going down recursively
+	/// with depth-first search. On each node, <seealso cref="ParseTreeWalker#enterRule"/> is called before
```
ericvergnaud: why not use block comments?

runtime/CSharp/runtime/CSharp/Antlr4.Runtime/Tree/ParseTreeWalker.cs
```diff
@@ -9,6 +9,16 @@ public class ParseTreeWalker
     {
         public static readonly ParseTreeWalker Default = new ParseTreeWalker();
 
+        /// <summary>
+        /// Performs a walk on the given parse tree starting at the root and going down recursively
+	    /// with depth-first search. On each node, 
+        /// <see cref="ParseTreeWalker.EnterRule(IParseTreeListener, IRuleNode)"/> is called before
+	    /// recursively walking down into child nodes, then
+	    /// <see cref="ParseTreeWalker.ExitRule(IParseTreeListener, IRuleNode)"/>
```
ericvergnaud: why not use block comments?

sharwell: `///` line comments are the de-facto standard for C# code. Java-style `/** */` are technically allowed, but nearly non-existent in the wild.

XsongyangX: microsoft uses triple slashes on their [c# documentation](https://docs.microsoft.com/en-us/dotnet/csharp/language-reference/language-specification/documentation-comments#an-example) so I thought of doing the same

runtime/Cpp/runtime/src/tree/ParseTreeWalker.h
XsongyangX: ill change this into blocks

ericvergnaud:
thanks song yang
@parrt blessed

PR #2797, opened 2020-04-07
refactored AntlrInputStream.cs by removing a redundand if-statement

deniskyashif:
signed contributors.txt


runtime/CSharp/runtime/CSharp/Antlr4.Runtime/AntlrInputStream.cs
```diff
@@ -48,11 +48,7 @@ public virtual void Consume()
                 System.Diagnostics.Debug.Assert(LA(1) == IntStreamConstants.EOF);
                 throw new InvalidOperationException("cannot consume EOF");
             }
-            //System.out.println("prev p="+p+", c="+(char)data[p]);
-            if (p < n)
-            {
-                p++;
-            }
+            p++;
```
ericvergnaud, line 51: Can you make it an else ?
It would keep the logic readable at no execution cost

deniskyashif: Sure, I have updated my commit.

ericvergnaud:
@parrt blessed

PR #2805, opened 2020-04-19
C++ target: escape ? character to prevent accidental trigraphs

felixn:
If grammars contain something like ?? as a lexer rule (e.g. `CONDITION_CONVERSION: '??';`), the generated C++ code will contain something like `"'??'"` in _literalNames. This can cause compiler warnings or even unintended behaviour, because sequences starting with two ? characters are interpreted as trigraphs.

Using an override of shouldUseUnicodeEscapeForCodePointInDoubleQuotedString in CppTarget
shouldUseUnicodeEscapeForCodePointInDoubleQuotedString in Target was changed from private static to protected to allow overriding

(It's possible to avoid this issue by using C++17 (drops support for trigraphs) and `-Wno-trigraphs`, but escaping the ? character is a universal fix).

### DIFF
--- a/tool/src/org/antlr/v4/codegen/Target.java
+++ b/tool/src/org/antlr/v4/codegen/Target.java
@@ -282,7 +282,7 @@ public abstract class Target {
 		return sb.toString();
 	}
 
-	private static boolean shouldUseUnicodeEscapeForCodePointInDoubleQuotedString(int codePoint) {
+	protected boolean shouldUseUnicodeEscapeForCodePointInDoubleQuotedString(int codePoint) {
 		// We don't want anyone passing 0x0A (newline) or 0x22
 		// (double-quote) here because Java treats \\u000A as
 		// a literal newline and \\u0022 as a literal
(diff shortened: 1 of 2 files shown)


felixn:
Ping @mike-lischke

mike-lischke:
Can't you switch off trigraphs in every C++ compiler? This is a feature so rarely used that it seems unnecessary to waste CPU cycles to provide a fix in ANTLR4. But if you deem it important enough then we can take it in.

felixn:
Hi Mike,
it looks like it's only possible to disable trigraphs by selecting C++17 or higher or by selecting a "std" mode instead of a "gnu" mode. See
https://clang.llvm.org/docs/UsersManual.html#differences-between-various-standard-modes
or
https://gcc.gnu.org/onlinedocs/cpp/Initial-processing.html

Either way, my suggestion (implemented in this PR) is for ANTLR to never generate trigraphs - so we don't force the user to use certain compiler settings / disable certain warning.
@parrt - what's your take on this?

parrt:
I'll defer to @mike-lischke

mike-lischke:
A fix for a really marginal problem, but OK.

PR #2808, opened 2020-04-21
Update cpp-target.md

standbyme:
Use dot instead of -> operator


mike-lischke:
I just approved a similar patch, so this one can be closed.

PR #2809, opened 2020-04-23
[Swift] Update codegen template

martinvw:
Update Swift codegen template to resolve warning:

> ~~Variable '_localctx' was never mutated; consider changing to 'let' constant~~
> Variable '_parentState' was never mutated; consider changing to 'let' constant
> Variable '_startState' was never mutated; consider changing to 'let' constant

Since there is no path where they can be changed I would say its best to just generate without the warnings.

Edit: `_localctx` has been reversed because it is needed for a test


martinvw:
CC: @ewanmellor

martinvw:
@ewanmellor the current build-error is unrelated (PHP) so the swift tests are green.

PR #2814, opened 2020-04-28
Add s390x support for travis

srajmane:
As Travis CI [officially supports](https://blog.travis-ci.com/2019-11-12-multi-cpu-architecture-ibm-power-ibm-z) s390x builds, adding support for same.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -204,7 +204,127 @@ matrix:
       jdk: openjdk8
       env: TARGET=go
       stage: main-test
-
+    - os: linux
+      arch: s390x
+      dist: bionic
+      compiler: clang
+      jdk: openjdk11
+      env:
+        - TARGET=cpp
+        - CXX=g++
+        - GROUP=LEXER
+      stage: main-test
+      addons:
+        apt:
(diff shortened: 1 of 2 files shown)


srajmane:
@parrt Could you please have a look ?

parrt:
Hmm...I'm not well-versed in the travis config but it seems like a really big addition. Does s390x often have compatibility issues compared to other linux?  Maybe @ericvergnaud has some idea.

ericvergnaud:
I don’t have any idea to be honest
@srajmane what we worry about  is maintenance
when this starts breaking (it will we just don’t know when) who will be able to fix the build?

srajmane:
Hi @ericvergnaud Yes, we already build and maintain Anlr4 on s390x. Our build instructions are published [here](https://github.com/linux-on-ibm-z/docs/wiki/Building-ANTLR-4) .

parrt:
I'm not sure this is a huge market and I'm loath to make a major change like this to the configs. It complicates everyones job for the whole future of the tool. sorry!

PR #2820, opened 2020-05-11
fix code smells

XenoAmess:


tool/src/org/antlr/v4/tool/ErrorManager.java
```diff
@@ -193,15 +193,13 @@ private static StackTraceElement getLastNonErrorManagerCodeLocation(Throwable e)
 	public void emit(ErrorType etype, ANTLRMessage msg) {
 		switch ( etype.severity ) {
 			case WARNING_ONE_OFF:
+			case ERROR_ONE_OFF:
```
sharwell, line 196: ❗ this is not correct

XenoAmess: @sharwell Sorry, my bad.

runtime-testsuite/test/org/antlr/v4/test/runtime/csharp/BaseCSharpTest.java
```diff
@@ -245,8 +245,7 @@ public String execLexer(String grammarFileName,
 	Set<String> sourceFiles = new HashSet<String>();
 
 	private void addSourceFiles(String ... files) {
-		for(String file : files)
-			this.sourceFiles.add(file);
+		this.sourceFiles.addAll(Arrays.asList(files));
```
NathanQingyangXu, line 248: There might be better optimization here:
```
Collections.addAll(this.sourceFiles, files);
```
As per Javadoc, its function remains the same, but it is likely to run significantly faster under most implementations

XenoAmess: @NathanQingyangXu 
Hi.
After further test I notice you are right, this is a wrong optimize, it runs slightly slower in HashSet implementation.
Should revoke to original.

parrt:
Hi.  I agree with Sam that it's too big of a change. Sorry.

PR #2833, opened 2020-05-31
Avoid stepping through ANTLR generated code when debugging (C#)

d-markey:


ericvergnaud:
Hi,
thanks for this but I'm not sure everyone would want this by default, rather the opposite imho...
Eric

d-markey:
I see the point, maybe setting the attribute at class level is a too general. My use-case is debugging post-parsing code with custom listeners: I just don't want to go through each "XXXContext.Accept()" and "BaseVisitor.VisitXXX()" methods. I can revise my changes for the parser part and apply the DebuggerStepThrough attribute only for "Accept()" methods. Would that seem more legitimate?

ericvergnaud:
DebuggerNonUserCodeAttribute would be more acceptable

ericvergnaud:
@parrt blessed

PR #2839, opened 2020-06-05
[C++ runtime] Improvements of Vocabulary::fromTokenNames

IohannRabeson:
This commit doesn't change the behavior. It contains few small improvements to the class Vocabulary:
- use a reference for the variable tokenName to prevent a useless copy
- do not to check twice if tokenName is empty
- removed useless variable firstChar and use std::string::front() when possible
- use std::string::empty() instead of creating an empty string to compare with
- use std::string::clear() instead of assigning an empty string to clear a string
- use defaulted functions when possible


mike-lischke (approved):

mike-lischke:
@IohannRabeson Please pull the branch again and fix the conflict.

@parrt After the conflict has been solve this patch can be merged.

PR #2843, opened 2020-06-07
Dlang target

BayernEV:


ericvergnaud:
Hi Bayern

wow, you managed to achieve all this without support from the core team... very impressive!

I'm not the final approver here, but I can already see various improvements required before we can consider this PR:
 - the appveyor test suite fails (seems a trivial merge conflict, you might simply need to rebase)
 - there is no travis build

happy to give you some hints if needed.

BayernEV:
Hi Eric,

thanks for this quick response. I was simply not aware of the CI. Does it seem a have to look into some examples because of missing stuff in the guideline https://github.com/antlr/antlr4/blob/master/doc/creating-a-language-target.md?

Egbert

ericvergnaud:
you need to add dlang to the list of builds in .travis.yml
then there are 2 script per language target, 1 for install and 1 for testing
you could follow examples from the .travis directory

ericvergnaud:
@BayernEV FYI we had to reshape the CI
so you would need to rebase and look at the 4 (no less) CIs we are now using

PR #2847, opened 2020-06-12
[C++ runtime] Add AntlrInputStream(std::string_view) constructor

nburles:
Guarded by the `__cplusplus` macro as it is only available in C++17 or newer.

This avoids a copy if the input is actually `char const *` (due to implicit conversion to `std::string` in the pre-C++ 17 version), and performs the same if the input is actually `std::string`.


mike-lischke (approved):
I like that. We should probably migrate over to C++17 in the mid term unless someone objects.

mike-lischke:
@parrt A nice C++ addition, ready to be merged.

nburles:
Yeah, there's definitely some areas that could benefit from the new functionality in C++17.

PR #2851, opened 2020-06-18
Be more precise in documentation of char set when escaping is needed

akosthekiss:


akosthekiss:
Result of discussion with / comment from @KvanTTT at https://github.com/renatahodovan/grammarinator/pull/20#discussion_r441775176

PR #2854, opened 2020-06-22
[C++ codegen] Remove use of the u8"" string literal prefix

### LINKED ISSUE #2844: C++20 can not construct a std::string with a UTF-8 encoded string literal
Problem is in this [file](https://github.com/antlr/antlr4/blob/f6c6217804189b406a2daca718e3b4464e25e302/tool/resources/org/antlr/v4/tool/templates/codegen/Cpp/Cpp.stg)

```
std::vector\<std::string> <lexer.name>::_ruleNames = {
  <lexer.ruleNames: {r | u8"<r>"}; separator = ", ", wrap, anchor>
};
```

In C++20 this is ill formed, see http://www.open-std.org/jtc1/sc22/wg21/docs/papers/2019/p1423r2.html for more details.

nburles:
Prior to C++20, the type of a "UTF-8 encoded string literal" is `char const *` - i.e. it is exactly the same as a non-prefixed string literal (they are semantically different, but technically the same).

Since C++20, the type of a UTF-8 encoded string literal is changed to `char8_t const *` - which is not convertable to `char const *`.  Even in C++20, there is no actual change to how characters are stored, only the type is changed (they are now semantically different, and supposed to be technically different - but nothing in the language actually uses the semantics).

In short, removing the u8"" prefix has no effect prior to C++20, and simply allows compilation to succeed since C++20.

Fixes #2844

### DIFF
--- a/tool/resources/org/antlr/v4/tool/templates/codegen/Cpp/Cpp.stg
+++ b/tool/resources/org/antlr/v4/tool/templates/codegen/Cpp/Cpp.stg
@@ -192,23 +192,23 @@ atn::ATN <lexer.name>::_atn;
 std::vector\<uint16_t> <lexer.name>::_serializedATN;
 
 std::vector\<std::string> <lexer.name>::_ruleNames = {
-  <lexer.ruleNames: {r | u8"<r>"}; separator = ", ", wrap, anchor>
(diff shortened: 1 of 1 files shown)


mike-lischke (approved):
I'm fine with that. Looks like a left over to me.

mike-lischke:
@parrt Small C++ change, ready to be merged.

PR #2859, opened 2020-07-02
Java: shrink critical section in LexerATNSimulator

dhalperi:
When using the same Lexer in many threads, the LexerATNSimulator.addDFAState
can be a bottleneck. The most expensive work in this section is computing the
hashCode of the ATNConfigSet. This happens twice, once when checking if the
state is already in the map, and once when inserting it. If we make the configs
readonly before both operations, the hashCode will be reused. Second, if we
compute the hashCode outside of the critical section it will be cached,
shortening the total time in that section.

Aside: in Java8, there are functions like `putIfAbsent` that will simplify this code
and reduce the number of operations. When if at all is Antlr considering bumping
the minimal language version?

### DIFF
--- a/runtime/Java/src/org/antlr/v4/runtime/atn/LexerATNSimulator.java
+++ b/runtime/Java/src/org/antlr/v4/runtime/atn/LexerATNSimulator.java
@@ -687,17 +687,16 @@ public class LexerATNSimulator extends ATNSimulator {
 		}
 
 		DFA dfa = decisionToDFA[mode];
+		configs.setReadonly(true);
+		configs.hashCode(); // compute and cache outside of the synchronized block.
+
 		synchronized (dfa.states) {
 			DFAState existing = dfa.states.get(proposed);
 			if ( existing!=null ) return existing;
 
-			DFAState newState = proposed;
-
-			newState.stateNumber = dfa.states.size();
-			configs.setReadonly(true);
-			newState.configs = configs;
-			dfa.states.put(newState, newState);
-			return newState;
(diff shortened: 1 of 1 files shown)


dhalperi:
@parrt @sharwell - do you have any opinions here?

parrt:
unfortunately it's probably too much of a risk to change the threading mechanism at this point.  sorry to hear about the bottleneck :(

sharwell:
My fork of the project ([tunnelvisionlabs/antlr4](https://github.com/tunnelvisionlabs/antlr4)) uses a fully lock-free fast path (i.e. the case where a DFA state already exists). The change to implement that was deemed too complex for this project, since it only impacts a small number of users with very specific demands.

sharwell:
@parrt I'm hesitant to make this change here, at least for the reason indicated above. If the goal is reducing contention in highly-parallel environments, anything we do here won't address the problem like the one in my fork does. The code paths for DFA state updates are completely different.

dhalperi:
I agree that the fork likely has a better fundamental approach, but it doesn't quite follow that changes to improve the performance of this implementation are therefore useless.

Also strongly respect risk aversion. What about just moving the `readonly` setting up, so that the hashCode is only computed once. Is that kind of thing considered low enough risk?

parrt:
unfortunately, the reality is my brain is dedicated elsewhere and it would require swapping this project back into my head and then thinking hard about threading. Sorry!  Sam's fork is extremely performant and works the same so maybe try that out.

PR #2864, opened 2020-07-12
Fix a NPE

ericvergnaud:
@parrt 
bumped into this bug with the JS runtime, caused I suspect by a bug which may affect all targets.

The root cause would come from PredictionContext.merge, lines 169 and 172, where an ArrayPredictionContext is created from a SingletonPredictionContext, which is some cases is an EmptyPredictionContext, with a null parent.
This translates into ArrayPredictionContext.parents being initialized to { null }, which is a nasty code smell...

I'll happily fix it if you think it makes sense, just let me know.

### DIFF
--- a/runtime/JavaScript/src/antlr4/Utils.js
+++ b/runtime/JavaScript/src/antlr4/Utils.js
@@ -66,11 +66,11 @@ String.prototype.hashCode = function () {
 };
 
 function standardEqualsFunction(a, b) {
-    return a.equals(b);
+    return a ? a.equals(b) : a==b;
 }
 
 function standardHashCodeFunction(a) {
-    return a.hashCode();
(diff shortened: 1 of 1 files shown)


ericvergnaud:
Forgot to mention that the JS bug only appears starting with 4.8, thanks to a dramatic improvement by @lmy269 in PredictionContext. combineCommonParents, which on significant inputs (like 500 lines of code) improves performance by 40% <- yes

parrt:
Wow. Good catch. I'm *really* nervous about touching this stuff in Java target after so many years though..

ericvergnaud:
I'll give a try to see how it affects tests and performance, just don't accept the PR

parrt:
Mostly concerned with breaking something! nervous about changing that stuff haha

PR #2866, opened 2020-07-16
Added version constants to Python runtimes

### LINKED ISSUE #2775: Add version constant to python runtime
It would be nice if you could get the version of the python runtime from a constant in the base package. May python packages uses `__version__` for this.

Currently, the version is embedded in Recognizer.checkVersion and is not easily accessible except through some sort of hack involving capturing stdout.

omry:
Fixes #2775

Also updated Python version in setup.py and readme.

### DIFF
--- a/contributors.txt
+++ b/contributors.txt
@@ -249,3 +249,4 @@ YYYY/MM/DD, github id, Full name, email
 2020/04/30, TristonianJones, Tristan Swadell, tswadell@google.com
 2020/05/31, d-markey, David Markey, dmarkey@free.fr
 2020/07/01, sha-N, Shan M Mathews, admin@bluestarqatar.com
+2020/07/16, omry, Omry Yadan, omry@yadan.net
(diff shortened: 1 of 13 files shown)


ericvergnaud:
Hi,

thanks for this.
I'm supportive of the requirement, maybe not so much of the proposed implementation, which requires keeping the version in sync in 2 places.
Any chance the version in setup.py could be used to populate __version__ ?
(that's the only one terence will ever update)
Would also need this for Python 2

analog-cbarber:
Yes, I think it should be implemented in such a way that the official version lives in only one place. Typically in Python projects, you have setup.py get the version from the code or a file, not the other way around because the setup.py file will not be included in the distributed package.

Typically, I just put the version in a file named VERSION in the shipped package where it can be read either by setup.py or by the code, but there are other ways to do it as well.

omry:
Sure, I can make setup.py take the version from runtime/Python2/src/antlr4/version.py.

omry:
Looks like I do not have the time to do that, so it will be great if someone picks it up.
otherwise, feel free to close.

PR #2883, opened 2020-08-18
Fix travis ci for Dart pin version

lingyv-li:


lingyv-li:
The Travis CI job failed but it doesn't block PR merge. Maybe the Travis CI integration is broken? @parrt

PR #2892, opened 2020-08-31
[C++ runtime] Remove codecvt header (deprecated in C++17) in cpp runtime.

Clcanny:
1. I would like to remove codecvt header for two reasons:
    1.1. codecvt header is deprecated in C++17.
    1.2. It stop me to use gcc 4.9 to compile file.
2. And I would like to let cpp runtime support gcc 4.9.


Clcanny:
@mike-lischke Would you take some times to review this change? Because Most changes are in runtime/Cpp/runtime/support/Strings.cpp. Thanks you very much! (I am not native English speaker, please ignore my bad english.)

runtime/Cpp/runtime/src/support/StringUtils.h
```diff
@@ -6,44 +6,33 @@
 #pragma once
 
 #include "antlr4-common.h"
+#include "utf8.h"
```
mike-lischke, line 9: What lib is that? Is it available on Windows and macOS?

mike-lischke (commented):
Note: we are not striving to make the build compatible with older cmake or C++ versions. Instead our goal is to use newer features if we can to drive adoption of new C++ versions. So, adding support for older cmake versions is counter productive and I will accept that only if there's strong reason, supported by multiple community members.

Additionally: pls ensure your patch works on all normally supported platforms and compilers (i.e. gcc, clang, XCode clang/llvm, Microsoft Visual C++). Not all of them use cmake for building!

runtime/Cpp/runtime/src/support/StringUtils.h
Clcanny: Thanks a lot for you reviewing!

1. According to https://en.cppreference.com/w/cpp/locale/wstring_convert, std::wstring_convert is deprecated in C++17. More details are on https://dbj.org/c17-codecvt-deprecated-panic/ and http://www.open-std.org/jtc1/sc22/wg21/docs/papers/2017/p0618r0.html. So I think it is time to remove codecvt dependency even if we want to move on C++20.
2. utf8 is a portable header-only (only according my usage experience) library avaible at https://github.com/nemtrif/utfcpp. According to the introduction, author of utf8 says: "The library has been used a lot in the past ten years both in commercial and open-source projects and is considered feature-complete now". So I think utf8 is a good substitution for codecvt.
3. Thanks for you remind.
  3.1. Do you think it is a good idea to import utf8 headers directly if license supports?
  3.2. And I will make this change work on Microsoft Visual C++ and other platforms if this change has the opportunity 
         to be merged.

mike-lischke: >     1. According to https://en.cppreference.com/w/cpp/locale/wstring_convert, std::wstring_convert is deprecated in C++17. More details are on https://dbj.org/c17-codecvt-deprecated-panic/ and http://www.open-std.org/jtc1/sc22/wg21/docs/papers/2017/p0618r0.html. So I think it is time to remove codecvt dependency even if we want to move on C++20.

Right, this is a pity. I guess going ICU is then the best choice.

>     2. utf8 is a portable header-only (only according my usage experience) library avaible at https://github.com/nemtrif/utfcpp. According to the introduction, author of utf8 says: "The library has been used a lot in the past ten years both in commercial and open-source projects and is considered feature-complete now". So I think utf8 is a good substitution for codecvt.

The problem is that in big companies every 3rd party lib has to go through a review and approval process, which can be pretty lengthy. So people prefer not to pull in a 4th party dependency, if possible. And if it really necessary to pull in such a dep then better use something that is known to be used in the industry (like IBM's ICU). On the other hand that's a pretty big extra chunk, just for that small task, which not everybody will like. So that's why I'm a bit iffy about that.

>     3. Thanks for you remind.
>        
>        1. Do you think it is a good idea to import utf8 headers directly if license supports?

I guess that's unavoidable, otherwise you cannot compile. Maybe we should add a switch and keep the current codecvt solution as long as it compiles. For those who don't want it, they can switch to utf8.h.

>        2. And I will make this change work on Microsoft Visual C++ and other platforms if this change has the opportunity to be merged.

If all platforms compile the code with the same tools as before then I see no reason it shouldn't be merged.

mike-lischke (approved):
Perfect, thanks.

mike-lischke:
@parrt Here's another C++ patch to merge.

Clcanny:
> @parrt Here's another C++ patch to merge.

Thank you very much! Actually, I still have some work to do about MacOS platform and Windows platform. Please wait for me for 1 some time. I will @ you if this change is ready. Thanks!

Clcanny:
@mike-lischke I have test this change on macOS and Linux, but I haven't test it on Windows.
I met some trouble when trying to build antlr4 runtime on Windows, VS2017 complains wrong version of Windows SDK.
So I don't test it on Windows. Is it okay? Will CI progress try to build antlr4 runtime on Windows, macOS and Linux platform?

parrt:
okay, I will wait to hear from you about merging.

Clcanny:
@parrt I think it is okay to merge this change. Thanks!

Issue #2898, opened 2020-09-06
[C++] Runtime compile error with clang.


ArthurSonzogni:
See the following execution on a cloned ANTLR repository.
[![asciicast](https://asciinema.org/a/358080.svg)](https://asciinema.org/a/358080)
I am using clang 10.

I will take a look today to understand the error. Maybe some CMake configuration went wrong?

The CMake-generate command is:
```bash
cd /tmp/antlr4/runtime/Cpp/build/runtime && /usr/bin/c++  -I/tmp/antlr4/runtime/Cpp/runtime/src -I/tmp/antlr4/runtime/Cpp/runtime/src/atn -I/tmp/antlr4/runtime/Cpp/runtime/src/dfa -I/tmp/antlr4/runtime/Cpp/runtime/src/misc -I/tmp/antlr4/runtime/Cpp/runtime/src/support -I/tmp/antlr4/runtime/Cpp/runtime/src/tree -I/tmp/antlr4/runtime/Cpp/runtime/src/tree/pattern -I/tmp/antlr4/runtime/Cpp/runtime/src/tree/xpath -Wall -pedantic -W -stdlib=libc++ -O3 -DNDEBUG -O3 -DNDEBUG  -Wno-overloaded-virtual -Wno-dollar-in-identifier-extension -Wno-four-char-constants  -std=gnu++11 -o CMakeFiles/antlr4_static.dir/src/ANTLRErrorListener.cpp.o -c /tmp/antlr4/runtime/Cpp/runtime/src/ANTLRErrorListener.cpp
```

ArthurSonzogni:
This was quick:

We need not to add the "-stdlib=libc++" flag"
My problem can be solved by disabling the "WITH_LIBCXX" option (enabled by default)

I am wondering what this is about. I will git blame to understand.

ArthurSonzogni:
The option: "WITH_LIBCXX" was introduced by:
https://github.com/antlr/antlr4/commit/d46ef90aa03066300752895d57fc57a1c5aa8ee6

I believe its default value should have been "Off".
It will not append "--stdlib=libc++" by default and won't cause a compilation error on clang users who do not have libc++ installed. Clang will just use the default library.

I will prepare a PR.

Moreover, I am seeing many cmake warning. I will likely make another PR for this too.

PR #2899, opened 2020-09-06
Fix clang compile error with --stdlib=libc++

ArthurSonzogni:
On linux. Most users haven't installed libc++-dev, but ANTLR is passing
the "--stdlib=libc++" argument. As a result, this won't compile. Users
will see:
```
/tmp/antlr4/runtime/Cpp/runtime/src/antlr4-common.h:8:10: fatal error:
'algorithm' file not found
```

This is caused by the "WITH_LIBCXX" option.
It was introduced by:
https://github.com/antlr/antlr4/commit/d46ef90aa03066300752895d57fc57a1c5aa8ee6

It causes the option "--stdlib=libc++" to be appended by default.

I believe its default value should have been left as "Off".

With "off" by default, clang will use its default C++ library, which is
always available.

The WITH_LIBCXX option is kept, being able to change the C++ library
might be useful?

BUG=https://github.com/antlr/antlr4/issues/2898

### DIFF
--- a/contributors.txt
+++ b/contributors.txt
@@ -251,3 +251,4 @@ YYYY/MM/DD, github id, Full name, email
 2020/05/25, graknol, Sindre van der Linden, graknol@gmail.com
 2020/05/31, d-markey, David Markey, dmarkey@free.fr
 2020/07/01, sha-N, Shan M Mathews, admin@bluestarqatar.com
+2020/09/06, ArthurSonzogni, Sonzogni Arthur, arthursonzogni@gmail.com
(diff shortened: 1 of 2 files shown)


ArthurSonzogni:
Hi everyone,

Could you please take a look?
@mike-lischke FYI (original author of the WITH_LIBCXX option)

mike-lischke (approved):

mike-lischke:
@parrt This is a small C++ build patch and can be merged.

PR #2900, opened 2020-09-06
Fix one warning on non-MSVC compilers.

ArthurSonzogni:
Some Microsoft's specific #pragma were left in the code. They causes the
following errors to be printed:
```
/tmp/antlr4/runtime/Cpp/runtime/src/atn/ParserATNSimulator.cpp:1351:9:
warning: unknown pragma ignored [-Wunknown-pragmas]
may be unsafe. Consider using _dupenv_s instead.
        ^
/tmp/antlr4/runtime/Cpp/runtime/src/atn/ParserATNSimulator.cpp:1361:9:
warning: unknown pragma ignored [-Wunknown-pragmas]
```

This patch removes the #pragma macros and uses the matching compiler
definition only on MSVC compiler.

### DIFF
--- a/runtime/Cpp/CMakeLists.txt
+++ b/runtime/Cpp/CMakeLists.txt
@@ -66,7 +66,7 @@ if(WITH_DEMO)
 endif(WITH_DEMO)
 
 if(MSVC_VERSION)
-  set(MY_CXX_WARNING_FLAGS "  /W4")
+  set(MY_CXX_WARNING_FLAGS "  /W4 /wd4996")
 else()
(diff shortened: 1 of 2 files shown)


ArthurSonzogni:
Note: The "contributors.txt" will be updated first in: https://github.com/antlr/antlr4/pull/2899. So I won't do it here.

mike-lischke:
Not all users build the runtime using cmake. So your change will leave them then with new warnings.

ArthurSonzogni:
> Not all users build the runtime using cmake. So your change will leave them then with new warnings.

Indeed! Thanks!
I haven't thought about this requirement.

PR #2904, opened 2020-09-10
Update getting started guide to use antlr4.8

Khalian:
Updating getting started to use antlr4.8 complete instead of 4.7.1 complete.


ericvergnaud:
@parrt need to change to 4.9

parrt:
@Khalian Could you make that 4.9?  we are going to be releasing soon. thanks!

PR #2905, opened 2020-09-13
Fix #2902

### LINKED ISSUE #2902: [JavaScript runtime] Implementation of ArrayPredictionContext.prototype.equals is incorrect
I checked that the JavaScript runtime is very slow compared to the Java runtime.
And I noticed that `ArrayPredictionContext.prototype.equals` rarely returns `true`.

Looking at the Java and JavaScript implementations, the Java implementation is comparing the array elements correctly, while the JavaScript implementation is comparing the array instances.

https://github.com/antlr/antlr4/blob/a4c66dc863bb6d5de1daf2ecac59bf5342d8dd98/runtime/JavaScript/src/antlr4/PredictionContext.js#L252-L253
https://github.com/antlr/antlr4/blob/be881fa6b91d1980936f8dcab902a9dc26ecd310/runtime/Java/src/org/antlr/v4/runtime/atn/ArrayPredictionContext.java#L76-L78

I think perhaps JS needs to be modified as follows:

```js
    return equalArrays(this.returnStates, other.returnStates) &&
        equalArrays(this.parents, other.parents);
```
We can also use `equalArrays` in `Utils.js`, but we also need to change `equalArrays` in `Utils.js` because `equalArrays` in `Utils.js` doesn't allow arrays of primitives.

ericvergnaud:
Fixes a naive array comparison which improves performance dramatically

### DIFF
--- a/runtime/JavaScript/src/antlr4/PredictionContext.js
+++ b/runtime/JavaScript/src/antlr4/PredictionContext.js
@@ -4,7 +4,7 @@
  */
 
 const RuleContext = require('./RuleContext');
-const {Hash, Map} = require('./Utils');
+const {Hash, Map, equalArrays} = require('./Utils');
 
 class PredictionContext {
 	constructor(cachedHashCode) {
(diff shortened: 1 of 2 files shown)


ericvergnaud:
@parrt blessed

PR #2906, opened 2020-09-13
Fix template issue

### LINKED ISSUE #2745: Python3 parser generation incorrect by ANTLR 4.7.1 and problem in Parser.py
ANTLR 4.7.1 is generating incorrect Python3 code for the grammer :
```
var_defs
locals [ varlist = [] ]
   : 'var' var_def ( ',' var_def )* ';' {print($varlist)}
   ;

var_def
   : v=VARIABLE ':' var_type {$var_defs::varlist.append($v.text)} 
   ;
```
The code generated for the parser is : 
```
...
self.state = 45
self.var_type()
getInvokingContext(1).varlist.append((None if localctx.v is None else localctx.v.text))
...
```
instead of : 
```
...
self.state = 45
self.var_type()
self.getInvokingContext(1).varlist.append((None if localctx.v is None else localctx.v.text))
...
```
When the getInvokingContext function is called then the following error pops up in Parser.py: ctx.ruleIndex is not defined.
```
    def getInvokingContext(self, ruleIndex:int):
        ctx = self._ctx
        while ctx is not None:
>>>>>> if ctx.ruleIndex == ruleIndex:
                return ctx
            ctx = ctx.parentCtx
        return None
```
I had to change it to:
```
    def getInvokingContext(self, ruleIndex:int):
        ctx = self._ctx
        while ctx is not None:
>>>>>> if ctx.getRuleIndex() == ruleIndex:
                return ctx
            ctx = ctx.parentCtx
        return None
```
I am using Python 3.8.1, vs code ANTLR4 grammer plugin. with ANTLR 4.7.1

ericvergnaud:
Fix #2745 (nb would be great to have a test that reproduces the issue)

### DIFF
--- a/runtime/JavaScript/src/antlr4/PredictionContext.js
+++ b/runtime/JavaScript/src/antlr4/PredictionContext.js
@@ -4,7 +4,7 @@
  */
 
 const RuleContext = require('./RuleContext');
-const {Hash, Map} = require('./Utils');
+const {Hash, Map, equalArrays} = require('./Utils');
 
 class PredictionContext {
 	constructor(cachedHashCode) {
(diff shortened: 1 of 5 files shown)


ericvergnaud:
The test would be a small grammar that triggers NonLocalAttrRef and SetNonLocalAttr string template atoms

ericvergnaud:
@parrt blessed

PR #2806, opened 2020-04-19
C++ target: don't generate extra/duplicate semicolons

felixn:
Currently, the generated C++ source code contains sporadic duplicate semicolons, e.g.:
```
Parser::FactorContext *factorContext = nullptr;;
std::vector<FactorContext *> f;;
```
This requires the use of `-Wno-extra-semi` with `-Werror`.

This PR removes the duplicate semicolons.
(Full disclosure: I couldn't figure out where the second semicolon is generated...)

PR #2907, opened 2020-09-13
Fix regressions in travis Cpp tests

felixn:
The testcase org.antlr.v4.test.runtime.cpp.TestParserExec was failing after #2806 removed duplicate semicolons from the declarations in generated Cpp header files.

The bug was triggered by TestParserExec because it includes a grammar with alternative labels:
```Antlr
grammar T;
start : a* EOF;
a
  : label=subrule {std::cout << $label.text << std::endl;} #One
  | label='y' {std::cout << $label.text << std::endl;} #Two
  ;
subrule : 'x';
WS : (' '|'\n') -> skip ;
```
Grammars without alternative labels (most?) were not affected.

The templates for StructDeclHeader and AltLabelStructDeclHeader differed in whether they appended semicolons to attributes:
StructDeclHeader: `<attrs: {a | <a>;}; separator="\n">`
AltLabelStructDeclHeader: git `<attrs: {a | <a>}; separator = "\n">`


felixn:
@mike-lischke @ericvergnaud @parrt 
This fixes the travis build (for now), ready to merge!

(#2908 will deal with further build regressions when newer gcc/clang versions are used)

Issue #802, opened 2015-01-23
enterXXX for alt label never gets called


martint:
The test case below prints:

```
exiting alt1
```

but I'd expect it to print:

```
entering alt1
exiting alt1
```

```
grammar Test;

a
    : 'b'   #alt1
    | 'c'   #alt2
    ;
```

```java
import org.antlr.v4.runtime.ANTLRInputStream;
import org.antlr.v4.runtime.CommonTokenStream;
import org.antlr.v4.runtime.misc.NotNull;

public class Main
{
    public static void main(String[] args)
    {
        TestParser parser = new TestParser(new CommonTokenStream(new TestLexer(new ANTLRInputStream("b"))));

        parser.addParseListener(new TestBaseListener() {
            @Override
            public void enterAlt1(@NotNull TestParser.Alt1Context ctx)
            {
                System.out.println("entering alt1");
            }

            @Override
            public void exitAlt1(@NotNull TestParser.Alt1Context ctx)
            {
                System.out.println("exiting alt1");
            }
        });

        parser.a();
    }
}
```

sharwell:
The [documentation for `addParseListener`](http://www.antlr.org/api/Java/org/antlr/v4/runtime/Parser.html#addParseListener(org.antlr.v4.runtime.tree.ParseTreeListener)) *explicitly* states that calls to `enterRule` methods may be omitted. You should read this documentation very carefully because the method certainly does not do what most new users would expect. Whenever possible, you should create your parse tree first and then use a listener or visitor after the parse is complete, since this will provide the expected deterministic results.

That said, it *might* be possible to include the call you were expecting by generating a call to the enter method after the outermost decision is made for a rule containing labeled alternatives.

parrt:
interesting. I would've expected the parser listener to drop enter calls only four left recursive rules. I will look into this.

parrt:
Crap. reopening. harder than I thought to do it properly. Waiting until I can *really* fix for interp as well before incorporating.

PR #2910, opened 2020-09-14
Fix enterXXX for alternative labels

3c1u:
Moved `[tT]riggerEnterRuleEvent` from `enterRule` to `enterOuterAlt`. This will fix antlr/antlr4#802 for targets except Swift (of which target already has `triggerEnterRuleEvent` in `enterOuterAlt`).


ericvergnaud:
Hi, thanks for this, but it can potentially break a lot of code
@parrt any view?

parrt:
Hi, yeah, I'm too afraid to change the listening mechanism. Sorry! Thanks for trying to help though!

PR #2911, opened 2020-09-14
separate py3 runtime tests from src tree + actually run said tests in CI

### LINKED ISSUE #2909: [Python3] Large import time cost paid for inline unittests
Hey,

The python3 antlr runtime spend a large proportion of its import load time importing `unittest`.
This can profiled with pythons newish `-X importtime` option.
[importtime.txt](https://github.com/antlr/antlr4/files/5215411/importtime.txt)

My on system it takes ~150ms to `import antlr4` and ~100ms of that time is spent importing `unittest`.
This seems to be like quite a high time cost to pay just the keep the tests inline with the source.

My current project involing creating a python command line tool, which utilises antlr, and I would ideally like to keep the startup time as low as possible.

Would you consider moving the unitests out of the source tree?

Cheers,
Rob

rmcgregor1990:
Move py3 runtime tests out of the src files into the `test` tree. 
This removes `unittest` from the runtime imports, which speeds up import time significantly.
Included the newly seperated tests to actually be run as part of the testsuite, as they weren't previously.

Closes #2909


ericvergnaud:
@parrt blessed, does not affect the actual runtime

parrt:
@rmcgregor1990  could you sign the contributors file?  thanks

rmcgregor1990:
Done, Cheers.

PR #2933, opened 2020-10-04
Correct the csharp-target example for invoking a custom listener

RexJaeschke:
Correct the spelling of 2 members: `Default` and `Walk`


ericvergnaud:
@parrt blessed

PR #2934, opened 2020-10-08
Made C# Trace Listener usable for most cases

Marti2203:
Previously, when using the internal constructor, the trace listener had no text stream and it threw a null pointer exception. I have chosen the default stream to be Console.Out If somebody wants to create an instance of Trace Listener they will have to include a Parser and a Text Writer such that no such errors are thrown.


ericvergnaud:
Thanks for this
@parrt blessed

PR #2942, opened 2020-10-16
Corrected Webpack configuration in JavaScript documentation

adarshbhat:
When using the JavaScript code generated by antlr4 in a web browser environment, we have to make exceptions for certain node.js based global modules. The configuration provided in documentation is incorrect and no longer works on new versions of Webpack. This PR attempts to correct the documentation with new configuration that works. Thanks to the Webpack team for pointing out the correct approach.

My understanding is that even after switching to ES6, we will still need this configuration.

### DIFF
--- a/doc/javascript-target.md
+++ b/doc/javascript-target.md
@@ -68,7 +68,7 @@ The steps to create your parsing code are the following:
  
 You are now ready to bundle your parsing code as follows:
  - following webpack specs, create a webpack.config file
- - in the `webpack.config` file, exclude node.js only modules using: `node: { module: "empty", net: "empty", fs: "empty" }`
+ - in the `webpack.config` file, exclude node.js only modules using: `resolve: { fallback: { fs: false } }`
  - from the cmd line, navigate to the directory containing webpack.config and type: webpack
  
 This will produce a single js file containing all your parsing code. Easy to include in your web pages!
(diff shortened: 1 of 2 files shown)


doc/javascript-target.md
```diff
@@ -68,7 +68,7 @@ The steps to create your parsing code are the following:
  
 You are now ready to bundle your parsing code as follows:
  - following webpack specs, create a webpack.config file
- - in the `webpack.config` file, exclude node.js only modules using: `node: { module: "empty", net: "empty", fs: "empty" }`
+ - in the `webpack.config` file, exclude node.js only modules using: `resolve: { fallback: { fs: false } }`
```
ericvergnaud, line 71: Not everybody is keen to move to latest webpack, so suggest we say l71 is for webpack vn whilst l72 is for webpack vn+1

adarshbhat: Let me check whether or not the configuration works in the previous version of Webpack. Based on other people's conversations I found on Google, it doesn't work there either. It probably worked 2 major versions ago. But let me confirm.

adarshbhat: You are right. Someone else confirmed that the configuration fails only in the latest version of webpack. I will update this PR to reflect that.

ericvergnaud:
@parrt blessed

PR #2956, opened 2020-11-09
fix bug for javascript runtime: 'getRuleContext()' not exist.

xiaocj:
JavaScript runtime for ANTLR4 failed to run because  'getRuleContext()' property is not exist for 'TerminalNodeImpl'.

### DIFF
--- a/runtime/JavaScript/src/antlr4/tree/Tree.js
+++ b/runtime/JavaScript/src/antlr4/tree/Tree.js
@@ -195,9 +195,11 @@ class ParseTreeWalker {
 	 * @param r The grammar rule containing the rule context
 	 */
 	enterRule(listener, r) {
-		const ctx = r.getRuleContext();
-		listener.enterEveryRule(ctx);
-		ctx.enterRule(listener);
+		if(r.getRuleContext) {
+			const ctx = r.getRuleContext();
+			listener.enterEveryRule(ctx);
+			ctx.enterRule(listener);
+		}
 	}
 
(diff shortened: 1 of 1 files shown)


ericvergnaud:
Hi,

thanks for this, however I very much doubt that this is a genuine bug.
Rather, may I suggest you are misusing ParseTreeWalker ?

xiaocj:
@ericvergnaud the bug is triggered in a very basic scenario like below:
`
        const chars = new antlr4.InputStream(input);
        const lexer = new wslex.WsLang2Lexer(chars);
        const tokens  = new antlr4.CommonTokenStream(lexer);
        const parser: wsp.WsLang2Parser = new wsp.WsLang2Parser(tokens);
        parser.buildParseTrees = true;

        const listener: WsListener = new WsListener();
        const tree = parser.exp();
        antlr4.tree.ParseTreeWalker.DEFAULT.walk(listener, tree);
`

ericvergnaud:
@xiaocj There are literally thousands of developers out there using the runtime the same way with no issue...

ParseTreeWalker.walk checks for TerminalNode and does not call enterRule/exitRule for them:
`if (t instanceof TerminalNode) {
		listener.visitTerminal(t);
	}`

It seems you are using the TypeScript runtime (which we do not support).
So I suspect the issue could lie in the compilation, with TerminalNodeImpl losing its inheritance somehow?

Or it could be your listener doing some uncontrolled walking?

Issue #2968, opened 2020-11-25
antlr4 4.9.0 - ERR! Error: Cannot find module 'antlr4/index'


jantimon:
antlr4 4.9.0 is causing the following error in my webpack build: `ERR! Error: Cannot find module 'antlr4/index'`

maybe its a duplicate of #2965 but unfortunately I don't speak chinese..

In 4.8.0 the index file was directly at root:
https://unpkg.com/antlr4@4.8.0/index.js

In 4.9.0 the index file is now in the src folder:
https://unpkg.com/antlr4@4.9.0/index.js

That's a breaking change and according to the npm guidlines it must be released as 5.0 instead of 4.9 - otherwise many packages are broken

yarn & npm follow semantic versioning:

> 
> Given a version number MAJOR.MINOR.PATCH, increment the:
> 
> MAJOR version when you make incompatible API changes,
> MINOR version when you add functionality in a backwards compatible manner, and
> PATCH version when you make backwards compatible bug fixes.
> Additional labels for pre-release and build metadata are available as extensions to the MAJOR.MINOR.PATCH format.

all packages which use `"antlr4": "^4.8.0"` as a dependency will automatically install 4.9.0 and break

It breaks my dependencies because they are importing antlr4 like this: `var antlr4 = require('antlr4/index');` 

Broken Packages:

- https://github.com/adobe/htlengine/blob/ffcfeecc1d60e0f8f47c15be68ce794a8620b6d6/src/parser/generated/SightlyLexer.js#L3
- https://github.com/asbjornh/json-parser/blob/2665a97823a9d8e3f67d372d42f021b58753eb97/lib/jsonParser.js#L3
- https://github.com/RicardoGeek/lua-antlr/blob/a9a687b2150ffec3e855ee2f1471d47b17d81cf2/LuaParser.js#L3
- https://github.com/tagomaru/truffle-sca2t/blob/531d91074359f52e24d8f66b911a47d758b6a28f/packages/truffle-sca2t-eventgen/lib/SolidityParser.js#L3
- https://github.com/asbjornh/math-parser/blob/017d75a9d827fb1cd209330d456fa0b02df5b662/lib/mathParser.js#L3
- and [many many more](https://github.com/search?l=JavaScript&q=%22antlr4%2Findex%22&type=Code)

Can you please revert that change or place an additional index.js in the root with the following content: `module.exports = require('./src/index')`? I added a pr for that here: #2969

cmoesel:
[...]

PR #2969, opened 2020-11-25
add index.js back to root

jantimon:
this change intends to revert a breaking change in 4.9.0 as described in https://github.com/antlr/antlr4/issues/2968

### DIFF
--- a/runtime/JavaScript/index.js
+++ b/runtime/JavaScript/index.js
@@ -1,0 +1,1 @@
+module.exports = require('./src/antlr4/index.js');


ericvergnaud:
Hi,
4.9 follows module semantics.
Replace your code with:
` import antlr4 from 'antlr4';`

jantimon:
@ericvergnaud that's a breaking change and must be released as 5.0 instead of 4.9 - otherwise many many packages are broken as described in #2968

jantimon:
@ericvergnaud please reopen this pull request

ericvergnaud:
it's very easy to fix so not worth rolling back

jantimon:
It is automatically breaking production builds!

GordonSmith:
@ericvergnaud the issue is in the antl4 generated code not user code - are you suggesting we patch the generated code each time the grammar changes (not an "easy fix")?

ericvergnaud:
@GordonSmith sorry I think you are incorrect, the newly generated code refers only to the module, not to files within it.

GordonSmith:
@ericvergnaud - you are correct I was generating with an older version of antlr4

PR #2972, opened 2020-11-26
tweak to avoid tests when kicking off appveyor. just trying to get a nupkg

parrt:


parrt:
Weird. appveyor not kicking off.

PR #2908, opened 2020-09-13
Travis: use newer versions

felixn:
During investigation of https://github.com/antlr/antlr4/pull/2806#issuecomment-691701866, I noticed that the Cpp runtime tests were failing locally (with clang 10) with the following errors:
```
In file included from /tmp/BaseCppTest-main-1600443341009/Test.cpp:3:
In file included from /home/felix/src/antlr4/runtime-testsuite/target/classes/Cpp/runtime/src/antlr4-runtime.h:126:
/home/felix/src/antlr4/runtime-testsuite/target/classes/Cpp/runtime/src/misc/InterpreterDataReader.h:20:5: error: call to implicitly-deleted default constructor of 'dfa::Vocabulary'
    InterpreterData() {}; // For invalid content.
    ^
/home/felix/src/antlr4/runtime-testsuite/target/classes/Cpp/runtime/src/Vocabulary.h:25:5: note: explicitly defaulted function was implicitly deleted here
    Vocabulary() = default;
    ^
/home/felix/src/antlr4/runtime-testsuite/target/classes/Cpp/runtime/src/Vocabulary.h:185:36: note: default constructor of 'Vocabulary' is implicitly deleted because field '_literalNames' of const-qualified type 'const std::vector<std::string>' (aka 'const vector<basic_string<char> >') would not be initialized
    std::vector<std::string> const _literalNames;
                                   ^
1 warning and 1 error generated.
In file included from /tmp/BaseCppTest-main-1600443341009/TBaseVisitor.cpp:5:
In file included from /tmp/BaseCppTest-main-1600443341009/TBaseVisitor.h:7:
In file included from /home/felix/src/antlr4/runtime-testsuite/target/classes/Cpp/runtime/src/antlr4-runtime.h:32:
In file included from /home/felix/src/antlr4/runtime-testsuite/target/classes/Cpp/runtime/src/LexerInterpreter.h:10:
/home/felix/src/antlr4/runtime-testsuite/target/classes/Cpp/runtime/src/Vocabulary.h:25:5: warning: explicitly defaulted default constructor is implicitly deleted [-Wdefaulted-function-deleted]
    Vocabulary() = default;
    ^
/home/felix/src/antlr4/runtime-testsuite/target/classes/Cpp/runtime/src/Vocabulary.h:185:36: note: default constructor of 'Vocabulary' is implicitly deleted because field '_literalNames' of const-qualified type 'const std::vector<std::string>' (aka 'const vector<basic_string<char> >') would not be initialized
[...]

PR #2983, opened 2020-11-30
travis: upgrade c++ build to focal & llvm 10

felixn:
Alternative PR to #2908, upgrading only the c++ build

The first build shows that openjdk8 is not available with Ubuntu focal:
https://travis-ci.org/github/antlr/antlr4/builds/746825671
```
The command "~/bin/install-jdk.sh --target "/home/travis/openjdk8" --workspace "/home/travis/.cache/install-jdk" --feature "8" --license "GPL" --cacerts" failed and exited with 3 during .
```

The second build shows the issues in the C++ runtime:
https://travis-ci.org/github/antlr/antlr4/builds/746909160
```
In file included from /tmp/BaseCppTest-main-1606808268300/TLexer.cpp:5:
In file included from /tmp/BaseCppTest-main-1606808268300/TLexer.h:7:
In file included from /home/travis/build/antlr/antlr4/runtime-testsuite/target/classes/Cpp/runtime/src/antlr4-runtime.h:126:
/home/travis/build/antlr/antlr4/runtime-testsuite/target/classes/Cpp/runtime/src/misc/InterpreterDataReader.h:20:5: error: call to implicitly-deleted default constructor of 'dfa::Vocabulary'
    InterpreterData() {}; // For invalid content.
    ^
/home/travis/build/antlr/antlr4/runtime-testsuite/target/classes/Cpp/runtime/src/Vocabulary.h:25:5: note: explicitly defaulted function was implicitly deleted here
    Vocabulary() = default;
    ^
/home/travis/build/antlr/antlr4/runtime-testsuite/target/classes/Cpp/runtime/src/Vocabulary.h:185:36: note: default constructor of 'Vocabulary' is implicitly deleted because field '_literalNames' of const-qualified type 'const std::vector<std::string>' (aka 'const vector<basic_string<char> >') would not be initialized
    std::vector<std::string> const _literalNames;
                                   ^
1 error generated.
```

The third build passes:
https://travis-ci.org/github/antlr/antlr4/builds/747024792?utm_source=github_status&utm_medium=notification

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -21,7 +21,7 @@ matrix:
     - os: linux
       dist: focal
       compiler: clang
-      jdk: openjdk8
+      jdk: openjdk11
       env:
         - TARGET=cpp
         - CXX=g++-10
@@ -38,7 +38,7 @@ matrix:
     - os: linux
       dist: focal
       compiler: clang
-      jdk: openjdk8
+      jdk: openjdk11
       env:
         - TARGET=cpp
         - CXX=g++-10
@@ -55,7 +55,7 @@ matrix:
     - os: linux
       dist: focal
       compiler: clang
-      jdk: openjdk8
+      jdk: openjdk11
       env:
         - TARGET=cpp
         - CXX=g++-10
--- a/.travis.yml
+++ b/.travis.yml
@@ -19,59 +19,56 @@ stages:
 matrix:
   include:
     - os: linux
-      dist: trusty
+      dist: focal
       compiler: clang
       jdk: openjdk8
       env:
         - TARGET=cpp
-        - CXX=g++-5
+        - CXX=g++-10
         - GROUP=LEXER
       stage: main-test
       addons:
         apt:
           sources:
-            - ubuntu-toolchain-r-test
-            - llvm-toolchain-precise-3.7
+            - sourceline: 'deb http://apt.llvm.org/focal/ llvm-toolchain-focal-10 main'
           packages:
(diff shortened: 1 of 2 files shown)


felixn:
This is ready to merge, from my point of view.

mike-lischke (approved):

mike-lischke:
@parrt That's a minimal change, so I believe it is safe to merge.

parrt:
Hmm... I think it would feel better if it didn't also change the Java version being used to test.

mike-lischke:
All tests passed, so what suprise do you expect by that change?

parrt:
well, we claim java 7 compatibility. Seems like we should stick to that version for tests.

felixn:
Hi Terence,

unfortunately, the current Ubuntu LTS (20.04 / focal) Travis image does not support openjkd8 - that's why the first build failed.

Also, from my point of view, it's actually a good thing if we test with some varying Java versions (C++ with Java 11, others with Java 8) to increase test coverage.

Plus, Java 8 is "almost" EOL: "Java SE 8 has gone through the End of Public Updates process for legacy releases."
https://www.oracle.com/java/technologies/java-se-support-roadmap.html

But you're the big boss ;)
If you want, I can look for a Ubuntu version that supports the old Java 8 and the newer LLVM releases.

Cheers Felix

parrt:
Well if there is no java 8 then I guess we have no choice. Ha ha. Merging. thanks for your efforts.

PR #2985, opened 2020-12-01
:ambulance: fix bad import

maxence-lefebvre:
Hi,

As of today, `InputStream` is the default export of its module and not a named one.

image

The module `CharStreams` was trying to load the named one which does not exist.

Test case 

```javascript
console.log(require('antlr4').CharStreams.fromString('test'))
```

Expected

```
InputStream {
  name: '<empty>',
  strdata: 'test',
  decodeToUnicodeCodePoints: true,
  _index: 0,
  data: [ 116, 101, 115, 116 ],
  _size: 4
}
```

Actual

```
TypeError: InputStream is not a constructor
    at Object.fromString (node_modules\antlr4\src\antlr4\CharStreams.js:19:12)
```


ericvergnaud:
Would you mind adding a test? It's a pity we didn't have one to detect that break...

ericvergnaud:
Thanks!

Issue #1741, opened 2017-03-09
JavaScript target cannot use rule exceptions


ChristophWedler:
The `try` block around the rule body is not properly closed by `}` when using rule exceptions:

```antlr
grammar noRuleExceptions;
options { language = JavaScript; }

start
    :
        'start'
    ;
    catch [re] {
        throw re;
    }
```

In `antlr4/tool/resources/org/antlr/v4/tool/templates/codegen/JavaScript/JavaScript.stg`,
the `}` must move from _after_ to _before_ the following block (starting in line 269),
as it is done for the Java target:

```
    <if(exceptions)>
    <exceptions; separator="\n">
    <else>
```

ericvergnaud:
Hi,

thanks for this.
can you submit a PR so we can understand how this would behave?

Eric

ChristophWedler:
I have understood that I also need to sign contributors.txt - still need some days to get an official OK from the company I work at.

ericvergnaud:
Closing as already fixed

PR #2995, opened 2020-12-06
Fixed exception clauses in JS template.

tehbone:
The generated exception clause block was missing a leading '}' to terminate the try block.

This addresses Issue #1741.


ericvergnaud:
Thanks for this!

Issue #2982, opened 2020-11-30
JS Runtime Target for Antlr 4.9 runs TOO SLOW


yulingtianxia:
I'm generating JS runtime target with [my Objc grammer files](https://github.com/dart-native/codegen/tree/feature/antlr4.9/parser/objc) using antlr 4.9. On antlr 4.8, it works well. But when I update to 4.9, It runs too slow for a 80 lines header file.(It didn't even end for hours)

ericvergnaud:
There has been no change in the algorithm, so something else must be going on.

ericvergnaud:
@yulingtianxia can you confirm that 4.9.1 fixes the issue and close?

yulingtianxia:
4.9.1 fixed.

PR #2996, opened 2020-12-06
rollback partially incorrect fix for #2902

ericvergnaud:


ericvergnaud:
Maybe a fix for #2982

PR #2999, opened 2020-12-09
Clarify why mismatched input occurred by showing recognized token type

wbbradley:
Users are encountering the mismatched input error and this is by design. However, the error message could be constructed to simplify the user experience. This PR suggests a fix in the Go runtime to always show the token type.

Example issue that folks are encountering: https://stackoverflow.com/questions/16364046/antlr4-mismatched-input


wbbradley:
(Haven't had a chance to look at this error. Will be back at some point to do so.)

PR #3017, opened 2020-12-29
Improve python3 performance by adding __slots__

amykyta3:
# Summary
Python isn't exactly known for its performance, but this PR is my attempt at squeezing out a few more percent of speedup from the Python3 runtime.
This change adds the `__slots__` variable to most classes in the Antlr runtime in order to explicitly declare class members and avoid the use of the slower `__dict__`-based member storage mechanism. This results in faster Python class instance creation, as well as member access.
A benchmark I ran showed roughly 10% improvement in performance.

# Compatibility considerations
For normal "sane" uses of the Antlr runtime, this change is backwards compatible with the current runtime API and should have no effect on behavior.
The side effect of `__slots__` usage is that one is no longer able to dynamically assign new variables to class instances. There are also some caveats with Python weakref usage. I can't quite imagine why someone may need to stuff their own variables on-top of runtime classes, but then again, you never know what other users will do :smile:.

This limitation can be removed by adding `__dict__` and `__weakref__` back into the `__slots__` definition. (possibly at a small performance cost)
The current implementation omits this for now.

See python docs for more details: https://docs.python.org/3/reference/datamodel.html#slots

# Other changes made
The following other changes were made to correct issues and allow `__slots__` to be used:
* In `ATN.py`, fixed capitalization of an assignment to `<IntervalSet>.readOnly`
  * The class `IntervalSet` initializes a member `readOnly`, but this was being assigned as `readonly` in `ATN.py`
  * Looks like a typo.
* In `LexerATNSimulator.py`, removed the `match_calls` counter.
  * This was initialized as a class variable but incremented in an instance variable which is incompatible with `__slots__`.
  * Looks like this was used while debugging at one point and not actually necessary, as it is not referenced anywhere.

# Benchmark
I ran some comparison benchmarks on some sample code from one of my projects. Also including stats using the [speedy-antlr-tool](https://speedy-antlr-tool.readthedocs.io/en/latest/) accelerator.
Inputs are using the lexer/parser for [SystemRDL](https://github.com/SystemRDL/systemrdl-compiler) as well as one of the testcase files. Measured time elapsed for 1000 iterations via timeit.

| Benchmark              | seconds |
|------------------------|---------|
| Baseline               | 134.42  |
| Baseline + slots       | 121.65  |
| "speedy-antlr"         | 14.58   |
| "speedy-antlr" + slots | 13.47   |

It isn't much, but the addition of `__slots__` improves performance by roughly 10% in both the baseline Python3 parser, as well as the accelerated version. I expect this also depends somewhat on the grammar used.

# Other notes
I ran the unit-tests in `runtime/Python3/test/run.py` without issues. Not sure how exhaustive these are.
My name is already signed in `contributors.txt` which is why you don't see it in this PR's diff.

### DIFF
--- a/runtime/Python3/src/antlr4/atn/ATN.py
+++ b/runtime/Python3/src/antlr4/atn/ATN.py
@@ -12,6 +12,11 @@ from antlr4.atn.ATNState import ATNState, DecisionState
 
 
 class ATN(object):
+    __slots__ = (
+        'grammarType', 'maxTokenType', 'states', 'decisionToState',
+        'ruleToStartState', 'ruleToStopState', 'modeNameToStartState',
+        'ruleToTokenType', 'lexerActions', 'modeToStartState'
+    )
 
     INVALID_ALT_NUMBER = 0
 
@@ -58,7 +63,7 @@ class ATN(object):
         if s.nextTokenWithinRule is not None:
             return s.nextTokenWithinRule
         s.nextTokenWithinRule = self.nextTokensInContext(s, None)
-        s.nextTokenWithinRule.readonly = True
+        s.nextTokenWithinRule.readOnly = True
         return s.nextTokenWithinRule
 
     def nextTokens(self, s:ATNState, ctx:RuleContext = None):
--- a/runtime/Python3/src/antlr4/atn/LexerATNSimulator.py
+++ b/runtime/Python3/src/antlr4/atn/LexerATNSimulator.py
@@ -34,6 +34,7 @@ from antlr4.dfa.DFAState import DFAState
 from antlr4.error.Errors import LexerNoViableAltException, UnsupportedOperationException
 
 class SimState(object):
+    __slots__ = ('index', 'line', 'column', 'dfaState')
 
     def __init__(self):
         self.reset()
@@ -49,6 +50,10 @@ Lexer = None
 LexerATNSimulator = None
 
 class LexerATNSimulator(ATNSimulator):
+    __slots__ = (
+        'decisionToDFA', 'recog', 'startIndex', 'line', 'column', 'mode',
+        'DEFAULT_MODE', 'MAX_CHAR_VALUE', 'prevAccept'
+    )
 
     debug = False
     dfa_debug = False
@@ -57,8 +62,6 @@ class LexerATNSimulator(ATNSimulator):
     MAX_DFA_EDGE = 127 # forces unicode to stay in ATN
 
     ERROR = None
-
-    match_calls = 0
 
     def __init__(self, recog:Lexer, atn:ATN, decisionToDFA:list, sharedContextCache:PredictionContextCache):
         super().__init__(atn, sharedContextCache)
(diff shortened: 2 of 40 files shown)


tool/resources/org/antlr/v4/tool/templates/codegen/Python3/Python3.stg
```diff
@@ -652,6 +652,7 @@ CaptureNextTokenType(d) ::= "<d.varName> = self._input.LA(1)"
 
 StructDecl(struct,ctorAttrs,attrs,getters,dispatchMethods,interfaces,extensionMembers) ::= <<
 class <struct.name>(<if(contextSuperClass)><contextSuperClass><else>ParserRuleContext<endif>):
+    __slots__ = ('parser', '__dict__')
```
amykyta3, line 655: Adding `__dict__` here as a catch-all for any other class members (labeled grammar rule members?)
Ideally these would be emitted here too, but I'm not familiar enough with how the parser generator templates work to implement this, so using `__dict__` is sufficient.

runtime/Python3/src/antlr4/atn/ATN.py
```diff
@@ -58,7 +63,7 @@ def nextTokensNoContext(self, s:ATNState):
         if s.nextTokenWithinRule is not None:
             return s.nextTokenWithinRule
         s.nextTokenWithinRule = self.nextTokensInContext(s, None)
-        s.nextTokenWithinRule.readonly = True
+        s.nextTokenWithinRule.readOnly = True
```
amykyta3, line 66: This is fixing what appears to be a capitalization typo. The `IntervalSet` class initializes this using camelCase instead of lowercase. Not sure which is preferred since I see both usages in the runtime.

runtime/Python3/src/antlr4/atn/LexerATNSimulator.py
```diff
@@ -89,7 +92,6 @@ def copyState(self, simulator:LexerATNSimulator ):
         self.startIndex = simulator.startIndex
 
     def match(self, input:InputStream , mode:int):
-        self.match_calls += 1
```
amykyta3, line 92: Removing a debug counter that conflicts with the ability to use `__slots__`. Nothing else references this.

ericvergnaud:
Hi,

looks great but it currently breaks the Travis CI build...

tool/resources/org/antlr/v4/tool/templates/codegen/Python3/Python3.stg
ericvergnaud: this should be as simple as:
__slots__ = ('parser', <attrs:{a | '<a>'}; separator=",">) 
can you try ?then we can leverage your perf improvement measures

ericvergnaud:
am I correct in thinking this improvement would also work for Python 2.x?

amykyta3:
Yeah. Slots have been around since early python 2: https://docs.python.org/2.7/reference/datamodel.html#__slots__
Is it worth implementing for 2.x? Use of py2 has been plummeting since it was deprecated.

ericvergnaud:
I've merged this PR to have it part of 4.9.1 being released this weekend.
Python2 runtime is still widely used, see https://pypistats.org/api/packages/antlr4-python2-runtime/recent
Also would be great if you could try out my suggestion for __slots__ in the stg file

amykyta3:
Will do! I'll try to prep another PR today

PR #3018, opened 2020-12-29
Fragments can be used only in another lexer rule error

### LINKED ISSUE #2886: Lex rule fragment used directly in parse causes parse fail. Should be disallowed.
See discussion here <https://groups.google.com/forum/?utm_medium=email&utm_source=footer#!topic/antlr-discussion/rA7U5t0RUBU>
Thanks to Mike Cargal for explanation.

lex rule declared as a fragment messes up parse. This is understandable but confusing and should probably be picked up by ANTLR.
To reproduce (tested):

```
HELLO : 'hello' ;
WORLD : 'world' ;
start_parse : HELLO WORLD  ;
```

input:

**hello world**

This succeeds. Now instead

```
fragment HELLO : 'hello' ;
WORLD : 'world' ;
start_parse : HELLO WORLD  ;
```

This builds ok but when run with same input it errors:

_line 1:0 mismatched input 'hello' expecting HELLO
Parse error line/col 1/0, expecting HELLO
error in parse._

I have a strong suspicion this didn't used to happen (error raised at compile time) so it may be a regression.

setup is MS windows, antlr 4.8, C# 8, vis studio + emacs

### LINKED ISSUE #3016: Fragments do not work as expected
I'm working on a plantuml grammar and fragments do not work as expected:

This works well:
```
relationType
:
	(
		'|o'
		| '||'
		| '}o'
		| '}|'
		| '}'
	)
	(
		(
			'-'+
			(
				'left'
				| 'right'
				| 'top'
				| 'down'
			) '-'+
		)
		| '-'+
	)
	(
		'{'
		| 'o{'
		| 'o|'
		| '|{'
		| '||'
	)
	| '-->'
;
```

This gives `line 34:16 no viable alternative at input '-------down'` :

```
relationType
:
	(
		'|o'
		| '||'
		| '}o'
		| '}|'
		| '}'
	)
	(
		(
			'-'+ DIRECTION '-'+
		)
		| '-'+
	)
	(
		'{'
		| 'o{'
		| 'o|'
		| '|{'
		| '||'
	)
	| '-->'
;

fragment
DIRECTION
:
	'left'
	| 'right'
	| 'top'
	| 'down'
;
```

Line:
`    A7 }o-------down----------{ B7`

KvanTTT:
A more clear error message to avoid confusion if fragment lexer rules are used in parser rules; fix #3016, fix #2886

### DIFF
--- a/runtime/Java/src/org/antlr/v4/runtime/Token.java
+++ b/runtime/Java/src/org/antlr/v4/runtime/Token.java
@@ -13,23 +13,28 @@ package org.antlr.v4.runtime;
 public interface Token {
 	public static final int INVALID_TYPE = 0;
 
-    /** During lookahead operations, this "token" signifies we hit rule end ATN state
-     *  and did not follow it despite needing to.
-     */
-    public static final int EPSILON = -2;
+	/**
+	 * During lookahead operations, this "token" signifies we hit rule end ATN state
+	 * and did not follow it despite needing to.
+	 */
+	public static final int EPSILON = -2;
+
+	public static final int FRAGMENT = -3;
 
 	public static final int MIN_USER_TOKEN_TYPE = 1;
 
-    public static final int EOF = IntStream.EOF;
+	public static final int EOF = IntStream.EOF;
 
-	/** All tokens go to the parser (unless skip() is called in that rule)
-	 *  on a particular "channel".  The parser tunes to a particular channel
-	 *  so that whitespace etc... can go to the parser on a "hidden" channel.
+	/**
+	 * All tokens go to the parser (unless skip() is called in that rule)
+	 * on a particular "channel".  The parser tunes to a particular channel
+	 * so that whitespace etc... can go to the parser on a "hidden" channel.
 	 */
 	public static final int DEFAULT_CHANNEL = 0;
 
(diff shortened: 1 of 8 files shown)


parrt:
hi.  This is a bit hard to evaluate since there seem to be a lot of changes but really they are just comment changes. Can you undo all of the unnecessary changes so I can look at the difference? BTW, I'm pretty nervous about introducing a new imaginary token.

KvanTTT:
> Can you undo all of the unnecessary changes so I can look at the difference?

Ok.

> BTW, I'm pretty nervous about introducing a new imaginary token.

It's possible to implement without introducing a new imaginary token type but such error will work only with combined grammar, i.e. the test `testFragmentRuleInParserRuleInSeparatedGrammar` will not work.

The current implementation adds fragment tokens to `.tokens` file and analyzes them in the parser generation stage if grammars are separated:

```
TOKEN=1
FRAGMENT=-3
FRAGMENT2=-3
```

parrt:
Hmm... in the end I don't have the motivation to think carefully about these things and any kind of changes that affect end-users should be done after a lot of thought.  I think we should focus on the targets.

Issue #2361, opened 2018-09-11
Deduce token name from value at least for primitive string literals 


KvanTTT:
Consider the following grammar:

```antlr
grammar G;
root: 'value';
```

We got the name unclear name `T__0`  for `value`. I suggest extracting the token name from value: `VALUE`.

vlsi:
:+1:

Infering `'::'` as `T_PAAMAYIM_NEKUDOTAYIM` would be a plus

KvanTTT:
> Infering '::' as T_PAAMAYIM_NEKUDOTAYIM would be a plus

There is no semantics info in lexer that's why it makes sense to handle words char by char. Also, it's easier in implementation.

For instance, lexer does not know about "logical or" `||`, and in some languages, two vertical bars may mean something else. That's why the name `BAR_BAR` is more correct.

PR #3021, opened 2020-12-30
Do not throw an error if defined token name conflicts with autogenerated

### LINKED ISSUE #3020: Unclear error message for tokens that start with T__ suffix since they may conflict with autogenerated tokens
I'm running the tool on the following grammar:

```g4
grammar Test;
root: 'root';
T__0: 't0';
```

And getting the following error:

```
Test.g4:5:0: rule T__0 redefinition; previous at line 0
```

This error is totally unclear for beginner users and occurs because ANTLR creates token for `root` string literal with the same name.

I suggest either throwing of special error for such a case or generate another name for `root` literal, maybe `T__0_`.

KvanTTT:
Fix #3020


parrt:
Why wouldn't it be simpler to simply check that the users rule name does not start with our magic `T__` prefix?

KvanTTT:
Just check if the defined token starts with `T__` and restrict such names? I don't like it because:

* In my opinion, such a suffix is not required at all because all token names can be deducted automatically based on string literals, see #2361
* Yet another error message should be introduced `token name starts with reserved prefix T__`
* Tokens that start with `T__` are not presented in classes that extend `ParseRuleContext`. For instance, for grammar

```antlr
grammar test;
root: Token;
Token: 'token';
```

The following code is generated:

```java
public static class RootContext extends ParserRuleContext {
	public TerminalNode Token() { return getToken(testParser.Token, 0); }
	public RootContext(ParserRuleContext parent, int invokingState) {
		super(parent, invokingState);
	}
	@Override public int getRuleIndex() { return RULE_root; }
}
```

But if token is defined in implicit way:

```antlr
grammar test;
root: 'token';
```

Then `Token()` method is missing in generated code:

```java
public static class RootContext extends ParserRuleContext {
	public RootContext(ParserRuleContext parent, int invokingState) {
		super(parent, invokingState);
	}
	@Override public int getRuleIndex() { return RULE_root; }
}
```

parrt:
well this is not a big change, but any change at all in the tool has to have a really good reason given my low involvement and the huge community out there that could be relying on something. If you'd like to add an error message I'm okay with that but I'd rather not introduce any changes to the interface to methods etc..

KvanTTT:
> If you'd like to add an error message I'm okay with that

Anyway, this change does not require introducing the new error message at all.

> but I'd rather not introduce any changes to the interface to methods etc..

It's extending, not changing. But if even it's changing, covering by bug number of tests almost ensures that everything is fine.

PR #3036, opened 2021-01-11
Experiment self hosted GitHub CI

ericvergnaud:


KvanTTT:
BTW, it's better to use [git amend](https://git-scm.com/book/en/v2/Git-Tools-Rewriting-History) to get rid of garbage commits. CI will be triggered on every git push anyway.

ericvergnaud:
All these are genuine commits.
I’m able to trigger manual builds from the GitHub UI

PR #3037, opened 2021-01-11
Cpp target: build antlr4-runtime as 3 Nuget packages.

zosrothko:
Hello

This PR is adding a directory `antlr4\runtime\Cpp\runtime\nuget `which contains a DOS script `pack.cmd` that is building the `c++ antlr4-runtime` as 3 Nuget packages.

usage: `pack vsversion release [prerelease] `
as for example :

1. pack antlr4-runtime compiled with vs2019 for version 4.9.1 with prerelease '-rc': `pack 2019 4.9.1 -rc`
2. pack antlr4-runtime compiled with vs2019 for version 4.9.1 : `pack 2019 4.9.1`

updated the `contributor.txt` by my signature


mike-lischke (changes requested):
Interesting idea, I like it, however I can barely review the patch as I don't know nuget. Maybe @sharwell could do that?

The only change that I request is to fix the conflict in contributors.txt (and any build that fails related to this patch).

zosrothko:
By the way, you can already dowload or install the ANTLR4 4.9.1 Nuget package for VS2019 at [int.nugettest.org](https://int.nugettest.org/packages?q=ANTLR4.Runtime.vs2019)

Issue #3038, opened 2021-01-13
Annotation processor broken in IntelliJ


ericvergnaud:
Tests work from cmd-line when i set jdk-7 but not from intellij.  It hangs in the annotation preprocessing every time; it appears that even if you say to use Java seven, it insists upon using its own Java for those preprocessing steps, which no longer work with our annotation code.  can't remember exactly, but there was some proxy class type mismatch.

I should point out that this is even during build not running the tests actually. firing up the development environment now to test again...  I believe this is also after I upgraded the development environment because a few months ago it was okay.

During java compilation, I get the following error message:

```
...
Adding @NotNull assertions... [antlr4-runtime-test-annotations]
Parsing java... [antlr4-runtime-test-annotation-processors]
java: User-specified option "-proc:none" is ignored for "antlr4-runtime-test-annotation-processors". This compilation parameter is set automatically according to project settings.
...

java: java.lang.ClassCastException: com.sun.proxy.$Proxy11 cannot be cast to com.sun.tools.javac.processing.JavacProcessingEnvironment
Abnormal build process termination: 
[...]

PR #3039, opened 2021-01-13
Potential fix to #3038

ericvergnaud:
@parrt can you check if this works for you?


parrt:
Confirmed!!  I just recompiled and it seemed to work. Nice job, thanks @ericvergnaud !

PR #3040, opened 2021-01-19
Cpp target: No building tests/samples for external utfcpp

bigerl:
It is unnecessary that each time the external dependency `utfcpp` is pulled, its tests are build and run. Also its samples are unnecessarily generated.
I suggest to use the provided cmake to disable this behavior.
You can find the documentation of the cmake options here: https://github.com/nemtrif/utfcpp/blob/v3.1.1/CMakeLists.txt#L10-L12

Let's make this build a little bit more green and not use unnecessary resources.


mike-lischke (approved):
@parrt Another C++ build change ready to be merged. This might create a conflict with the other build change for utf8cpp, but is the older one, so merge this first please. The other patch might have to be updated then.

mike-lischke:
@bigerl Please fix the contributors file in your patch.

mike-lischke:
@bigerl Your patch broke the build with this error:

> Building ANTLR4 C++ runtime (if necessary) at /Users/macosx-ci-4/actions-runner/_work/antlr4/antlr4/runtime-testsuite/target/classes/Cpp
Cloning into 'utfcpp'...
fatal: could not create work tree dir 'utfcpp': File exists
fatal: destination path 'utfcpp' already exists and is not an empty directory.
fatal: destination path 'utfcpp' already exists and is not an empty directory.
CMake Error at /Users/macosx-ci-4/actions-runner/_work/antlr4/antlr4/runtime-testsuite/target/classes/Cpp/runtime/utfcpp-prefix/tmp/utfcpp-gitclone.cmake:31 (message):
  Failed to clone repository: 'git://github.com/nemtrif/utfcpp'

Can you fix that?

bigerl:
I've checked the delta between my fork's master and the current master. There have been several changes to cmake files since I created the pull request which might influence the behavior I patched. 
Also there is additional CI code now in the current master I must test against and which might fail right now. 

I do not have time to look into that immediately. So I would suggest the following:
- revert the merge of this pull-request
- I'll fork the current master and apply and test the patch again; ETA 2 weeks
- When I am done, I'll create another pull request.

mike-lischke:
OK @bigerl I created a patch to revert yours and the builds are green again. Just create a new PR for your case when you are ready.

PR #3053, opened 2021-01-25
Fix 3052

### LINKED ISSUE #3052: Python3 runtime raises AttributeError in Parser.py
I met an `AttributeError` when trying codes in the 'Actions and Attributes' section.

I used the following grammar which is translated into a Python3 version from the sample grammar shown in that section.

```
grammar DynScope;
 
prog: block ;
 
block
	/* List of symbols defined within this block */
	locals [
        symbols = list()
	]
	: '{' decl* stat+ '}'
	// print out all symbols found in block
	// $block::symbols evaluates to a List as defined in scope
	{print("symbols=%s" % ($symbols))}
	;
 
/** Match a declaration and add identifier name to list of symbols */
decl: 'int' ID {$block::symbols.append($ID.text);} ';' ;
 
/** Match an assignment then test list of symbols to verify
 * that it contains the variable on the left side of the assignment.
 * Method contains() is List.contains() because $block::symbols
 * is a List.
 */
stat: ID '=' INT ';'
	{
if not $ID.text in $block::symbols:
    print("Undefined variable: %s" % ($ID.text))
	}
	| block
	;
 
ID : [a-z]+ ;
INT : [0-9]+ ;
WS : [ \t\r\n]+ -> skip ;
```

Executing the Python3 target generated from the above grammar leads to an `AttributeError`, saying that the Context object has no `ruleIndex` attribute.

Below is the last part of the traceback.

```
...
  File "/usr/local/lib/python3.9/site-packages/antlr4/Parser.py", line 456, in getInvokingContext
    if ctx.ruleIndex == ruleIndex:
AttributeError: 'DeclContext' object has no attribute 'ruleIndex'
```

There was a previous issue #2745 talking about this, however, it seems that this problem is still remaining unsolved.

Manually changing the line into below solves this problem.

```
if ctx.getRuleIndex() == ruleIndex:
```

l215884529:
Fix #3052.

The problem is caused by an undefined `ruleIndex` attribute.
Calling `getRuleIndex()` method instead of accessing the `ruleIndex` attribute solves this problem.

### DIFF
--- a/contributors.txt
+++ b/contributors.txt
@@ -284,3 +284,4 @@ YYYY/MM/DD, github id, Full name, email
 2020/11/26, mr-c, Michael R. Crusoe, 1330696+mr-c@users.noreply.github.com
 2020/12/01, maxence-lefebvre, Maxence Lefebvre, maxence-lefebvre@users.noreply.github.com
 2020/12/03, electrum, David Phillips, david@acz.org
+2021/01/25, l215884529, Qiheng Liu, 13607681+l215884529@users.noreply.github.com
--- a/runtime/Python3/src/antlr4/Parser.py
+++ b/runtime/Python3/src/antlr4/Parser.py
@@ -453,7 +453,7 @@ class Parser (Recognizer):
     def getInvokingContext(self, ruleIndex:int):
         ctx = self._ctx
         while ctx is not None:
-            if ctx.ruleIndex == ruleIndex:
+            if ctx.getRuleIndex() == ruleIndex:
                 return ctx
             ctx = ctx.parentCtx
         return None


ericvergnaud:
Hi,
thanks for this but unfortunately our CI is currently down, so can't assess the impact of this proposed change.
Is it ok to park this for a week or so until the CI is fixed?

l215884529:
Thank you for the quick reply.
It's totally fine for me :)

ericvergnaud:
Hi,
the CI is fixed
can you touch the PR at your end in order to trigger checks?
thanks

PR #3073, opened 2021-02-10
Cpp/runtime: use utfcpp from system if possible

jirislaby:
Try to find system's `utf8cpp` first, instead of downloading it from github. `utf8cpp` 3 defines a cmake package, so we can use `find_package`. Older `utf8cpp`s don't, so we try also `find_path`. If all that fails, we still fall back to github download.

This is needed as distributions cannot download stuff from their build systems. And people want to use libraries/headers from the distribution anyway.

I had to reorder the file a bit, so that the libraries are defined through the `utf8cpp` detection.


mike-lischke (approved):

mike-lischke:
I started disliking the approach to use a submodule for an optional part like this, but it seems we cannot get away from this easily, as we don't have automatic dependency handling outside of cmake.

@parrt This is a C++ build change and can be merged.

Issue #212, opened 2013-03-29
A new lexer command: Less


massaad:
It will be helpful if we can do some look-ahead while in the lexer without tying the grammar to a particular target language.

So, I am proposing adding a new lexer command that is almost the opposite of `more`. Using an example derived from the example in the book (in section 12.4):

```
SPECIAL_OPEN : '<?' Name -> less, pushMode(PROC_INSTR);
```

when matched, will switch to `PROC_INSTR` mode, but then will start matching again starting from `<?`.

gagern:
In the question [Syntactic predicates in ANTLR lexer rules](http://stackoverflow.com/q/35724082/1468366) on Stack Overflow I've looked at the `INT '..'` vs. `REAL: INT '.'` ambiguity from [the Antlr v2 manual](http://www.antlr2.org/doc/lexer.html#Predicated-LL%28k%29_Lexing) and [the Antlr 3 faq](https://theantlrguy.atlassian.net/wiki/pages/viewpage.action?pageId=2687092), looking for ways to address this in Antlr 4. The way I see it, the methods suggested here, both `less` and `~&'.'`, could be used to address that, too. It would be good to have *some* way of looking ahead and then consuming only part of the matched text.

Groostav:
I'm writing a yaml parser, and to determine if the root is a horizontal or vertical style yaml document, I'd like the `DEFAULT_MODE` to do a quick switch for me:

two equivalent yaml documents:
```yaml
#horizontal
{ x: { y : z} }
#vertical
x:
  y: z
```

as you might imagine, this maps to two lexer modes, vertical and horizontal. So what Id like the DEAFAULT_MODE to contain is simply

```g4
INDENT_H_INIT : '{' -> type(INDENT_H), mode(HORIZONTAL) ;
INDENT_V_INIT : . -> less, type(INDENT_V_0), mode(VERTICAL);
```

without that `less` rule, I'm actually having a really hard time doing this.

so I'm :+1: here.

Groostav:
huh, and github's own syntax highlighter doesn't like it, looks like they might need to use `less` also :p

parrt:
why don't modes alone solve this? can't see it right away.

Groostav:
my problem is picking the initial mode, which I don't want to do in java, I want it in my lexer. So far the rule that I'm trying right now is 

```
INITIAL_INDENT_V : {} -> skip, mode(value), pushMode(vertical);
```

[...]

PR #3074, opened 2021-02-10
WIP: 212 a new lexer command less

dkleszyk:
Work in progress for #212

This adds an entirely new token type instead of overloading `MORE`.

An alternative I considered was adding a 'rewind' field and having the less action set `_rewind = true; _type = MORE`.

I mostly just duplicated the code for the 'more' action and updated the behavior while necessary.

Still needs design review, documentation, tests, and the contributors update.


dkleszyk:
@sharwell Is this an addition you would still be interested in?

FWIW, the checks did pass for the last meaningful commit (e91e700). It's just after I pulled in the updated contributors file (60cdbfe) that the checks failed.

runtime/Cpp/runtime/src/atn/LexerLessAction.h
```diff
@@ -0,0 +1,57 @@
...
+namespace antlr4 {
+namespace atn {
+
+  /// <summary>
+  /// Implements the {@code less} lexer action by calling <seealso cref="Lexer#less"/>.
+  ///
+  /// <para>The {@code less} command does not have any parameters, so this action is
+  /// implemented as a singleton instance exposed by <seealso cref="#INSTANCE"/>.</para>
+  ///
+  /// @author Sam Harwell
```
mike-lischke, line 20: You should actually set yourself as the author, if at all. Even if most of the code is copied. You created this file.

runtime/Cpp/runtime/src/atn/LexerLessAction.h
```diff
@@ -0,0 +1,57 @@
...
+  /// @since 4.2
```
mike-lischke, line 21: This must also be updated (to 9.3).

runtime/Java/src/org/antlr/v4/runtime/atn/ATNSerializer.java
```diff
@@ -314,6 +314,11 @@ public void serializeCodePoint(IntegerList data, int cp) {
 					data.add(actionIndex != -1 ? actionIndex : 0xFFFF);
 					break;
 
+				case LESS:
```
mike-lischke, line 317: That's a problem. You should add a new UUID to enable the LESS lexer mode, just like it had been done with the lexer actions. Otherwise there can be conflicts with generated parser/lexer files, using a runtime that doesn't support the LESS command. We could require that their versions always match, but traditionally this kind of counter measure has been used and we should continue with that approach.

mike-lischke (changes requested):
This PR is definitely worth to be considered. A lot of effort has been put into it to cover all supported runtimes @parrt, @KvanTTT

What I request to change is that:
- the correct copyrights are used in the new files (author, version, year)
- the support for the LESS command in all ATN deserializer classes is covered by a feature UUID and
- at least one test is added for the new feature

dkleszyk:
@mike-lischke Thanks for the feedback.

I've updated the copyright and version info. I'm working on the feature UUID for the ATN serialization.

As for testing, I did include [a test for the new command](https://github.com/antlr/antlr4/pull/3074/files#diff-51eacd42c7a45f1fb753b2f2d8c206f6ec9e88d9a017d9f86292228b6afff501R181). Are you looking for something additional? (Such as testing ATN serialization/deserialization for the new command?)

parrt:
Guys, I'm not able to add new features at this time...just a heads up. Especially something that would monkey with the state machines. sorry!

KvanTTT:
Maybe somebody else can maintain the repository?  As I understand @ericvergnaud has privileges that allow merging. Otherwise, ANTLR development is stagnating a bit...

I don't have objections to such a feature, but tests that cover the new functionality are strictly required.

parrt:
hi @KvanTTT  well unfortunately a successful project is not about having somebody evaluate PRs for correctness and so on. ANTLR has been successful because I have been ruthless to my vision over the last 30+ years in the feature set and direction, so it's hard to just start letting people sticking features, such as this one.   A project that works might stagnate as you say in terms of adding new features, but Microsoft Word was totally fine in 1995 and I'm not sure anything they added since has been useful to me.  Sometimes products just work and sometimes they need new stuff. You'll notice that we have new targets, but I'm not gonna start adding new features simply because somebody thought they were cool.

mike-lischke:
Maybe it's time to start a fork and take over responsibility, if someone wants to do that. I fully understand Terence here, but on the other hand if a project doesn't evolve further, it's going to become obsolete over time. I'm pretty sure I wrote this before, but here we go: maybe it's time to hand ANTLR completely over to the community and let it develop to something which is no longer in your hands Ter? Or are there aspects we don't know, which would not allow that?

parrt:
Forks are always welcome, for example Sam Harwell's.   At this time I will maintain leadership even if it is going to asymptotically approach stagnation. People are always welcome to build a better tool, which is what I did because I didn't like yacc. There are lots of developers out there, not all of them make the right design and strategy decisions.

dkleszyk:
@parrt Is there a less intrusive way of incorporating this functionality that you would be willing to support?

From my perspective, the core change is allowing rewinding the lexer and rematching a token in a different mode. The goal being lexer lookahead in a way that is more flexible than predicates, and is not tied to a specific language runtime.

If not, I'll try to keep the branch up-to-date with master.

Thanks.

parrt:
@dkleszyk Unfortunately I have to take into consideration all targets and any additional complexity and possible introduction of bugs etc. It is extremely hard to shepherd a project to success; one of the most difficult decisions a designer has to make is what to include and what to exclude. At this point I more or less excluding just about anything that touches the critical $ALL(*)$ algorithm. so in the end, I can't support this officially, certainly not just for one target.

PR #3080, opened 2021-02-13
fix javascript semantic predicates

### LINKED ISSUE #3078: [JS] Utils.js: arrayToString(null) should check null arg
The code for [arrayToString](https://github.com/antlr/antlr4/blob/master/runtime/JavaScript/src/antlr4/Utils.js#L7) should check whether parameter "a" is null. This function is called when debugging for ParserATNInterpreter.debug=true is set. Unfortunately, the debug code throws an exception when evaluating "a.join()". Also, if the array contains null values, the output string contains no value. In C#, a null element of the array is outputted as "null".

ericvergnaud:
fixes #2007 
fixes #3078

### DIFF
--- a/runtime/JavaScript/src/antlr4/Utils.js
+++ b/runtime/JavaScript/src/antlr4/Utils.js
@@ -4,7 +4,7 @@
  */
 
 function arrayToString(a) {
-    return "[" + a.join(", ") + "]";
+    return Array.isArray(a) ? ("[" + a.join(", ") + "]") : "null";
 }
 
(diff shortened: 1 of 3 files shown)


kaby76:
@ericvergnaud  Thank you for all these changes. They were important because the JS grammar had semantic predicates that weren't updated for JS. But I didn't even know the code was bad because they weren't even called until your fixes here to the JS runtime. The parse trees for C#, Java, and JS on my little example are all the same.

PR #3094, opened 2021-02-27
Fix #3093 - The antlr4-maven-plugin is not marked as thread safe

### LINKED ISSUE #3093: The antlr4-maven-plugin is not marked as thread safe
If you start a build which contains antlr4-maven-plugin with multi thread support (parallel build) you get the following warning:
```
[INFO] --- maven-jar-plugin:2.4:jar (default-jar) @ cql ---
[INFO] Building jar: ..../grammars-v4/cql/target/cql-1.0-SNAPSHOT.jar
[INFO] 
[INFO] ----------------------< org.antlr.grammars:cql3 >-----------------------
[INFO] Building Apache Cassandra CQL 3 grammar 1.0-SNAPSHOT            [34/256]
[INFO] --------------------------------[ jar ]---------------------------------
[WARNING] *****************************************************************
[WARNING] * Your build is requesting parallel execution, but project      *
[WARNING] * contains the following plugin(s) that have goals not marked   *
[WARNING] * as @threadSafe to support parallel building.                  *
[WARNING] * While this /may/ work fine, please look for plugin updates    *
[WARNING] * and/or request plugins be made thread-safe.                   *
[WARNING] * If reporting an issue, report it against the plugin in        *
[WARNING] * question, not against maven-core                              *
[WARNING] *****************************************************************
[WARNING] The following plugins are not marked @threadSafe in Apache Cassandra CQL 3 grammar:
[WARNING] com.khubla.antlr:antlr4test-maven-plugin:1.17
[WARNING] org.antlr:antlr4-maven-plugin:4.9.1
[WARNING] Enable debug to see more precisely which goals are not marked @threadSafe.
[WARNING] *****************************************************************
[INFO] 
[INFO] --- maven-clean-plugin:2.5:clean (default-clean) @ cql3 ---
```
It would be nice to mark the antlr4-maven-plugin `@threadSafe` to suppress this warning.

khmarbaise:

### DIFF
--- a/antlr4-maven-plugin/src/main/java/org/antlr/mojo/antlr4/Antlr4Mojo.java
+++ b/antlr4-maven-plugin/src/main/java/org/antlr/mojo/antlr4/Antlr4Mojo.java
@@ -55,7 +55,7 @@ import java.util.Set;
 	name = "antlr4",
 	defaultPhase = LifecyclePhase.GENERATE_SOURCES,
 	requiresDependencyResolution = ResolutionScope.COMPILE,
-	requiresProject = true)
+	requiresProject = true, threadSafe = true)
 public class Antlr4Mojo extends AbstractMojo {
 
     // First, let's deal with the options that the ANTLR tool itself
--- a/contributors.txt
+++ b/contributors.txt
@@ -287,3 +287,4 @@ YYYY/MM/DD, github id, Full name, email
 2021/01/25, l215884529, Qiheng Liu, 13607681+l215884529@users.noreply.github.com
 2021/02/02, tsotnikov, Taras Sotnikov, taras.sotnikov@gmail.com
 2021/02/21, namasikanam, Xingyu Xie, namasikanam@gmail.com
+2021/02/27, khmarbaise, Karl Heinz Marbaise, github@soebes.com


ericvergnaud:
@parrt any concerns?

parrt:
are we sure antlr is threadsafe at maven level? not sure what it means.

khmarbaise:
Thread safety means in this relationship that you can correctly build with `-T XXX`. (Excerpt [from the docs](https://maven.apache.org/developers/mojo-api-specification.html): `Marks this mojo as being thread-safe, i.e. the mojo safely supports concurrent execution during parallel builds`...)

parrt:
Hi. thanks. I am not sure what requirements are for thread safety in this context though.

khmarbaise:
The point is that during a build process for each module the antl4-maven-plugin is called once (ok you can do several executes via execution blocks but those will be sequential). Executing a build in parallel means on the level of modules in parallel... for example if you have a build with 20 modules you could theoretically execute it with 20 thread via `mvn package -T20` and it would be extremely fast but of course you have usually dependencies between modules which results in the module level tree where each level is on it's own thread. In this case it's safe to define the plugin `threadsafe=true` which in the end does not produce the WARNING.
Technically a plugin which uses an external resource (for example a file somewhere in the file system) would not be thread safe because it could happen that two instances of the plugin try to access that resource simultaneously which would result in an issue. But based on the antlr4-maven-plugin I don't see that.

PR #3097, opened 2021-02-27
Fix #3096 - antlr4-maven-plugin is using maven-project version 2.2.1 …

### LINKED ISSUE #3096: antlr4-maven-plugin is using maven-project version 2.2.1 dependency
The antlr4-maven-plugin is using a dependency to maven-project version 2.2.1 where as everything else is using Maven 3.X

khmarbaise:
cleanup the dependency for Maven Project.


parrt:
this does look like it simplifies the maven pom but I literally don't understand maven; I did check that it compiles and installs the same way however :)

PR #3098, opened 2021-02-28
Swift InterpreterDataReader and deserialize fixes

ahooper:
Swift runtime could not deserialize a Grammar.interp file. ATNDeserialize deserialize() failed initializing a Character with a -1 value. The data representation is changed to Int.

A class InterpreterDataReader is added to read Grammar.interp files, with unit tests.


ahooper:
Integration test "ci/circleci: test_runtime-cpp-RECURSION" is failing, but this does not seem related to this change.

PR #3099, opened 2021-02-28
Arrays are not classes in Swift

Dante-Broggi:
unlike in Java, where this was transliterated from.

It might also be reasonable to create a request for a warning for a variable which is not read after it's last write,
with the previous code as the example, on <https://bugs.swift.org/>.


hanjoes:
neat fix. the original port was done verbatim from java so lots of places not clinging to the swift idiom. will merge it as part of #3301

Issue #3079, opened 2021-02-12
[C#] Setting ParserATNInterpreter.debug = true cause crashes


kaby76:
I'm am trying to debug in a side-by-side manner the actions of the ATN interpreter of C# with JS. So, I am trying to leverage the debug output information when setting ParserATNSimulator.debug = true. Unfortunately, debug output is shuffled to ConsoleWriteLine(fmt, args) which summarily crashes because it calls System.Console.WriteLine(fmt, args), and it is not a formated call because it is called with only one arg, and it is not to be intrepreted as a format. Either the method needs just call System.Console.WriteLine(fmt), or the applied references to the method changed to System.Console.WriteLine(). I thought I raised an issue years ago for this, even supplied the diffs, but it was neglected. This is important. I will submit a PR, but please merged the fix asap.

kaby76:
The problem was introduced with [this commit](https://github.com/antlr/antlr4/commit/ef49021c8b5e84e409b2a52b7cf7de59fda18348#diff-e4b2755a11fc5fdfaaa170fbab8c65f336360f94800879c0c3258c146cf7f62e). Interpretation of the first string into ConsoleWriteLine() is the problem. As far as I can tell, there are no calls for formatted output.

KvanTTT:
@kaby76 @ericvergnaud please close since it's fixed

PR #3104, opened 2021-03-03
Fix for #3079

kaby76:
This change request is a fix for #3079. Basically, if you want to debug the C# runtime, you will want to set ParserATNSimulator.debug to true, then run you tests. Unfortunately, the runtime crashes if you try. As I write this, I am currently chasing down yet another problem with the C# runtime. So, this is important to fix.

This problem goes back to [this change](https://github.com/antlr/antlr4/commit/ef49021c8b5e84e409b2a52b7cf7de59fda18348#diff-e4b2755a11fc5fdfaaa170fbab8c65f336360f94800879c0c3258c146cf7f62e) with the introduction of the wrapper function ConsoleWriteLine(), which interprets the first parameter as a format string. Unfortunately, "ToString()" of states and sets in the Antlr runtime are interpreted as C# formatting.

I looked over every other runtime target, and not a single one has a wrapper around the native runtime call. While it makes sense to encapsulate the native calls, C# should not be unique here. And, even if you do wrap Console.WriteLine(), you should have tested your code!

I changed all the calls to "ConsoleWriteLine()" back to "Console.WriteLine()" and added a regression test for one for this code (antlr4/runtime/CSharp/tests/issue-3079). The readme.md should be self explanitory.

This change request also includes a .sln file for antlr4/runtime/CSharp/tests/issue-2693. "dotnet build" does not work without relating the Antlr4 C# runtime with this program without a .sln.

--Ken

### DIFF
--- a/runtime/CSharp/src/Atn/ATNSimulator.cs
+++ b/runtime/CSharp/src/Atn/ATNSimulator.cs
@@ -75,11 +75,6 @@ namespace Antlr4.Runtime.Atn
 			throw new Exception("This ATN simulator does not support clearing the DFA.");
 		}
 
-        protected void ConsoleWriteLine(string format, params object[] arg)
-        {
-            System.Console.WriteLine(format, arg);
-        }
-
         public PredictionContextCache getSharedContextCache()
 		{
 			return sharedContextCache;
--- a/runtime/CSharp/src/Atn/LexerATNSimulator.cs
+++ b/runtime/CSharp/src/Atn/LexerATNSimulator.cs
@@ -119,7 +119,7 @@ namespace Antlr4.Runtime.Atn
 			ATNState startState = atn.modeToStartState[mode];
             if (debug)
 			{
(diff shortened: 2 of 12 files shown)


ericvergnaud:
Hey Ken, don't be too harsh. From the original PR it seems there was a bit of a mess created accidentally, and we had to fix things very rapidly. 
@lecode-official would you be able to opine on this PR vs your original one (#1229)?

kaby76:
I should note that I have tried multiple ways to not have a "test.sh" script to sed edit the source code. And, I should have a check for a zero return value from the "dotnet run", since that is what is the whole point here--not crashing. The script for the test is grotesque, but I cannot find a way to mutate the value of ParserATNSimulator.debug to true without changing the source for the test--which should not be run in CI less you cut a release with debug=true! I have tried subclassing, SetValue(), and dynamically load the assembly plus SetValue(). I have not tried a code injection of a static initializer to set debug=true, but I will try that at some point. SetValue() of a static readonly is discussed [here](https://github.com/dotnet/runtime/issues/37796).

Issue #3109, opened 2021-03-06
[C++] string_view for all target compilations


xTachyon:
`std::string_view` is a great addition to the standard. It provides a great and cheap way to pass const strings around that could come from a `std::string`, from a c string or even from some in-memory buffer that doesn't even have to be null terminated.

I noticed there was an attempt to add usages of this type in #2847 by checking in a macro for 17. However, since this still is C++11 library, this poses some problems:
1. Firstly, the obvious one that I'm the most concerned about is that for builds made on C++11 and C++14 there is still the unnecessary string copy that compilers can't yet optimize out. 
For the `ANTLRInputStream` class, even if it has a constructor that takes `const char*, size_t` that looks like it doesn't make a copy, that actually creates a `std::string` and it passes down to the other constructor. So even on a C++17 build this class is not saved from extra copies.

2. As someone pointed out in another PR, this creates 2 separate ABIs for the library, making it impossible to link antlr built with 17 and a binary built with 11 or vice versa.

3. This significantly reduces the ability detect compilation errors and/or runtime errors without building and running tests with the library compiled with both 17 and pre-17.

The solutions that I thought of involve replacing `std::string_view` with a substitute or removing it at all. This can be done because `std::string_view` doesn't use any language feature introduced in 17.
1. [string-view-lite](https://github.com/martinmoene/string-view-lite) looks like a full implementation of `string_view` and can be configured to not use `std::string_view` even when building with 17 in order to not break ABI again.

2. `string_view` is a trivial class to implement and so far it doesn't look like Antlr will need anything more the constructors, `.size()` and `.data()`. One can implement a small `StringView` that can be embedded directly in this project.

3. Maybe it would be a good idea to take a step back and just provide overloads that take `const char*, size_t` for any constructor/method that doesn't copy the input string and remove `string_view` altogether.

To be fair, I am not a big fan of the last suggestion. I think either suggestion 1 or 2 would be a good fit for the project, but I like 2 the most.

What do you think about this?

xTachyon:
Sorry for pinging but it seems you're the guy who I should talk to @mike-lischke .

mike-lischke:
I'm trying to figure out what you are after here and, to my shame, I currently fail at that. The C++ runtime is not per se a C++11 lib. I use it with C++17 already long before PR #2847 was added.

[...]

PR #3113, opened 2021-03-09
Fixed ANTLRInputStream and ANTLRFileStream

mike-lischke:
A previous change to add std::string_view support to ANTLRInputStream for C++17 caused some trouble because of ABI changes. This has been changed to define 2 constructors, one for std::string_view (for C++17) and the original one for std::string.
This is turn caused an error in ANTLRFileStream, which also takes a string in the constructor (but handling it as file name instead of input). To make this clearer the c-tor taking a std::string has been deleted in ANTLRFileStream and the class now requires to load input via the loadFile() function. This might cause some trouble for those users who had used the std::string constructor of ANTLRFileStream, but I think the better error reporting outweighs the little annoyance.

This is supposed to handle issue #3109.


ericvergnaud:
@mike-lischke not one of my boxes
I restarted it, not sure why you're not able to do that...

mike-lischke:
I see the dropdown but all entries are disabled for me. And it looks as if the restarted build will again fail.

mike-lischke:
@xTachyon I picked up your suggestions and also fixed a number of issues I found.

xTachyon:
Looks good except that comment I made.

runtime/Cpp/runtime/src/ANTLRInputStream.cpp
```diff
@@ -17,44 +17,41 @@ using namespace antlrcpp;
...
 void ANTLRInputStream::load(const std::string &input) {
+  load(input.data(), input.size());
+}
+
+void ANTLRInputStream::load(const char *data, size_t length) {
   // Remove the UTF-8 BOM if present.
   const char bom[4] = "\xef\xbb\xbf";
-  if (input.compare(0, 3, bom, 3) == 0)
-    _data = antlrcpp::utf8_to_utf32(input.data() + 3, input.data() + input.size());
+  if (strncmp(data, bom, 3) == 0)
```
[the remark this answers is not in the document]
mike-lischke: I don't move the head (input) so `input.data() + input.size()` still points to the original end of the string. It's not a size value!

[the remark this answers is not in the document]
xTachyon: Not sure where your comment went but here's an example of what I'm saying:
```
        auto ptr = new char[1]{ '\xef' };
        return strncmp(ptr, "\xef\xbb\xbf", 3);
```
If just that character is passed as a string to `load`, `strncmp` will happily try to read 2 bytes past where the character is allocated. I just tried this with an end of page allocator and on my platform it crashes.

Maybe this function is only supposed to accept correct utf8? I guess the problem wouldn't happen then.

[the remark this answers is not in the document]
mike-lischke: I misunderstood your comment first. Now I think the patch should be complete.

runtime/Cpp/runtime/src/ANTLRInputStream.cpp
```diff
@@ -17,44 +17,41 @@ using namespace antlrcpp;
...
+  load(input.data(), input.size());
+}
+
+void ANTLRInputStream::load(const char *data, size_t length) {
   // Remove the UTF-8 BOM if present.
-  const char bom[4] = "\xef\xbb\xbf";
-  if (input.compare(0, 3, bom, 3) == 0)
-    _data = antlrcpp::utf8_to_utf32(input.data() + 3, input.data() + input.size());
+  const char *bom = "\xef\xbb\xbf";
+  if (length > 3 && strncmp(data, bom, 3) == 0)
```
xTachyon, line 51: Maybe I'm just nitpicking at this point but wouldn't an empty bom be a valid empty string that could be valid for some grammars? So the check would be `length >= 3`. Or maybe I don't know my unicode. Sorry for all the disturbance.

mike-lischke: No worries, a good code review is often inconvenient. 

The approach there is that if there's enough to have a possible BOM then check it, if not just go ahead and load what you got. So even an empty string is "loaded" correctly in the else branch.

I'm assuming here that by "empty bom" you actually mean an empty input string.

xTachyon: Here's what I mean: if only `"\xef\xbb\xbf"` is passed in the function, `length` will be 3 and the `else` branch will be taken. Then, at least when used with utfcpp conversion functions, it will throw `utf8::invalid_code_point` because it can't parse the bom.

mike-lischke: Yes, I also thought about this case and found it would not be worth to be considered as it would mean no useful input was given. However, there could be a workflow where a BOM is always attached automatically, regardless of what input was given and empty input is often valid input. So it would make sense to strip off the BOM and deal with the empty input anyway. I change the check therefore.

xTachyon (approved):

mike-lischke:
@parrt Here's a new C++ patch to fix ABI problems for file + input stream. Ready to be merged.

Issue #3127, opened 2021-03-17
Not possible to turn Dart's debug logging off


canastro:
Currently the dart runtime is too lousy in terms of logging, which makes it impossible to read other logs defined in the user codebase.

PR #3128, opened 2021-03-17
Configure dart debug flags through environment variables

canastro:
https://github.com/antlr/antlr4/issues/3127


canastro:
@lingyv-li please take a look if this is an ok change.

lingyv-li (approved):
Changes looks good to me, thanks for adding this.

kaby76:
This would be nice feature to apply across all targets so as to be a consistent feature. If work on all targets, and when I am trying to debug the runtime, I would like to be able to turn on debug first and get some quick feedback. Otherwise, I will have to build a version of the runtime. In C#, and likely Java, it's basically impossible to turn this on because the value is static const final. The value is used at program load time to JIT compile out the debugging output code, and impossible to turn on from a switch to Parser. --Ken

ericvergnaud:
> This would be nice feature to apply across all targets so as to be a consistent feature. I'm working with all targets. When I am trying to debug the runtime, I would like to be able to turn on debug first and get some quick feedback. Otherwise, I will have to build a version of the runtime. In C#, and likely Java, it's basically impossible to set debug dynamically because the value is static final const. The value is used at program load time to JIT compile out the debugging output code, and thus impossible to turn on from a switch to Parser. --Ken

Unfortunately it would impact performance. Using a static final lets the compiler drop dead code, which would not be possible if the flag was read from outside.

kaby76:
> Unfortunately it would impact performance. Using a static final lets the compiler drop dead code, which would not be possible if the flag was read from outside.

From what I've read of this in an Issue posted on the compiler for C#, the assembly initializers are performed first, then jited for exactly this reason. I would point you to what Ayers at MS said, but i can't remember. So it would have to be checked.

lingyv-li:
> But why would the Dart code have debug = true to begin with?

That clearly is a mistake by me! Thanks for catching the bug too @canastro  😃 

On the performance note for Dart, would it be better to also change the flags to const for parser?

kaby76:
> On the performance note for Dart, would it be better to also change the flags to const for parser?

I don't know what the Dart compiler generates. Does the debug code compile away with or without const? That would be the deciding guide I suppose. I would think it should just be like all the other targets. It's hard for me to context switch between 7 different runtimes now.

doc/dart-target.md
```diff
@@ -114,4 +114,15 @@ The expected output is:
 {"a":1}
 "a":1
 1
-```
\ No newline at end of file
+```
+
+### Debug
+We have some logs in place that can ease the debugging process, in order to turn these logs on you can enable the following environment variables:
```
lingyv-li, line 120:
```suggestion
We have some logs in place that can ease the debugging process, in order to turn these logs on you can enable the following environment declarations:
```

runtime/Dart/lib/src/atn/src/parser_atn_simulator.dart
```diff
@@ -257,10 +257,15 @@ import 'transition.dart';
...
-  static final bool retry_debug = false;
+  static final bool debug =
+      bool.fromEnvironment('PARSER_DEBUG', defaultValue: false);
+  static final bool debug_list_atn_decisions = bool.fromEnvironment(
+      'PARSER_LIST_ATN_DECISIONS_DEBUG',
+      defaultValue: false);
+  static final bool dfa_debug =
+      bool.fromEnvironment('PARSER_DFA_DEBUG', defaultValue: false);
+  static final bool retry_debug =
+      bool.fromEnvironment('PARSER_RETRY_DEBUG', defaultValue: false);
```
lingyv-li, line 268:
```suggestion
  static const bool debug =
      bool.fromEnvironment('PARSER_DEBUG', defaultValue: false);
  static const bool debug_list_atn_decisions = bool.fromEnvironment(
      'PARSER_LIST_ATN_DECISIONS_DEBUG',
      defaultValue: false);
  static const bool dfa_debug =
      bool.fromEnvironment('PARSER_DFA_DEBUG', defaultValue: false);
  static const bool retry_debug =
      bool.fromEnvironment('PARSER_RETRY_DEBUG', defaultValue: false);
```

lingyv-li (changes requested):
I was misled by the name `fromEnvironment`, it turns out to be not "from the environment variables" but "the environment declaration" provided when [compiling (AOT) or running](https://api.dart.dev/stable/1.24.2/dart-core/String/String.fromEnvironment.html). And that's why it's returns a compile-time constant value.

With this I believe at least for Dart target it won't affect performance. Though I don't think other languages can have this without impacting performance. Good news is there is an explicit defaultValue so you don't need to remember how to set it if you forgot.

See: https://github.com/dart-lang/sdk/issues/27585

lingyv-li (approved):
lgtm!

stefanschaller:
@ericvergnaud Same here. The PR looks great, @lingyv-li approved the PR, all steps from the pipeline are passing (Of course not the go-runtime). 
We definitely need those features in the next version on pub.dev that should be published ASAP. 
Would be so great, because I really need those features and fixes.

ansiemens:
fixes #3127

lingyv-li:
@parrt can you please check and merge this? Thanks.

Issue #3126, opened 2021-03-17
Parsing Java with JavaScript parser fails while CSharp/Java parser succeeds


kaby76:
I have been parsing some Java code using grammars-v4/java/java/*.g4, and I noticed that for the following input, the JavaScript parser fails but the CSharp and Java parsers succeed.

    public class HelloWorld { 
       public static void main(String[] args) { 
          System.out.println("Hello, World" + "hi There");
       }
    }

Note, in all parsers I generate for the grammar, I used the default listeners.

Debugging the JavaScript parser side-by-side with the CSharp parser, I've determined that the JS parser fails to parse the expression "Hello, World" + "hi There" at the '+' operator.

Looking over the generated parsers JavaParser.cs and JavaParser.js, I have noticed that the likely culprit is the generated code:

CSharp:

    if ( ParseListeners!=null )
        TriggerExitRuleEvent();

JavaScript:

    if(this._parseListeners!==null) {
        this.triggerExitRuleEvent();
    }

These codes are not the same, and they execute differently.

In CSharp, [ParseListeners](https://github.com/antlr/antlr4/blob/4b649103f31e0c0bc3bd0bb964efd0dea6b9f5a0/runtime/CSharp/src/Parser.cs#L404) in the snippet is a `get` property call in the `Parser` class that returns an empty list if the [backing field](https://github.com/antlr/antlr4/blob/4b649103f31e0c0bc3bd0bb964efd0dea6b9f5a0/runtime/CSharp/src/Parser.cs#L158) is null. In this case, it returns an empty list, which is for sure not equal to null, and `this.triggerExitRuleEvent()` is called.

In JavaScript, `this._parseListeners` is a reference to the [backing field](https://github.com/antlr/antlr4/blob/4de9e80109530a72ccae8997bdb305c07dcbcd23/runtime/JavaScript/src/antlr4/Parser.js#L71) itself, not the getter [getParseListeners()](https://github.com/antlr/antlr4/blob/4de9e80109530a72ccae8997bdb305c07dcbcd23/runtime/JavaScript/src/antlr4/Parser.js#L165). In JavaScript, `triggerExitRuleEvent()` is never called.

Also, just scanning through the generated parsers, I've also noticed constants `~0x3f` used in CSharp vs `~0x1f` used in JavaScript. While it's likely not important in this example, it is rather irritating to see magic number differences between the runtimes.

--Ken

ericvergnaud:
[...]

PR #3130, opened 2021-03-18
Fix 3126

ericvergnaud:


ericvergnaud:
Fixes #3126

ericvergnaud:
@pboyer @jwkohnen would you ming checking the consistent failure of the go build on AppVeyor for the last week?
I suspect it could be caused by:
```private static final String GO_RUNTIME_IMPORT_PATH = "github.com/antlr/antlr4/runtime/Go/antlr"; // TODO: Change this before merging with upstream```
but would rather let you opine

PR #2816, opened 2020-05-01
[Go] Thread-safe ANTLR codegen

TristonianJones:
The ANTLR Go codegen allocates mutable global variables which can result in data races as evidenced in https://github.com/google/cel-go/pull/177 and reported in #2040. As a local workaround, global variables in the generated source were shifted into local variables within the `NewLexer` and `NewParser` functions. 

While the change made the generated code thread-safe, it came at a significant CPU and memory overhead in parse-heavy call paths. However, exporting the `SetInputStream` method on the `lexer.go` makes it possible to use `sync.Pool` objects with the modified codegen artifacts and reduces the parsing overhead to within striking distance of the prior non-thread-safe code.

To take advantage of such thread-safety and object reuse within ANTLR Go applications, developers would need to regen their grammars and incorporate the following pattern of use within their own code base:

```go
import (
  "sync"
  "github.com/antlr/antlr4/runtime/Go/antlr"
)

var(
  lexerPool *sync.Pool = &sync.Pool{ New: func() interface{} {
    return New<LexerName>(nil)
  }}
  parserPool *sync.Pool = &sync.Pool{ New: func() interface{} {
    return New<ParserName>(nil)
  }} 
)

func Parse(text string) antlr.ParseTree {
   lex := lexerPool.Get().(*<LexerName>)
   defer lexerPool.Put(lex)
   par := parserPool.Get().(*<ParserName>)
   defer parserPool.Put(par)

   lex.SetInputStream(antlr.NewInputStream(text))
   par.SetInputStream(antlr.NewCommonTokenStream(lex, 0))
   return par.<StartRule>()
}
```

All Go-related tests have been run and executed successfully following this change:

```
-------------------------------------------------------
 T E S T S
-------------------------------------------------------
Running org.antlr.v4.test.runtime.go.TestListeners
..Tests run: 7, Failures: 0, Errors: 0, Skipped: 0, Time elapsed: 7.444 sec - in org.antlr.v4.test.runtime.go.TestListeners
Running org.antlr.v4.test.runtime.go.TestSemPredEvalLexer
.Tests run: 8, Failures: 0, Errors: 0, Skipped: 0, Time elapsed: 5.974 sec - in org.antlr.v4.test.runtime.go.TestSemPredEvalLexer
Running org.antlr.v4.test.runtime.go.TestCompositeLexers
[...]

PR #3137, opened 2021-03-26
Fix for #2940: fix data race warning in Go

### LINKED ISSUE #2940: go: data race in dfa.go
I recently observed a data race in a generated parser (while running my own tests under the thread sanitizer).

```
// Write, in dfa.go.
func (d *DFA) setS0(s *DFAState) {
	d.s0Mu.Lock()
	defer d.s0Mu.Unlock()
	d.s0 = s
}

// Read, in lexer_atn_simulator.go.
dfa := l.decisionToDFA[mode]

if dfa.s0 == nil {
  return l.MatchATN(input)
}
```

The problem is that `dfa.s0` is being accessed without a read lock.

minjoosur:
This fixes #2940.
There was some missing and incorrect usage of locks which caused Data Race Warning on Go Target. 
I solved the problem by putting a lock on the correct spots and by making sure read/write are coming through getter and setters.

### DIFF
--- a/contributors.txt
+++ b/contributors.txt
@@ -290,3 +290,4 @@ YYYY/MM/DD, github id, Full name, email
 2021/02/27, khmarbaise, Karl Heinz Marbaise, github@soebes.com
 2021/03/02, hackeris
 2021/03/03, xTachyon, Damian Andrei, xTachyon@users.noreply.github.com
+2021/03/26, minjoosur, Minjoo Sur, msur@salesforce.com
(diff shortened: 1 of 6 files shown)


happygiraffe:
@pboyer are you the right person to review this? (based on comments in #2816)

Joakker (approved):
Looks good to me! I do help with maintaining the go runtime, but I don't have write permissions it seems. So maybe ping someone with write access 😅 

Also, please make sure to resolve the conflicts in `contributors.txt`

PR #3144, opened 2021-04-03
[C++] Performance improvement in finally

xTachyon:
Before, to use `finally`, the compiler had to:
- set up the lambda;
- construct a `std::function` that could allocate if it wants to;
- call `finaly()`;
- the dtor of the `std::function` would call the lambda;

Neither MSVC nor GCC could inline any of the functions of `std::function`, the `finally` function, nor the lambda, even with LTO.

With this patch both MSVC and GCC can see through the object and inline the lambda directly into the function code, making it essentially free.
MSVC would occasionally still set up the `FinalAction` object even if it doesn't use it, but would inline the lambda anyway.

Should we migrate to a non-movable `FinalAction` wrapper that wouldn't need the `bool` field so MSVC would get the idea to not set up the object?

Anyway, this should still be faster and generate smaller code because the simpler logic.
@mike-lischke

### DIFF
--- a/runtime/Cpp/runtime/src/antlr4-common.h
+++ b/runtime/Cpp/runtime/src/antlr4-common.h
@@ -34,7 +34,6 @@
 #include <exception>
 #include <bitset>
 #include <condition_variable>
-#include <functional>
 
 #ifndef USE_UTF8_INSTEAD_OF_CODECVT
   #include <codecvt>
--- a/runtime/Cpp/runtime/src/support/CPPUtils.cpp
+++ b/runtime/Cpp/runtime/src/support/CPPUtils.cpp
@@ -202,12 +202,6 @@ namespace antlrcpp {
     return result;
   }
 
-  //----------------- FinallyAction ------------------------------------------------------------------------------------
-
-  FinalAction finally(std::function<void ()> f) {
-    return FinalAction(f);
-  }
-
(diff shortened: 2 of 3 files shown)


mike-lischke:
Do I see that right, the real optimization is the `std::move` call in the `FinalAction` c-tor? Do you have any numbers that can prove the performance improvement? I'm not against that patch, just wanna know how much it brings.

xTachyon:
The move has little to do with the optimization. The real optimization is taking the lambda as a template instead wrapping it in a `std::function`. This allows the compiler to inline it directly in the calling function.

I don't have any numbers right now; what I've done is check what asm is generated by msvc and gcc.
See https://godbolt.org/z/nsvnYx3eY .

With `std::function`, there's multiple function calls, including virtual method calls and the posibility for dynamic allocation or thrown exceptions.
With the template version, the resulting function is 4 lines long and it includes 2 simple calls(that could be inlined further) with the functions I needed to make the example.

Given that `finally()` is used in a lot of autogenerated code, I'd say the optimization is significant for how easy it is.

mike-lischke (approved):

mike-lischke:
Makes sense, thanks.

@parrt Another C++ patch to merge. As usual, some of the tests fail due to timeouts.

PR #3148, opened 2021-04-08
[Go] Expose BaseParser.removeParseListeners

jcking:
This method is exposed in the Java and C++ runtimes, at least. It is useful for using sync.Pool and being able to reset the parse listeners.


runtime/Go/antlr/parser.go
```diff
@@ -274,7 +274,7 @@ func (p *BaseParser) RemoveParseListener(listener ParseTreeListener) {
 }
 
 // Remove all parse listeners.
```
DavidGamba, line 276: Documentation should be updated to match go standards:

`// RemoveParseListeners removes all parse listeners.`

DavidGamba (changes requested):
Fix doc

PR #3149, opened 2021-04-09
Revert "Cpp target: No building tests/samples for external utfcpp"

mike-lischke:
Reverts antlr/antlr4#3040

This seems to conflicts more than expected with previous patches for utfcpp.


mike-lischke:
@parrt A patch to fix the C++ test builds.

PR #3163, opened 2021-04-21
Allow Existing utf8cpp Dependency

mattpaletta:


jcking:
This should no longer be an issue, as ANTLR has its own UTF-8 handling now and no longer has any external dependencies.

PR #3164, opened 2021-04-24
Patch utfcpp build

bigerl:
As promised to @mike-lischke , here is the rebase of #3040 :

> It is unnecessary that each time the external dependency utfcpp is pulled, its tests are built and run. Also its samples are unnecessarily generated.
> I suggest to use the provided cmake to disable this behavior.
> You can find the documentation of the cmake options here: https://github.com/nemtrif/utfcpp/blob/v3.1.1/CMakeLists.txt#L10-L12
> 
> Let's make this build a little bit more green and not use unnecessary resources.


mike-lischke (approved):

mike-lischke:
Thanks @bigerl.

@parrt This is a small C++ build change and ready to be merged.

PR #3172, opened 2021-05-03
Javascript runtime: InputStream performance optimization

redexp:
If you know feature array size then it faster to init array with `new Array(size)` instead of `array.push`. Here comparison https://jsbench.me/67ko818eiq/1


ericvergnaud:
@redexp thanks

PR #3176, opened 2021-05-04
Add `go.mod` to go runtime module

### LINKED ISSUE #3175: Fix Go test suite
Currently the Go target has some problems that prevented from passing tests.  @Joakker and @prnvbn have agreed to help with the Go target so let's see if we can fix that first and then come back to https://github.com/antlr/antlr4/issues/2504

Here's a link to the [failing appveyor for Go](https://ci.appveyor.com/project/parrt/antlr4/build/job/t7k58ax1tidlimng):

```
...
 T E S T S
Running org.antlr.v4.test.runtime.go.TestCompositeLexers
Tests run: 1, Failures: 0, Errors: 1, Skipped: 0, Time elapsed: 0.406 sec <<< FAILURE! - in org.antlr.v4.test.runtime.go.TestCompositeLexers
org.antlr.v4.test.runtime.go.TestCompositeLexers  Time elapsed: 0.406 sec  <<< ERROR!
java.lang.Exception: 
Non-zero exit while caching go runtime, output: go: go.mod file not found in current directory or any parent directory; see 'go help modules'
	at org.antlr.v4.test.runtime.go.TestCompositeLexers.groupSetUp(TestCompositeLexers.java:24)
...
```

Seems `go.mod` is missing whatever that is!

Joakker:
Hopefully fixes #3175

### DIFF
--- a/contributors.txt
+++ b/contributors.txt
@@ -298,3 +298,4 @@ YYYY/MM/DD, github id, Full name, email
 2021/04/24, bigerl, Alexander Bigerl, alexander [äät] bigerl [pkt] eu
 2021/05/02, michalharakal, Michal Harakal, michal.harakal@users.noreply.github.com
 2021/05/03, redexp, Sergii Kliuchnyk, redexp@users.noreply.github.com
+2021/05/04, joakker, Joaquín León, joaquinandresleon108@gmail.com
--- a/runtime/Go/antlr/go.mod
+++ b/runtime/Go/antlr/go.mod
@@ -1,0 +1,3 @@
+module github.com/antlr/antlr4/runtime/Go/antlr
+
+go 1.16


parrt:
Weird. still failing: go.mod not found still. Must be in wrong dir?  Our test does something weird when running go?

Joakker:
I'm inclined to believe it's the CIitself that's the problem, because on circle-ci, the go runtime tests are successful.

I have the suspicion that it's pulling from the github repo instead of the forked copy with the module file. Since go imports by url (`import "github.com/antlr/antlr4/runtime/Go/antlr"`), and the repo that the url is pointing to doesn't have the module file, it's not gonna find it like that.

parrt:
Hmm...ok, so I'll incorporate and retest.

Joakker:
Ok, I don't know why it keeps failing. We might have to rewrite the test

Joakker:
Actually, I think I do. The go proxy (which is used by the go tool to download packages from the internet), is a bit slow to update, especially for pseudo-releases, which would be the case here. That can be attested by the fact that the [documentation page](https://pkg.go.dev/github.com/antlr/antlr4/runtime/Go/antlr) still says that it lacks a `go.mod` file @parrt

parrt:
Ok, it now seems to think we have one: `checked Valid go.mod file` :)    Still seems to be failing though I think

Joakker:
At least it's progress :D

PR #3177, opened 2021-05-05
Add LICENSE to go runtime

Joakker:
For the docs page to show up in https://pkg.go.dev/github.com/antlr/antlr4/runtime/Go/antlr, a LICENSE file or similar has to be present in the module directory. This commit adds the same license that is at the top of the repo, so there won't be any confusion.

This is merely a formality to improve the documentation


Joakker:
cc @parrt

PR #3184, opened 2021-05-15
Avoid function unnecessary bind call

### LINKED ISSUE #3183: Allow Antlr Javascript runtime to be loaded into Kindle Touch
Currently if I want to load the antlr4 JS runtime into a constrained environment _everything works flalwesly_, but there is a small change that I have to monkey-patch on my side each time: the `this.readInt.bind(this)` call fails because `bind` doesn't exist.

Since this issue only happens on 2 lines (and all others just work without issue) I thought maybe an alternative solution might help remove this last problem by using a very easy straight-forward replacement of those 2 lines?

Thank you for the great work on this library, it really helps a lot and I am a great fan!

milanaleksic:
(I have signed the contributors.txt already previously)

closes #3183

This is a monkey patch I apply on my side each time before I include the JS code into my web application and it allows Antlr4 to be started on my Kindle Touch integrated browser (and probably other constrained environments)

### DIFF
--- a/runtime/JavaScript/src/antlr4/atn/ATNDeserializer.js
+++ b/runtime/JavaScript/src/antlr4/atn/ATNDeserializer.js
@@ -123,11 +123,11 @@ class ATNDeserializer {
         this.readModes(atn);
         const sets = [];
         // First, deserialize sets with 16-bit arguments <= U+FFFF.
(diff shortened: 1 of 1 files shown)


milanaleksic:
Closing this PR since it will be fixed on my side using a polyfill

Issue #2504, opened 2019-03-06
Is antlr4's golang visitor pattern in a usable state?


ianamason:
I asked this [question](https://stackoverflow.com/questions/54992660/is-antlr4s-golang-visitor-pattern-in-a-usable-state?noredirect=1#comment96742736_54992660) over on stack overflow, without much enlightenment. I was told to ask it
here. 

### Is antlr4's golang visitor pattern in a usable state?

I see no working examples, and I see a few pull requests that remain open. I would like to avoid going down the garden path.

https://github.com/antlr/antlr4/pull/1807

https://github.com/antlr/antlr4/issues/1843

There are also pull requests that make me think the golang target might be dead in the water.

https://github.com/antlr/antlr4/issues/2152

The fact that the [antlr4 documentation page for golang](https://github.com/antlr/antlr4/blob/master/doc/go-target.md) is over two years stale is also not 
encouraging.

So all I am really asking is should I steer clear of golang, or is the documentation just somewhere that google doesn't see :-)

nikunjy:
Looking to use antlr with golang. Can you address some of these @parrt ?

parrt:
Sorry. I don't know Go.

ianamason:
Might want to be honest then, and remove go from the supported language targets. At least warn the gullible user that they may be disappointed.

parrt:
Are you claiming it doesn't support Go in any way?

ianamason:
I am claiming that I could not get  the --visitor pattern to work. Nor could I find any  working examples
of other people getting it to work. What I did find were issues in this repository that seem to  indicate that it never worked, and that people have tried and failed to fix it.  I ended up having to fall back on the listener pattern for the task I had at hand, but given what I know now, would probably 
have no chosen go for the project I am/was working on.

parrt:
I see. ok, well, my apologies. There are lots of different language targets for this project and it's very hard to keep track of the current state of every piece of every API. It's also no longer my focus.

ianamason:
Might be an idea then to keep the issue open, so that someone with time and inclination 
may succeed where others have failed.

ocurr:
@ianamason According to @pboyer 's [comment](https://github.com/antlr/antlr4/pull/1841#pullrequestreview-66582914) on #1841 the Go visitor implementation is in a usable state; however, the user is currently required to put in more work then, for instance, the java implementation.

ianamason:
[...]

PR #3186, opened 2021-05-23
Tidy up go runtime

Joakker:
Partially addresses https://github.com/antlr/antlr4/issues/2504#issuecomment-846481915

This PR normalizes the doc comments in the go runtime sources to one more compatible with the go documentation tool, `godoc`, so the auto-generated docs over at [pkg.go.dev](https://pkg.go.dev/github.com/antlr/antlr4/runtime/Go/antlr) look more readable, since people familiar with go will look there first for the documentation instead of in this repo.

It also hides some elements that don't need to be exported as they don't have to do strictly with the generated parser.


Joakker:
Aight, the go tests are passing, so that's what matters

Joakker:
cc @parrt

parrt:
Maybe @davesisson @pboyer @willfaught could take a look and bless the changes.

Joakker:
Okie, thanks

runtime/Go/antlr/atn.go
```diff
@@ -4,8 +4,11 @@
 
 package antlr
 
+// ATNInvalidAltNumber represents an invalid ATN alternative
```
willfaught, line 7: Thanks for filling out the doc, that's good. To fit Go convention, doc comments should be full sentences: https://golang.org/doc/effective_go#commentary. This one needs ending punctuation.

runtime/Go/antlr/atn_config.go
```diff
@@ -47,6 +48,7 @@ type BaseATNConfig struct {
 	reachesIntoOuterContext    int
 }
 
+// NewBaseATNConfig7 creates a new instance of BaseATNConfig.
```
willfaught, line 51: It's unclear to me what the "7" means. Same for 6, 5, 4, etc. below.

runtime/Go/antlr/atn_config.go
```diff
@@ -108,29 +117,39 @@ func (b *BaseATNConfig) setPrecedenceFilterSuppressed(v bool) {
...
 	return b.context
 }
 
+// GetSemanticContext returns the semantic context of this config.
 func (b *BaseATNConfig) GetSemanticContext() SemanticContext {
 	return b.semanticContext
 }
 
+// GetReachesIntoOuterContext returns whether this config reaches into the outer
+// context.
```
willfaught, line 146: It's unclear how the int will indicate that. Same for Set* below.

runtime/Go/antlr/atn_config.go
```diff
@@ -198,23 +218,27 @@ func (b *BaseATNConfig) String() string {
 	return fmt.Sprintf("(%v,%v%v%v%v)", b.state, b.alt, s1, s2, s3)
 }
 
+// LexerATNConfig extends BaseATNConfig for lexers.
```
willfaught, line 221: Nit: "extends" is the wrong term/concept for Go. "embeds" would technically be the correct term.

runtime/Go/antlr/atn_config.go
```diff
@@ -180,6 +199,7 @@ func (b *BaseATNConfig) hash() int {
 	return murmurFinish(h, 4)
 }
 
+// String implements the Stringer interface.
```
willfaught, line 202: What info is returned, and in what format?

runtime/Go/antlr/atn_config_set.go
```diff
@@ -6,12 +6,13 @@ package antlr
 
 import "fmt"
 
+// ATNConfigSet extends ATNConfig in the form of an unordered set.
 type ATNConfigSet interface {
 	hash() int
-	Add(ATNConfig, *DoubleDict) bool
+	Add(ATNConfig, *doubleDict) bool
```
willfaught, line 12: It's unusual to expose unexported names in exported declarations. If Add itself can't be made entirely unexported, the typical design is to leave DoubleDict exported, since the user is responsible for constructing one. I'm not familiar with this code anymore, so perhaps the user has a way to get a doubleDict from somewhere, and knows to plug it in here?

runtime/Go/antlr/atn_config_set.go
```diff
@@ -6,12 +6,13 @@ package antlr
...
 	AddAll([]ATNConfig) bool
 
-	GetStates() *Set
+	getStates() *_set
```
willfaught, line 15: Leading underscores aren't conventional. If I recall correctly, golint would flag this. It would be better to just name it "set".

runtime/Go/antlr/atn_config_set.go
```diff
@@ -155,14 +158,17 @@ func (b *BaseATNConfigSet) GetStates() *Set {
 	return states
 }
 
+// HasSemanticContext returns true if this config has a semantic context.
```
willfaught, line 161: I like your "whether" wording in SetHasSemanticContext below, it explains both Boolean cases. Here, the doc only explains what happens if it's true.

willfaught (changes requested):
I got a few files in and then noticed how many changes there are. I won't be able to cover it all in one day. Hopefully the others can pitch in too. :)

runtime/Go/antlr/atn_deserialization_options.go
```diff
@@ -4,22 +4,28 @@
 
 package antlr
 
+// ATNDeserializationOptionsdefaultOptions is the default settings for
```
willfaught, line 7: "default" in ATNDeserializationOptionsdefaultOptions is lowercase.

runtime/Go/antlr/atn_config_set.go
```diff
@@ -351,14 +378,16 @@ func (b *BaseATNConfigSet) String() string {
 	return s
 }
 
+// OrderedATNConfigSet extends BaseATNConfigSet
```
willfaught, line 381: No ending punctuation.

runtime/Go/antlr/atn_config_set.go
```diff
@@ -319,6 +345,7 @@ func (b *BaseATNConfigSet) SetReadOnly(readOnly bool) {
 	}
 }
 
+// String implements the stringer interface.
```
willfaught, line 348: stringer -> Stringer

What's the format?

runtime/Go/antlr/atn_config_set.go
```diff
@@ -269,48 +284,59 @@ func (b *BaseATNConfigSet) ContainsFast(item ATNConfig) bool {
...
+func (b *BaseATNConfigSet) setConflictingAlts(v *bitSet) {
 	b.conflictingAlts = v
 }
 
+// ReadOnly returns true if this set is read-only.
 func (b *BaseATNConfigSet) ReadOnly() bool {
 	return b.readOnly
 }
 
+// SetReadOnly controls whether this set is read-only or not.
```
willfaught, line 339: Nit: "or not" can be removed.

runtime/Go/antlr/atn_config_set.go
```diff
@@ -269,48 +284,59 @@ func (b *BaseATNConfigSet) ContainsFast(item ATNConfig) bool {
...
 	return b.conflictingAlts
 }
 
-func (b *BaseATNConfigSet) SetConflictingAlts(v *BitSet) {
+// setConflictingAlts sets the conflicting alternatives in this config.
+func (b *BaseATNConfigSet) setConflictingAlts(v *bitSet) {
 	b.conflictingAlts = v
 }
 
+// ReadOnly returns true if this set is read-only.
```
willfaught, line 334: Nit: true if -> whether

runtime/Go/antlr/atn_config_set.go
```diff
@@ -269,48 +284,59 @@ func (b *BaseATNConfigSet) ContainsFast(item ATNConfig) bool {
...
 func (b *BaseATNConfigSet) SetDipsIntoOuterContext(v bool) {
 	b.dipsIntoOuterContext = v
 }
 
+// GetUniqueAlt returns this set's unique alternative.
 func (b *BaseATNConfigSet) GetUniqueAlt() int {
 	return b.uniqueAlt
 }
 
+// SetUniqueAlt sets the unique alternative for this set.
```
willfaught, line 319: It's unclear from this what a unique alternative is, and how an int represents that. Same for Get* above.

runtime/Go/antlr/atn_config_set.go
```diff
@@ -269,48 +284,59 @@ func (b *BaseATNConfigSet) ContainsFast(item ATNConfig) bool {
...
-	b.configLookup = NewSet(nil, equalATNConfigs)
+	b.configLookup = newSet(nil, equalATNConfigs)
 }
 
+// FullContext returns true if this set is a full context
 func (b *BaseATNConfigSet) FullContext() bool {
 	return b.fullCtx
 }
 
+// GetDipsIntoOuterContext returns true is this set dips into it's outer
```
willfaught, line 303: is -> if
it's -> its

runtime/Go/antlr/atn_config_set.go
```diff
@@ -177,10 +183,13 @@ func (b *BaseATNConfigSet) GetPredicates() []SemanticContext {
 	return preds
 }
 
+// GetItems returns the configurations contained in this object.
 func (b *BaseATNConfigSet) GetItems() []ATNConfig {
 	return b.configs
 }
 
+// OptimizeConfigs optimizes the elements in this object. The method will panic
```
willfaught, line 191: Optimizes how?

Nit: While this wording is fine, you don't have to explicitly say it will panic, you can just say the object must (keyword: must) be writeable. Users will/should know that means getting a panic.

willfaught (changes requested):

runtime/Go/antlr/atn_config.go
Joakker: Yeah, I'll be renaming them according to what they actually do to create the ATN configs

Joakker:
Aight, that should do it

parrt:
Hi @Joakker  it looks like we are getting a Go language failure:

```
testOne[Go:JavaExpressions_8](org.antlr.v4.test.runtime.go.TestLeftRecursion)  Time elapsed: 0.98 sec  <<< FAILURE!
junit.framework.AssertionFailedError: 
[Go:JavaExpressions_8] Parse output is incorrect: expectedOutput:<(s (e (e new (typespec A) ( )) . b) <EOF>)
>; actualOutput:<>; expectedParseErrors:<>; actualParseErrors:<# antlr.org/test/parser/antlr
parser\antlr\semantic_context.go:163:55: undefined: _set
>; expectedToolErrors:<>; actualToolErrors:<>.
```

Joakker:
Oop, will fix in the morning

runtime/Go/antlr/atn_config.go
```diff
@@ -57,35 +56,56 @@ func NewBaseATNConfig7(old *BaseATNConfig) *BaseATNConfig { // TODO: Dup
...
 }
 
-func NewBaseATNConfig2(c ATNConfig, semanticContext SemanticContext) *BaseATNConfig {
+// ATNConfigWithContext creates a new instance of BaseATNConfig.
+func ATNConfigWithContext(c ATNConfig, semanticContext SemanticContext) *BaseATNConfig {
 	return NewBaseATNConfig(c, c.GetState(), c.GetContext(), semanticContext)
 }
 
-func NewBaseATNConfig1(c ATNConfig, state ATNState, context PredictionContext) *BaseATNConfig {
+// ATNConfigWithStateContext creates a new instance of BaseATNConfig.
```
parrt, line 93: I wonder if these comments actually add anything. I don't know go very well at all but it seems like these comments are not adding very much to constructor functions.

parrt:
seems like a huge change; perhaps @jcking  or someone more knowledgeable can scan through but many of those comments seem unnecessary to me per my comment I just made.

runtime/Go/antlr/atn_config.go
KvanTTT: Actually, they look like useless comments because they don't give additional information, don't explain how an algorithm works, etc.

parrt:
Guys trying to clean up. I think I'll close as there is controversy here. sorry!

PR #3197, opened 2021-06-06
Don't restrict supported types for clone in antlrcpp::Any

### LINKED ISSUE #3194: [C++] antlrcpp::Any incorrect behavior when storing std::string
The `antlrcpp::Any` class does not properly handle storing std::string types in copy constructor/assignment. For example, this code:

```c++
#include <iostream>
#include <string>
#include "antlr4-runtime.h"

int main()
{
    antlrcpp::Any a = std::string("hello");
    antlrcpp::Any b = a;

    std::cout << "a = ";
    if (a.isNotNull()) {
        if (a.is<std::string>()) {
            std::cout << a.as<std::string>() << "\n";
        }
    } else {
        std::cout << "null\n";
    }

    std::cout << "b = ";
    if (b.isNotNull()) {
        if (b.is<std::string>()) {
            std::cout << b.as<std::string>() << "\n";
        }
    } else {
        std::cout << "null\n";
    }
    return 0;
}
```

Would be expected to print (and does if changed to use std::any):
```
a = hello
b = hello
```

However b does not get assigned the correct value from a and the above code actually prints:
```
a = hello
b = null
```

Wrapping the std::string in a pointer like std::shared_ptr yields the expected behavior; however it seems like the [condition on clone in Any](https://github.com/antlr/antlr4/blob/master/runtime/Cpp/runtime/src/support/Any.h#L135-L138) is too restrictive:

```c++
    template<int N = 0, typename std::enable_if<N == N && std::is_nothrow_copy_constructible<T>::value, int>::type = 0>
    Base* clone() const {
      return new Derived<T>(value);
    }
```

In looking into this I noticed that Boost's Any actually has [no restriction applied](https://www.boost.org/doc/libs/1_67_0/boost/any.hpp), which makes sense to me in terms of cloning any type to itself that is copy/move/etc constructible:

```c++
virtual placeholder * clone() const
{
    return new holder(held);
}
```

So could clone in Any just become this?
```c++
    Base* clone() const {
      return new Derived<T>(value);
    }
```

I changed this in my local build and it fixes the original issue, but I'm not sure if this has some potential side effects since I'm testing in a small codebase. If it seems ok, I can make this change and open a PR

Twinklebear:
# Issue

The `antlrcpp::Any` class does not properly handle storing std::string types in copy constructor/assignment. For example, this code (see #3194 )

```c++
#include <iostream>
#include <string>
#include "antlr4-runtime.h"

int main()
{
    antlrcpp::Any a = std::string("hello");
    antlrcpp::Any b = a;

    std::cout << "a = ";
    if (a.isNotNull()) {
        if (a.is<std::string>()) {
            std::cout << a.as<std::string>() << "\n";
        }
    } else {
        std::cout << "null\n";
    }

    std::cout << "b = ";
    if (b.isNotNull()) {
        if (b.is<std::string>()) {
            std::cout << b.as<std::string>() << "\n";
        }
    } else {
        std::cout << "null\n";
    }
    return 0;
}
```

Would be expected to print (and does if changed to use std::any):
```
a = hello
b = hello
```

However b does not get assigned the correct value from a and the above code actually prints:
```
a = hello
b = null
```

Wrapping the std::string in a pointer like std::shared_ptr yields the expected behavior; however it seems like the [condition on clone in Any](https://github.com/antlr/antlr4/blob/master/runtime/Cpp/runtime/src/support/Any.h#L135-L138) is too restrictive:

```c++
    template<int N = 0, typename std::enable_if<N == N && std::is_nothrow_copy_constructible<T>::value, int>::type = 0>
    Base* clone() const {
      return new Derived<T>(value);
    }
```

# Modification in this PR

This PR resolves #3194  by removing the template restrictions. In looking into this I noticed that Boost's Any actually has [no restriction applied](https://www.boost.org/doc/libs/1_67_0/boost/any.hpp), which makes sense to me in terms of cloning any type to itself that is copy/move/etc constructible:

```c++
virtual placeholder * clone() const
{
    return new holder(held);
}
```

So this PR modifies clone to set no restrictions:
```c++
    Base* clone() const {
      return new Derived<T>(value);
    }
```

I've just been testing this in my own project which is quite small, so I'm not sure if this opens some wider issue. It seems like it should be fine, as this how Boost's Any works. The C++ runtime test suite also looks to be passing:

```
[INFO] Scanning for projects...
[WARNING] The project org.antlr:antlr4-runtime-testsuite:jar:4.9.3-SNAPSHOT uses prerequisites which is only intended for maven-plugin projects but not for non maven-plugin projects. For such purposes you should use the maven-enforcer-plugin. See https://maven.apache.org/enforcer/enforcer-rules/requireMavenVersion.html
[INFO] 
[INFO] -----------------< org.antlr:antlr4-runtime-testsuite >-----------------
[INFO] Building ANTLR 4 Runtime Tests (2nd generation) 4.9.3-SNAPSHOT
[INFO] --------------------------------[ jar ]---------------------------------
[INFO] 
[INFO] --- maven-enforcer-plugin:1.2:enforce (enforce-maven) @ antlr4-runtime-testsuite ---
[INFO] 
[INFO] --- antlr4-maven-plugin:4.9.3-SNAPSHOT:antlr4 (default) @ antlr4-runtime-testsuite ---
[INFO] No grammars to process
[INFO] ANTLR 4: Processing source directory /home/will/repos/antlr4/runtime-testsuite/test
[INFO] 
[INFO] --- maven-resources-plugin:2.6:resources (default-resources) @ antlr4-runtime-testsuite ---
[INFO] Using 'UTF-8' encoding to copy filtered resources.
[INFO] Copying 15 resources
[INFO] Copying 1242 resources
[INFO] 
[INFO] --- maven-compiler-plugin:3.8.1:compile (default-compile) @ antlr4-runtime-testsuite ---
[INFO] No sources to compile
[INFO] 
[INFO] --- maven-resources-plugin:2.6:testResources (default-testResources) @ antlr4-runtime-testsuite ---
[INFO] Using 'UTF-8' encoding to copy filtered resources.
[INFO] Copying 203 resources
[INFO] 
[INFO] --- maven-compiler-plugin:3.8.1:testCompile (default-testCompile) @ antlr4-runtime-testsuite ---
[INFO] Nothing to compile - all classes are up to date
[INFO] 
[INFO] --- maven-surefire-plugin:2.19.1:test (default-test) @ antlr4-runtime-testsuite ---

 T E S T S
Running org.antlr.v4.test.runtime.cpp.TestLexerExec
Compiler version is: clang version 10.0.0-4ubuntu1 
Target: x86_64-pc-linux-gnu
Thread model: posix
InstalledDir: /usr/bin

Building ANTLR4 C++ runtime (if necessary) at /home/will/repos/antlr4/runtime-testsuite/target/classes/Cpp
C++ runtime build succeeded

Tests run: 37, Failures: 0, Errors: 0, Skipped: 0, Time elapsed: 50.258 sec - in org.antlr.v4.test.runtime.cpp.TestLexerExec
Running org.antlr.v4.test.runtime.cpp.TestParseTrees
Tests run: 9, Failures: 0, Errors: 0, Skipped: 0, Time elapsed: 36.744 sec - in org.antlr.v4.test.runtime.cpp.TestParseTrees
Running org.antlr.v4.test.runtime.cpp.TestSemPredEvalLexer
Tests run: 8, Failures: 0, Errors: 0, Skipped: 0, Time elapsed: 10.048 sec - in org.antlr.v4.test.runtime.cpp.TestSemPredEvalLexer
Running org.antlr.v4.test.runtime.cpp.TestCompositeLexers
Tests run: 2, Failures: 0, Errors: 0, Skipped: 0, Time elapsed: 2.514 sec - in org.antlr.v4.test.runtime.cpp.TestCompositeLexers
Running org.antlr.v4.test.runtime.cpp.TestPerformance
Tests run: 7, Failures: 0, Errors: 0, Skipped: 0, Time elapsed: 31.145 sec - in org.antlr.v4.test.runtime.cpp.TestPerformance
Running org.antlr.v4.test.runtime.cpp.TestFullContextParsing
Tests run: 15, Failures: 0, Errors: 0, Skipped: 0, Time elapsed: 60.92 sec - in org.antlr.v4.test.runtime.cpp.TestFullContextParsing
Running org.antlr.v4.test.runtime.cpp.TestParserExec
Tests run: 38, Failures: 0, Errors: 0, Skipped: 0, Time elapsed: 154.39 sec - in org.antlr.v4.test.runtime.cpp.TestParserExec
Running org.antlr.v4.test.runtime.cpp.TestLexerErrors
Tests run: 12, Failures: 0, Errors: 0, Skipped: 0, Time elapsed: 15.032 sec - in org.antlr.v4.test.runtime.cpp.TestLexerErrors
Running org.antlr.v4.test.runtime.cpp.TestListeners
Tests run: 7, Failures: 0, Errors: 0, Skipped: 0, Time elapsed: 28.649 sec - in org.antlr.v4.test.runtime.cpp.TestListeners
Running org.antlr.v4.test.runtime.cpp.TestCompositeParsers
Tests run: 15, Failures: 0, Errors: 0, Skipped: 0, Time elapsed: 61.146 sec - in org.antlr.v4.test.runtime.cpp.TestCompositeParsers
Running org.antlr.v4.test.runtime.cpp.TestSemPredEvalParser
Tests run: 25, Failures: 0, Errors: 0, Skipped: 0, Time elapsed: 101.993 sec - in org.antlr.v4.test.runtime.cpp.TestSemPredEvalParser
Running org.antlr.v4.test.runtime.cpp.TestLeftRecursion
Tests run: 98, Failures: 0, Errors: 0, Skipped: 0, Time elapsed: 401.897 sec - in org.antlr.v4.test.runtime.cpp.TestLeftRecursion
Running org.antlr.v4.test.runtime.cpp.TestSets
Tests run: 32, Failures: 0, Errors: 0, Skipped: 0, Time elapsed: 129.207 sec - in org.antlr.v4.test.runtime.cpp.TestSets
Running org.antlr.v4.test.runtime.cpp.TestParserErrors
Tests run: 33, Failures: 0, Errors: 0, Skipped: 0, Time elapsed: 133.091 sec - in org.antlr.v4.test.runtime.cpp.TestParserErrors

Results :

Tests run: 338, Failures: 0, Errors: 0, Skipped: 0

[INFO] ------------------------------------------------------------------------
[INFO] BUILD SUCCESS
[INFO] ------------------------------------------------------------------------
[INFO] Total time:  20:18 min
[INFO] Finished at: 2021-06-06T10:49:56-06:00
[INFO] ------------------------------------------------------------------------
```


mike-lischke (approved):

mike-lischke:
I approved the patch, but I would much more like a C++17 version with `std::any` instead.

Twinklebear:
I can move to std::any, if folks are ok pushing the requirements up to C++17. Personally, I'd also prefer std::any

mike-lischke:
Yes, I think it is time to raise the required C++ version to C++17. However, changing that should include all the build scripts and projects (Xcode, Visual Studio, cmake scripts). Can you do that? Unfortunately, I have no bandwidth to assist with that.

Twinklebear:
I can update CMake stuff, but Xcode, Visual Studio might be tougher or just more time consuming than I could also commit to. I saw that there's a bunch of visual studio project files in the repo and an Xcode project: https://github.com/antlr/antlr4/tree/master/runtime/Cpp/runtime . 

Would it actually be worth just moving to CMake only? CMake can generate VS and XCode projects so it may not really be necessary to keep them in the repo like they are now, but I'm not familiar enough with this project to know about previous motivation or reasons for keeping them.

mike-lischke:
@Twinklebear You could at least start the port to std::any and update the cmake scripts. I can later check with Xcode and may find someone to do the Visual Studio changes.

For going cmake only: as long as the IDEs do not support cmake natively as projects, this will not help much. You will always have to run that extra step to generate the project files and would have to manually maintain the cmake files if you want to change something in these project files. Maybe we can do that at a later date, but not now.

mike-lischke (changes requested):
Is this change really what you intended? You removed the last author from the contributors list.

Twinklebear:
> Is this change really what you intended? You removed the last author from the contributors list.

Oh it's not, I thought they were on both branches of the merge but it seems like I misread it. I'll fix that.

I've updated the rest of the code to use std::any instead of antlrcpp::Any and increased the C++ standard in CMake to 17

mike-lischke (commented):
Looks good so far. Have you checked the other cmake files?

Twinklebear:
It seems like the files in:

```
tool/target/classes/org/antlr/v4/tool/templates/codegen/Cpp/
```

Are generated using antlrcpp::Any , and this is causing the tests to fail since they need to be updated. But, I'm not sure where this code is generated from to update it.

mike-lischke:
These two files are the templates for the code generation of the C++ target. They are not generated, but hand written. You have to change them as well.

Twinklebear:
Aha yes, looks like I did something wrong when editing them the first time since now all checks are passing. I'll squash down my commits to clean up these working commits from the history

mike-lischke:
@Twinklebear Sorry for coming back to you so late. I really forgot about this PR. Meanwhile @jcking has filed a number of new PRs and one of them is the std::any move (#3395, which required a move to C++17).

So I think this PR can be closed now.

Twinklebear:
No worries, I saw the move to std::any/C++17 PR and that's also my preferred solution :D 

I'll close this PR now

Issue #3189, opened 2021-05-25
Long pathnames when building with ExternalAntlr4Cpp.cmake


skef:
I apologize for being *that person*

I'm involved with project that's adding an antlr4 parser written in C++ to a project ultimately wrapped in a python package. At the end of all the wrapping we face this when building on Windows:

```
C:/Users/TheUser/AppData/Local/Temp/pip-req-build-0v67w40v/_skbuild/win-amd64-3.7/cmake-build/antlr4_runtime/src/antlr4_runtime/runtime/Cpp/runtime/utfcpp-prefix/src/utfcpp-build/extern/gtest/googlemock/CMakeFiles/gmock_main.dir/./
    has 230 characters.  The maximum full path to an object file is 250 characters (see CMAKE_OBJECT_PATH_MAX).  Object file
      __/googletest/src/gtest-all.cc.obj
     cannot be safely placed under this directory.  The build may not work correctly.
```

And indeed it *doesn't* work correctly, at least not without changing a Windows 10 registry flag, which is beyond what many users can manage.

Obviously there's a lot of shared responsibility here. However, the Antlr part of this with the out-of-the-box ExternalAntlr4Cpp.cmake is 172 characters. That *is* kind of a lot. 

I already tried knocking one of those antlr4_runtime/src's off but it (unsurprisingly) didn't work.

Does utfcpp need to be added on so deep? Could these lengths be mitigated somewhat?

PR #3200, opened 2021-06-08
ExternalAntlr4Cpp: Mitigate build problems with restricted path lengths and zip files

skef:
This deals with #3189 and fixes a problem with specifying a zip file of the Cpp runtime from the website (e.g. https://www.antlr.org/download/antlr4-cpp-runtime-4.9.2-source.zip ).

Given that you've turned the utfcpp tests off in a more recent commit there may no longer be critical path length problems, but it may be better to be safe than sorry.

Accept or close this as you will -- I'll answer questions and do fixes as requested.

### DIFF
--- a/runtime/Cpp/cmake/ExternalAntlr4Cpp.cmake
+++ b/runtime/Cpp/cmake/ExternalAntlr4Cpp.cmake
@@ -2,7 +2,13 @@ cmake_minimum_required(VERSION 3.7)
 
 include(ExternalProject)
 
-set(ANTLR4_ROOT ${CMAKE_CURRENT_BINARY_DIR}/antlr4_runtime/src/antlr4_runtime)
+if(CMAKE_OBJECT_PATH_MAX LESS "300")
+    set(ANTLR4_PATH_ID a4r)
+else()
+    set(ANTLR4_PATH_ID antlr4_runtime)
+endif()
+
+set(ANTLR4_ROOT ${CMAKE_CURRENT_BINARY_DIR}/${ANTLR4_PATH_ID}/src/${ANTLR4_PATH_ID})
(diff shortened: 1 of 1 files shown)


mike-lischke:
`a4r` is a path provided by the caller of the script? It's only used if the path is too long, so it might cause some grief.

skef:
> `a4r` is a path provided by the caller of the script? It's only used if the path is too long, so it might cause some grief.

No I believe the string is only used internal to what is patched.

mike-lischke:
We can merge the other 2 patches first, no problem. But the issue with `a4r` still exists and I want to know what this is about.

skef:
Sorry, my explanation wound up mistakenly quoted so you probably didn't see it. Here is the plain text:

The only place `a4r` comes from line six of this patch. Although the longer `antlr4_runtime` string is used in other places in the repository, the four places it's used in the lines changed in this patch just have to be self-consistent. So this replaces those hard-coded strings with a variable and makes it shorter when path length may be an issue.

Note that this PR *may* not be needed anymore. The ultimate problem was with building the utfcpp test suite, which simply went too deep into sub-directories on Windows if built somewhere not very near the root (by path length). Since that suite has changed there may be no further issue. However, it isn't likely to hurt either. 

If you read this could you take a look at and potentially update issue #3280? I'm willing to do a PR along the lines I describe if the other utfcpp issues can be worked out (it seems to be downloaded, compiled, but not used, at least on Linux, related to `USE_UTF8_INSTEAD_OF_CODECVT` not being defined).

mike-lischke:
OK, so what do we want to do with this PR then? Will you make a new PR with the utfcpp changes and close this one? I rather no like to add a PR, just because it doesn't hurt :-)

skef:
Now that utf8cpp is no longer a dependency it's much less likely a build will run into the path length problem. Closing.

PR #3205, opened 2021-06-10
[Python] Fix index when combining inserts

JamesHutch:
prevInserts list is the subset of rewrites, so indices taken from
enumerating its elements do not correspond to the correct indices in
rewrites. Use prevIop.instructionIndex instead, which matches the other
runtimes.


ericvergnaud:
@lygav as original writer of this code, can you check whether this fix is ok ?

lygav:
Hi @JamesHutch , can you write a test to demonstrate the issue and verify the solution?

parrt:
I'm going to leave this for after 4.10

Issue #3226, opened 2021-07-14
Typo in python 2 runtime


renzhentaxi:
In `IntervalSet.py`, we have a property called `readOnly`
```python
		__init__(...):
				...
				self.readOnly = False
```
however, we are using `readonly` in `atn/ATN.py` (lowercase o)
```
atn/ATN.py:56:        s.nextTokenWithinRule.readonly = True
```

Here is a grep search: `grep -rin "\.readonly" *`

```
atn/ATN.py:56:        s.nextTokenWithinRule.readonly = True
atn/ATNConfigSet.py:40:        self.readonly = False
atn/ATNConfigSet.py:69:        if self.readonly:
atn/ATNConfigSet.py:117:        if self.readonly:
atn/ATNConfigSet.py:146:        if self.readonly:
atn/ATNConfigSet.py:173:        if self.readonly:
atn/ATNConfigSet.py:180:        self.readonly = readonly
atn/ATNDeserializationOptions.py:10:        self.readOnly = False
atn/ATNDeserializationOptions.py:15:        if key!="readOnly" and self.readOnly:
atn/ATNDeserializationOptions.py:20:ATNDeserializationOptions.defaultOptions.readOnly = True
atn/ParserATNSimulator.py:1596:        if not D.configs.readonly:
IntervalSet.py:32:        self.readOnly = False
```
based on my investigation:
* the reference in  `atn/ATN.py` is a typo since nextTokenWithinRule is an `IntervalSet`
* the references in `atn/ATNConfigSet.py` are not typos
* the references in `atn/ATNDeserializationOptions.py` are not typos
* the references in `atn/ParserATNSimulator.py` are not typos since the configs is an `ATNConfigSet`

suggested fix:
change `self.readOnly = False` in `IntervalSet` to `self.readonly = False`

since the java runtime uses lowercased readonly https://www.antlr.org/api/Java/org/antlr/v4/runtime/misc/IntervalSet.html#readonly

ericvergnaud:
Thanks,
suggest using lowercase everywhere.
could you submit a PR ?

PR #3228, opened 2021-07-16
[Python 2, Python 3] Change all "readOnly" to "readonly"

renzhentaxi:
Previously, in python2's runtime there was a typo where we used `readonly` when the property was actually `readOnly`.

After discussing with Eric on #3226, We decided to change all properties named `readOnly` to `readonly` since that is what the java runtime does.


ericvergnaud:
Fixes #3226

PR #3229, opened 2021-07-19
Add pretty print to Javascript toStringTree

tom-james-watson:
Non-pretty (existing functionality):

```
> console.log(antlr4.tree.Trees.toStringTree(tree, tree.parser.ruleNames));
(parse (sql_stmt_list (sql_stmt (select_stmt (select_core select (result_column (expr (column_name (any_name id)))) , (result_column (expr (column_name (any_name first_name)))) from (table_or_subquery (table_name (any_name users)))))) ;) <EOF>)
```

Pretty:

```
> console.log(antlr4.tree.Trees.toStringTree(tree, tree.parser.ruleNames, null, true));
(parse 
 (sql_stmt_list 
  (sql_stmt 
   (select_stmt 
    (select_core 
     select    
     (result_column 
      (expr 
       (column_name 
        (any_name 
         id
        )
       )
      )
     )    
     ,    
     (result_column 
      (expr 
       (column_name 
        (any_name 
         first_name
        )
       )
      )
     )    
     from    
     (table_or_subquery 
      (table_name 
       (any_name 
        users
       )
      )
     )
    )
   )
  ) 
  ;
 )
 <EOF>
)
```

### DIFF
--- a/runtime/JavaScript/src/antlr4/tree/Trees.js
+++ b/runtime/JavaScript/src/antlr4/tree/Trees.js
@@ -14,26 +14,52 @@ const Trees = {
      *  node payloads to get the text for the nodes.  Detect
      *  parse trees and extract data appropriately.
      */
-    toStringTree: function(tree, ruleNames, recog) {
+    toStringTree: function(tree, ruleNames, recog, prettyPrint, indentLvl) {
         ruleNames = ruleNames || null;
         recog = recog || null;
+        prettyPrint = prettyPrint || false;
+        indentLvl = indentLvl || 0;
+
         if(recog!==null) {
             ruleNames = recog.ruleNames;
         }
+
         let s = Trees.getNodeText(tree, ruleNames);
         s = Utils.escapeWhitespace(s, false);
         const c = tree.getChildCount();
+
         if(c===0) {
+            if (prettyPrint) {
+                return `\n${" ".repeat(indentLvl)}${s}`;
+            }
             return s;
         }
-        let res = "(" + s + ' ';
+
+        let res = "(" + s + " ";
+        if (prettyPrint) {
+            res = `\n${" ".repeat(indentLvl)}${res}`;
+        }
+
         if(c>0) {
-            s = Trees.toStringTree(tree.getChild(0), ruleNames);
+            s = Trees.toStringTree(
+                tree.getChild(0), ruleNames, recog, prettyPrint, indentLvl + 1
+            );
             res = res.concat(s);
         }
+
         for(let i=1;i<c;i++) {
-            s = Trees.toStringTree(tree.getChild(i), ruleNames);
-            res = res.concat(' ' + s);
+            s = Trees.toStringTree(
+                tree.getChild(i), ruleNames, recog, prettyPrint, indentLvl + 1
+            );
+            if (prettyPrint) {
+                res = `${res}${" ".repeat(indentLvl)}${s}`;
+            } else {
+                res = res.concat(' ' + s);
+            }
+        }
+
+        if (prettyPrint) {
+            res = `${res}\n${" ".repeat(indentLvl)}`;
         }
         res = res.concat(")");
         return res;
--- a/contributors.txt
+++ b/contributors.txt
@@ -304,3 +304,4 @@ YYYY/MM/DD, github id, Full name, email
 2021/05/06, canastro, Ricardo Canastro, ricardocanastro@users.noreply.github.com
 2021/07/01, marcauberer, Marc Auberer, marc.auberer@chillibits.com
 2021/07/14, renzhentaxibaerde, Renzhentaxi Baerde, renzhentaxibaerde@gmail.com
+2021/07/19, tom-james-watson, Thomas Watson, antlr.contributors@tomjwatson.com


runtime/JavaScript/src/antlr4/tree/Trees.js
```diff
@@ -14,26 +14,52 @@ const Trees = {
...
+            if (prettyPrint) {
+                res = `${res}${" ".repeat(indentLvl)}${s}`;
+            } else {
+                res = res.concat(' ' + s);
+            }
+        }
+
+        if (prettyPrint) {
+            res = `${res}\n${" ".repeat(indentLvl)}`;
         }
```
ericvergnaud, line 63: missing else

runtime/JavaScript/src/antlr4/tree/Trees.js
```diff
@@ -14,26 +14,52 @@ const Trees = {
...
         if(c===0) {
+            if (prettyPrint) {
+                return `\n${" ".repeat(indentLvl)}${s}`;
+            }
             return s;
         }
-        let res = "(" + s + ' ';
+
+        let res = "(" + s + " ";
+        if (prettyPrint) {
```
ericvergnaud, line 39: missing else

runtime/JavaScript/src/antlr4/tree/Trees.js
```diff
@@ -14,26 +14,52 @@ const Trees = {
...
         }
+
         let s = Trees.getNodeText(tree, ruleNames);
         s = Utils.escapeWhitespace(s, false);
         const c = tree.getChildCount();
+
         if(c===0) {
+            if (prettyPrint) {
+                return `\n${" ".repeat(indentLvl)}${s}`;
+            }
```
ericvergnaud, line 34: missing else

tom-james-watson:
None of those instances need an `else`. They are simply adding newlines/indentation where necessary.

ericvergnaud:
> None of those instances need an `else`. They are simply adding newlines/indentation where necessary.

Hi, they do for clarity.

tom-james-watson:
I can't say I agree that that would make anything clearer. I can do it, but it will simply mean that there is unnecessary duplication.

tom-james-watson:
Instead of

```
let res = "(" + s + " ";
if (prettyPrint) {
  res = `\n${" ".repeat(indentLvl)}${res}`;
}
```

You'd have

```
let res;
if (prettyPrint) {
  res = `\n${" ".repeat(indentLvl)}(${s} `;
} else {
  res = "(" + s + " ";
}
```

ericvergnaud:
> Instead of
> 
> ```
> let res = "(" + s + " ";
> if (prettyPrint) {
>   res = `\n${" ".repeat(indentLvl)}${res}`;
> }
> ```
> 
> You'd have
> 
> ```
> let res;
> if (prettyPrint) {
>   res = `\n${" ".repeat(indentLvl)}(${s} `;
> } else {
>   res = "(" + s + " ";
> }
> ```

or:

const res = prettyPrint ? `\n${" ".repeat(indentLvl)}(${s} `: "(" + s + " ";

another option would be to have toStringTree dispatch to toUglyStringTree/toPrettyStringTree such that the maintainer does not have to think twice about side effects when evolving the formatting

KvanTTT:
I suggest not using parentheses form for the prettified tree:

```
parse 
 sql_stmt_list 
  sql_stmt 
   select_stmt 
    select_core 
     select    
     result_column 
      expr 
       column_name 
        any_name 
         id
     ,    
     result_column 
      expr 
       column_name 
        any_name 
         first_name
     from    
     table_or_subquery 
      table_name 
       any_name 
        users
  ;
 <EOF>
```

Also, I suggest adding a similar method to all runtimes, not only to JavaScript.

tom-james-watson:
I was just going off the similar PR that was merged for C++: https://github.com/antlr/antlr4/pull/2505

KvanTTT:
Maybe it makes sense to add a method for prettified output (my suggestion above) and method that outputs resulting JSON that can be easy serialized/deserialized.

parrt:
Have you guys come to a consensus?

KvanTTT:
If add the feature to JavaScript target it should be added to all other targets as well for consistency.

parrt:
> If add the feature to JavaScript target it should be added to all other targets as well for consistency.

yeah that sounds like a big change that I don't want to focus on. Maybe we just make that function available as part of the documentation for JavaScript?

KvanTTT:
> Maybe we just make that function available as part of the documentation for JavaScript?

I don't know but I'm not sure it's a very demanded feature for the current time. Especially considering that tree in any format can be obtained using a utility method in runtime.

Also, not sure about lisp-like format. I would think about something serializable at first (JSON).

parrt:
> Also, not sure about lisp-like format. I would think about something serializable at first (JSON).

The lisp thing is good for showing the structure but I still prefer the GUI visualization using the intellij plug-in (all hail jetbrains!)

KvanTTT:
> The lisp thing is good for showing the structure

Maybe it's better for showing than JSON. But [my suggested format](https://github.com/antlr/antlr4/pull/3229#issuecomment-884191247) with indents and without parentheses is also good and is shorter. That's why I don't want to rush with accepting a similar request because pretty print formats may differ and may be subjective. On the other hand, JSON format can be converted to almost everything else.

tom-james-watson:
I'm no longer using antlr at all and have no bandwidth to address this, so will close out. Of course feel free to repurpose my changes if you want them.

Issue #3230, opened 2021-07-20
Python3 lexer file has implicit relative import of parser, which is not supported by Python3


SkittlePox:
When antlr4 generates python3 lexer and parser files, it uses an *implicit* relative import of the parser from the lexer file:
```
# RLangLexer.py
...
from RLangParser import RLangParser
...
```
Apparently, this kind of import is a no-go for Python3, and at least `pytest` breaks when a test uses an imported file which itself tries to make an implicit relative import. The solution is simple, add a period to make it an *explicit* relative import:
```
# RLangLexer.py
...
from .RLangParser import RLangParser
...
```
This makes running `pytest` possible (and I presume running a listener which uses the lexer from outside the package). However, it breaks `pygrun` with an `ImportError`:
```
...
ImportError: attempted relative import with no known parent package
```
No fear, a simple fix to the`pygrun` file which updates the `__package__` global with the current working directory allows for `pygrun` to work with both implicit and explicit relative imports:
```
# pygrun
102     sys.path.append('.')
103     #print(sys.path)
104     p = os.path.basename(os.getcwd())        # New Code
105     globals().update({'__package__': p})     # New Code
106     
107     # print("Load Lexer {}".format(lexerName))
108     module_lexer = __import__(lexerName, globals(), locals(), lexerName, 1)
109     class_lexer = getattr(module_lexer, lexerName)
```

I can add the changes to `pygrun` to support these imports with a pull request, but I'm not sure how to change the initial generation of implicit relative imports in the lexer (which would add a `.` to the beginning of the parser module) without a post-processing script which adds a `.`. That would be quite messy. Is there an internal fix?

ericvergnaud:
Hi,

the file to modify is:
 tool/resources/org/antlr/v4/tool/templates/codegen/Python3/Python3.stg

Eric

SkittlePox:
Thanks Eric,

It appears that the listener template in Python3.stg has a working import rule:
```
...
ListenerFile(file, header, namedActions) ::= <<
<fileHeader(file.grammarFileName, file.ANTLRVersion)>
from antlr4 import *
if __name__ is not None and "." in __name__:
    from .<file.parserName> import <file.parserName>
else:
    from <file.parserName> import <file.parserName>
<header>
...
```
However the lexer file template doesn't even look like it imports the parser:
```
...
LexerFile(lexerFile, lexer, namedActions) ::= <<
[...]

PR #3232, opened 2021-07-21
Python local import fixes

SkittlePox:
This should allow the `pygrun` executable to run lexer files with both implicit and explicit relative imports. It is a partial fix to issue #3230. This **does not** fix the `Python3.stg` template which causes the implicit relative import in the generated lexer file.


SkittlePox:
This indeed does fix #3230 in full. Checks not passed because of `Failed to download action` errors:
```
Download action repository 'actions/checkout@v2'
Download action repository 'actions/setup-java@v1'
Download action repository 'stCarolas/setup-maven@v4'
Warning: Failed to download action 'https://api.github.com/repos/stCarolas/setup-maven/tarball/1d56b37995622db66cce1214d81014b09807fb5a'. Error: A task was canceled.
Warning: Back off 26.298 seconds before retry.
Warning: Failed to download action 'https://api.github.com/repos/stCarolas/setup-maven/tarball/1d56b37995622db66cce1214d81014b09807fb5a'. Error: The operation was canceled.
Warning: Back off 21.457 seconds before retry.
Error: The operation was canceled.
```
Re-run tests and they should pass if the actions download.

ericvergnaud:
Can you also fix Python 2 ?

SkittlePox:
Okay, looks like I fixed a bug which is only an issue for Python3 and not Python2. Implicit relative imports were removed in Python3, but are acceptable in Python2, meaning the `pygrun` in Python2 should work just fine.

parrt:
Is this blessed @ericvergnaud ?

ericvergnaud:
@parrt blessed

parrt:
@SkittlePox  could you resolve the conflict, that just got introduced, and then I can merge? thank you!

Issue #3236, opened 2021-07-28
Python runtime is inconsistent with Java


Vladimir-Chan:
Hello, I am analyzing c# source code with the grammar files in [grammar repository](https://github.com/antlr/grammars-v4/tree/master/csharp) by Python target. I converted the grammar actions to Python source. When I tested the grammar by a simple example, I got ParseCancellationException. After debugging, I find that the method `nextTokenOnChannel` in class `BufferedTokenStream` is inconsistent with Java. In Python, the comment says

> Return -1 if there are no tokens on channel between i and EOF.

While in Java, it says

> Return the index of the EOF token if there are no tokens on channel between {@code i} and EOF.

So, which is correct? Please give a consistent version, thx.

ericvergnaud:
It's an inconsistency between the various runtimes.
Seems the 16/06/2014 change in the Java version was never ported to Python.
Might affect other runtimes, depending on which code base they used as a starting point.

ericvergnaud:
Feel free to submit a PR

Vladimir-Chan:
Ok I have pulled request

PR #3241, opened 2021-07-31
keep consistent with Java

Vladimir-Chan:
Fix for #3236


ericvergnaud:
Hi
can you check the Python2 version too ?

Vladimir-Chan:
> Hi
> can you check the Python2 version too ?

I don't check the whole Python3 runtime. I just got a exception and found this issue. I can only check if same issue appear in the Python2 runtime.

ericvergnaud:
> > Hi
> > can you check the Python2 version too ?
> 
> I don't check the whole Python3 runtime. I just got a exception and found this issue. I can only check if same issue appear in the Python2 runtime.

Yes please, let's ensure both Python runtimes are consistent

Issue #2152, opened 2017-12-07
Golang target extremly slow, while Java target is fast on same input


Lacuno:
I need to parse MySQL for a project and use https://github.com/antlr/grammars-v4/tree/master/mysql
this grammar with the newest ANTLR from the master branch (4.7.1-SNAPSHOT).
On certain queries the parser spends a lot of time in the golang target (25-27sec), while the Java target is much faster (700ms-1sec). The IntelliJ Plugin even parses faster (~300ms).
Example query: 
  
    SELECT SUM((IFNULL(X, 0) - IFNULL(Y, 0) - IFNULL(Z, 0) - (IFNULL(XX, 0) - IFNULL(YY, 0) - IFNULL(ZZ, 0))) * IFNULL(ASDF, 0)) AS `REVENUE` FROM SALES_FLAT

I used SLL PredicionMode on both targets, but that does not help.
I also ran the query multiple times, but warming up also didn't help much.
Afterwards I profiled the Go target and got the following output for my query mentioned above:

      flat  flat%   sum%        cum   cum%
    5980ms 14.28% 14.28%    16020ms 38.26%  runtime.scanobject C:\dev\Go\src\runtime\mgcmark.go
    4970ms 11.87% 26.15%     5430ms 12.97%  runtime.heapBitsForObject C:\dev\Go\src\runtime\mbitmap.go
    3280ms  7.83% 33.99%     8070ms 19.27%  github.com/antlr/antlr4/runtime/Go/antlr.(*BaseSingletonPredictionContext).hash C:\dev\projects\go\src\github.com\antlr\antlr4\runtime\Go\antlr\prediction_context.go
    2620ms  6.26% 40.24%     2620ms  6.26%  github.com/antlr/antlr4/runtime/Go/antlr.murmurUpdate C:\dev\projects\go\src\github.com\antlr\antlr4\runtime\Go\antlr\utils.go (inline)
    2070ms  4.94% 45.19%     2070ms  4.94%  github.com/antlr/antlr4/runtime/Go/antlr.murmurFinish C:\dev\projects\go\src\github.com\antlr\antlr4\runtime\Go\antlr\utils.go (inline)
    1240ms  2.96% 48.15%     1240ms  2.96%  runtime.markBits.isMarked C:\dev\Go\src\runtime\mbitmap.go (inline)
     980ms  2.34% 50.49%     1540ms  3.68%  runtime.evacuate C:\dev\Go\src\runtime\hashmap.go
[...]

PR #3243, opened 2021-08-05
fix: use 32bit murmur3 same with java, implement array2dhashset which…

jjeffcaii:
Hi All:

This PR fixes several serious performance problem in Go runtime:
1. the `murmur3` implementation in Go SDK has some problem, it's not consistent with Java.
2. some hash code implementation is not correct, some of them are not consistent with Java.
3. the `Set` implementation which based on `map[int][]interface{}` has poor performance, it is not consistent with Java,  I implement the `Array2DHashSet` instead, this implementation is same with Java.
4. the hash cache is now fixed for `PredictionContext`.

Many problems have been raised, and I think this PR will fix them, please check these issues:
https://github.com/antlr/antlr4/issues/2888
https://github.com/antlr/grammars-v4/issues/2272
https://github.com/antlr/antlr4/issues/2152

The core problem in Golang runtime: it wastes too much time on hash code computing because of the incorrect `Murmur3` and `Set` and `hasher` implementation. Compare the Java and the original Go runtime, Go computes murmur3 *~80,000,000* times, but the Java runtime only *~600,000* times.

Here's my benchmark in my MBP '16 2019, the g4 files are copied from https://github.com/antlr/grammars-v4/tree/master/sql/mysql/Positive-Technologies, the original Go runtime cost *4.8s*, and this PR costs *171ms*.

Benchmark:

```go
	begin := time.Now()
	var sql string

	sql = "SELECT DATE_FORMAT(IF(COUNT(*)>0,'2021-01-02','2020-01-01'),'%Y') AS cc FROM student WHERE uid BETWEEN 1 AND 5"
	//sql = "SELECT 1"

	is := antlr.NewInputStream(sql)
	lexer := p.NewMySqlLexer(antlrx.NewCaseChangingStream(is, true))
	//lexer := p.NewMySqlLexer(is)

	ts := antlr.NewCommonTokenStream(lexer, antlr.TokenDefaultChannel)

	mp := p.NewMySqlParser(ts)
	//mp.GetInterpreter().SetPredictionMode(antlr.PredictionModeSLL)
	mp.RemoveErrorListeners()
	mp.SetErrorHandler(antlr.NewBailErrorStrategy())

	_ = mp.Root()

	fmt.Println("cost:", time.Since(begin)) 
        // COST:
        // BEFORE: 4.853549529s
        // AFTER: 171ms
        // JAVA: 20ms
```

### DIFF
--- a/contributors.txt
+++ b/contributors.txt
@@ -305,3 +305,4 @@ YYYY/MM/DD, github id, Full name, email
 2021/07/01, marcauberer, Marc Auberer, marc.auberer@chillibits.com
 2021/07/14, renzhentaxibaerde, Renzhentaxi Baerde, renzhentaxibaerde@gmail.com
 2021/08/02, minjoosur, Minjoo Sur, msur@salesforce.com
+2021/08/05, jjeffcaii, Jeff Tsai, caiweiwei.cww@alibaba-inc.com
--- a/runtime/Go/antlr/atn_config.go
+++ b/runtime/Go/antlr/atn_config.go
@@ -251,7 +251,7 @@ func (l *LexerATNConfig) hash() int {
 		f = 0
 	}
 	h := murmurInit(7)
-	h = murmurUpdate(h, l.state.hash())
+	h = murmurUpdate(h, l.state.GetStateNumber())
 	h = murmurUpdate(h, l.alt)
 	h = murmurUpdate(h, l.context.hash())
 	h = murmurUpdate(h, l.semanticContext.hash())
--- a/runtime/Go/antlr/atn_config_set.go
+++ b/runtime/Go/antlr/atn_config_set.go
@@ -11,7 +11,7 @@ type ATNConfigSet interface {
 	Add(ATNConfig, *DoubleDict) bool
 	AddAll([]ATNConfig) bool
 
-	GetStates() *Set
+	GetStates() Set
 	GetPredicates() []SemanticContext
 	GetItems() []ATNConfig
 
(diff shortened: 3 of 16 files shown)


parrt:
Can somebody in Go area review? @davesisson @easonlin404 @pboyer ?

jjeffcaii:
Is anyone focusing on it?

parrt:
Can any of the Go folks take a look @davesisson @easonlin404 @pboyer ?   I'm afraid I'm not qualified

KvanTTT:
Maybe somebody else can review, not necessary Go folks? @ericvergnaud ?

ericvergnaud:
I'm a bit skeptical re the inconsistent use of MurmurHash (why is not used by hashATNConfig ?).
That said, I believe in tests ans as long as the tests pass and there is a performance improvement I can't think of any reason to reject this PR.

jjeffcaii:
@ericvergnaud Thanks for the review.

The main difference of murmur3 is that I changed it to 32bit like the java codes, and the `hashATNConfig` is copied from java runtime:

https://github.com/antlr/antlr4/blob/ce3c483ec50592d8e265c644e31ec5e8261c87fb/runtime/Java/src/org/antlr/v4/runtime/atn/ATNConfigSet.java#L40-L63

To be honest, I am not familiar with the core antlr implementation, most codes were translated from the Java runtime, try my best. Hope it can help those who are using the Go runtime.

parrt:
I concur with @ericvergnaud .  The tests pass and improves performance; it also brings it in line with what ANTLR core does. Sam Harwell worked really hard to get that performance up, and using that hash function was important. merging.

PR #3246, opened 2021-08-08
[Dart] Add case changing char stream sample

lingyv-li:

### DIFF
--- a/doc/resources/case_changing_char_stream.dart
+++ b/doc/resources/case_changing_char_stream.dart
@@ -1,0 +1,56 @@
+// @dart=2.12
+
+import '../../runtime/Dart/lib/antlr4.dart';
+import '../../runtime/Dart/lib/src/interval_set.dart';
+
+/// This class supports case-insensitive lexing by wrapping an existing
+/// {@link CharStream} and forcing the lexer to see either upper or
+/// lowercase characters. Grammar literals should then be either upper or
+/// lower case such as 'BEGIN' or 'begin'. The text of the character
+/// stream is unaffected. Example: input 'BeGiN' would match lexer rule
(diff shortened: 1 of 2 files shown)


doc/resources/case_changing_char_stream.dart
```diff
@@ -0,0 +1,56 @@
...
+  CaseChangingCharStream(this.stream, this.upper);
+
+  @override
+  int? LA(int i) {
+    int? c = stream.LA(i);
+    if (c == null || c <= 0) {
+      return c;
+    }
+    if (upper) {
+      return String.fromCharCode(c).toUpperCase().codeUnitAt(0);
```
KvanTTT, line 28: Please use something that doesn't change the length of the output character because it creates problems for mapping to input characters and actually such characters are not used in programming languages.

Your approach change `ß` to `SS` (2 chars), I've checked in sandbox:

```dart
void main() {
  print("ß".toUpperCase());
}
```

For instance, you can check the length of the output character, and if it is more than 1 then return the original char. See detail in [my comment](https://github.com/antlr/antlr4/pull/2046#issuecomment-917650686) in a thread about case insensitivity.

KvanTTT (changes requested):

doc/resources/case_changing_char_stream.dart
lingyv-li: Thanks for the comment, that's a interesting discussion.
In this case it doesn't change the string length because of the `.codeUnitAt(0)` at the end, but your solution is better. Dart is also a bit special where its behaviour is inconsistent between JS and VM: https://github.com/dart-lang/sdk/issues/6706

PR #3249, opened 2021-08-08
Support Dart multiple platforms builds.

ansiemens:
[Dart] Support Dart multiple platforms.

- Uses conditional imports / exports on `dart:io`.
- Defines 32bit version bit set operations as JS version does not operate on 64bit integers.
- Migrates from `dart2native` to `dart compile exe`.
- Runs `dartfmt` on Dart runtime.


ansiemens:
> Support Dart multiple platforms.
> 
> * Uses conditional imports / exports on `dart:io`.
> * Defines 32bit version bit set operations as JS version does not operate on 64bit integers.
> * Migrates from `dart2native` to `dart compile exe`.
> * Runs `dartfmt` on Dart runtime.

PR #3256, opened 2021-08-15
[Cpp] Add option for different ANTLRErrorListener

jnooree:
Currently, a `Parser` or `Lexer` could only be instantiated with the default `ConsoleErrorListener` instance. However, it would be better to have an option for a custom `ErrorListener` object at the instantiation time, for both performance and readability.

This commit will add a parameter named `listener`, which has the default parameter of `&ConsoleErrorListener::INSTANCE` for API stability.


mike-lischke:
This patch will change the recognizer (parser/lexer) signatures in a way that is not compatible with the other targets, for a pretty small benefit. Also, it will lead to 2 paths of adding an error listener (via c-tor and via member function). I'd vote for it if there was more benefit, but the effort outweighs that. Keep in mind, you should also add tests for this patch.

mike-lischke:
@parrt This patch is rejected and can be closed.

PR #3257, opened 2021-08-17
Improve Travis CI build Performance

YunLemon:
According to [Build times out because no output was received](https://docs.travis-ci.com/user/common-build-problems/#build-times-out-because-no-output-was-received), we should carefully use travis_wait, as it may make the build unstable and extend the build time.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -32,7 +32,7 @@ before_install:
 script:
   - |
     cd runtime-testsuite;
-    travis_wait 40 ../.travis/run-tests-$TARGET.sh
+ ../.travis/run-tests-$TARGET.sh
     rc=$?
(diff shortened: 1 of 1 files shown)


parrt:
Interesting. I didn't know that there was such an option. @ericvergnaud I think we mostly have the build completing now. Should we increase wait time a little bit? Actually, I turned off the quiet option to make sure output is generated which might be the issue what time out.

ericvergnaud:
afaik the build generally fails because of insufficient build credits
my recommendation would be to get Swift/Linux running on CircleCI

parrt:
Thanks @YunLemon  I think I will close this because we are working on the build system at the moment and will continue trying to get it to finish correctly.

PR #3261, opened 2021-08-25
Upgrade com.ibm.icu:icu4j to latest (69.1) to avoid vulnerabilities it brings

XenoAmess:
Also change the template to suite this upgrade, as this upgrade makes the java class larger than jvm can accept.


XenoAmess:
> Looks like some errors in the Java tests...

@part

I don't think the errors are from this pr

will rebase it on latest master and try again.

Issue #3272, opened 2021-09-12
Option `-package` allows identifier with space and other not allowed chars


KvanTTT:
For instance, `-package "invalid package"` generates not compilable code.

I suggest using the following regex: `^([a-zA-Z_][a-zA-Z\d_]*)$` for packages validation.

haeungun:
Can I take a look?

KvanTTT:
You wanna fix the issue? I'm all for it, but I'm not a maintainer of this repository.

PR #3273, opened 2021-09-13
Option `encoding` only allows valid input

haeungun:
Issue: https://github.com/antlr/antlr4/issues/3272

### DIFF
--- a/tool/src/org/antlr/v4/Tool.java
+++ b/tool/src/org/antlr/v4/Tool.java
@@ -64,6 +64,7 @@ import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
 import java.util.concurrent.CopyOnWriteArrayList;
+import java.util.regex.Pattern;
 
 public class Tool {
 	public static final String VERSION;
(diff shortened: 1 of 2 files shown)


tool/src/org/antlr/v4/Tool.java
```diff
@@ -79,6 +80,8 @@
 	public static final List<String> ALL_GRAMMAR_EXTENSIONS =
 		Collections.unmodifiableList(Arrays.asList(GRAMMAR_EXTENSION, LEGACY_GRAMMAR_EXTENSION));
 
+	public static final Pattern ENCODING_PATTERN = Pattern.compile("([a-zA-Z]+[a-zA-Z0-9_-]*)\\w+");
```
KvanTTT, line 83: This regex is only suitable for `-package` option, not for `-encoding`

tool/src/org/antlr/v4/Tool.java
```diff
@@ -224,6 +227,12 @@ protected void handleArgs() {
 						argValue = args[i];
 						i++;
 					}
+					if ( "grammarEncoding".equals(o.fieldName) ) {
```
KvanTTT, line 230: This is not `encoding` nor `package`.

KvanTTT (changes requested):
Also, tests are missing.

PR #3276, opened 2021-09-14
[Swift] Changes to Mutex-es to guard the staticly shared [DFA]

### LINKED ISSUE #3271: Swift Target Crashes with Multi-Threading

# Reproducing Error

Running via Xcode for iOS 13
Project Swift Version: 4.2
Compiled swift target from 4.9.2 release

## Grammar
```
grammar math;

NUMBER : [0-9] ;
WS : [ \r\n\t] + -> skip ;

operation : l=NUMBER op='+' r=NUMBER  ;
```

## Swift Code
```swift
for _ in 0...10 {
    DispatchQueue.global().async {
        let chars = ANTLRInputStream("1+1")
        let lexer = mathLexer(chars)
        
        let tokenStream = CommonTokenStream(lexer)
        let parser = try! mathParser(tokenStream)
        
        let node = try? parser.operation()
    }
}
```

# What Happened

The program crashes in LexerATNSimulator.swift on line 721 with EXC_BAD_ACCESS. There was a double free at some point in the stack.

Any help is greatly appreciated! Hopefully this is just user error.

martinvw:
Request for review: @janyou, @ewanmellor, @hanjoes, @rmehta33 

By sharing the mutex in the same way as the DFA array it becomes possible to run in parallel

Fixes #3271

### DIFF
--- a/runtime/Swift/Sources/Antlr4/atn/LexerATNSimulator.swift
+++ b/runtime/Swift/Sources/Antlr4/atn/LexerATNSimulator.swift
@@ -72,16 +72,6 @@ open class LexerATNSimulator: ATNSimulator {
     public private(set) final var decisionToDFA: [DFA]
     
     internal var mode = Lexer.DEFAULT_MODE
-    
-    /// 
-    /// mutex for DFAState change
-    /// 
-    private let dfaStateMutex = Mutex()
-    
-    /// 
-    /// mutex for changes to all DFAStates map
-    /// 
-    private let dfaStatesMutex = Mutex()
 
     /// 
     /// Used during DFA/ATN exec to record the most recent accept configuration info
@@ -675,7 +665,7 @@ open class LexerATNSimulator: ATNSimulator {
             print("EDGE \(p) -> \(q) upon \(t)")
         }
 
-        dfaStateMutex.synchronized {
+        p.mutex.synchronized {
             if p.edges == nil {
                 //  make room for tokens 1..n and -1 masquerading as index 0
                 p.edges = [DFAState?](repeating: nil, count: LexerATNSimulator.MAX_DFA_EDGE - LexerATNSimulator.MIN_DFA_EDGE + 1)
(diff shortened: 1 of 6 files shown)


ericvergnaud:
Hi,

this sounds useful but I'm not sure your test is 'hungry' enough to prove that it works.
there is a genuine possibility that the first iteration ends before the second one starts,
plus they both generate the same DFA enrichment so might be missing important scenarios.
May I suggest that you try parsing longer and different inputs in parallel such that the test does proves that it works in the problematic scenarios too ?

martinvw:
Without my changes this test always fails and with it always succeeds. I do believe we should aim for a bit more code coverage but I doubt that multiple or much more complex grammar is needed, however a longer text of course make sense.

I do not completely understand the second part of your comment, I might need a hint about that one.

Thanks for your time!

ericvergnaud:
The reason a mutex is required is because originally the DFA is empty and gets populated gradually as the runtimes bumps into new grammar fragments. No mutex is required for reading.
Since the current test always parses the same expression, the DFA only gets populated once with a single exception where 2 threads are populating it the same way at the same time (which you seem ti have encountered)
In real situations It’s much more likely that 2 threads will populate it in different ways, thus going through slightly different code paths, hence my ask that you use longer and different expressions 

Envoyé de mon iPhone

martinvw:
Thanks for the explanation!

That makes a lot of sense, thanks for insight. I’ll try to look into it
later this week.

Could you also shed some light on when the ParserATNSimulator code path is
triggered? I do believe my current tests might not cover this as well.

martinvw:
@ericvergnaud I updated the test with a bit more complex expressions, do you think this would be enough:

https://github.com/antlr/antlr4/pull/3276/files#diff-3a18ea2989cadcaf86cc213acaeeac01f23da4167ce1a7031bd19ff3aa27ef86R16-R38

I check the call trace and I see that we actually touch all the changed files with the current test, so that looks good.

ericvergnaud:
@parrt blessed, the failing test is unrelated and is fixed through another PR

martinvw:
Thanks!

PR #3284, opened 2021-09-24
[Go] Fix null ptr dereference

### LINKED ISSUE #3254: Go target, cannot use superClass for the lexer grammar!
I'm trying to use the superClass option in a lexer grammar. It works fine for a parser grammar, but for a lexer grammar, it fails at runtime with a null pointer dereference in the generated constructor NewFOOBARLexer().

For example,

	type CSharpLexer struct {
		*CSharpLexerBase
		channelNames []string
		modeNames    []string
		// TODO: EOF string
	}

	func NewCSharpLexer(input antlr.CharStream) *CSharpLexer {
		l := new(CSharpLexer)
		...
		l.BaseLexer = antlr.NewBaseLexer(input)  <<<<<< CRASHES HERE WITH NULL PTR DEREF.
		...
		return l
	}

It crashes on `l.BaseLexer = antlr.NewBaseLexer(input)`, specifically on the assignment to `l.BaseLexer`.

Contrast this with the parser code:

	type CSharpParser struct {
		CSharpParserBase
	}

	func NewCSharpParser(input antlr.TokenStream) *CSharpParser {
		this := new(CSharpParser)
		...
		this.BaseParser = antlr.NewBaseParser(input)
		...
		return this
	}

This code fails because the `type CSharpLexer struct` is wrong. It defines it to contain a pointer `*CSharpLexerBase`, whereas in the parser `type CSharpParser struct`, it defines it directly `CSharpParserBase`. Changing the type struct decl for the generated lexer to `CSharpLexerBase` instead of `*CSharpLexerBase` fixes the crash, and my parser/lexer work just fine.

The problem is [here](CSharpLexerBase)-- the line erroneously contains `*`, whereas for the parser [here](https://github.com/antlr/antlr4/blob/b915242cd638fbb3ff5c14a05600abc3b2e2533c/tool/resources/org/antlr/v4/tool/templates/codegen/Go/Go.stg#L180), it is declared directly.

SKalt:
Fixes #3254; this is the exact fix recommended by @kaby76.


kaby76:
This looks good--it does what is exactly done for superClass in the parser. Thank you for fixing this.

SKalt:
I'm getting one failing test with the message

```sh
Running org.antlr.v4.test.runtime.csharp.TestLeftRecursion

Too long with no output (exceeded 10m0s): context deadline exceeded
```

I don't think that test is actually running my changes (I'd expect `runtime.csharp` is using the c# runtime, not  parsing c# using go). I'm going to rebase on the development branch and retry the tests.

PR #3286, opened 2021-09-26
Remove `late` modifier for Dart generated context classes

idraper:
When antlr generates code using the Dart runtime, it creates context classes that use the `late` modifier on labeled grammar elements. For example:

```
term : EOF_=EOF?;
```

would generate a term context class where `EOF_` would be declared as:

```
late Token? EOF_;
```

However, in the above example, there is the possibility that `EOF_` would not be set. This is an error in Dart at runtime when trying to access the `EOF_` member since the `late` keyword means Dart expects the value to be assigned before use. The `late` modifier should be removed since the type is already nullable and thus is not required to be assigned (see [here](https://dart.dev/null-safety/understanding-null-safety) for details about null safety with `late`).

Thus, the runtime should just generate:

```
Token? EOF_;
```


lingyv-li:
Thanks for your contribution. Can you please provide a reproducible example of when this would fail?

@renancaraujo @canastro can you please take a look at this PR? Thanks

parrt:
@lingyv-li  if you think it looks correct and useful, we can merge.

lingyv-li:
@parrt Yes this LGTM

PR #3290, opened 2021-09-28
add heartbeat to csharp/recursion

ericvergnaud:

### DIFF
--- a/runtime-testsuite/test/org/antlr/v4/test/runtime/BaseRuntimeTest.java
+++ b/runtime-testsuite/test/org/antlr/v4/test/runtime/BaseRuntimeTest.java
@@ -62,7 +62,7 @@ public abstract class BaseRuntimeTest {
 	}
 
 	private static boolean requiresHeartbeat() {
-		return isTravisCI() || isAppVeyorCI() || (isCPP() && isRecursion()) || (isCircleCI() && isGo());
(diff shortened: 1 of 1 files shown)


runtime-testsuite/test/org/antlr/v4/test/runtime/BaseRuntimeTest.java
```diff
@@ -62,7 +62,7 @@ public static void startHeartbeatToAvoidTimeout() {
 	}
 
 	private static boolean requiresHeartbeat() {
-		return isTravisCI() || isAppVeyorCI() || (isCPP() && isRecursion()) || (isCircleCI() && isGo());
+		return isTravisCI() || isAppVeyorCI() || (isCPP() && isRecursion()) || (isCircleCI() && isGo())|| (isCircleCI() && isCSharp() && isRecursion());
```
KvanTTT, line 65: Minor: `&&` has higher priority than `||`. Thus, parentheses are excess:

Also, it looks more clear if it's splitted on several lines:

```suggestion
		return isTravisCI() ||
		       isAppVeyorCI() ||
		       isCPP() && isRecursion() ||
		       isCircleCI() && isGo() ||
		       isCircleCI() && isCSharp() && isRecursion();
```

PR #3299, opened 2021-10-10
Go visitor pattern

sheik:
Having a Go at enabling the visitor pattern. 

New output does not require you to write the Visit() function, and the Visit() function does not consist of a switch statement. Instead, the type system is used to pick which function to distribute the visitor to.

### DIFF
--- a/contributors.txt
+++ b/contributors.txt
@@ -286,10 +286,12 @@ YYYY/MM/DD, github id, Full name, email
 2020/11/26, mr-c, Michael R. Crusoe, 1330696+mr-c@users.noreply.github.com
 2020/12/01, maxence-lefebvre, Maxence Lefebvre, maxence-lefebvre@users.noreply.github.com
 2020/12/03, electrum, David Phillips, david@acz.org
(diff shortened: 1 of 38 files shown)


sheik:
Here is an example project showing how to use the output:

https://github.com/sheik/antlr4-go-visitor-test

This is an example grammar from the definitive ANTLR4 reference.

The main files of interest are main.go and parser/expr_visitor.go. 

You can see the implementation of Visit functions in parser/expr_base_visitor.go.

parrt:
@jcking do you feel like taking a look at this PR?

jcking:
Hmmm.

So "generics" are landing in Go in the next couple of months I believe. I would hate to use `interface{}` and then have some nice generic approach available and force migration. We should double check whether generics will be usable in this situation.

parrt:
Yeah, let's wait.

dragonfax:
I started a comment on this and just made it an issue. Since it seems thats where this discussion is heading.
https://github.com/antlr/antlr4/issues/3457

PR #3302, opened 2021-10-12
[Swift] improvements to equality functions

Dante-Broggi:
NFC, except for fixing reflexivity for pattern/Chunk


parrt:
Hi @Dante-Broggi  some of these changes to equality look to cut out a lot of functionality. Is that because swift has a more sophisticated equality notion?

Dante-Broggi:
Yes, Swift Optional propagates `==` just as the code was doing longhand.

PR #3304, opened 2021-10-12
[Swift] Use proper constants

Dante-Broggi:
changes SUPPRESS_PRECEDENCE_FILTER to be a static (type) variable, not an instance variable.


parrt:
Hi! Thanks. Are these Constants consistent with the other target?

Dante-Broggi:
These constants all have the same integer values in other language runtimes, yes.

PR #3309, opened 2021-10-15
Fix some thread sanitizer warnings in C++ runtime

jcking:
There are some shared global states in the C++ runtime, specifically in the ATN, that cause thread sanitizer warnings. This pull request switches those use `std::atomic`.


mike-lischke (approved):

mike-lischke:
@parrt Another C++ only patch, ready to be merged.

PR #3313, opened 2021-10-18
Use atomic instead of a shared mutex for DFA.s0

jcking:
A shared mutex is kind of heavyweight when we are essentially replicating `volatile` from Java which is ultimately an atomic load/store. This change switches Go to use the equivalent, atomic.StorePointer and atomic.LoadPointer.


parrt:
that sounds right to me as we did very little locking in java version.  Particularly when all we did was switch a pointer.

PR #3327, opened 2021-10-28
[C++] Upstream various performance, build, and best practice fixes

jcking:
Pushing upstream a bunch of performance, build, and best practice fixes.

Most of the performance fixes are related to incorrect/misused auto. Lots of places simply use `auto` which causes values to be copied. They have been switched to `const auto&` which avoids copying-by-value. The best practice fix is switching to throw `RuntimeException` instead of a string literal.

### DIFF
--- a/runtime/Cpp/runtime/src/TokenStreamRewriter.cpp
+++ b/runtime/Cpp/runtime/src/TokenStreamRewriter.cpp
@@ -93,7 +93,7 @@ TokenStreamRewriter::TokenStreamRewriter(TokenStream *tokens_) : tokens(tokens_)
 }
 
 TokenStreamRewriter::~TokenStreamRewriter() {
-  for (auto program : _programs) {
+  for (const auto &program : _programs) {
     for (auto *operation : program.second) {
       delete operation;
     }
--- a/runtime/Cpp/runtime/src/TokenStreamRewriter.h
+++ b/runtime/Cpp/runtime/src/TokenStreamRewriter.h
@@ -5,6 +5,8 @@
 
 #pragma once
 
+#include "antlr4-common.h"
+
 namespace antlr4 {
 
   /**
--- a/runtime/Cpp/runtime/src/atn/ATNConfig.h
+++ b/runtime/Cpp/runtime/src/atn/ATNConfig.h
@@ -5,6 +5,8 @@
 
 #pragma once
 
+#include "antlr4-common.h"
+
 namespace antlr4 {
 namespace atn {
 
--- a/runtime/Cpp/runtime/src/atn/ATNSerializer.h
+++ b/runtime/Cpp/runtime/src/atn/ATNSerializer.h
@@ -5,6 +5,8 @@
 
 #pragma once
 
+#include "antlr4-common.h"
+
 namespace antlr4 {
 namespace atn {
 
--- a/runtime/Cpp/runtime/src/misc/InterpreterDataReader.h
+++ b/runtime/Cpp/runtime/src/misc/InterpreterDataReader.h
@@ -6,6 +6,8 @@
 #pragma once
 
 #include "antlr4-common.h"
+#include "atn/ATN.h"
+#include "Vocabulary.h"
 
 namespace antlr4 {
 namespace misc {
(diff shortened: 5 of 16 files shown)


runtime/Cpp/runtime/src/TokenStreamRewriter.h
```diff
@@ -5,6 +5,8 @@
 
 #pragma once
 
+#include "antlr4-common.h"
```
mike-lischke, line 8: Why's that here now?

runtime/Cpp/runtime/src/atn/ATNConfig.h
```diff
@@ -5,6 +5,8 @@
 
 #pragma once
 
+#include "antlr4-common.h"
```
mike-lischke, line 8: Why's that here now?

runtime/Cpp/runtime/src/TokenStreamRewriter.cpp
```diff
@@ -93,7 +93,7 @@ TokenStreamRewriter::TokenStreamRewriter(TokenStream *tokens_) : tokens(tokens_)
 }
 
 TokenStreamRewriter::~TokenStreamRewriter() {
-  for (auto program : _programs) {
+  for (const auto& program : _programs) {
```
mike-lischke, line 96: Coding convention: that should be: `for (const auto &program : _program) {`

runtime/Cpp/runtime/src/atn/ATNSerializer.h
```diff
@@ -5,6 +5,8 @@
 
 #pragma once
 
+#include "antlr4-common.h"
```
mike-lischke, line 8: Why's that here now?

runtime/Cpp/runtime/src/misc/InterpreterDataReader.h
```diff
@@ -6,6 +6,8 @@
 #pragma once
 
 #include "antlr4-common.h"
+#include "atn/ATN.h"
```
mike-lischke, line 9: Why's that here now?

runtime/Cpp/runtime/src/support/CPPUtils.cpp
```diff
@@ -72,8 +72,13 @@ namespace antlrcpp {
 
   std::string arrayToString(const std::vector<std::string> &data) {
     std::string answer;
-    for (auto sub: data) {
-      answer += sub;
+    size_t to_reserve = 0;
```
mike-lischke, line 75: Coding convention: should be `size_t toReserve = 0;`

runtime/Cpp/runtime/src/tree/AbstractParseTreeVisitor.h
```diff
@@ -5,6 +5,7 @@
 
 #pragma once
 
+#include "tree/ParseTree.h"
```
mike-lischke, line 8: Why's that here now?

mike-lischke (changes requested):

runtime/Cpp/runtime/src/TokenStreamRewriter.h
jcking: This is to make the header compilable by itself, rather than relying on include order. This is necessary for some tooling.

runtime/Cpp/runtime/src/TokenStreamRewriter.cpp
jcking: Missed this one, thanks.

runtime/Cpp/runtime/src/atn/ATNSerializer.h
jcking: This is to make the header compilable by itself, rather than relying on include order. This is necessary for some tooling.

runtime/Cpp/runtime/src/misc/InterpreterDataReader.h
jcking: This is to make the header compilable by itself, rather than relying on include order. This is necessary for some tooling.

runtime/Cpp/runtime/src/tree/AbstractParseTreeVisitor.h
jcking: Same as other comments, missed this one.

jcking:
Friendly poke.

mike-lischke (approved):

mike-lischke:
@parrt Another pure C++ patch ready for merge.

parrt:
Thanks!

PR #3329, opened 2021-10-29
Fix stack overflow exception when using Lexer.type in JavaScript

### LINKED ISSUE #3328: Critical exception when using Lexer.type() in JavaScript
This is because of infinite recursion: https://github.com/antlr/antlr4/blob/master/runtime/JavaScript/src/antlr4/Lexer.js#L322-L324

Probably should be changed to:

```js
get type(){
    return this._type;
}
```

KvanTTT:
fixes #3328


KvanTTT:
@ericvergnaud please merge

PR #3337, opened 2021-11-02
Add .clang-format file to ensure consistent style of C++

jcking:
This change adds a `.clang-format` file to the root directory to enable consistent styling of C++. It disables all languages except C++ for now (unless we decide to use clang-format for other languages as well). I based it off of LLVM and forced right alignment for pointers and references. It seems to be pretty close, however we can always tweak it going forward as it is very configurable. https://clang.llvm.org/docs/ClangFormatStyleOptions.html

### DIFF
--- a/.clang-format
+++ b/.clang-format
@@ -1,0 +1,20 @@
+---
+# This section defines defaults for all languages. Currently we derive ANTLR style from LLVM.
+BasedOnStyle: LLVM
+# Only use clang-format for C++ for now.
+DisableFormat: true
+
+---
+# This section configures C++ formatting.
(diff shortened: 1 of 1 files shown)


jcking:
@mike-lischke Adds `.clang-format` configuration so we can get consistent formatting without having to think to much about it. ANTLR can tweak it as we go along and identify things we do not like.

mike-lischke:
Does adding that file change anything in the current formatting?

mike-lischke (approved):

mike-lischke:
@parrt Another C++ patch ready for merging.

Issue #2885, opened 2020-08-20
Escape issue for characeters


rogerfar:
My parser contains the following characters:
```
WS                      : (' '|'\r'|'\t'|'\u000C'|'\n') -> skip;
SQ                      : ['];
DQ                      : '\u0022';
SP                      : '\u0020';
HTAB                    : '\u0009';
CR                      : '\u000D';
LF                      : '\u000A';
```
When I target C# the code is good and compiles fine, no errors in C# or Antlr.

But when I compile it against Java, I get:
```
error: unclosed character literal "'false'", "'null'", null, null, "'\u0022'", "'\u0020'", "'\u0009'",
```

When I look in the generated Java file I see:
```
private static String[] makeLiteralNames() {
	return new String[] {
		null, null, null, "'datetimeoffset'", "'datetime'", "'guid'", "'true'", 
		"'false'", "'null'", null, null, "'\u0022'", "'\u0020'", "'\u0009'", 
		"'\u000D'", "'\u000A'", null, null, null, "'$'", null, null, null, null, 
		"'('", "')'", "'['", "']'", "'{'", "'}'", "'~'", null, null, "'/'", "'.'", 
		"':'", "'%'", "'@'", "'!'", "'?'", "'_'", null, null, null, null, null, 
		null, "'asc'", "'desc'", "'and'", "'or'", "'eq'", "'ne'", "'lt'", "'le'", 
		"'gt'", "'ge'", "'in'", "'nin'", "'nlike'", "'like'"
	};
}
```

The issue is that `"'\u0022'"` is not valid Java, this should be '\u0022'.

Now this can be mitigated by changing the grammer to:
```
WS                      : (' '|'\r'|'\t'|'\u000C'|'\n') -> skip;
SQ                      : ['];
DQ                      : [\u0022];
SP                      : [\u0020];
HTAB                    : [\u0009];
CR                      : [\u000D];
LF                      : [\u000A];
```

But is this the recommended approach or a bug?

KvanTTT:
It looks duplicate to #2281

KvanTTT:
@parrt also close

Issue #2281, opened 2018-04-24
Newline characters in grammars can produce illegal escape codes when targeting Java


timmc:
A grammar containing Unicode code point references for newlines (carriage return and line feed) may produce invalid Java code.

Sample grammar:
```
grammar Demo;
linebreak: LF
         | CR
         ;
CR : '\u000D';
LF : '\u000A';
```

Snippet from resulting DemoLexer.java:
```
	private static final String[] _LITERAL_NAMES = {
		null, "'\u000D'", "'\u000A'"
	};
	private static final String[] _SYMBOLIC_NAMES = {
		null, "CR", "LF"
	};
```

This will fail to compile. It turns out that since the Java compiler interprets Unicode character escapes *before* parsing, `"\u000D"` is equivalent to having a literal carriage return in the middle of a string. Instead of `\u000D` and `\u000A`, ANTLR should emit `\r` and `\n`.

- [x] I am not submitting a question on how to use ANTLR; instead, go to [antlr4-discussion google group](https://groups.google.com/forum/#!forum/antlr-discussion) or ask at [stackoverflow](http://stackoverflow.com/questions/tagged/antlr4)
- [x] I have done a search of the existing issues to make sure I'm not sending in a duplicate

davesisson:
Looks like the generated code includes your single quotes from the
grammar.  Should you be using double quotes in your grammar instead?

timmc-bcov:
Double quotes? Is that even allowed? I don't see it mentioned in the ANTLR reference, and antlr4 doesn't accept it.

Anyway, the single quotes in the generated code aren't relevant; `String x = "\u000A"` is invalid Java as well.

KvanTTT:
In generated code backslah should be escaped:

```java
	private static final String[] _LITERAL_NAMES = {
		null, "'\\u000D'", "'\\u000A'"
	};
	private static final String[] _SYMBOLIC_NAMES = {
		null, "CR", "LF"
	};
```

PR #3349, opened 2021-11-06
Some tool bugfixes

KvanTTT:
* I removed `TREE_GRAMMAR : 'tree' WSNLCHARS* 'grammar' ;` rule from ANTLR 4 grammar because it was created for back compatibility with ANTLR 3. But at the current moment, it's not actual and causes another bug that is fixed by this PR.
* `INVALID_ESCAPE_SEQUENCE` should be an error because the rest charset is undefined if it's a warning and it may cause unwanted behavior. Moreover, it's covered by `EMPTY_STRINGS_AND_SETS_NOT_ALLOWED` error.
* `EMOJI_MODIFIER: [\p{Grapheme_Cluster_Break=E_Base}]` (unsupported emoji modifier) throws `INVALID_ESCAPE_SEQUENCE` instead of `EMPTY_STRINGS_AND_SETS_NOT_ALLOWED`
* Fixed a couple of critical errors related to stack overflow using lexer.
* I introduced `TargetType` enum that is used instead of language string, refactored, and simplified targets code.
* Added `EOF_CLOSURE` error for preventing stack overflow errors on runtime.


KvanTTT:
@parrt please review

KvanTTT:
All tests are passing.

ericvergnaud:
Will switching Target to an enum not break all unofficial targets: Kotlin, Typescript… ?

Envoyé de mon iPhone

KvanTTT:
I don't think so because it's an internal enum that is not used from the external environment. To support unofficial target source code should be modified anyway, specifically, yet another enum value should be added together with related target class and template.

tool/src/org/antlr/v4/codegen/CodeGenerator.java
```diff
@@ -37,70 +36,77 @@
...
+			return target;
 		}
-		return false;
-	}
 
-
-	public Target getTarget() {
-		if ( target == null && targetExists(language) ) {
-			loadLanguageTarget(language);
+		switch (targetType) {
```
parrt, line 66: Hmm... Why is reflection bad in this case? If we happen to create a new target we have to modify this code.

tool/src/org/antlr/v4/codegen/Target.java
```diff
@@ -57,17 +54,14 @@ protected Target(CodeGenerator gen, String language) {
 		targetCharValueEscape['\''] = "\\'";
 		targetCharValueEscape['"'] = "\\\"";
 		this.gen = gen;
-		this.language = language;
 	}
 
+	protected abstract TargetType getTargetType();
```
parrt, line 59: Hmm... Maybe Target language is better than target type?

tool/src/org/antlr/v4/codegen/UnicodeEscapes.java
```diff
@@ -8,31 +8,42 @@
...
-		if (Character.isSupplementaryCodePoint(codePoint)) {
-			// char is not an 'integral' type, so we have to explicitly convert
-			// to int before passing to the %X formatter or else it throws.
-			sb.append(String.format("\\u%04X", (int)Character.highSurrogate(codePoint)));
-			sb.append(String.format("\\u%04X", (int)Character.lowSurrogate(codePoint)));
-		}
-		else {
-			sb.append(String.format("\\u%04X", codePoint));
-		}
+public class UnicodeEscapes {
```
parrt, line 13: oh I see. You have made it an enumerator type. I'm not sure I see what the advantages here and I see a disadvantage and that we now have to update switch statements for new target in a number of locations. And target type it's not the right word because we mean target language. Type sounds like data type or something.

tool/src/org/antlr/v4/codegen/UnicodeEscapes.java
```diff
@@ -8,31 +8,42 @@
...
+		return result.toString();
 	}
 
-	static public void appendPythonStyleEscapedCodePoint(int codePoint, StringBuilder sb) {
-		if (Character.isSupplementaryCodePoint(codePoint)) {
-			sb.append(String.format("\\U%08X", codePoint));
-		}
-		else {
-			sb.append(String.format("\\u%04X", codePoint));
+	public static void appendEscapedCodePoint(StringBuilder sb, int codePoint, TargetType targetType) {
```
parrt, line 20: This looks a lot more inconvenient but it appears that we need more flexibility; is that correct? In another words, it looks like you're saying that different targets need to generate different things. I wonder why that was working before.

tool/src/org/antlr/v4/codegen/CodeGenerator.java
KvanTTT: It's better to get rid of refection in static languages like Java if it's possible because it breaks type guarantees. Yes, we have to add a couple of lines if we add a new target, but it does not look like a big problem.

tool/src/org/antlr/v4/codegen/Target.java
KvanTTT: Ok, `targetLanguage`. In theory, it could be a platform (.NET, JVM) but let's leave it for the future.

tool/src/org/antlr/v4/codegen/UnicodeEscapes.java
KvanTTT: I'm not sure what disadvantage you mean. With `String` it will be the same `switch` statement. Also, some branches of `switch` are common for several targets.

I've refactored code, separated responsibilities. For instance, JavaScript target now does not call unrelated Java method as it [was before](https://github.com/antlr/antlr4/blob/master/tool/src/org/antlr/v4/codegen/target/JavaScriptTarget.java#L146), but has the common branch with Java instead.

tool/src/org/antlr/v4/codegen/UnicodeEscapes.java
KvanTTT: Different targets may have different string literal syntax (escaping), but some of them have the same syntax. It's ok that some branches are common for different targets.

tool/src/org/antlr/v4/codegen/Target.java
KvanTTT: Maybe just `language` instead of `targetLanguage` for clarity (it's clear what this language means)?

KvanTTT:
> Will switching Target to an enum not break all unofficial targets: Kotlin, Typescript… ?

It looks like at least Kotlin target relies on the current structure, reflection, and contains target files in certain directories: https://github.com/Strumenta/antlr-kotlin/tree/master/antlr-kotlin-target/src/main. It's not clear structure but it works. That's why `TargetType` (Language) field should be reverted to `String` field instead of `Enum`.

tool/src/org/antlr/v4/codegen/CodeGenerator.java
KvanTTT: Ok, I'll revert reflection back together with `TargetType` -> `String`, because unofficial targets have its own unknown values that can be accessed only by using reflection.

tool/src/org/antlr/v4/codegen/UnicodeEscapes.java
KvanTTT: This method does not need fixing because it's used in [Target.appendUnicodeEscapedCodePoint](https://github.com/KvanTTT/antlr4/blob/tool-bugfixes/tool/src/org/antlr/v4/codegen/Target.java#L173) that can be easily overridden in unofficial targets if it's needed.

tool/src/org/antlr/v4/codegen/UnicodeEscapes.java
```diff
@@ -8,31 +8,42 @@
...
+			case Python3:
+			case Cpp:
+			case Go:
+			case PHP:
+				String format = Character.isSupplementaryCodePoint(codePoint) ? "\\U%08X" : "\\u%04X";
+				sb.append(String.format(format, codePoint));
+				break;
+			case Swift:
+				sb.append(String.format("\\u{%04X}", codePoint));
+				break;
```
KvanTTT, line 46: Probably we also need `default` branch here for unofficial targets (use `Java` implementation or throw an exception).

tool/src/org/antlr/v4/codegen/CodeGenerator.java
parrt: I think you are focusing on a minor code cleanliness issue, but there are bigger principles at stake here. The biggest issue is now it is less flexible and users cannot create their own custom target object.  also you've created a situation where  we have increased the number of places to change when we add a new target.  I will say that it looks like your code suggests that different targets should handle string encoding differently, which the original code didn't do.  Either we do need to fix this or this spuriously introduced flexibility we don't need.

There are two other important considerations when you are considering changes to code:

1.   You have a limited budget of my time and you spend some of it every time I have to review something; not sure this is worth spending treasure on.

2. The code change has no effect to the user and yet you are risking breaking something by changing the code for some small principle of static types that, as I said, also reduces flexibility.  Unless something is broken or you need to add a new feature, I just can't see any reason to alter code. Remember we have hundreds of thousands of users around the world; who knows what we will break.

I'd like to revert this change from a string to an enumerated type. thanks!

Currently we don't have a stable build system so I think we should focus on that.  Also, I can't possibly listen to all of the issues and PRs that command so it would be great if you could triage those for my attention. I think this was your original suggestion.

Another interesting area you want a bigger task is to think about measuring performance across targets. One simple way might be to have target templates for capturing the current time in nanoseconds or whatever. Then a performance variant of the test rig could inject that into the tests as they are generated.  We'd also need a way to run each test multiple times from within the same process rather than rebuilding and rerunning.

tool/src/org/antlr/v4/codegen/UnicodeEscapes.java
parrt: We do not need this functionality now so you are introducing complexity and flexibility that we don't currently need. I really think there's a large principle here that you never change code unless you have to for fear of breaking something.  When you remove the enumerated type, this issue will go away I guess.

tool/src/org/antlr/v4/codegen/UnicodeEscapes.java
parrt: as you've discovered, a big disadvantage is no one can make their own target.

parrt:
> * Added `EOF_CLOSURE` error for preventing stack overflow errors on runtime.

I think you have a separate PR for this.  Should we kill the old one?

KvanTTT:
> > * Added `EOF_CLOSURE` error for preventing stack overflow errors on runtime.
> 
> I think you have a separate PR for this. Should we kill the old one?

The fix is included in this pull request, specifically the commit https://github.com/antlr/antlr4/pull/3349/commits/bab5a9e24924239e9bec039de70d15048bfc1769

tool/src/org/antlr/v4/codegen/UnicodeEscapes.java
KvanTTT: > We do not need this functionality now so you are introducing complexity and flexibility that we don't currently need.

I just noted that such functionality had been implemented before my changes (other targets overrides `appendUnicodeEscapedCodePoint`) and my changes don't break this.

tool/src/org/antlr/v4/codegen/CodeGenerator.java
KvanTTT: > The code change has no effect to the user and yet you are risking breaking something by changing the code for some small principle of static types that, as I said, also reduces flexibility.

Originally I was fixing bugs https://github.com/antlr/antlr4/issues/2281 and https://github.com/antlr/antlr4/issues/2885. During fixing, I noticed some inconsistency (JavaScript uses the call to Java method) and redundancy. That's why I decided to refactor it together with fixing.

> Remember we have hundreds of thousands of users around the world; who knows what we will break.

I understand it quite well because I'm working on Kotlin and there are much more users. We have a lot of tests (thousands) and anyway some code is not covered, bugs sometimes appear. Maybe it makes sense to think about alpha or beta version publishing. Bugs appears always during product development.

parrt:
Ok. cool. is it ready for a final review, @KvanTTT ?  Thanks for all your efforts!

tool/src/org/antlr/v4/codegen/target/JavaTarget.java
```diff
@@ -20,6 +19,11 @@
 import java.util.Set;
 
 public class JavaTarget extends Target {
+	public final static String key = "Java";
```
parrt, line 22: Seems like we should automate this to use the required convention we have that XTarget uses X as key.  I should have done this earlier rather than hardcoding.

tool/src/org/antlr/v4/codegen/Target.java
```diff
@@ -172,11 +177,20 @@ else if (shouldUseUnicodeEscapeForCodePointInDoubleQuotedString(c)) {
...
+		}
+		appendUnicodeEscapedCodePoint(codePoint, sb);
+	}
+
 	/**
 	 * Escape the Unicode code point appropriately for this language
 	 * and append the escaped value to {@code sb}.
 	 */
-	abstract protected void appendUnicodeEscapedCodePoint(int codePoint, StringBuilder sb);
+	protected void appendUnicodeEscapedCodePoint(int codePoint, StringBuilder sb) {
```
parrt, line 191: Ok, so now we group escaping to a central method but leave ability for target to override? makes sense but then we have a target overriding this and a diff implementation than what is in UnicodeEscapes.appendEscapedCodePoint. Isn't this what virtual method calls were all about? (removing IF/switches?)

parrt: Seems like we should either do switch and call directly or use virtual methods but not both.

tool/src/org/antlr/v4/codegen/target/JavaTarget.java
KvanTTT: I like automatization but I don't how to implement it in a better way with Java. The key value is being checked below in the static constructor (it should be X from XTarget). These constants are used in the following switch: https://github.com/antlr/antlr4/pull/3349/files#diff-0bbbf52a2f6fceb14a08812a3c96f56f6069ba4a8acab5032ac75ca5c1b0c181R24-R39

tool/src/org/antlr/v4/codegen/Target.java
KvanTTT: The problem is in the static method `appendEscapedCodePoint`. It's used both in another static method `escapeCodePoint` and in `Target.appendUnicodeEscapedCodePoint`. Static `escapeCodePoint` is used in [tests](https://github.com/antlr/antlr4/blob/master/tool-testsuite/test/org/antlr/v4/test/tool/TestUnicodeEscapes.java#L19) and there is no access to Target in these tests.

```
@Test
public void latinJavaEscape() {
	checkUnicodeEscape("\\u0061", 0x0061, JavaTarget.key);
}
```

`Target` requires `CodeGenerator` that requires tool and grammar as constructor parameters. It's possible to use fake grammar, tool in tests for creating fake `CodeGenerator` and `Target` but it does not look clear.

That's why I've split methods in such a way.

tool/src/org/antlr/v4/codegen/target/JavaTarget.java
parrt: let me experiment. Might be another reason to avoid the switch.

KvanTTT: Without the switch, some logic is duplicated across several targets, that I'd like to avoid:

```java
class JavaTarget {
    ...
    protected override void appendUnicodeEscapedCodePoint(codePoint) {
        if (Character.isSupplementaryCodePoint(codePoint)) {
			// char is not an 'integral' type, so we have to explicitly convert
			// to int before passing to the %X formatter or else it throws.
			sb.append(String.format("\\u%04X", (int)Character.highSurrogate(codePoint)));
			sb.append(String.format("\\u%04X", (int)Character.lowSurrogate(codePoint)));
		}
		else {
			sb.append(String.format("\\u%04X", codePoint));
		}
    }
}

class JavaScriptTarget {
    ...
    protected override void appendUnicodeEscapedCodePoint(codePoint) {
        if (Character.isSupplementaryCodePoint(codePoint)) {
			// char is not an 'integral' type, so we have to explicitly convert
			// to int before passing to the %X formatter or else it throws.
			sb.append(String.format("\\u%04X", (int)Character.highSurrogate(codePoint)));
			sb.append(String.format("\\u%04X", (int)Character.lowSurrogate(codePoint)));
		}
		else {
			sb.append(String.format("\\u%04X", codePoint));
		}
    }
}
```

The common code can be extracted into a separate method:

```java
class JavaTarget {
    ...
    protected override void appendUnicodeEscapedCodePoint(codePoint) {
        appendUnicodeEscapedCodePointForJavaAndJavaScript(codePoint);
    }
}

class JavaScriptTarget {
    ...
    protected override void appendUnicodeEscapedCodePoint(codePoint) {
        appendUnicodeEscapedCodePointForJavaAndJavaScript(codePoint);
    }
}
```

but it's not clear how to name it. `appendUnicodeEscapedCodePointForJavaAndJavaScript` does not look clear.

tool/src/org/antlr/v4/codegen/Target.java
parrt: Good thoughts. yes. hmm...well, this argues for going completely to your switch but then we need an enum or const string.

tool/src/org/antlr/v4/codegen/target/JavaTarget.java
parrt: Certainly this removes need to specify "Java" etc... in subclasses:

```java
protected Target(CodeGenerator gen) {
  ...
  String clazz = this.getClass().getSimpleName();
  this.language = clazz.substring(0,clazz.indexOf("Target"));
```

but can't do string switch. :(

parrt: `public final static String language = "Java";` is probably ok in each target to get constant.  What does it look like to remove refs to virtual method? We would replace it with direct calls to appendEscapedCodePoint().

KvanTTT: Yes, I've also tried similar tricks, but switch accepts only with constants (final static) :(

KvanTTT: > What does it look like to remove refs to virtual method? We would replace it with direct calls to appendEscapedCodePoint().

I am afraid it could break compatibility with other unofficial targets because it will not be possible to override escaping behavior. It's not possible to call an instance method from a static method.

parrt: Ok, maybe we just add a comment to the virtual method to indicate we are leaving it in for flexibility and backward compatibility with external targets but that target implementations are free to call the static method as a support function if they like.

parrt:
@KvanTTT reviewing PR now.  it looks like you have deleted:

runtime-testsuite/resources/org/antlr/v4/test/runtime/descriptors/ParserExec/EOFInClosure.txt

 is that because you've placed it elsewhere? I added it recently I think because it was missing. can't remember.

runtime-testsuite/test/org/antlr/v4/test/runtime/java/BaseJavaTest.java
```diff
@@ -461,7 +461,7 @@ public void testActions(String templates, String actionName, String action, Stri
             AnalysisPipeline anal = new AnalysisPipeline(g);
             anal.process();
 
-			CodeGenerator gen = new CodeGenerator(g);
+			CodeGenerator gen = CodeGenerator.createCodeGenerator(g);
```
parrt, line 464: just curious why you converted this to a factory method.  again I'm looking for gratuitous changes ;)

KvanTTT:
> is that because you've placed it elsewhere?

Because it's not relevant runtime test now. `stat : 'x' ('y' | EOF)*?;` now throws error `EOF_CLOSURE` on generation step, see https://github.com/antlr/antlr4/pull/3349/files#diff-84648ab3df4fefe24aa050658ef718c13f75b5621e18060d3118460d44c7caceR374

parrt:
awesome. Thanks

runtime-testsuite/test/org/antlr/v4/test/runtime/java/BaseJavaTest.java
KvanTTT: Because it returns `null` if `CodeGenerator` can not be created (and throws error `CANNOT_CREATE_TARGET_GENERATOR`). A contructor can not return not null value. Also, I removed duplication of some logic in `CodeGnerator`.

parrt: cool. thanks.

KvanTTT: Maybe it makes sense to rename to `CodeGenerator.create` instead of `CodeGenerator.createCodeGenerator` (a bit redundancy).

parrt: oh. yeah.  that makes sense.

tool/src/org/antlr/v4/codegen/UnicodeEscapes.java
```diff
@@ -6,33 +6,47 @@
...
+		appendEscapedCodePoint(result, codePoint, language);
+		return result.toString();
 	}
 
-	static public void appendPythonStyleEscapedCodePoint(int codePoint, StringBuilder sb) {
-		if (Character.isSupplementaryCodePoint(codePoint)) {
-			sb.append(String.format("\\U%08X", codePoint));
+	public static void appendEscapedCodePoint(StringBuilder sb, int codePoint, String language) {
+		switch (language) {
+			case CSharpTarget.key:
```
parrt, line 24: actually, hang on a second.  why isn't this just

```
switch ( language ) {
    case "Cpp": ...
    case "Java": ...
...
}
```

We have a requirement that XTarget be language X.  That way we can simplify creating the  constants by looking at the class name in the target constructor.  In other words, `Python3Target.key` must always be `Python3`.

tool/src/org/antlr/v4/codegen/target/JavaTarget.java
parrt: per my other comment made just now, seems like we should use `case "Java":` here :)

parrt:
Ok, this is looking really good @KvanTTT  let's merge it after we change the switch statement to use literals. I.e., `Python3Target.key` is always `Python3` so we don't need anything complicated like a constant. after that we can merge!

KvanTTT:
> I.e., Python3Target.key is always Python3

These constants can be used in [`BaseRuntimeTest`](https://github.com/antlr/antlr4/blob/master/runtime-testsuite/test/org/antlr/v4/test/runtime/BaseRuntimeTest.java) because runtime tests also have a lot of duplicated logic that can be merged. And in other places with runtimes where it's more convenient to use `switch` expression instead of virtual methods. But since such a place is only single now you insist on constants removing for simplicity?

tool/src/org/antlr/v4/codegen/UnicodeEscapes.java
KvanTTT: To prevent future potential mistakes in string literals if these constants will be used something else. For instance, `BaseRuntimeTest` and other similar classes can use these constants as well.

parrt:
I think you are combining multiple arguments here.  I agreed to your recommendation that we leave both the virtual method and the switch statement.  I'm simply trying to remove multiple changes across all targets and make the switch statement simpler.  Not sure what you are referring to as to my inconsistency.

KvanTTT:
Sorry, maybe I picked up the wrong words and it was a bit rude. I just want to add those constants can be used in other classes as well. And you still think it's better to remove them for the current moment?

tool/src/org/antlr/v4/codegen/UnicodeEscapes.java
parrt: in general you're following a great principle here which is to use the type system versus strings, but code generators are specifically about flexibility and besides. People are always able to make errors and strings even here:

```
	public final static String key = "CSharp";
```

As long as we are going for flexibility, we might as well remove all of the extra code you've added to each target and simplify the switch statement. 

literally the only place we use the string is in the switch statement, if intellij is giving me the straight story. I can delete many lines of code from your PR and simplify multiple expressions by going back to a string.

parrt:
> Sorry, maybe I picked up the wrong words and it was a bit rude. I just want to add those constants can be used in other classes as well. And you still think it's better to remove them for the current moment?

 no worries. yes, I think that since the only place we ever refer to that constant is in the switch statement, it's worth getting rid of it along with the extra code needed to support it.

tool/src/org/antlr/v4/codegen/target/JavaTarget.java
KvanTTT: I forgot that these constants are also already used in tests besides `switch` statement: https://github.com/antlr/antlr4/pull/3349/files#diff-453e4cf66eb5c1ed759da1d5a64049c1307117fc10c10e0b2fc4a2dcec17b9e1R27 Should I also replace `JavaTarget.key` with `Java` there? I still think it's error-prone to use literals instead of string constants when it's possible, but I accept your point here.

parrt: Yep, let's converted to a string and simplify this PR.  You can tease me later when we decide we need a constant here. haha.

parrt:
looking good @KvanTTT  are we ready to merge?

KvanTTT:
Done. Also, I've restored API of Target, now it's exactly the same as it was before (except for some methods that are not abstract now).

parrt:
good work, thanks!!

PR #3351, opened 2021-11-12
Use `AnyObject` instead of `class`.

niw:
**Problem**

Using the latest Swift to build Swift runtime will emit warnings about `class` keyword
which is deprecated.

```
warning: using 'class' keyword to define a class-constrained protocol is deprecated; use 'AnyObject' instead
```

**Solution**

Use `AnyObject` instead.


parrt:
@hanjoes ?

parrt:
thanks @hanjoes  and happy new year!

hanjoes:
> thanks @hanjoes and happy new year!

Happy new year to you too!

PR #3371, opened 2021-11-29
Avoid NPE in DFAState

### LINKED ISSUE #3368: Method "equals()" in DFAState may produce NullPointerException
The "equals()" method in DFAState.java may produce NullPointerException when this.configs is null.
(BTW, the TODO in the code was considering about the same trouble.)

Tinker1024:
Avoid NPE in DFAState

fix #3368


Tinker1024:
> Avoid NPE in DFAState

# 3368

Tinker1024:
> Could you also add a test that reproduces the problem?

Thank you for your attention and I will commit the test later.

PR #3376, opened 2021-11-30
Fix Dart linter errors under strict mode

jcking:


jcking:
@parrt Not sure who maintains the Dart runtime. But this change addresses some Dart linter errors that surface when running in strict mode with null safety turned on.

KvanTTT:
@lingyv-li maintains Dart runtime, at least it was before.

KvanTTT:
Unfortunately, Dart tests are failing.

jcking:
Whoops, I did not realize the underlying variable was plural but the setter/getter was singular. Fixed the variable to be singular as well.

PR #3377, opened 2021-11-30
[C++] Switch from NOEXCEPT to noexcept

jcking:
Get rid of `NOEXCEPT` macro as we are on C++17 and `noexcept` is available.


jcking:
@mike-lischke

mike-lischke (approved):

mike-lischke:
@parrt And another C++ patch to merge.

PR #3378, opened 2021-11-30
[C++] Modernize Vocabulary using std::string_view

jcking:
Update `Vocabulary` to use `std::string_view` where possible. This is part 1 of multiple patches which will migrate most uses to std::string_view where possible for generated grammars.


jcking:
@mike-lischke

runtime/Cpp/runtime/src/Vocabulary.h
```diff
@@ -37,7 +37,7 @@ namespace dfa {
     /// </param>
     /// <seealso cref= #getLiteralName(int) </seealso>
     /// <seealso cref= #getSymbolicName(int) </seealso>
-    Vocabulary(const std::vector<std::string> &literalNames, const std::vector<std::string> &symbolicNames);
+    Vocabulary(std::vector<std::string> literalNames, std::vector<std::string> symbolicNames);
```
mike-lischke, line 40: What's the reason for removing the const references here?

runtime/Cpp/runtime/src/Vocabulary.h
```diff
@@ -55,29 +55,14 @@ namespace dfa {
...
-    /// <para>The resulting vocabulary instance returns {@code null} for
-    /// <seealso cref="#getLiteralName(int)"/> and <seealso cref="#getSymbolicName(int)"/>, and the
-    /// value from {@code tokenNames} for the display names.</para>
-    /// </summary>
-    /// <param name="tokenNames"> The token names, or {@code null} if no token names are
-    /// available. </param>
-    /// <returns> A <seealso cref="Vocabulary"/> instance which uses {@code tokenNames} for
-    /// the display names of tokens. </returns>
-    static Vocabulary fromTokenNames(const std::vector<std::string> &tokenNames);
+    Vocabulary(std::vector<std::string> literalNames, std::vector<std::string> symbolicNames,
```
mike-lischke, line 58: Same here.

runtime/Cpp/runtime/src/Vocabulary.h
```diff
@@ -179,7 +164,7 @@ namespace dfa {
     /// </param>
     /// <returns> The display name of the token type, for use in error reporting or
     /// other user-visible messages which reference specific token types. </returns>
-    virtual std::string getDisplayName(size_t tokenType) const;
+    std::string getDisplayName(size_t tokenType) const;
```
mike-lischke, line 167: A string_view as result would also make sense here, right?

mike-lischke (commented):
Looks good, just a few minor comments/questions.

runtime/Cpp/runtime/src/Vocabulary.h
jcking: We cannot unfortunately because the implementation uses std::to_string which return std::string. std::string_view is non-owning, so the backing storage for std::string would disappear leading to std::string_view pointing to garbage.

runtime/Cpp/runtime/src/Vocabulary.h
jcking: Removing the const reference allows the usage of std::move to avoid copying the vector.

```
std::vector<std::string> foo;
Vocabulary(foo); // copy

std::vector<std::string> bar;
Vocabulary(std::move(bar)); // move
```

runtime/Cpp/runtime/src/Vocabulary.h
jcking: Same as the other.

mike-lischke (approved):

mike-lischke:
@parrt A C++ patch ready for merge.

PR #3379, opened 2021-11-30
[Python3] rename pygrun, add entry_point

### LINKED ISSUE #3350: Modernize python scripts packaging, python versioning
Hi folks, thanks for the antlrverse!

The current use of [`scripts`](https://github.com/antlr/antlr4/blob/master/runtime/Python3/setup.py#L15) in the `PythonX` runtimes makes it harder to generate otherwise-solid cross-platform downstream packages that have different constraints than e.g. PyPI. Adopting a newer approach, would look like: 
```python
    entry_points={
        'console_scripts': [
            "pygrun=some-importable.some-module:some-method"
        ]
    },
```` 
and would means that...
  - the functionality would move either into `antlr4-pythonX-runtime.pygrun:main` or some other thing
  - the `.bat` wrapper script would no longer be needed
  - downstreams could reason more about the generated files and where to put them
    - selfishly, for [`conda-forge`](https://github.com/conda-forge/antlr-python-runtime-feedstock/pull/28), this would mean going from a matrix of 20+ builds to a single build. 

Furthermore, as the description _suggests_ a bottom python pin, formally declaring this would further improve the metadata for both `pip` users and others.

```python
    python_requires='>=3.7',
```` 

Thanks again!

bollwyvl:
- [x] fixes #3350
  - [x] replace `scripts` with `entry_points` in `setup.py`
  - [x] moved `pygrun` to `src/antlr/_pygrun.py` 
- [x] signed contributor file


parrt:
I will have to leave this to @ericvergnaud for his expertise.

ericvergnaud:
@parrt blessed - requires setup tools v60+

bollwyvl:
Thanks all!

PR #3380, opened 2021-12-01
[Java] Fix complaints from ErrorProne static analysis

jcking:
This pull request addresses issues identified by https://github.com/google/error-prone.

### DIFF
--- a/tool/src/org/antlr/v4/codegen/target/GoTarget.java
+++ b/tool/src/org/antlr/v4/codegen/target/GoTarget.java
@@ -134,7 +134,9 @@ public class GoTarget extends Target {
 
 	@Override
 	public int getSerializedATNSegmentLimit() {
-		return 2 ^ 31;
+		// 65535 is the class file format byte limit for a UTF-8 encoded string literal
+		// 3 is the maximum number of bytes it takes to encode a value in the range 0-0xFFFF
+		return 65535 / 3;
 	}
 
 	@Override
--- a/tool/src/org/antlr/v4/tool/ErrorManager.java
+++ b/tool/src/org/antlr/v4/tool/ErrorManager.java
@@ -61,7 +61,7 @@ public class ErrorManager {
 		}
 		if (msg.fileName != null) {
 			String displayFileName = msg.fileName;
-			if (format.equals("antlr")) {
+			if (formatName.equals("antlr")) {
 				// Don't show path to file in messages in ANTLR format;
 				// they're too long.
 				File f = new File(msg.fileName);
@@ -206,6 +206,8 @@ public class ErrorManager {
 				errors++;
 				tool.error(msg);
 				break;
+      default:
+        break;
 		}
 		errorTypes.add(etype);
 	}
--- a/runtime/Java/src/org/antlr/v4/runtime/ANTLRFileStream.java
+++ b/runtime/Java/src/org/antlr/v4/runtime/ANTLRFileStream.java
@@ -15,6 +15,7 @@ import java.io.IOException;
  *
  * @deprecated as of 4.7 Please use {@link CharStreams} interface.
  */
+@Deprecated
 public class ANTLRFileStream extends ANTLRInputStream {
 	protected String fileName;
 
--- a/runtime/Java/src/org/antlr/v4/runtime/ANTLRInputStream.java
+++ b/runtime/Java/src/org/antlr/v4/runtime/ANTLRInputStream.java
@@ -22,6 +22,7 @@ import java.util.Arrays;
  *
  * @deprecated as of 4.7 Please use {@link CharStreams} interface.
  */
+@Deprecated
 public class ANTLRInputStream implements CharStream {
     public static final int READ_BUFFER_SIZE = 1024;
    	public static final int INITIAL_BUFFER_SIZE = 1024;
--- a/runtime/Java/src/org/antlr/v4/runtime/CommonToken.java
+++ b/runtime/Java/src/org/antlr/v4/runtime/CommonToken.java
@@ -261,8 +261,7 @@ public class CommonToken implements WritableToken, Serializable {
 		return toString(null);
 	}
 
-	public String toString(Recognizer r) {
-
+	public String toString(Recognizer<?, ?> r) {
 		String channelStr = "";
 		if ( channel>0 ) {
 			channelStr=",channel="+channel;
--- a/runtime/Java/src/org/antlr/v4/runtime/atn/ATNDeserializer.java
+++ b/runtime/Java/src/org/antlr/v4/runtime/atn/ATNDeserializer.java
@@ -789,8 +789,7 @@ public class ATNDeserializer {
 			return new LexerTypeAction(data1);
 
 		default:
-			String message = String.format(Locale.getDefault(), "The specified lexer action type %d is not valid.", type);
-			throw new IllegalArgumentException(message);
+			throw new IllegalArgumentException(String.format(Locale.getDefault(), "The specified lexer action type %s is not valid.", type));
 		}
 	}
 }
--- a/runtime/Java/src/org/antlr/v4/runtime/misc/NotNull.java
+++ b/runtime/Java/src/org/antlr/v4/runtime/misc/NotNull.java
@@ -17,5 +17,6 @@ import java.lang.annotation.Target;
 @Documented
 @Retention(RetentionPolicy.CLASS)
 @Target({ElementType.FIELD, ElementType.METHOD, ElementType.PARAMETER, ElementType.LOCAL_VARIABLE})
+@Deprecated
 public @interface NotNull {
 }
--- a/runtime/Java/src/org/antlr/v4/runtime/misc/TestRig.java
+++ b/runtime/Java/src/org/antlr/v4/runtime/misc/TestRig.java
@@ -14,6 +14,7 @@ import java.lang.reflect.Method;
  *  @deprecated
  *  @since 4.5.1
  */
+@Deprecated
 public class TestRig {
 	public static void main(String[] args) {
 		try {
--- a/runtime/Java/src/org/antlr/v4/runtime/tree/Trees.java
+++ b/runtime/Java/src/org/antlr/v4/runtime/tree/Trees.java
@@ -185,6 +185,7 @@ public class Trees {
 	}
 
 	/** @deprecated */
+  @Deprecated
 	public static List<ParseTree> descendants(ParseTree t) {
 		return getDescendants(t);
 	}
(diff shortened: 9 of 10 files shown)


jcking:
@parrt PTAL

tool/src/org/antlr/v4/codegen/target/GoTarget.java
```diff
@@ -134,7 +134,7 @@ public String encodeIntAsCharEscape(int v) {
 
 	@Override
 	public int getSerializedATNSegmentLimit() {
-		return 2 ^ 31;
+		return 29;
```
KvanTTT, line 137: Constant folding does not improve code readability. Moreover, compiler does it itself.

Also, I'm not sure it's the correct value here. There are much greater values for other runtimes: [Python3Target](https://github.com/antlr/antlr4/blob/7b32c9b169a2d9dd2375d4fc9c5e678f0ed0039c/tool/src/org/antlr/v4/codegen/target/Python3Target.java#L60), [SwiftTarget.java](https://github.com/antlr/antlr4/blob/7b32c9b169a2d9dd2375d4fc9c5e678f0ed0039c/tool/src/org/antlr/v4/codegen/target/SwiftTarget.java#L111), and [others](https://github.com/antlr/antlr4/search?q=getSerializedATNSegmentLimit).

jcking: Correct. The problem is that `2 ^ 31` is XOR and not power. The static analyzer identified that it is confusing as written as it always evaluates to 29 and is likely incorrect. I simply changed it so it is clear. Since most other targets use `65535 / 3`, I've switched it to that.

parrt: for the life of me I can't remember why we are doing 2^31...XOR? weird.  Oh right, yeah, let's use:
`return 65535 / 3` like Java.

tool/src/org/antlr/v4/tool/ErrorManager.java
```diff
@@ -61,7 +61,7 @@ public ST getMessageTemplate(ANTLRMessage msg) {
 		}
 		if (msg.fileName != null) {
 			String displayFileName = msg.fileName;
-			if (format.equals("antlr")) {
```
parrt, line 64: wow.  weird that we never ran into this.  seems like this would never be true.

parrt:
thanks guys...

tool/src/org/antlr/v4/codegen/target/GoTarget.java
KvanTTT: I think it makes sense to fix this value for the [JavaScript](https://github.com/antlr/antlr4/blob/7b32c9b169a2d9dd2375d4fc9c5e678f0ed0039c/tool/src/org/antlr/v4/codegen/target/JavaScriptTarget.java#L91) target. It's also incorrect.

parrt:
I just push to commit to mimic the 2^31 fix for JavaScript; added a comment that java class file format might not require the same limit for other targets..

ericvergnaud:
Hey, there is no such limit in JavaScript, it's plain text! I put 2 ^31 as max positive int = 2 gb

KvanTTT:
But `^` operation is `xor` in JavaScript, it's not exponentiation. I think you meant `1 << 31`

ericvergnaud:
This code is in the tool, in Java

ericvergnaud:
The big lexer test shows that this value is not causing any problem.

KvanTTT:
Yes, my mistake. But in Java it's also `xor` but not exponentiation.

KvanTTT:
Yes, my mistake. But in Java it's also `xor` but not exponentiation.

KvanTTT:
Yes, my mistake. But in Java it's also `xor` but not exponentiation.

ericvergnaud:
I don't quite remember how I ended up writing this misleading code, but what's certain is that the result is desired i.e. an array of readable strings, as opposed to a veeeeeryyyyy long string that no IDE is able to display nicely.
I kind of recall an horrible moment where splitting the veeeeeryyyyy long string into a + b + c ... small strings was producing a stack overflow.
So suggest to change the code to return 32

KvanTTT:
> I don't quite remember how I ended up writing this misleading code

It was not you: https://github.com/antlr/antlr4/blame/2d445f5466e096c39af315697fd4d5f531020281/tool/src/org/antlr/v4/codegen/target/JavaScriptTarget.java#L91

> So suggest to change the code to return 32

In this case, I don't understand such a big difference between JavaScript (32) and other runtimes (65535 / 3). It looks weird.

parrt:
Should I just make it `Math.pow(2, 31)` for javaScript? I can make a comment to explain

ericvergnaud:
I would just return 32 or 64.
Deserializing the ATN only happens once, and optimizing the serialized data matters less than being able to load the generated file in dev tools.

parrt:
Do you mean 2 to the power 32 or 64? It's the len of the string not num bits right?

ericvergnaud:
yes: return 32; not return Math.pow(2, 32)

parrt:
sorry i am not following. why would we limit strings to 32 char?

ericvergnaud:
Because it's much more convenient when opening the generated file.

Most words require 6 chars.
So you get (actual example):
const serializedATN = ["\u0003\u608b\ua72a\u8133\ub9ed\u417c\u3be7\u7786",
    "\u5964\u0002\u00bc\u06fc\b\u0001\u0004\u0002\t\u0002\u0004\u0003\t\u0003",
    "\u0004\u0004\t\u0004\u0004\u0005\t\u0005\u0004\u0006\t\u0006\u0004\u0007",
... 2000 lines here
    "\t\u0007\u0004\b\t\b\u0004\t\t\t\u0004\n\t\n\u0004\u000b\t\u000b\u0004",
    "\u069b\u069e\u06a3\u06a7\u06ac\u06b2\u06b8\u06be\u06c4\u06ca\u06d0\u06d7",
    "\u06dc\u06f5\u06fa\u0003\u0002\u0003\u0002"].join("");

The above is much easier to load by editors than if we don't limit:
const serializedATN = "\u0003\u608b\ua72a\u8133\ub9ed\u417c\u3be7\u7786\u5964\u0002\u00bc\u06fc\b\u0001\u0004\u0002\t\u0002\u0004\u0003\t\u0003\u0004\u0004\t\u0004\u0004\u0005\t\u0005\u0004\u0006\t\u0006\u0004\u0007\t\u0007\u00... 100000 characters here .... 04\b\t\b\u0004\t\t\t\u0004\n\t\n\u0004\u000b\t\u000b\u0004\u069b\u069e\u06a3\u06a7\u06ac\u06b2\u06b8\u06be\u06c4\u06ca\u06d0\u06d7\u06dc\u06f5\u06fa\u0003\u0002\u0003\u0002";
and comes with no measurable penalty (join is ultrafast)

KvanTTT:
In this case, I suggest using the most widespread values of recommended line length in text editors: 80 or 120.

parrt:
okay, I spent a few minutes and looked at the source code again. I had completely forgotten what this does. It is a segment not the entire maximum length so it's fine to do whatever the target developer wants for a  particular target I guess.

@ericvergnaud I think that Java is a special case here because it has such fundamental limits on strings in the class file. Should we go for consistency across all targets other than Java and then make Java a special case?   there's definitely something about the maximum length of a single string in a class file, which I guess is why we had to break them up... it also might slow things down if we had, say, 2000 strings in the class file versus one big one.  The initialization of strings pulled from the class file is weird if I remember.

KvanTTT:
> I think that Java is a special case here because it has such fundamental limits on strings in the class file

ANTLR generator returns concatenation for serialized ATN for Java. Java compiler folds constant string concatenation to one single string for efficiency. So, class file contains one big string instead of concatenation. Thus, it does not make sense to take care of string length in Java source files.

Also, string concatenation could be replaced with StringBuilder. But actually, I don't think it's a good idea. I don't think there is a limit in class file since we haven't encountered such a problem before. Maybe it's a limit in Java compiler itself.

parrt:
I remember the class file format limiting the length of any single string and I doubt that has changed for backward compatibility reasons. Maybe take a look at this to see if I'm still correct? 

https://stackoverflow.com/questions/8323082/size-of-initialisation-string-in-java

https://stackoverflow.com/questions/45275732/increase-string-literal-length-limit

KvanTTT:
The second link doesn't relate to Java because it's about the C language. The first link is also unlikely related to string size in class file, it's about bytecode:

> It shows that the problem does not relate to string literals, but to array initializers.

Also, there are other quotes from the answer:

> So, a pure ASCII string literal can have up to 65535 characters, while a string consisting of characters in the range U+0800 ...U+FFFF have only one third of these. And the ones encoded as surrogate pairs in UTF-8 (i.e. U+10000 to U+10FFFF) take up 6 bytes each.
> The Java Language Specification does not mention any limit for string literals:

So, the current value of `65535 / 3` looks correct for Java (ideally it should be checked). But if it's a limit in class file I'm not sure this limit affects result generation because of compiler string folding.

I've found another answer: https://stackoverflow.com/a/18361275/1046374:

> Note that a String object can have up to 2^31 - 1 characters. The 2^16 -1 limit is for String literals; e.g. String constants that are embedded in the source code of a Java program.

ericvergnaud:
4.4.7. The CONSTANT_Utf8_info Structure

The CONSTANT_Utf8_info structure is used to represent constant string values:

CONSTANT_Utf8_info {
    u1 tag;
    u2 length;
    u1 bytes[length];
}
The items of the CONSTANT_Utf8_info structure are the following:

tag
The tag item of the CONSTANT_Utf8_info structure has the value CONSTANT_Utf8 (1).

length
The value of the length item gives the number of bytes in the bytes array (not the length of the resulting string). The strings in the CONSTANT_Utf8_info structure are not null-terminated.

bytes[]
The bytes array contains the bytes of the string. No byte may have the value (byte)0 or lie in the range (byte)0xf0 - (byte)0xff.


Envoyé de mon iPhone

ericvergnaud:
That converts into max string size = 65535 / 3

Envoyé de mon iPhone

parrt:
Heh, I think we left this discussion incomplete. we need to update the JavaScript per @ericvergnaud  right? just trying to get my to do list together.

ericvergnaud:
I think a small size of strings is highly desirable because the ability to read code in an idea is worth the microseconds spent joining those strings once at startup.
So I'd return 64 or 80 or whatever small number that fits on screen.

parrt:
For python3 it says:

```
// set to something stupid to avoid segmentation
return Integer.MAX_VALUE;
```

Should all but java be about 80 char?

parrt:
Please check out https://github.com/antlr/antlr4/pull/3438

PR #3382, opened 2021-12-02
cpp: support/CPPUtils.cpp: replace clang attribute [[clang::fallthrou…

enedil:
…gh]]

[[clang::fallthrough]] has been used to mark a switch branch target to
be falling through, however just if `clang` has been used.
Since C++17, [[fallthrough]] is a part of the standard:
https://en.cppreference.com/w/cpp/language/attributes/fallthrough

It's recoginzed by MSVC, clang, GCC and respected when using all the C++
versions at least from C++11 upwards.

This change also removes warning in the switch, which would happen when
compiling with GCC.


mike-lischke:
@enedil Thanks for the patch, however this problem will be solved by PR #3395.

enedil:
In that case, I'll close the PR.

PR #3383, opened 2021-12-02
[Java] Fix remaining complaints from ErrorProne static analysis

jcking:
Fix remaining complaints related to non-exhaustive switch statements.


jcking:
@parrt Sorry, missed these two originally. Static analysis for Java is clean after these.

PR #3387, opened 2021-12-03
Update Swift template to silence warnings about unmodified class reference

eneko:
When generating language parsers in Swift, the resulting code triggers some warnings due to a variable not being modified in the generated function bodies.

Example generated code:
```swift
	@discardableResult
	 internal func typeCondition() throws -> TypeConditionContext {
		var _localctx: TypeConditionContext = TypeConditionContext(_ctx, getState())
		try enterRule(_localctx, 28, GraphQLParser.RULE_typeCondition)
		defer {
	    		try! exitRule()
	    }
		do {
		 	try enterOuterAlt(_localctx, 1)
		 	setState(237)
		 	try match(GraphQLParser.Tokens.ON.rawValue)
		 	setState(238)
		 	try namedType()

		}
		catch ANTLRException.recognition(let re) {
			_localctx.exception = re
			_errHandler.reportError(self, re)
			try _errHandler.recover(self, re)
		}

		return _localctx
	}
```

### Compiler warning
```
warning: variable '_localctx' was never mutated; consider changing to 'let' constant
```

## Solution
<s>
All possible types asignable to `_localctx` are classes that inherit from [`RuleContext`](https://github.com/antlr/antlr4/blob/7da07c73a3dc40d4ed24d90c6ac52702050295b1/runtime/Swift/Sources/Antlr4/RuleContext.swift#L58), which conforms to protocol [`Tree`](https://github.com/antlr/antlr4/blob/7da07c73a3dc40d4ed24d90c6ac52702050295b1/runtime/Swift/Sources/Antlr4/tree/Tree.swift#L11). This protocol conforms to `class`/`AnyObject` which requires any conforming types to be Swift classes (reference types). Because of this, it is safe to use `let` instead of `var` in the generated code.

Class/protocol hierarchy:
`ParserRuleContext` -> `RuleContext` -> `RuleNode` -> `ParseTree` -> `SyntaxTree` -> `Tree` -> `AnyObject`

This pull requests updates the Swift template to use `let` instead of `var`, in order to silence this warning.
</s>

Swift template has been updated to break this statements into two. By doing this, we prevent the Swift compiler from raising warnings when the variable is not modified in the generated function body.

The code from the above example would now look like this:
```swift
var _localctx: TypeConditionContext 
_localctx = TypeConditionContext(_ctx, getState())
```


eneko:
Seems like the assumptions above were incorrect and there are instances where `_localctx` is override by a new instance, in which cases `let` wouldn't work.

eneko:
Found another way to silence the warnings, see updated PR description.

PR #3392, opened 2021-12-05
To java 8

parrt:
@KvanTTT @ericvergnaud trying to see if 1.8 works.

### DIFF
--- a/doc/getting-started.md
+++ b/doc/getting-started.md
@@ -12,7 +12,7 @@ If you are going to integrate ANTLR into your existing build system using mvn, a
 
 ### UNIX
 
-0. Install Java (version 1.7 or higher)
+0. Install Java (version 1.8 or higher)
 1. Download
 ```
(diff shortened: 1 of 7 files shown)


KvanTTT:
Unfortunately, there are problems with tests due to compilation failure.

parrt:
yep, this is the error in maven i can't solve

```
Error:  Failed to execute goal org.apache.maven.plugins:maven-compiler-plugin:3.8.1:compile (default-compile) on project antlr4-runtime: Compilation failure -> [Help 1]
```

Please note javac and intellij compile code fine with java 8

parrt:
I am abandoning this because I cannot get 1.8 or above to work due to incompatibility with the use of internal javac lib objects.  I was able to compile with 1.8 and 11 but not 17 prior to this attempt, when the source and target were left as 1.7 in the maven pom.

PR #3393, opened 2021-12-06
[C++] Partially undo commit a8d828016e54adfc403326d8e56d93e7c75c27fc

jcking:
Partially undo a previous commit. I misread the original preprocessor condition. It is actually testing for C++ 20, not C++ 17. So it needs to stay otherwise compilers conformant to C++ 17 without any extensions will fail.


jcking:
@parrt PTAL somewhat quickly. This is a simple revert of a previous change. That preprocessor condition needs to remain. Currently C++20 is required, when we intended C++17. The error is `explicit capture of 'this' with a capture default of '=' is a C++20 extension [-Werror,-Wc++20-extensions]`.

KvanTTT:
Thank you for the patch. The last weekend I tried to set up msbuild C++ compilation but it was failing with a similar error. I wanted to create an issue but it seems this patch fixes the issue.

PR #3394, opened 2021-12-06
[C++] Fix ambiguous ANTLRInputStream constructor overload

jcking:
Not sure when this was added, but std::string is implicitly convertible to std::string_view. So this can cause compilation errors as the compiler does not know which one to select.


jcking:
@parrt Compilation fix. `std::string` is implicitly convertible to `std::string_view`. So the current code causes the compiler to be confused when using something like a string literal.

PR #3395, opened 2021-12-06
[C++] Switch to std::any and deprecate antlrcpp::Any

jcking:
Remove `antlrcpp::Any` in favor of `std::any`. Unfortunately there is no good story for backwards compatibility as `std::any` is a smaller feature set and only supports `static_cast`.


jcking:
@mike-lischke I kept `antlrcpp::Any` as a type alias to `std::any` to at least make the error people will get more clear and require less changes overall. We can remove `antlrcpp::Any` at some point in the future.

mike-lischke (changes requested):
The patch looks good, however in the build output I see 3 warnings, that should be fixed. No idea when this creeped in, but I see now:

> /tmp/BaseCppTest-main-1638301471082/TParser.cpp:79:29: warning: explicit capture of 'this' with a capture default of '=' is a C++20 extension [-Wc++20-extensions]
  auto onExit = finally([=, this] {

Then we get  a number of warnings because of the recently deprecated `EMPTY_VOCABULARY`. 

And finally a (probably old) problem with an implicit switch fall through. See the C++ build output in CircleCI.

Can you fix those too?

KvanTTT:
> /tmp/BaseCppTest-main-1638301471082/TParser.cpp:79:29: warning: explicit capture of 'this' with a capture default of '=' is a C++20 extension [-Wc++20-extensions]

It looks like fixed recently: https://github.com/antlr/antlr4/pull/3393

jcking:
Fixed everything except EMPTY_VOCABULARY. I'll fix that in a standalone CL. The C++20 warning was fixed in a separate CL after I noticed it. I had misread a preprocessor condition as checking for C++ 17 when it was checking for C++ 20.

mike-lischke (approved):

parrt:
Ok, yeah, I think I broke everything yesterday outside of Circleci

PR #3396, opened 2021-12-06
Update maven plugin versions

parrt:
trying to do some updates to the version numbers to get cleaner maven compiles / installs.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -13,7 +13,7 @@ jobs:
           name: test runtime
           command: |
             cd runtime-testsuite
-            mvn -q -Dparallel=methods -DthreadCount=4 -Dtest=java.* test
+            mvn -q -Dparallel=methods -DthreadCount=4 -Dtest=java.** test
             cd ..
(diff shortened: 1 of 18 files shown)


parrt:
Ok, I have updated all of the versions that I could easily find and updated the documentation with a reliable build sequence from an empty maven repository.  we have some nasty cyclic dependency. Try wiping out `.m2` and then trying to build... the runtime tests rely on the software already being built with a jar in the repository cache. doh!!!

parrt:
@KvanTTT and @ericvergnaud  please take a look...  Note that I'm using jdk8 and will turn my attention back to the java 8 PR now.

parrt:
I had to update the scripts that run the tests also use `java.**` not `java.*`.  that was why everything was failing but it looks better now!

KvanTTT:
Unfortunately, now all our tests are red on master. I don't recommend merging without passing all pipelines.

parrt:
They passed all of the circlei

PR #3397, opened 2021-12-07
[WIP] Require source java 8

parrt:
set version to 1.8 for the source code but leave it as 1.7 for class file code generation purposes.


parrt:
See https://github.com/antlr/antlr4/pull/3392

PR #3398, opened 2021-12-07
[C++] Implement standalone Unicode encoding and decoding handling

jcking:
Historically ANTLRv4 C++ runtime has relied on deprecated facilities provided by the standard library to convert between UTF-8, UTF-16 (Windows only) and UTF-32. More recently https://github.com/nemtrif/utfcpp was added as an option. ANTLR has historically kept external dependencies for its runtimes to a minimum, basically nothing except the languages standard library. This change restores that by implementing full UTF-8 and UTF-32 handling based on a port off CEL's C++ implementation which is itself based on Go's `unicode/utf8` package. UTF-16 support (indirectly via `ws2s` and `s2ws`) has been dropped as it was only relevant for Windows. On Windows users should use `MultiByteToWideChar` and `WideCharToMultiByte`.

With this change ANTLRv4 C++ is now dependency free, has full UTF-8 and UTF-32 handling, and `ANTLRInputStream` can support invalid UTF-8 strings by replacing illegal bytes with the Unicode replacement character.

### DIFF
--- a/runtime/Cpp/runtime/src/ANTLRInputStream.cpp
+++ b/runtime/Cpp/runtime/src/ANTLRInputStream.cpp
@@ -9,7 +9,7 @@
 #include "misc/Interval.h"
 #include "IntStream.h"
 
-#include "support/StringUtils.h"
+#include "support/Utf8.h"
 #include "support/CPPUtils.h"
 
 #include "ANTLRInputStream.h"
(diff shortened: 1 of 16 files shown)


jcking:
@parrt FYI Will hopefuly be ready to review later this week.

To make life easier for everyone, we are contributing a port of CEL's C++ UTF-8 decoding and encoding logic.

jcking:
@mike-lischke This is ready now. I added unit tests as well that use googletest and hooked up the scripts to run them.

runtime/Cpp/runtime/src/ANTLRInputStream.cpp
```diff
@@ -144,7 +153,11 @@ std::string ANTLRInputStream::getText(const Interval &interval) {
     return "";
   }
 
-  return antlrcpp::utf32_to_utf8(_data.substr(start, count));
+  auto maybe_utf8 = Utf8::strictEncode(std::u32string_view(_data).substr(start, count));
+  if (!maybe_utf8.has_value()) {
+    throw IllegalArgumentException("Input stream contains invalid Unicode code points");
+  }
+  return std::move(maybe_utf8).value();
```
mike-lischke, line 160: Please stay with the naming convention used in the rest of the code: camelCase.

runtime/Cpp/runtime/src/UnbufferedCharStream.cpp
```diff
@@ -195,7 +195,11 @@ std::string UnbufferedCharStream::getText(const misc::Interval &interval) {
   }
   // convert from absolute to local index
   size_t i = interval.a - bufferStartIndex;
-  return utf32_to_utf8(_data.substr(i, interval.length()));
+  auto maybe_utf8 = Utf8::strictEncode(std::u32string_view(_data).substr(i, interval.length()));
+  if (!maybe_utf8.has_value()) {
+    throw IllegalArgumentException("Unbuffered stream contains invalid Unicode code points");
+  }
+  return std::move(maybe_utf8).value();
```
mike-lischke, line 202: also here

runtime/Cpp/runtime/src/ANTLRInputStream.cpp
```diff
@@ -155,7 +168,11 @@ std::string ANTLRInputStream::getSourceName() const {
 }
 
 std::string ANTLRInputStream::toString() const {
-  return antlrcpp::utf32_to_utf8(_data);
+  auto maybe_utf8 = Utf8::strictEncode(_data);
+  if (!maybe_utf8.has_value()) {
+    throw IllegalArgumentException("Input stream contains invalid Unicode code points");
+  }
+  return std::move(maybe_utf8).value();
```
mike-lischke, line 175: also here

mike-lischke (changes requested):
I love it! Just fix the small casing issues please.

runtime/Cpp/runtime/src/UnbufferedCharStream.cpp
jcking: Done.

runtime/Cpp/runtime/src/ANTLRInputStream.cpp
jcking: Done. Sorry, we use a different style and sometimes I forget.

runtime/Cpp/runtime/src/ANTLRInputStream.cpp
jcking: Done.

mike-lischke (approved):

mike-lischke:
@jcking Not sure if that would show up in the patch: did you also remove the utf8cpp submodule from git?

KvanTTT:
There are some problems with tests. Do they related to these changes?

jcking:
@mike-lischke I do not see any submodule utf8cpp present. What path is the submodule at?

mike-lischke:
@jcking You are right, there's no submodule in git. I didn't check first, but answered from memory. So from this side we should be fine here.

@KvanTTT The tests are odd. All those marked as failure actually seem to hang. It's difficult to say what's wrong there. Also Swift tests failed, which are not affected by this patch. The CircleCI tests succeed, however, I think the failures have nothing to do with this patch.

@parrt Even though some tests fail, I'm pretty sure this patch is ok. So I'd say we can merge it.

PR #3399, opened 2021-12-09
Implement caseInsensitive option

KvanTTT:
Yet another try to suggest `caseInsensitive` option (the old is [here](https://github.com/antlr/antlr4/pull/1092)).

There are [a lot of issues](https://github.com/antlr/grammars-v4/issues?q=label%3Acase-insensitive) related to case insensitiveness in grammars-v4 repository. Actually, it's very inconvenient to write clear grammars without fragment rules and use [CaseChangingCharStream](https://github.com/antlr/antlr4/tree/master/doc/resources) for each runtime. Because it's yet another dependency that even is not available from ANTLR runtime. Moreover, different runtimes may have their own rules of changing input streams. It leads to the situation when a lot of users don't know how to do it and create useless and duplicated issues on GitHub. 

I suggest a new option that resolves all the abovementioned problems. I tried to implement it in a less intrusive way to make reviewing easier (grammar is not affected). Yet another advantage of the new option: it allows to use ANTLR semantics checks, unlike runtime approach.

In short, just one line could resolve a lot of problems:

```g4
options { caseInsensitive=true; }
```

Later we can implement a case insensitive option for each mode if it will be required.

@parrt please take a look at this again, it's very important for grammars community :)


parrt:
Hiya. As you know, my philosophy at this point is that any change to the parsing strategy really should be an emergency.  Wouldn't this all be simpler if they passed in an uppercase or lowercase version of the input and then set the character source to the original unmodified stream for creating the token objects? This wouldn't require much in the way of changes at all... In fact it's probably just an idiom how to use a tool. You need two copies of the input stream but that usually won't be a dealbreaker.

KvanTTT:
> my philosophy at this point is that any change to the parsing strategy really should be an emergency

It's quite emergency because a lot of users still report about not working grammars despite the fact they actually work but users have to use an additional runtime class that normalizes the input stream. Now there are near 30 questions related to case insensitivity: https://github.com/antlr/grammars-v4/issues?q=label%3Acase-insensitive And two recently postponed merge requests: https://github.com/antlr/grammars-v4/pull/2400, https://github.com/antlr/grammars-v4/pull/2417

> Wouldn't this all be simpler if they passed in an uppercase or lowercase version of the input and then set the character source to the original unmodified stream for creating the token objects?

It's not simpler than using the suggested option. Also, maybe it's not fully correct because different runtimes work in different ways. For instance, in Java `Character.toUpperCase('ß')` returns just `ß` and it's correct.  But in JavaScript `'ß'.toUpperCase()` returns `SS` and it's not okay. Moreover, the runtime approach does not allow checking additional diagnostics such as `CHARACTERS_COLLISION_IN_SET`. Anyway, it will be still possible to use input stream normalization if you want but it's not recommended.

> This wouldn't require much in the way of changes at all...

It's not a big change but it's still a change that should be ideally eliminated to make ANTLR using simpler.

BTW, the new option does not affect the parsing strategy and everything is fully back compatible with previous versions. All tests are passing (except for not stable Swift and C++).

KvanTTT:
Also, I've tested the performance of grammar and runtimes case changing approaches and found out there is almost no difference between them: https://github.com/antlr/antlr4/pull/2046#issuecomment-917650686 (grammar approach is a bit better).

parrt:
i'll take a closer look. thanks.

mike-lischke:
I support the idea of handling case insensitivity directly in ANTLR4, however, it should either be limited to ANSI characters (where simple up case / down case operations work) or the full Unicode case mapping process must be implemented. See also: https://www.unicode.org/versions/Unicode4.0.0/ch05.pdf#G21180

IMO, all we need is case insensitive keywords, which are always (?) written using ASCII letters (but the full ANSI script would work too). We don't need to match any possible string. 

So, a better solution is probably to add the case sensitivity flag to a lexer rule (an option, to make only this rule case insensitive) and ANTLR4 can check if the case mapping round trip works (`a.toLower().toUpper().toLower() == a.toLower()`). If it does not then an error is shown. Otherwise we can take both upper and lower variants into the ATN transition's label set.

KvanTTT:
> it should either be limited to ANSI characters (where simple up case / down case operations work) or the full Unicode case mapping process must be implemented.

I think not only to ANSI characters. As [I've already written](https://github.com/antlr/antlr4/pull/2046#issuecomment-917650686), there are case insensitive languages with non-ANSI characters, for instance, well-known in Russia [1C](https://github.com/1c-syntax/bsl-parser/blob/c0f9c91b119a1d4341e90a64a7aa8a4d9dc5d585/src/main/antlr/SDBLLexer.g4#L229-L261). But characters with working round trip as you suggested (`a.toLower().toUpper().toLower() == a.toLower()`) or I suggested (do not change case if length of resulting char (actually string) more than 1). Also, in this PR request I've written tests for different cultures:

The parser from the following grammar:

```g4
lexer grammar L;
ENGLISH_TOKEN:   [a-z]+;
GERMAN_TOKEN:    [äéöüß]+;
FRENCH_TOKEN:    [àâæ-ëîïôœùûüÿ]+;
CROATIAN_TOKEN:  [ćčđšž]+;
ITALIAN_TOKEN:   [àèéìòù]+;
SPANISH_TOKEN:   [áéíñóúü¡¿]+;
GREEK_TOKEN:     [α-ω]+;
RUSSIAN_TOKEN:   [а-я]+;
WS:              [ ]+ -> skip;
```

Matches the following sequence of words:

```
abcXYZ äéöüßÄÉÖÜß àâæçÙÛÜŸ ćčđĐŠŽ àèéÌÒÙ áéÚÜ¡¿ αβγΧΨΩ абвЭЮЯ
```

> So, a better solution is probably to add the case sensitivity flag to a lexer rule (an option, to make only this rule case insensitive)

I don't think it's a better solution because it's excess. In most languages, all tokens are case insensitive, at least modes. Also, it requires ANTLR grammar changing. All SQL dialects (MySql, T-Sql, PlSql, SQLite) are fully case insensitive, Pascal-based languages (Delphi). Only PHP has different modes that may be marked with different case sensitivity options (because it may contain JavaScript, HTML, and other "islands"): https://github.com/antlr/grammars-v4/blob/master/php/PhpLexer.g4

> if the case mapping round trip works (a.toLower().toUpper().toLower() == a.toLower()). If it does not then an error is shown

I think the solution with checking the length of the resulting char is better because it guarantees length equality of input and actual input stream. Also, it looks like Java (that is responsible for grammar conversion) works as expected without such tricks.

`ß` != `ss`, but `ю` == `Ю`.

parrt:
Great discussion and thoughts, @mike-lischke and @KvanTTT. Thanks!  Hmm...ok, let me review the implementation here. From user point of view, I like simple `options { caseInsensitive=true; }` idea if it covers almost all cases.  

Just so I'm clear, we change auto convert in grammar 'a' -> [aA] and leave input char stream alone, right?

parrt:
Adding note that we should update doc and/or remove https://github.com/antlr/antlr4/blob/master/doc/case-insensitive-lexing.md if we merge this.

parrt:
I think we need to delete `* [Case-Insensitive Lexing](case-insensitive-lexing.md)` from index.md too.

ericvergnaud:
This will require enhancing every target. I’m supportive but just raising awareness.

Envoyé de mon iPhone

parrt:
My hope is to avoid changing targets at all. If grammar gets converted ala `'a' -> [aA]` everywhere then targets should be ok as-is.

tool/src/org/antlr/v4/automata/LexerATNFactory.java
```diff
@@ -466,10 +470,10 @@ public IntervalSet getSetFromCharSetLiteral(GrammarAST charSetAST) {
 						state = CharSetParseState.ERROR;
 						break;
 					case CODE_POINT:
-						state = applyPrevStateAndMoveToCodePoint(charSetAST, set, state, escapeParseResult.codePoint);
+						state = applyPrevStateAndMoveToCodePoint(charSetAST, set, state, escapeParseResult.codePoint, caseInsensitive);
```
parrt, line 473: Given `caseInsensitive` is a field, can we remove some of these arguments? At first glance, `applyPrevStateAndMoveToCodePoint()` is a method in same factory class so we could avoid the arg.

tool/src/org/antlr/v4/automata/LexerATNFactory.java
```diff
@@ -546,52 +551,64 @@ private CharSetParseState applyPrevStateAndMoveToProperty(
...
-							sb.append(((GrammarAST)child).getText());
+	private void checkRangeAndAddToSet(GrammarAST ast, IntervalSet set, int a, int b, boolean caseInsensitive) {
+		if (caseInsensitive) {
+			int lowerA = Character.toLowerCase(a);
+			int upperA = Character.toUpperCase(a);
+			int lowerB = Character.toLowerCase(b);
+			int upperB = Character.toUpperCase(b);
+			if (lowerA == upperA && upperB == lowerB) {
+				checkRangeAndAddToSet(ast, set, a, b, false);
+			} else {
```
parrt, line 580: hi. can we keep `}` and `else {` on diff lines? thanks :)

runtime/Java/src/org/antlr/v4/runtime/atn/CodePointTransitions.java
```diff
@@ -40,13 +35,30 @@ public static Transition createWithCodePoint(ATNState target, int codePoint) {
...
+			int codePointTo,
+			boolean caseInsensitive) {
+		if (caseInsensitive) {
+			int lowerCodePointFrom = Character.toLowerCase(codePointFrom);
+			int upperCodePointFrom = Character.toUpperCase(codePointFrom);
+			int lowerCodePointTo = Character.toLowerCase(codePointTo);
+			int upperCodePointTo = Character.toUpperCase(codePointTo);
+			if (lowerCodePointFrom == upperCodePointFrom && lowerCodePointTo == upperCodePointTo) {
+				return createWithCodePointRange(target, lowerCodePointFrom, lowerCodePointTo, false);
+			} else  {
```
parrt, line 47: more `} else {` :)

parrt:
Code is looking go so far...tool tests all pass. checking other stuff.

tool/src/org/antlr/v4/automata/LexerATNFactory.java
KvanTTT: Yes, we can. I left them for the furture if single-token case insensitivity is implemented in some way. I'll remove for simplicity.

tool/src/org/antlr/v4/automata/LexerATNFactory.java
KvanTTT: IDEA reformats code block in such way. Does it matter a lot?

parrt: At this point I think it's late in the game to change the style. That code is used the separate line style for 25 years. Some people have let that drift, but we might as well keep it to the original style as I notice it.

parrt:
@KvanTTT Looking good, but i think `checkRangeAndAddToSet()` also has arg we can remove.

KvanTTT:
It's not true because this method calls itself and pass another value, see https://github.com/antlr/antlr4/pull/3399/files#diff-44489d61ffc263da31b35a8ca18d048cffb043b290fc929e79c5a3202262c487R579

parrt:
Ok, I'm ready to merge if you're happy with it also now.

KvanTTT:
I'm happy with merging, some tests are failing but it looks like yet another problem with CircleCI.

parrt:
Yeah, i'm fiddling with the tests to fix.

parrt:
woohoo! thanks, @KvanTTT  :)

PR #3401, opened 2021-12-11
[C++] Optimize ATN deserialization

jcking:
Improve ATN deserialization by avoiding unnecessary copies and right-sizing `std::vector` when the size is known ahead of time.

### DIFF
--- a/runtime/Cpp/runtime/src/atn/ATNDeserializer.cpp
+++ b/runtime/Cpp/runtime/src/atn/ATNDeserializer.cpp
@@ -39,6 +39,7 @@
 #include "Exceptions.h"
 #include "support/CPPUtils.h"
 #include "support/StringUtils.h"
+#include "support/Casts.h"
 
 #include "atn/LexerCustomAction.h"
 #include "atn/LexerChannelAction.h"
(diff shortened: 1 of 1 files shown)


jcking:
@mike-lischke

runtime/Cpp/runtime/src/atn/ATNDeserializer.cpp
```diff
@@ -39,6 +39,7 @@
 #include "Exceptions.h"
 #include "support/CPPUtils.h"
 #include "support/StringUtils.h"
+#include "support/Casts.h"
```
mike-lischke, line 42: Seems this file is missing.

mike-lischke (commented):
I like the clean up of the code (casts in particular), but I don't really see the advantage of use the `SerializedATNView` class. Does it really speed up deserialization? Do you have any evidence?

jcking:
`SerializedATNView` avoids having to make yet another copy of the serialized ATN. This is already done once during static initialization (before `main`), and this would require it be done again resulting in 2 copies of the serialized ATN. Allocations are slow and avoiding them improves speed, especially when we really only need to subtract 2 for each element. The `SerializedATNView` basically just wraps the vector and subtracts 2 at each access, to keep code simple.

mike-lischke (approved):
Increase complexity to make things smaller and faster. Yeah, that's what C++ is known for 🤣

mike-lischke:
@parrt OK, this one is also ready for merge.

PR #3402, opened 2021-12-11
Fixed C++ runtime demo

mike-lischke:
Some small fixes to make the C++ runtime demo build in latest Xcode.


mike-lischke:
Some builds failed due to timeouts, @parrt, nothing I can fix ¯\_(ツ)_/¯

parrt:
Yeah, Normally it's just the C++ parser one that fails but this one looks like Swift also... can't be related to your C++ though.

PR #3403, opened 2021-12-12
[WIP] Remove tools.jar

parrt:
Working on this but files are translated to use block text from java 16.


runtime-testsuite/test/org/antlr/v4/test/runtime/descriptors/trans.py
```diff
@@ -0,0 +1,64 @@
+"""
```
KvanTTT, line 1: I think this script is excess in the repository because it's being used only one time for conversion.

tool/pom.xml
```diff
@@ -6,7 +6,10 @@
 
 <project xmlns="http://maven.apache.org/POM/4.0.0" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/maven-v4_0_0.xsd">
 	<modelVersion>4.0.0</modelVersion>
-	<parent>
+    <properties>
```
KvanTTT, line 9: Please use the same indentation style within one file (tabs here). If you use IDEA or another IDE you can turn on highlighting of leading spaces.

KvanTTT:
Looks great, but it's better to fix a couple of issues. Also, AppVeyor fails with the following error:

```
[ERROR] Failed to execute goal org.apache.maven.plugins:maven-compiler-plugin:3.8.1:compile (default-compile) on project antlr4-runtime: Fatal error compiling: invalid target release: 16 -> [Help 1]
```

It's a bit weird because our image (Visual Studio 2019) should support Java 16: https://www.appveyor.com/docs/windows-images-software/#java

Also, I recommend using [draft PR](https://github.blog/2019-02-14-introducing-draft-pull-requests/) instead of `[WIP]` suffix because it looks more clear.

KvanTTT (commented):
Looks great, but it's better to fix a couple of issues. Also, AppVeyor fails with the following error:

```
[ERROR] Failed to execute goal org.apache.maven.plugins:maven-compiler-plugin:3.8.1:compile (default-compile) on project antlr4-runtime: Fatal error compiling: invalid target release: 16 -> [Help 1]
```

It's a bit weird because our image (Visual Studio 2019) should support Java 16: https://www.appveyor.com/docs/windows-images-software/#java

Also, I recommend using [draft PR](https://github.blog/2019-02-14-introducing-draft-pull-requests/) instead of `[WIP]` suffix because it looks more clear.

KvanTTT:
Also, it would be great to use text blocks not only for runtime tests but for tool tests as well. There are a lot of multiline grammars and error messages. Take a look at this for example https://github.com/antlr/antlr4/blob/master/tool-testsuite/test/org/antlr/v4/test/tool/TestSymbolIssues.java

parrt:
@KvanTTT ah yes. Tool tests also are kind of old school. well, those at least work without the annotation mechanism so maybe I will leave that for another day.

Thanks for pointing me to the draft PR thing! I had no idea :)

As I mentioned, because of the requirement to build with Java 16, I think I will have to abandon this PR. At least I had fun yesterday and didn't watch TV all day.

PR #3408, opened 2021-12-14
[C++] Format all the things

jcking:
This also updated the `.clang-format` file to match `clang-format --dump-config`. This ensures future tweaks to the underlying LLVM style did not affect us. Lastly it removes all type definitions and aliases in the global namespace. Specifically `Ref`, `XPathLexer`, `ssize_t` on windows, and etc.


jcking:
@mike-lischke

KvanTTT:
Not sure it's a good solution to reformat the entire project because it contaminates git blame.

mike-lischke:
Well, it doesn't build anyway and I have a number of objections, but in order to discuss them we need to see the effect on the code. 

I also wouldn't want that big patch to go into ANTLR4 as a whole. But for now I need to find time to review the patch. 

@jcking What we can start with is to raise line length to 120, which is what I use for years already. I know people have very different opinions about that (and formatting in general), but let's start with this to keep in the current wrapping.

jcking:
Should build now. Switched to 120. The main idea is to just get the codebase consistently formatted. Blame shouldn't be interfered with to much, as its mostly just signatures and declarations being formatted.

runtime/Cpp/runtime/src/RuleContext.cpp
```diff
@@ -123,21 +104,15 @@ std::string RuleContext::toString(const std::vector<std::string> &ruleNames, Rul
...
-std::string RuleContext::toString(Recognizer *recog) {
-  return toString(recog, &ParserRuleContext::EMPTY);
-}
+std::string RuleContext::toString(Recognizer *recog) { return toString(recog, &ParserRuleContext::EMPTY); }
 
 std::string RuleContext::toString(Recognizer *recog, RuleContext *stop) {
   if (recog == nullptr)
-    return toString(std::vector<std::string>(), stop); // Don't use an initializer {} here or we end up calling ourselve recursivly.
+    return toString(std::vector<std::string>(),
+                    stop); // Don't use an initializer {} here or we end up calling ourselve recursivly.
```
mike-lischke, line 114: Same situation: the trailing comment makes the line too long. In such cases better move it to before the line it applies to, instead of wrapping that line. But of course this is something that needs manual editing. Has nothing to do with formatting rules.

runtime/Cpp/runtime/src/RuntimeMetaData.cpp
```diff
@@ -9,30 +9,34 @@ using namespace antlr4;
...
   std::string runtimeVersion = VERSION;
   bool runtimeConflictsWithGeneratingTool = false;
   bool runtimeConflictsWithCompileTimeTool = false;
 
   if (generatingToolVersion != "") {
-    runtimeConflictsWithGeneratingTool = runtimeVersion != generatingToolVersion
-      && getMajorMinorVersion(runtimeVersion) != getMajorMinorVersion(generatingToolVersion);
+    runtimeConflictsWithGeneratingTool =
+        runtimeVersion != generatingToolVersion &&
+        getMajorMinorVersion(runtimeVersion) != getMajorMinorVersion(generatingToolVersion);
```
mike-lischke, line 22: Here I would like to keep the original form: if a boolean expression gets too long for one line split it into groups (based on the operator precedence) so that a boolean operator starts a line, indicating so the grouping of the subexpressions (applies also to `if` statements).

runtime/Cpp/runtime/src/RuntimeMetaData.cpp
```diff
@@ -9,30 +9,34 @@ using namespace antlr4;
...
+      runtimeVersion != compileTimeVersion &&
+      getMajorMinorVersion(runtimeVersion) != getMajorMinorVersion(compileTimeVersion);
 
   if (runtimeConflictsWithGeneratingTool) {
-    std::cerr << "ANTLR Tool version " << generatingToolVersion << " used for code generation does not match "
-      "the current runtime version " << runtimeVersion << std::endl;
+    std::cerr << "ANTLR Tool version " << generatingToolVersion
+              << " used for code generation does not match "
+                 "the current runtime version "
+              << runtimeVersion << std::endl;
```
mike-lischke, line 33: Failed automatic formatting. This should manually be adjusted to match the line length, to avoid an automatic reformat.

runtime/Cpp/runtime/src/TokenStreamRewriter.cpp
```diff
@@ -15,27 +15,24 @@ using namespace antlr4;
 using antlr4::misc::Interval;
 
 TokenStreamRewriter::RewriteOperation::RewriteOperation(TokenStreamRewriter *outerInstance_, size_t index_)
-  : outerInstance(outerInstance_) {
+    : outerInstance(outerInstance_) {
```
mike-lischke, line 18: Why 4 spaces indentation here? Should be 2.

mike-lischke (changes requested):
I haven't checked the full patch. It's just too big. But I think I identified the most important problems. Fixing them will not only make the patch smaller, but also avoids unnecessary changes.

PR #3409, opened 2021-12-14
[C++] Clean up antlr4::misc::Interval and antlr4::dfa

jcking:
Fix const correctness, remove virtual where it makes sense (class is instantiated directly by antlr4 and there is no way for users to override), and make things constexpr where it is feasible.

### DIFF
--- a/runtime/Cpp/runtime/src/misc/Interval.h
+++ b/runtime/Cpp/runtime/src/misc/Interval.h
@@ -13,11 +13,11 @@ namespace misc {
   // Helpers to convert certain unsigned symbols (e.g. Token::EOF) to their original numeric value (e.g. -1)
   // and vice versa. This is needed mostly for intervals to keep their original order and for toString()
   // methods to print the original numeric value (e.g. for tests).
-  size_t numericToSymbol(ssize_t v);
-  ssize_t symbolToNumeric(size_t v);
+  constexpr size_t numericToSymbol(ssize_t v) { return static_cast<size_t>(v); }
+  constexpr ssize_t symbolToNumeric(size_t v) { return static_cast<ssize_t>(v); }
 
   /// An immutable inclusive interval a..b
-  class ANTLR4CPP_PUBLIC Interval {
+  class ANTLR4CPP_PUBLIC Interval final {
   public:
     static const Interval INVALID;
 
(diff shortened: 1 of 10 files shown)


jcking:
@mike-lischke Normal cleanup.

runtime/Cpp/runtime/src/misc/Interval.h
```diff
@@ -13,27 +13,30 @@ namespace misc {
   // Helpers to convert certain unsigned symbols (e.g. Token::EOF) to their original numeric value (e.g. -1)
   // and vice versa. This is needed mostly for intervals to keep their original order and for toString()
   // methods to print the original numeric value (e.g. for tests).
-  size_t numericToSymbol(ssize_t v);
-  ssize_t symbolToNumeric(size_t v);
+  constexpr size_t numericToSymbol(ssize_t v) { return static_cast<size_t>(v); }
+
+  constexpr ssize_t symbolToNumeric(size_t v) { return static_cast<ssize_t>(v); }
```
mike-lischke, line 18: Semantically these two belong together and the leading comment applies to both. Hence there's no empty line in the original code.

runtime/Cpp/runtime/src/misc/Interval.h
```diff
@@ -13,27 +13,30 @@ namespace misc {
...
+
+    constexpr Interval(ssize_t a_, ssize_t b_) : a(a_), b(b_) {}
 
     /// return number of elements between a and b inclusively. x..x is length 1.
     ///  if b < a, then length is 0.  9..10 has length 2.
-    size_t length() const;
+    constexpr size_t length() const { return b >= a ? static_cast<size_t>(b - a + 1) : 0; }
 
-    bool operator == (const Interval &other) const;
+    constexpr bool operator==(const Interval &other) const { return a == other.a && b == other.b; }
```
mike-lischke, line 39: I'm not so fond of having implementation code in header files. One single change (and be it for testing) and you might have to recompile everything that includes that header (in the worst case the entire application, which can take a while in C++ as you know).

mike-lischke (changes requested):

runtime/Cpp/runtime/src/misc/Interval.h
jcking: Putting the implementation in the header allows the compiler to perform optimizations without requiring link time optimization. Without it in the header file, the compiler is forced to emit a function call. With it in the header file, the compiler can boil down the Interval constructor to two simple move instructions.

I am not saying we should put everything in the header, but rather code that is called frequently and is relatively simple. Mostly accessors, logical operators, and etc. Things that boil down to a few instructions or simply forward to another function.

runtime/Cpp/runtime/src/misc/Interval.h
jcking: Done.

runtime/Cpp/runtime/src/misc/Interval.h
mike-lischke: OK, I follow your explanation.

mike-lischke (approved):

mike-lischke:
@parrt Another C++ patch to merge.

mike-lischke:
@jcking It just appeared to me that we may have something wrong here, so I need your input. A `constexpr` can be evaluated at compile time, but you introduced that for a number methods and even constructors (didn't know this is possible). However, the limit values `a` and `b` can be altered at runtime, so `length()` or `operator == ` are not constant. Do I miss something here?

jcking:
So `constexpr` is a little bit non-intuitive, as everybody has probably made that assumption. `constexpr` does not mean it must be evaluated at compile time, it simply restricts what can appear in the definition so that it could, if all inputs are known at compile time, be evaluated at compile time. It doesn't mean it actually is, as the compiler could choose not to technically AFAIK.

I believe a newer standard introduces something like `consteval` which enforces compile time evaluation, but `constexpr` doesn't.

PR #3413, opened 2021-12-15
[C++] Improve whitespace escaping

jcking:
`replaceAll` is the only function left and its usage is inefficient as its only use case is escaping whitespace. The current usage involves scanning the string 3 times. If any whitespace is actually encountered it updates the string in place, which is terribly slow as std::string is effectively just std::vector with an additional bookkeeping for the null terminator. So updating in place can involve may may memory moves and re-allocations. If there its lots of whitespace to escape, its performance is unacceptable.

Instead an optimal implementation which has a bounded worse case is written inline at its 3 call locations and StringUtils has been removed.

### DIFF
--- a/runtime/Cpp/runtime/src/CommonToken.cpp
+++ b/runtime/Cpp/runtime/src/CommonToken.cpp
@@ -10,8 +10,8 @@
 
 #include "misc/Interval.h"
 
-#include "support/StringUtils.h"
 #include "support/CPPUtils.h"
+#include "support/StringUtils.h"
 
 #include "CommonToken.h"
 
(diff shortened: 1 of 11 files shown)


jcking:
@mike-lischke

runtime/Cpp/runtime/src/CommonToken.cpp
```diff
@@ -165,9 +164,15 @@ std::string CommonToken::toString(Recognizer *r) const {
...
-    antlrcpp::replaceAll(txt, "\t", "\\t");
+    std::string escaped;
+    escaped.reserve(txt.size());  // Best case, nothing to escape.
+    for (const auto &c : txt) {
+      if (c == '\n' || c == '\r' || c == '\t') {
+        escaped.push_back('\\');
+      }
+      escaped.push_back(c);
+    }
+    txt = std::move(escaped);
```
mike-lischke, line 175: Is this really the better way? You now have duplicated the same code 3 times. Instead the existing code could be improved to do all replacements in one go. To be honest, I don't see an improvement in this patch.

jcking: Switched back to keeping StringUtils.{h,cpp} and instead replaced `replaceAll` with `escapeWhitespace` which does the logic. I kept the same complexity by having two versions, one which returns a new std::string and another which accepts an std::string and appends.

mike-lischke (approved):

mike-lischke:
@parrt A C++ patch ready for merge.

Issue #2852, opened 2020-06-19
Module not found: Error: Can't resolve 'fs' in '/Users/Saravanan/projects/frontend/widgets/web/node_modules/antlr4'


saravanan10393:
I am using antlr4 with javacript target since years, Now it is throwing error when i do a build using webpack
```
Module not found: Error: Can't resolve 'fs' in '/Users/Saravanan/projects/frontend/widgets/web/node_modules/antlr4'
```

It was working fine with webpack-4 with following [settings enabled](https://stackoverflow.com/questions/54754585/how-to-use-antlr-if-an-error-with-fs-occurs). But now with wbpack-5 it is throwing the error as above.

ericvergnaud:
Smells like a webpack issue more than an antlr one...
How about checking webpack specs and submitting a PR to update the antlr docs?

Envoyé de mon iPhone

toolgood:
Browser throws an error

require.js:178 GET http://localhost:8888/fs.js 404 (Not Found)


CharStreams.js
const fs = require("fs");

ArtemeeSenin:
@ericvergnaud Maybe it is possible to optionally include char and file stream files? Not to force its inclusion in index.js if antlr is used on frontend

ericvergnaud:
Suggest you stick to webpack 4 until webpack 5 provides an equivalent for the below config:

`   node: {
        module: "empty",
        net: "empty",
        fs: "empty"
    },
`

namratha-b:
> I am using antlr4 with javacript target since years, Now it is throwing error when i do a build using webpack
> 
> ```
> Module not found: Error: Can't resolve 'fs' in '/Users/Saravanan/projects/frontend/widgets/web/node_modules/antlr4'
> ```
> 
> It was working fine with webpack-4 with following [settings enabled](https://stackoverflow.com/questions/54754585/how-to-use-antlr-if-an-error-with-fs-occurs). But now with wbpack-5 it is throwing the error as above.

hi @saravanan10393 ,
any update on this issue?

kaiyoma:
Ah, there _is_ a solution for webpack 5, but for some reason it's not documented here. In your webpack config, add this:

```
resolve: {
  fallback: {
    fs: false
  }
}
```

Ketler13:
@saravanan10393 if you're on webpack5 could you please try suggestion I added here - https://github.com/antlr/antlr4/pull/3414 ? Just modify manually package.json file in node_modules/antlr4 and check how it works?

[...]

PR #3414, opened 2021-12-16
Add browser fs fallback

Ketler13:
This is an attempt to fix runtime browser issue - https://github.com/antlr/antlr4/issues/2852 for environments like create-react-app where there is no possibility to manipulate webpack config without ejecting the application. I did this locally for installed antlr4 package and it helped me. Hope this solution is good enough.

Please also read [webpack 5 migration notes](https://webpack.js.org/blog/2020-10-10-webpack-5-release/#automatic-nodejs-polyfills-removed) :
_Package authors: Use the browser field in package.json to make a package frontend-compatible. Provide alternative implementations/dependencies for the browser._

[Ignore module in package.json](https://github.com/defunctzombie/package-browser-field-spec#ignore-a-module)

Thank you for the great tool!


ericvergnaud:
@parrt blessed

MarkDGithub:
I am having the same issue, any idea when we might expect an updated npm package with the merged changes?

parrt:
We are actually doing quite a number of important changes so I don't have a date for you but I'm hoping within a month or so we can get 4.10 out the door.

MattNot:
Any news for this to be added in a release?

parrt:
Finally got in major changes that I needed to make so I will start the release process over the next weak or two. sorry for the delay. This has been merged so it will show up in the release 4.10

parrt:
(It's in dev branch now if you'd like to try out)

Issue #3412, opened 2021-12-15
Cpp : use UnbufferedCharStream throw antlr4::IllegalArgumentException


tangxing806:
I can parse it in the following way:
~~~cpp
    // inputLib : ./libPath/osu018_stdcells.lib
    std::ifstream infile(inputLib);
    antlr4::ANTLRInputStream input(infile);
    Liberty::libertyLexer lexer(&input);
    antlr4::CommonTokenStream tokens(&lexer);
    Liberty::libertyParser parser(&tokens);

    antlr4::tree::ParseTree* root = parser.root();
~~~

but when i use this way, program throw Exception:
~~~cpp
    std::wifstream infile(inputLib);
    antlr4::UnbufferedCharStream unbufferedCharStream(infile);
    Liberty::libertyLexer lexer(&unbufferedCharStream);
    lexer.setTokenFactory(new antlr4::CommonTokenFactory(true));
    antlr4::UnbufferedTokenStream tokens(&lexer);
    Liberty::libertyParser parser(&tokens);

    antlr4::tree::ParseTree* root = parser.root();
~~~

Exception information：
> terminate called after throwing an instance of 'antlr4::IllegalArgumentException'
>  what():  invalid interval

~~~cpp
std::string UnbufferedCharStream::getText(const misc::Interval &interval) {
  if (interval.a < 0 || interval.b >= interval.a - 1) {
    throw IllegalArgumentException("invalid interval");
  }

  ......
}
~~~

gdb information:
~~~shell
(gdb) r
Starting program: /mnt/d/WSL/work/antlr-local/build/st/liberty_st ./libPath/osu018_stdcells.lib ./test.log develop
start parsing
terminate called after throwing an instance of 'antlr4::IllegalArgumentException'
  what():  invalid interval

Program received signal SIGABRT, Aborted.
__GI_raise (sig=sig@entry=6) at ../sysdeps/unix/sysv/linux/raise.c:50
50      ../sysdeps/unix/sysv/linux/raise.c: No such file or directory.
(gdb) bt
#0  __GI_raise (sig=sig@entry=6) at ../sysdeps/unix/sysv/linux/raise.c:50
#1  0x00007ffffd495859 in __GI_abort () at abort.c:79
#2  0x00007ffffd72e911 in ?? () from /lib/x86_64-linux-gnu/libstdc++.so.6
#3  0x00007ffffd73a38c in ?? () from /lib/x86_64-linux-gnu/libstdc++.so.6
[...]

PR #3420, opened 2021-12-20
[C++] Fix bugs in UnbufferedCharStream

jcking:
As the name says, fix bugs in UnbufferedCharStream pointed out by #3412. I also cleaned up the class while I was there.

### DIFF
--- a/runtime/Cpp/runtime/src/UnbufferedCharStream.cpp
+++ b/runtime/Cpp/runtime/src/UnbufferedCharStream.cpp
@@ -13,9 +13,8 @@ using namespace antlrcpp;
 using namespace antlr4;
 using namespace antlr4::misc;
 
-UnbufferedCharStream::UnbufferedCharStream(std::wistream &input) : _input(input) {
-  InitializeInstanceFields();
-
+UnbufferedCharStream::UnbufferedCharStream(std::wistream &input)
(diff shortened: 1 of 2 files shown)


jcking:
@mike-lischke

jcking:
I also removed `virtual` prefixes where it is overriding, because it is unnecessary as override implies the method was virtual and will fail it it was not.

jcking:
I also chose to make `toString` throw, because it is not implemented in the Java version and would just return the Java representation which is usually the class name and an address/hash.

runtime/Cpp/runtime/src/UnbufferedCharStream.cpp
```diff
@@ -101,7 +98,7 @@ size_t UnbufferedCharStream::LA(ssize_t i) {
     return EOF;
   }
 
-  if (_data[static_cast<size_t>(index)] == 0xFFFF) {
+  if (_data[static_cast<size_t>(index)] == WEOF) {
```
mike-lischke, line 101: What is WEOF?

jcking: WEOF is the wide character version of EOF, defined in wchar.h. I switched it to `std::char_traits<wchar_t>::eof()` which is the preferred C++ equivalent.

mike-lischke (approved):
@parrt This C++ patch can be merged.

PR #3421, opened 2021-12-20
[Go] Improve BitSet implementation

jcking:
The current `BitSet` implementation uses a map. The ideal implementation uses a slice of integers and has significantly less overhead. This pull request switches `BitSet` to be backed by a slice of integers. This is very similar to how things like `std::bitset` in C++ are implemented.

### DIFF
--- a/runtime/Go/antlr/utils.go
+++ b/runtime/Go/antlr/utils.go
@@ -8,7 +8,7 @@ import (
 	"bytes"
 	"errors"
 	"fmt"
-	"sort"
+	"math/bits"
 	"strconv"
 	"strings"
 )
@@ -71,59 +71,87 @@ type hasher interface {
 	hash() int
 }
 
+const bitsPerWord = 64
+
+func indexForBit(bit int) int {
+	return bit / bitsPerWord
+}
+
+func wordForBit(data []uint64, bit int) uint64 {
+	idx := indexForBit(bit)
+	if idx >= len(data) {
+		return 0
+	}
+	return data[idx]
+}
+
+func maskForBit(bit int) uint64 {
+	return uint64(1) << (bit % bitsPerWord)
+}
+
+func wordsNeeded(bit int) int {
+	return indexForBit(bit) + 1
+}
+
 type BitSet struct {
-	data map[int]bool
+	data []uint64
 }
 
 func NewBitSet() *BitSet {
-	b := new(BitSet)
-	b.data = make(map[int]bool)
-	return b
+	return &BitSet{}
 }
 
 func (b *BitSet) add(value int) {
-	b.data[value] = true
+	idx := indexForBit(value)
+	if idx >= len(b.data) {
+		size := wordsNeeded(value)
+		data := make([]uint64, size)
+		copy(data, b.data)
+		b.data = data
+	}
+	b.data[idx] |= maskForBit(value)
 }
 
 func (b *BitSet) clear(index int) {
-	delete(b.data, index)
+	idx := indexForBit(index)
+	if idx >= len(b.data) {
+		return
+	}
+	b.data[idx] &= ^maskForBit(index)
 }
 
 func (b *BitSet) or(set *BitSet) {
-	for k := range set.data {
-		b.add(k)
+	size := intMax(b.minLen(), set.minLen())
+	if size > len(b.data) {
+		data := make([]uint64, size)
+		copy(data, b.data)
+		b.data = data
+	}
+	for i := 0; i < size; i++ {
+		b.data[i] |= set.data[i]
 	}
 }
 
 func (b *BitSet) remove(value int) {
-	delete(b.data, value)
+	b.clear(value)
 }
 
 func (b *BitSet) contains(value int) bool {
-	return b.data[value]
-}
-
-func (b *BitSet) values() []int {
-	ks := make([]int, len(b.data))
-	i := 0
-	for k := range b.data {
-		ks[i] = k
-		i++
+	idx := indexForBit(value)
+	if idx >= len(b.data) {
+		return false
 	}
-	sort.Ints(ks)
-	return ks
+	return (b.data[idx] & maskForBit(value)) != 0
 }
 
 func (b *BitSet) minValue() int {
-	min := 2147483647
-
-	for k := range b.data {
-		if k < min {
-			min = k
+	for i, v := range b.data {
+		if v == 0 {
+			continue
 		}
+		return i*bitsPerWord + bits.TrailingZeros64(v)
 	}
-
-	return min
+	return 2147483647
 }
 
 func (b *BitSet) equals(other interface{}) bool {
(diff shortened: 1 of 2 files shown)


jcking:
@parrt Not sure who deals with Go.

parrt:
I think @pboyer, original dev, isn't able to help anymore, but @KvanTTT has been working on perf also.

parrt:
Getting a speed check and then will check this one.  Go target is really slow at moment and this is likely culprit.

KvanTTT:
@jcking is it possible to look at a comparison table before/after optimization? Have you checked it?

parrt:
real	4m14.464s on my m1 mac mini and now I'm running with these tests in.

KvanTTT:
Have you tested performance on runtime tests? It is not reliable because of a lot of side effects and IO operation. The result may vary significantly. You can try to do it several times or take a look at two sequential AppVeyour builts: [1](https://ci.appveyor.com/project/parrt/antlr4/builds/41943576/job/lyywjsonc23m8r62), [2](https://ci.appveyor.com/project/parrt/antlr4/builds/41954992/job/s92n7tkjedok4gg9). There is a difference in 95 sec without code change at all.

parrt:
ah. yeah, good point. Not thinking straight today.  Maybe you can test using the rig you used for your patches?

KvanTTT:
I have not used any rig because it was obvious and significant changes (22 -> 6 min for javascript, 20 -> 8 min for Go). I have no objection to merging but ideally, it also should be tested using micro-benchmarks.

jcking:
I can get the Go tests to run in 1 min 50 sec instead of 3 min 44 sec by making them parallel. It looks like the Go tests copy the contents of GOROOT. It might be better to just override GOPATH and avoid the copy.

jcking:
Its also not really a micro-optimization. If you run in a for loop and set bits 0 through 63 on the map impl, its going to consume more memory and be less efficient then the slice implementation. The slice implementation will only allocate once, while the map one will have to resize and rehash at least once. It will also have to perform more allocations, depending on `map` underlying implementation. Using the slice reduces the amount of generate garbage for GC. A micro-optimization would be to take that further and avoid the slice for single element bitsets and use just a bare `uint64`.

parrt:
@jcking maybe a good excuse for me to learn some more Go.  wow. copying entire root? strange. ok, i'll look. maybe it was old-school Go before modules.

@KvanTTT i'll try to run a Go parser...

I'll finish https://github.com/antlr/antlr4/pull/3407 first though.

KvanTTT:
> I can get the Go tests to run in 1 min 50 sec instead of 3 min 44 sec by making them parallel. It looks like the Go tests copy the contents of GOROOT. It might be better to just override GOPATH and avoid the copy.

It's true. But I was not able to override GOPATH because of some problems, tests were not working. Anyway, single copying the content of GOPATH is better than the previous scheme with copying ALL runtime files for EVERY test.

KvanTTT:
I remembered it's not just copying, it's copying together with ANTLR runtime files. Redirection occurs but to the new directory that contains both go system files and go ANTLR runtime files: https://github.com/antlr/antlr4/blob/master/runtime-testsuite/test/org/antlr/v4/test/runtime/go/BaseGoTest.java#L182 I don't know how to implement it in a better way. Maybe copy go system files only one time but go version also can be updated.

parrt:
Yeah, maybe I shouldn't mess with the Go test rig until I know how it deals with modules and installing software better.

studentmain:
This seems caused grammars-v4 vb6 grammar fail. 

https://github.com/antlr/grammars-v4/pull/2429#issuecomment-999392241

https://github.com/antlr/grammars-v4/runs/4596478383?check_suite_focus=true#step:9:5206

KvanTTT:
I don't think so because grammars-v4 uses a stable version from the maven repository, this change is in the development branch.

studentmain:
> I don't think so because grammars-v4 uses a stable version from the maven repository, this change is in the development branch.

Due to how `go get` works, test for golang is using the runtime from the development branch.

https://github.com/antlr/grammars-v4/blob/d48d5e1cf74e5e73d3b1f57ae91509a18ed7d912/_scripts/templates/Go/tester.psm1#L11

KvanTTT:
Maybe I'm wrong. Is it possible to set up downloading from a certain commit?

studentmain:
> Maybe I'm wrong. Is it possible to set up downloading from a certain commit?

Yes, it's possible. I forgot to do that when I wrote the tester.

KvanTTT:
It looks like it's a good idea to run tests on grammar-v4 together with ordinary runtime tests.

jcking:
https://github.com/antlr/antlr4/pull/3427 should resolve the issue. I
missed that specific case.

parrt:
Hi. just did some testing. I built first with

```
a4.9.3 -Dlanguage=Go -o parser *.g4
```

(where a4.9.3 is an alias to run the tool.)

Not sure I'm actually using the `go.mod` file correctly, but here is what I have:

```
module go-parse-performance

go 1.17

require github.com/antlr/antlr4/runtime/Go/antlr v0.0.0-20220102222715-89f7760263f1 // indirect

replace github.com/antlr/antlr4/runtime/Go/antlr => /Users/parrt/tmp/antlr4.9.3/runtime/Go/antlr
```

That yields (subset of files in jdk 1.8) timing:

Parsed 501 files

real	0m13.971s
user	0m24.560s
sys	0m1.007s
```

Then `go.mod`:

```
module go-parse-performance

go 1.17

require github.com/antlr/antlr4/runtime/Go/antlr v0.0.0-20220102222715-89f7760263f1 // indirect

replace github.com/antlr/antlr4/runtime/Go/antlr => /Users/parrt/antlr/code/antlr4/runtime/Go/antlr
```

Gives:

```
Parsed 501 files (4.10 on disk I hope)

real	0m14.186s
user	0m23.983s
sys	0m1.123s
```

So, about the same.  Note it bounces around a bit depending on sys load of course.

```go
package main

import (
	"fmt"
	"github.com/antlr/antlr4/runtime/Go/antlr"
	"go-parse-performance/parser"
	"io/fs"
	"io/ioutil"
	"path/filepath"
	"strings"
)

func main() {
	// /Library/Java/JavaVirtualMachines/jdk-11.0.13.jdk/Contents/Home/lib/src.zip
	// /Library/Java/JavaVirtualMachines/jdk1.8.0_311.jdk/Contents/Home/src.zip
	// I unzip one of those into a local subdirectory
	root := "jdk8-src"
	files := []string{}
	filepath.WalkDir(root,
		func(path string, d fs.DirEntry, err error) error {
			files = append(files, path)
			return nil
		})
	n := 0
	for _, filename := range files {
		if n > 500 {
			break
		}
		if !strings.HasSuffix(filename, ".java") {
			continue
		}
		println(filename)
		content, err := ioutil.ReadFile(filename)
		if err != nil {
			panic("can't find file " + filename)
		}
		is := antlr.NewInputStream(string(content))
		lexer := parser.NewJavaLexer(is)

		stream := antlr.NewCommonTokenStream(lexer, antlr.TokenDefaultChannel)

		// Create the Parser
		p := parser.NewJavaParser(stream)

		p.CompilationUnit()
		n += 1
	}
	fmt.Printf("Parsed %d files\n", n)
}
```

PR #3422, opened 2021-12-20
[Go] Make BaseGoTest::initializeRuntime static

jcking:
After this its possible to update pom.xml to make the tests parallel, at least for Go. The other test suites may not support that.

```
diff --git a/runtime-testsuite/pom.xml b/runtime-testsuite/pom.xml
index 9c63282cd..c75dc195a 100644
--- a/runtime-testsuite/pom.xml
+++ b/runtime-testsuite/pom.xml
@@ -106,6 +106,8 @@
                         <include>**/dart/Test*.java</include>
                         <include>${antlr.tests.swift}</include>
                     </includes>
+          <parallel>all</parallel>
+          <useUnlimitedThreads>true</useUnlimitedThreads>
 				</configuration>
 			</plugin>
 			<plugin>

```

### DIFF
--- a/runtime-testsuite/test/org/antlr/v4/test/runtime/go/BaseGoTest.java
+++ b/runtime-testsuite/test/org/antlr/v4/test/runtime/go/BaseGoTest.java
@@ -203,7 +203,7 @@ public class BaseGoTest extends BaseRuntimeTestSupport implements RuntimeTestSup
 		return null;
 	}
 
-	private synchronized boolean initializeRuntime() {
+	private static synchronized boolean initializeRuntime() {
 		if (isRuntimeInitialized)
 			return true;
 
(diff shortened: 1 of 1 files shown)


jcking:
@parrt

jcking:
CSharp, Dart, and PHP at least do not support running in parallel. The others do, AFAICT.

jcking:
Oh, for Dart I might just be missing the runtime. Same for PHP.

Let me just create a pull request that enables parallelism and see what happens.

KvanTTT:
All runtimes should support parallel execution.

jcking:
Not specifically the runtimes, but the tests. Some of the setup is not static but per test case which breaks when you start running them parallely.

Issue #3370, opened 2021-11-29
Parallel runtime tests execution


KvanTTT:
I don't understand why parallel tests execution still is not enabled. It should be quite easy to implement since each test uses a separated directory and operates in an isolated process. Parallel tests can greatly improve performance because they may utilize all processor cores.

ericvergnaud:
Free CIs usually only provide single core containers so enabling parallel testing ends up being slower. This was definitely the case with Travis CI. Didn't try it on our latest CIs though.

KvanTTT:
Also, it is usefull on local machines without limitation on number of cores.

KvanTTT:
BTW, now I'm still not able to set up tests, I'm not a very experienced with maven.

KvanTTT:
It seems like at least CircleCI has 2 CPUs: https://app.circleci.com/pipelines/github/antlr/antlr4/678/workflows/005d374e-8b5b-40ae-a720-c04b03a6928b/jobs/7728/resources

KvanTTT:
AppVeyor also has 2 CPU cores: https://help.appveyor.com/discussions/questions/11179-how-many-cores-and-threads-can-be-used-in-free-appveyor-build

ericvergnaud:
you can try amending the test scripts, there’s a maven cmd line option I believe

KvanTTT:
Yes, I'll try to fix them after fixing dotnet and go pipelines on Linux. Also, it seems like parallel execution is already turned on for CircleCI.

PR #3423, opened 2021-12-20
Make runtime-testsuite tests run in parallel

jcking:


jcking:
@parrt Let's see what happens.

runtime-testsuite/pom.xml
```diff
@@ -106,6 +106,8 @@
                         <include>**/dart/Test*.java</include>
                         <include>${antlr.tests.swift}</include>
                     </includes>
+          <parallel>all</parallel>
+          <useUnlimitedThreads>true</useUnlimitedThreads>
```
KvanTTT, line 110: Please fix the unaligned indentation.

KvanTTT:
Related issue: https://github.com/antlr/antlr4/issues/3370

runtime-testsuite/pom.xml
jcking: Done. Editor uses spaces by default, file was using tabs.

.circleci/config.yml
```diff
@@ -3,7 +3,10 @@ version: 2.1
 jobs:
   test_tool_and_runtime_java:
     docker:
-      - image: cimg/openjdk:8.0
+      - image: cimg/openjdk:11.0.13
```
KvanTTT, line 6: I recommend waiting for the merging of that request https://github.com/antlr/antlr4/pull/3407 There are some troubles with JDK updating.

parrt: Almost ready.

jcking:
This is something to do with the circleci environment compared to running locally. I'll look into this a different time.

parrt:
Just a heads up that I think there's an issue with Swift + parallel tests + (possibly) M1 mac. I see an explosion of `swift-frontend` processes and it brings the machine to its knees; can't even bring a shell up to kill things.

KvanTTT:
Have you tried to restrict thread count (2, 4)? It might help.

parrt:
yep, I tried threadCount=4 and it's still host my machine despite having 8 core.

KvanTTT:
And single thread mode works fine?

parrt:
Yes, there is no problem running Single threaded mode Swift tests. I just completed one run and I'm now testing your release build version. It does seem faster by the way using release mode

.circleci/config.yml
parrt: looks like we are going to be only java 8 for this next release because there's no way to use java 11 and generate java 8 class files apparently, at least reliably.

parrt:
@jcking i think we can close this draft; i banged my head on a big CI clean up yesterday: https://github.com/antlr/antlr4/pull/3445  Merging shortly.

ericvergnaud:
I plan to look into it at some point, it would help to have more recent java for the tool and test rig, whilst sticking to Java 8 for the runtime.

KvanTTT:
> I plan to look into it at some point, it would help to have more recent java for the tool and test rig, whilst sticking to Java 8 for the runtime.

It's hardly possible. I've sent Terrence the following StackOverflow answer: https://stackoverflow.com/a/56064482/1046374 Java 8 and Java 11 have different JVM runtime API.

parrt:
oh right. forgot about that. We'll have to move in lock step src/trg :(

ericvergnaud:
> > I plan to look into it at some point, it would help to have more recent java for the tool and test rig, whilst sticking to Java 8 for the runtime.
> 
> It's hardly possible. I've sent Terrence the following StackOverflow answer: https://stackoverflow.com/a/56064482/1046374 Java 8 and Java 11 have different JVM runtime API.

That SO thread is about a different challenge i.e. running Java 11 source code on Java 8 JVM. 
What I'm proposing is run the tool using Java 11 (or 16), but keep the runtime in Java 8.

parrt:
We'd have to prevent java 11 code / lib calls in runtime somehow but I'm for it if we can.

parrt:
Yep, Thanks for the work on this @jcking.  Looks like the new PR covers this well... Please let me know if there's something else in this PR that we should carve out!

PR #3427, opened 2021-12-22
[Go] Fix BitSet OR operation

jcking:
I originally missed the case where `set.data` was small than `b.data`, which would lead to a panic. This resolves that.

### DIFF
--- a/runtime/Go/antlr/utils.go
+++ b/runtime/Go/antlr/utils.go
@@ -121,13 +121,18 @@ func (b *BitSet) clear(index int) {
 }
 
 func (b *BitSet) or(set *BitSet) {
-	size := intMax(b.minLen(), set.minLen())
-	if size > len(b.data) {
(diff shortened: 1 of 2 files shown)


jcking:
@parrt

KvanTTT:
Could you please cover your change with a unit test?

jcking:
> Could you please cover your change with a unit test?

Done.

jcking:
Actually, wait. Need to fix.

KvanTTT:
Maybe also add a grammar sample?

jcking:
Okay, now its good. The test is more exhaustive and I cleaned them up to be less repetative.

parrt:
looks like some conflicts after i merged my new test rig.  I merged https://github.com/antlr/antlr4/pull/3421 as a test in my test rig branch but then reversed with a commit. it looks like it left some cruft around.

PR #3431, opened 2021-12-23
[C++] Improve MurmurHash performance for Clang

jcking:


jcking:
@mike-lischke A small improvement to MurmurHash to enable clang to better optimize. Hashing is in the critical path, so the faster the better.

jcking:
Note that it looks like I changed the constants with the diff in Github, but I really just switched their places. The 64-bit murmur is the first one now instead of the second.

runtime/Cpp/runtime/src/misc/MurmurHash.h
```diff
@@ -64,12 +65,18 @@ namespace misc {
     template<typename T> // where T is C array type
     static size_t hashCode(const std::vector<Ref<T>> &data, size_t seed) {
       size_t hash = initialize(seed);
-      for (auto entry : data) {
-        hash = update(hash, entry->hashCode());
+      for (auto &entry : data) {
+        hash = update(hash, entry);
```
mike-lischke, line 69: By this change you removed the handling of sub elements hashes. This must stay. Certain objects generated hashes not only from their own state but that of their own data too.

runtime/Cpp/runtime/src/misc/MurmurHash.cpp
```diff
@@ -17,118 +20,83 @@ using namespace antlr4::misc;
 
 #if defined(_MSC_VER)
 
-#define FORCE_INLINE	__forceinline
```
mike-lischke, line 20: I assume FORCE_INLINE is no longer needed or even counter productive?

runtime/Cpp/runtime/src/misc/MurmurHash.cpp
```diff
@@ -17,118 +20,83 @@ using namespace antlr4::misc;
...
+
+#define ROTL32(x, y) __builtin_rotateleft32(x, y)
+#define ROTL64(x, y) __builtin_rotateleft64(x, y)
 
 #else	// defined(_MSC_VER)
 
 // Other compilers
 
-#define	FORCE_INLINE inline __attribute__((always_inline))
+namespace {
```
mike-lischke, line 37: Why that namespace? Nothing changed which would cause a violation of the ODR.

runtime/Cpp/runtime/src/misc/MurmurHash.cpp
jcking: Yeah, it was never referenced.

runtime/Cpp/runtime/src/misc/MurmurHash.h
jcking: This should be delegating to the `update` overload above, which accepts `Ref` and will actually check if its `nullptr`.

runtime/Cpp/runtime/src/misc/MurmurHash.cpp
jcking: We are defining it in the global namespace, so putting it in an anonymous namespace or declaring it static is best practice, makes it very obvious that this is only relevant for this translation unit. These days anonymous namespace are preferred over `static`.

mike-lischke (approved):
@parrt Another C++ patch to merge.

PR #3432, opened 2021-12-24
Add speed test

parrt:
add new java parsing speed test. i pasted in java grammar and about 9k lines of java to parse.  Required that I add flag to indicate grammar is not a ST template to avoid escaping \ and < chars.

Swift, Go, Java, Python all pass new test.

### DIFF
--- a/runtime-testsuite/resources/org/antlr/v4/test/runtime/descriptors/Performance/DropLoopEntryBranchInLRRule_4.txt
+++ b/runtime-testsuite/resources/org/antlr/v4/test/runtime/descriptors/Performance/DropLoopEntryBranchInLRRule_4.txt
@@ -49,5 +49,7 @@ between X1 and X2 or between X3 and X4
 
 [skip]
 Go
+Python2
+Python3
 PHP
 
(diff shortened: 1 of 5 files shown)


parrt:
Need a way to report average time to testers. Awkward using test rig but was easiest path.

parrt:
The more I think about this the more I don't think this is the right approach. We are never going to get reliable profiling information without Target specific knowledge and code. It's possible that we could make mechanism to do profiling similar to the testing infrastructure we have now, but I don't think it's worth the effort. Target developers can make their own specific code and stick it in the repository. For example, I have code that tests the Java runtime here: https://github.com/antlr/antlr4/tree/master/runtime-testsuite/test/org/antlr/v4/test/runtime/java/api/perf

KvanTTT:
I was working on runtime-specific performance tests in the previous year: [AntlrBenchmarks](https://github.com/KvanTTT/AntlrBenchmarks). I used runtime-specific tools such as [BenchmarkDotNet](https://github.com/dotnet/BenchmarkDotNet) for C#, [JMH](https://openjdk.java.net/projects/code-tools/jmh/) for Java, and other. Also, I used the warming-up stage and several runs. It was an interesting experience. I compared the performance of left recursive vs not left recursive rules: https://github.com/KvanTTT/AntlrBenchmarks#results

Actually, I don't think it's a bad idea and we can reach quite a reliable result if we'll implement everything correctly.

parrt:
Hmm...ok, yeah. It'd be very nice to have but would it be "easy" to create an infrastructure for perf as we did for testing various targets?

KvanTTT:
I have to think about it more and investigate performance tests in other projects. Also, I still think it would be great to simplify and unify runtime tests infrastructure because it'll simplify further steps.

Issue #3436, opened 2021-12-26
Introduce caseInsensitive lexer rule option


KvanTTT:
The new option `caseInsensitive` works great and it simplifies grammar a lot. But there are some lexer rule exceptions where it should be disabled. For instance, case-insensitivity affects lexing process for the following rule because a string starting with lower `n` is incorrect in TSql, MySql, but a string starting with upper `N` is correct. Enabled `caseInsensitive` allows both strings:

```g4
options { caseInsensitive=true; }
STRING: 'N'? '\'' (~'\'' | '\'\'')* '\'';
```

It's possible to check [online](https://sqliteonline.com/) the following string: `SELECT N'sample';`

I suggest adding a new lexer rule option `caseInsensitive` that naturally disables/enables case insensitivity for certain rules:

```g4
options { caseInsensitive=true; }
STRING options { caseInsensitive=false; } : 'N'? '\'' (~'\'' | '\'\'')* '\'';
```

PR #3437, opened 2021-12-26
caseInsensitive lexer rule option

KvanTTT:
It was quite simple to implement, but it should be merged after #3435

### DIFF
--- a/doc/lexer-rules.md
+++ b/doc/lexer-rules.md
@@ -308,3 +308,16 @@ As of 4.5, you can also define channel names like enumerations with the followin
 ```
 channels { WSCHANNEL, MYHIDDEN }
 ```
+
+## Lexer Rule Options
+
+### caseInsensitive
+
+Defines if the current lexer rule is case-insensitive.
+The argument can be `true` or `false`.
+The option rewrites `caseInsensitive` grammar option value if it's defined.
+
+```g4
+options { caseInsensitive=true; }
+STRING options { caseInsensitive=false; } : 'N'? '\'' (~'\'' | '\'\'')* '\''; // lower n is not allowed
+```
--- a/tool/src/org/antlr/v4/tool/Grammar.java
+++ b/tool/src/org/antlr/v4/tool/Grammar.java
@@ -74,6 +74,8 @@ public class Grammar implements AttributeResolver {
 	 */
 	public static final String INVALID_RULE_NAME = "<invalid>";
 
+	public static final String caseInsensitiveOptionName = "caseInsensitive";
+
 	public static final Set<String> parserOptions = new HashSet<String>();
 	static {
 		parserOptions.add("superClass");
@@ -83,16 +85,21 @@ public class Grammar implements AttributeResolver {
 		parserOptions.add("language");
 		parserOptions.add("accessLevel");
 		parserOptions.add("exportMacro");
-		parserOptions.add("caseInsensitive");
+		parserOptions.add(caseInsensitiveOptionName);
 	}
 
 	public static final Set<String> lexerOptions = parserOptions;
 
-	public static final Set<String> ruleOptions = new HashSet<String>();
+	public static final Set<String> lexerRuleOptions = new HashSet<>();
+	static {
+		lexerRuleOptions.add(caseInsensitiveOptionName);
+	}
 
-	public static final Set<String> ParserBlockOptions = new HashSet<String>();
+	public static final Set<String> parseRuleOptions = new HashSet<>();
 
-	public static final Set<String> LexerBlockOptions = new HashSet<String>();
+	public static final Set<String> parserBlockOptions = new HashSet<String>();
+
+	public static final Set<String> lexerBlockOptions = new HashSet<String>();
 
 	/** Legal options for rule refs like id&lt;key=value&gt; */
 	public static final Set<String> ruleRefOptions = new HashSet<String>();
--- a/tool/src/org/antlr/v4/tool/Rule.java
+++ b/tool/src/org/antlr/v4/tool/Rule.java
@@ -64,7 +64,10 @@ public class Rule implements AttributeResolver {
 	public Grammar g;
 
 	/** If we're in a lexer grammar, we might be in a mode */
-	public String mode;
+	public final String mode;
+
+	/** If null then use value from global option that is false by default */
+	public final boolean caseInsensitive;
 
     /** Map a name to an action for this rule like @init {...}.
      *  The code generator will use this to fill holes in the rule template.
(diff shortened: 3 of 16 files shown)


doc/lexer-rules.md
```diff
@@ -308,3 +309,21 @@ As of 4.5, you can also define channel names like enumerations with the followin
 ```
 channels { WSCHANNEL, MYHIDDEN }
 ```
+
+### caseInsensitive()
```
parrt, line 313: hmm... We should probably discuss change it such as this before you spend time on it. I'm not sure we need to add this part. Maybe we just fix the bug first and then worry about other changes?  What is the motivation? I think you were arguing against this previously

KvanTTT: > hmm... We should probably discuss change it such as this before you spend time on it.

No problem, it was interesting for me how it's simple to implement such a feature.

> Maybe we just fix the bug first and then worry about other changes?

It seems I've fixed all bugs by the previous PR (at least I've tested all case insensitive grammars).

> What is the motivation? I think you were arguing against this previously

I've encountered a couple of rules where `caseInsensitive` should be disabled in case insensitive grammars during testing all grammars-v4. The first one is described in related issue: #3436. The second one looks like [the following](https://github.com/antlr/grammars-v4/blob/master/sql/postgresql/PostgreSQLLexer.g4#L2577):

```g4
fragment EscapeStringText
    : 'u' letterOrDigit letterOrDigit letterOrDigit letterOrDigit
    | 'U' letterOrDigit letterOrDigit letterOrDigit letterOrDigit letterOrDigit letterOrDigit letterOrDigit letterOrDigit
    ;
fragment letterOrDigit: [0-9a-fA-F];
```

Also, there are some [Unicode ranges](https://github.com/antlr/grammars-v4/blob/master/sql/tsql/TSqlLexer.g4#L941-L956) that ideally should not be transformed (but they can be rewritten a bit to get rid of `CHARACTERS_COLLISION_IN_SET` warning):

```g4
fragment FullWidthLetter
    : '\u00c0'..'\u00d6'
    | '\u00d8'..'\u00f6'
    | '\u00f8'..'\u00ff'
    | '\u0100'..'\u1fff'
    | '\u2c00'..'\u2fff'
    | '\u3040'..'\u318f'
    | '\u3300'..'\u337f'
    | '\u3400'..'\u3fff'
    | '\u4e00'..'\u9fff'
    | '\ua000'..'\ud7ff'
    | '\uf900'..'\ufaff'
    | '\uff00'..'\ufff0'
    ;
```

parrt: I wonder if a case insensitive option on the rule isn't better because this really isn't an action.

KvanTTT: You are right, it's isn't an action. On the other side, it looks quite natural and clear in my option. Also, it's the simplest implementation without grammar changing (at least I don't know how it could be implemented better).

parrt: well I would disagree that it looks natural because only commands go after `->` :)    If we add an option I don't think we need to change the grammar right?

parrt: meaning rules already have options, if I remember correctly.

KvanTTT: Indeed, it seems like we have rule-level options "the tool still supports the following syntax for future use": https://github.com/antlr/antlr4/blob/master/doc/options.md#rule-options

Thanks, I'll try to reimplement the rule-level `caseInsensitive` option this way.

KvanTTT: Unfortunately, it looks like only parser rules support such options but not lexer rules, `parserRule` contains `rulePrequels` that contains `optionsSpec`, but `lexerRule` does not contain `rulePrequels`. That's why grammar should be changed a bit.

parrt: oh dang! Hmm...let me look. crap. That would take major work (tree grammar too). Not sure that's worth the risk. I'm not in favor of making an action though either. Maybe we hold off on this one and think about later. Might need lexer rule options at some point.

KvanTTT: I'll try to implement it using lexer rules. I've already estimated it seems not so major work. Anyway, I'll suggest a solution and you'll make the decision.

KvanTTT: I've remade it using the lexer rule option, and now it looks really more conscious than my first implementation with command, thanks @parrt! And it was not so complicated,

doc/lexer-rules.md
```diff
@@ -308,3 +308,16 @@ As of 4.5, you can also define channel names like enumerations with the followin
 ```
 channels { WSCHANNEL, MYHIDDEN }
 ```
+
+## Lexer Rule Options
+
+### caseInsensitive
```
KvanTTT, line 314: I'm not sure about the best location for `caseInsensitive` lexer rule. Maybe it makes sense to move to https://github.com/antlr/antlr4/blob/master/doc/options.md#rule-options

parrt: maybe a reference in one spot to the other?

tool/src/org/antlr/v4/tool/Grammar.java
```diff
@@ -74,21 +74,28 @@
 	 */
 	public static final String INVALID_RULE_NAME = "<invalid>";
 
-	public static final Set<String> parserOptions = new HashSet<String>();
+	public static final String caseInsensitiveOptionName = "caseInsensitive";
+
+	public static final Set<String> ParserOptions = new HashSet<String>();
```
parrt, line 79: I think for consistency we should keep this as initial lowercase, just like the others.

parrt:
sorry..  I messed up the build of this PR when I tried to resolve the conflict. I missed:

```
	private final boolean caseInsensitive;
```

 in the lexer atn factory. looking at it locally but maybe you could fix it in your branch.

KvanTTT:
> sorry.. I messed up the build of this PR when I tried to resolve the conflict. I missed:

No problem, let me resolve conflicts and repush it again.

parrt:
cool.  it's looking like a very simple change! kudos.

tool/src/org/antlr/v4/tool/Grammar.java
KvanTTT: I've encountered a mess with lower/UPPER case identifiers: https://github.com/antlr/antlr4/blob/master/tool/src/org/antlr/v4/tool/Grammar.java#L89-L95 So, you think it's better to use lower case for all options, do you?

tool/src/org/antlr/v4/tool/Rule.java
```diff
@@ -64,7 +64,10 @@
 	public Grammar g;
 
 	/** If we're in a lexer grammar, we might be in a mode */
-	public String mode;
+	public final String mode;
```
parrt, line 67: it's weird that I don't have this field already!!

parrt:
okay, everything looks good to me except for that minor capitalization thing.

tool/src/org/antlr/v4/tool/Grammar.java
parrt: well I reserve the first capital letter for types like classes... at the very least we should be consistent at this point

doc/lexer-rules.md
KvanTTT: Done, I made a reference in options.md: https://github.com/antlr/antlr4/pull/3437/files#diff-6420f854d462e04fe86b98f8bd495e3ec6f42f94121c39a37d228ba74ea91052R114

Issue #3443, opened 2021-12-29
Go runtime missing export of "String()"


kaby76:
Here's another missing method in the Go runtime: ToString(). Or should I say "[String()](https://github.com/antlr/antlr4/blob/33122ac442fdfd5884a5182dac559889de60c218/runtime/Go/antlr/token.go#L190)"? I'm not sure why it was not named ToString() to be consistent with all the other targets. Anyway, the method is already written, ready to be exported, but it is not in the interface for [Token](https://github.com/antlr/antlr4/blob/33122ac442fdfd5884a5182dac559889de60c218/runtime/Go/antlr/token.go#L21). It seems as though the code was literally translated from the C# or Java target but the person did not understand that [every object has a ToString()/toString() method](https://docs.oracle.com/javase/7/docs/api/java/lang/Object.html#toString()). So, it *must* be there to provide a consistent API across targets. The fix is to just add it to the interface.

PR #3444, opened 2021-12-29
Fix for #3443 -- missing export of String()

kaby76:
This is a fix for the missing API method String() on a Token interface. By analogy to [ToStringTree() in ParseTree](https://github.com/antlr/antlr4/blob/345cc17bf36c67d7be8b7775e9f3693334232b28/runtime/Go/antlr/tree.go#L34), String() should be in the interface Token.

### DIFF
--- a/runtime/Go/antlr/token.go
+++ b/runtime/Go/antlr/token.go
@@ -35,6 +35,7 @@ type Token interface {
 
 	GetTokenSource() TokenSource
 	GetInputStream() CharStream
+	String() string
 }
 
 type BaseToken struct {


kaby76:
There are significant problems with the build. But, the build of the Go runtime seems to have worked https://app.circleci.com/pipelines/github/antlr/antlr4/1082/workflows/953fb0cd-9726-4b49-bbbb-27cd5f68b490/jobs/12933. I do not know whether there is a "go test" of the Go runtime (probably not), but it works for me.

PR #3445, opened 2021-12-29
Simplify test rig related to timeouts

parrt:
The test rig has a bit of complexity related to making sure that our continuous integration servers don't timeout.  I'm trying an experiment where I turn off the quiet option so that we are always getting output.  I think as long as all TestX.java finish within 10 minutes we should be okay because he will generates an output.  Also, I have turned on `-Dparallel=classes -DthreadCount=4` so that 4 TestX will run the same time.  This seems a decent level of granularity rather than creating a new thread for possibly 100 tests within a TestX.java file.  If this works then I can try taking out the heartbeat mechanism which spits out `.` every once in a while to let the integration server know that we are still running.

### DIFF
--- a/.appveyor/workflow.yml
+++ b/.appveyor/workflow.yml
@@ -32,24 +32,24 @@ for:
       only:
         - job_name: java-tool-and-runtime
     build_script:
-      - mvn -q -DskipTests install --batch-mode
+      - mvn -DskipTests install --batch-mode
     test_script:
       - cd tool-testsuite
-      - mvn -q test
+      - mvn test
       - cd ..\runtime-testsuite
-      - mvn -q -Dtest=java.** test
+      - mvn -Dtest=java.** test
 
   - matrix:
       only:
         - job_name: csharp-runtime
     build_script:
-      - mvn -q -DskipTests install --batch-mode
+      - mvn -DskipTests install --batch-mode
       - dotnet build runtime/CSharp/src/Antlr4.csproj -c Release
     after_build:
       - dotnet pack runtime/CSharp/src/Antlr4.csproj -c Release
     test_script:
       - cd runtime-testsuite
-      - mvn -q -Dtest=csharp.** test
+      - mvn -Dtest=csharp.** test
     artifacts:
       - path: 'runtime\**\*.nupkg'
         name: NuGet
@@ -60,19 +60,19 @@ for:
     install:
       - cinst -y dart-sdk --version=2.12.1
     build_script:
-      - mvn -q -DskipTests install --batch-mode
+      - mvn -DskipTests install --batch-mode
     test_script:
       - cd runtime-testsuite
-      - mvn -q -Dtest=dart.** test -Dantlr-dart-dart="C:\tools\dart-sdk\bin\dart.exe" -Dantlr-dart-pub="C:\tools\dart-sdk\bin\pub.bat" -Dantlr-dart-dart2native="C:\tools\dart-sdk\bin\dart2native.bat"
+      - mvn -Dtest=dart.** test -Dantlr-dart-dart="C:\tools\dart-sdk\bin\dart.exe" -Dantlr-dart-pub="C:\tools\dart-sdk\bin\pub.bat" -Dantlr-dart-dart2native="C:\tools\dart-sdk\bin\dart2native.bat"
 
   - matrix:
       only:
         - job_name: go-runtime
     build_script:
-      - mvn -q -DskipTests install --batch-mode
+      - mvn -DskipTests install --batch-mode
     test_script:
       - cd runtime-testsuite
-      - mvn -q -Dtest=go.** test
+      - mvn -Dtest=go.** test
 
   - matrix:
       only:
@@ -87,13 +87,13 @@ for:
       - npm install
       - npm link
       - cd ..\..
-      - mvn -q -DskipTests install --batch-mode
+      - mvn -DskipTests install --batch-mode
     test_script:
       - cd runtime\JavaScript\
       - yarn test
       - cd ..\..
       - cd runtime-testsuite
-      - mvn -q -Dtest=javascript.** test -Dantlr-javascript-npm="C:\Program Files\nodejs\npm.cmd" -Dantlr-javascript-nodejs="C:\Program Files\nodejs\node.exe"
+      - mvn -Dtest=javascript.** test -Dantlr-javascript-npm="C:\Program Files\nodejs\npm.cmd" -Dantlr-javascript-nodejs="C:\Program Files\nodejs\node.exe"
 
   - matrix:
       only:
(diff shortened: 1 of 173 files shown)


parrt:
Wow. C++ test on my local box is much faster with 4 test classes running at once.  Single threaded is:

```
[INFO] Total time:  26:52 min
```

But with four classes running at once we get:

```
[INFO] Total time:  07:16 min
```

KvanTTT:
> I'm trying an experiment where I turn off the quiet option so that we are always getting output.

Maybe it makes sense to print the current test name after its completion? It's more clear than just a dot.

KvanTTT:
> This seems a decent level of granularity rather than creating a new thread for possibly 100 tests within a TestX.java file.

Thread is very cheap compare to the process that is created in every test (or several processes). I still don't understand why maven creates a lot of threads that halt your computer (you said before) with the defined option `-DthreadCount=4`.

parrt:
Interesting.  parallelism works for C++ on my mac and circleci but fails on github.  dotnet fails on circleci saying that multiple threads are working on the same artifact; is it rebuilding antlr runtime each test? waiting on results from appveyor and github.

btw, there are 1 billion of these messages for C++:

```
ld: warning: dylib (./libantlr4-runtime.dylib) was built for newer macOS version (11.2) than being linked (11.0)
```

parrt:
@KvanTTT  I can't reproduce the threading issue but it was only with Swift anyway.  I also saw that it did something very strange with threading=methods: it didn't actually launch multiple threads where as it does with classes.

getting different behavior on different integration servers is really driving me nuts.  I'm happy to throw away this PR / approach, but I would like to increase the speed and simplicity of the testing.

parrt:
instead of using something from the command line, perhaps it's better to update the poms like @jcking  [suggests](https://github.com/antlr/antlr4/pull/3423/files):

```
<parallel>all</parallel>
<useUnlimitedThreads>true</useUnlimitedThreads>
<forkCount>0</forkCount>
```

KvanTTT:
> is it rebuilding antlr runtime each test?

No, it's being built a single time during the test session. But C++ is very slow itself and even one test building is slow.

> btw, there are 1 billion of these messages for C++:

It's too many, maybe it makes sense to disable all or almost all warnings.

> instead of using something from the command line, perhaps it's better to update the poms like @jcking suggests:

It sounds good. I hope it also will reduce duplicated maven commands in config files.

> Maybe just dotnet and C++ are the issue, in which case, I'm happy to keep them single threaded until we can come up with a solution.

C# does not work with the following error:

```
The process cannot access the file '/tmp/ANTLR-runtime-testsuite-cache/CSharp/Antlr4.Runtime.Standard.deps.json' because it is being used by another process. [/home/circleci/project/runtime-testsuite/target/classes/CSharp/src/Antlr4.csproj]
```

I came up with two solutions to how it could be resolved. The good one is to isolate processes, the bad one is to retry failed test while the file is locked. Interesting, I'll take a look later. But I definitely want to make parallel C#.

>  I'm happy to throw away this PR / approach, but I would like to increase the speed and simplicity of the testing.

A month ago it was much worse with speed :)

parrt:
@KvanTTT I'm working on C#...might have a solution for parallel.

parrt:
Wow. clearly creating 40 processes to run with parallel classes. wth?!  I'll try methods next then modifying pom ala @jcking

parrt:
i really don't understand mave. Check this out

```
mvn -Dparallel=methods -DuseUnlimitedThreads -Dtest=csharp.** test
```

Screen Shot 2021-12-29 at 1 03 40 PM

So just 4 processes (with many threads?) but just 1 process running? i could have messed up the synch for building runtime but...

parrt:
Wow. Something is really really broken with maven or surefire or whatever.   The thread / parallel stuff makes zero sense in its behavior. Wasted 4 hours so far.

parrt:
Ah. I got it. The CI uses a matrix (launched in parallel I guess) that means multiple mvn tests which means multiple runtime lib builds at once in race condition:

```
      - test_runtime:
          matrix:
            parameters:
              target: [ dart, go, python2, python3, javascript, php ]
```

parrt:
Holy crap! @KvanTTT and @ericvergnaud all tests pass on appveyor, circle, github!  this is *after* I removed the fancy stuff with test groups like parser1, parser2.  All are parallel except swift.  Test scripts are reduced to this kinda thing:

```
pushd runtime-testsuite
mvn -Dparallel=classes -DthreadCount=4 -Dtest=cpp.** test
popd
```

ericvergnaud:
Good stuff. Appveyor only provides 1 concurrent build for free, we have 4 agents on the Github Mac, whereas CircleCI lets us have multiple agents with 2 threads each, so it all makes sense.

KvanTTT:
> 1 hour on appveyor

No improvement on AppVeyor, it was around 1 hour (before AOT changes): https://ci.appveyor.com/project/parrt/antlr4/builds/41984813

parrt:
My main concern here was to clean up the test rig and to get consistency not an improvement in speed.

@KvanTTT  are you okay with me squashing these commits and merging the PR?

KvanTTT:
Generally, I'm okay.

parrt:
ok, I think it's a good idea to use this as a steppingstone for the next round of cleanups.  Wow. appveyor is like 59$/month to get more than one concurrent job.

KvanTTT:
> ok, I think it's a good idea to use this as a steppingstone for the next round of cleanups. Wow. appveyor is like 59$/month to get more than one concurrent job.

Maybe it makes sense to ask them for free support because ANTLR is not a commercial but a valuable project. Also, I can try to ask JetBrains but probably TeamCity is not very suitable.

parrt:
hahah. if we rent a google cloud linux box, would this solve/help anything?

KvanTTT:
> hahah. if we rent a google cloud linux box, would this solve/help anything?

They don't have anything for Windows? :)

parrt:
Just to summarize a bit of what I learned.  We originally created the different groups of tests like parser1 and parser2 so that continuous integration servers would not timeout I have left in the heartbeat that spits out `.`s just in case that is still necessary. But it seems that as long as we get some output (not use quiet `-q` option on maven) at least every 10 minutes then I think we are good to go.

I also learned that this separate groups thing created a problem for parallel testing because group was executing a different maven instance which could not synchronize across VMs to prevent more than one runtime build.  This caused C# builds to complain when they see the same file being overwritten by another process.

I have turned on parallelism for all but the Swift target which I couldn't figure out:

```
mvn -Dparallel=classes -DthreadCount=4 -Dtest=cpp.** test
```

I also do it at the class level not the method level because that did not seem to actually make things go faster. I also had difficulty figuring out how many threads were running just by looking at the process list in activity monitor.  In the end, doing it class by class seemed to make things work.

I ran into a problem where 90 processes were created with `-Dparallel=classes -DthreadCount=4` on my M1 mac, or at least I think that was a combination. It really dragged the system down. Something to keep in mind that there might be some issue here.

I added `<forkCount>0</forkCount>` to the maven pom file in the runtime tests and it somehow made the code not compile. Very strange so I took it back out; note that it works from the command line with `-D`. ugh.

all of the CI are much simpler now because we don't have to test for the group and the code is much simpler because we don't have the annotations related to groups. parallelism makes at least circleCI complete much faster.

parrt:
> They don't have anything for Windows? :)

Azure is very expensive for anything more than a tiny 1 core box. probably wouldn't even run our stuff.

parrt:
Ok, tests all pass again... here we go!!!

ericvergnaud:
> > They don't have anything for Windows? :)
> 
> Azure is very expensive for anything more than a tiny 1 core box. probably wouldn't even run our stuff.

We could follow the same strategy we have for MacOS.

parrt:
That's what I was thinking. I would bet a $600 box even would work. I bought a laptop for $750.

ericvergnaud:
Yes, grab an Intel NUC, then you only spend on what this CI needs.

parrt:
So it'd connect to github actions like we do for mac os x now?  That would simplify our lives i think.

ericvergnaud:
Yes

PR #3447, opened 2021-12-30
cpp/cmake: Fix library install directories

Biswa96:
This installs DLLs in bin directory instead of lib.


mike-lischke (approved):
@parrt This can be merged.

PR #3450, opened 2021-12-31
use src 11 for tool, but 8 for plugin/runtime

parrt:
Wow. This appears to work.

### DIFF
--- a/.appveyor/workflow.yml
+++ b/.appveyor/workflow.yml
@@ -1,4 +1,6 @@
 environment:
+  JAVA_HOME: C:\Program Files\Java\jdk11
+  MAVEN_VERSION: 3.8
   matrix:
     - job_name: java-tool-and-runtime
     - job_name: csharp-runtime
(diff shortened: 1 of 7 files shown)


parrt:
Seems to use Java 8 for runtime!

```
$ od -x `find runtime/Java -name Token.class`|head
0000000      feca    beba    0000    3400    2500    0007    0722    2300
```

That `34` after 0xCAFEBABE means Java 8 (Java 7 is 33).

parrt:
Thanks to @ericvergnaud for pointing the obvious solution here to allow java 11 for tool while keeping java 8 runtime :)

KvanTTT (approved):

parrt:
Awesome. Updated appveyor as well. All java stuff passing...merging. woot!

ericvergnaud:
That's -really- cool!

wangyum:
Hi @parrt Why tool must use java 11?

parrt:
Because we wanted to use more recent versions of the language; I think there were some compatibility issues as well.

Issue #1614, opened 2017-01-18
Ability to invoke custom parse rule without reflection


KvanTTT:
At present time it's possible to use parse rule constants and names from `ruleNames` array. I can invoke any parse rule via reflection, but I don't want to do it. So, I suggest to add the following generated method to parser:

```CSharp
public ParserRuleContext customRule(int ruleId)
{
    switch (ruleId)
    {
        case MyParser.RULE_translationUnit:
            return translationUnit();
        case MyParser.RULE_expression:
            return expression();
        ...
        default:
            throw new InvalidOperationException("Invalid rule Id: " + ruleId);
    }
}
```

parrt:
Or perhaps just something like

```
	public interface Func0<TResult> {
		TResult exec();
	}
```

(or something from java 8) as a table of func ptrs?

Issue #1069, opened 2015-12-16
Optionally report unused lexer/parser rules


SimonSntPeter:
Sam Harwell points out this is dubious given that all parse rules can be considered roots from which a caller can start parsing from, and some rules deliberately left unused (his example <https://github.com/tunnelvisionlabs/antlr4-grammar-postgresql/blob/master/src/com/tunnelvisionlabs/postgresql/PostgreSqlLexer.g4#L609-L636>).
However it may be helpful to allow them to be reported. Definition of unused rule is any parse/lex rule unreferenced by any other. Put this onto stdout and let the user decide what to do with it.

KvanTTT:
Yes, it is especially usefull in grammar developing stage. Moreover, parsing from custom rule is rarely used feature.

Maybe additional attribute should be added to parser rule? Something like ```root```. If rule will have not this attribute and will have not any references, the warning will be generated.

KvanTTT:
What about such syntax for root rules?

```ANTLR
parserRule1, root
	: parserRule2
	| TOKEN
	...
	;
```

If at least one rule this "root" attribute exists, warnings will be generated for unreachable parser rules like this:

```ANTLR
parserRule1, root    // No warning, root.
	: parserRule2
	: parserRule3
	;

parserRule2          // Warning: this rule has not "root" attribute and references.
	: TOKEN1
	;

parserRule3          // No warning because it has reference from parserRule1.
	: parserRule4
	| TOKEN2
	;
	
parserRule5, root    // No warning here because it's root.
	: TOKEN3
	;
```

But if grammar has no root nodes, warnings will not be generated (for back compability issues with old format). In most cases only one root rule required in grammar development. But root node brings logical clarity.

KvanTTT:
Another idea: put "root" attribute before rule declaration (similary to token "fragment" modifier):
```ANTLR
root parserRule1
    : parserRule2
    | TOKEN
    ...
    ;
```
Moreover it can be extended to "caseSensitive" or "caseInsensitive" modifier in mode (implemented in [this PR](https://github.com/antlr/antlr4/pull/1092)):
```ANTLR
caseInsensitive mode PHP;
Abstract: 'abstract';
Array: 'array';
As: 'as';
```

KvanTTT:
If roots will be exist, they can be available in Parser class:
```C#
public override List<ParseRuleContext> Roots { get; }
```
Similary to ```Vocabulary```.

parrt:
[...]

PR #3451, opened 2022-01-01
Escape bad words during grammar generation

### LINKED ISSUE #1070: Have ANTLR4 prevent conflict with user rule names by behind-the-scenes renaming of its own variables
I have a grammar with rules named xif and xcontinue and others, which were renamed from if/continue/etc due to antlr creating variables with these names and clashing with java. It would be more polished if antlr distinguished its identifiers from any possible back-end reserved words somehow. 
A consistent prefixing scheme might suffice, preventing clashes but keep a human-comprehensible correspondence between the parser and the generated back-end code (perhaps have the prefix 'ANTLR4_' as a reserved prefix for antlr itself?).

KvanTTT:
fixes #1070 

Deprecate USE_OF_BAD_WORD

Ok, I'm converting this to a pull request since we've already started reviewing and discussing.

### DIFF
--- a/runtime-testsuite/resources/org/antlr/v4/test/runtime/descriptors/ParserExec/ReservedWordsEscaping.txt
+++ b/runtime-testsuite/resources/org/antlr/v4/test/runtime/descriptors/ParserExec/ReservedWordsEscaping.txt
@@ -1,0 +1,35 @@
+[notes]
+https://github.com/antlr/antlr4/issues/1070
+
+[type]
+Parser
+
+[grammar]
+grammar G;
+
+root
+    : {0==0}? continue+ {<writeln("$text")>}
+    ;
+
+continue returns [int return]
+    : for for? {1==1}?              #else
+    | break=BREAK BREAK+ (for | IF) #class
+    | if+=IF  if+=IF*               #int
+    | continue CONTINUE             #static
+    ;
+
+for: FOR;
+FOR: 'for_';
+BREAK: 'break_';
+IF: 'if_';
+CONTINUE: 'continue_';
+
+[start]
+root
+
+[input]
+for_for_break_break_for_if_if_for_continue_
+
+[output]
+"""for_for_break_break_for_if_if_for_continue_
+"""
--- a/runtime-testsuite/resources/org/antlr/v4/test/runtime/descriptors/ParserExec/ReservedWordsEscaping.txt
+++ b/runtime-testsuite/resources/org/antlr/v4/test/runtime/descriptors/ParserExec/ReservedWordsEscaping.txt
@@ -13,23 +13,27 @@ root
 
 continue returns [int return]
     : for for? {1==1}?              #else
-    | break=BREAK BREAK+ (for | IF) #class
+    | break=BREAK BREAK+ (for | IF) #else
     | if+=IF  if+=IF*               #int
-    | continue CONTINUE             #static
+    | continue CONTINUE_ {<AssignLocal("$return","0")>}   #class
+    ;
+
+args[int else] locals [int return]
+    : for
     ;
 
 for: FOR;
-FOR: 'for_';
-BREAK: 'break_';
-IF: 'if_';
-CONTINUE: 'continue_';
+FOR: 'for ';
+BREAK: 'break ';
+IF: 'if ';
+CONTINUE_: 'continue';
 
 [start]
 root
 
 [input]
-for_for_break_break_for_if_if_for_continue_
+for for break break for if if for continue
 
 [output]
-"""for_for_break_break_for_if_if_for_continue_
+"""for for break break for if if for continue
 """
--- a/tool/src/org/antlr/v4/codegen/ActionTranslator.java
+++ b/tool/src/org/antlr/v4/codegen/ActionTranslator.java
@@ -89,17 +89,19 @@ public class ActionTranslator implements ActionSplitterListener {
 		tokenPropToModelMap.put("int",   TokenPropertyRef_int.class);
 	}
 
-	CodeGenerator gen;
-	ActionAST node;
+	final CodeGenerator gen;
+	final Target target;
+	final ActionAST node;
 	RuleFunction rf;
-	List<ActionChunk> chunks = new ArrayList<ActionChunk>();
-	OutputModelFactory factory;
+	final List<ActionChunk> chunks = new ArrayList<ActionChunk>();
+	final OutputModelFactory factory;
 	StructDecl nodeContext;
 
 	public ActionTranslator(OutputModelFactory factory, ActionAST node) {
 		this.factory = factory;
 		this.node = node;
 		this.gen = factory.getGenerator();
+		this.target = gen.getTarget();
 	}
 
 	public static String toString(List<ActionChunk> chunks) {
@@ -151,31 +153,44 @@ public class ActionTranslator implements ActionSplitterListener {
 	public void attr(String expr, Token x) {
 		gen.g.tool.log("action-translator", "attr "+x);
 		Attribute a = node.resolver.resolveToAttribute(x.getText(), node);
+		String name = x.getText();
+		String escapedName = target.escapeIfNeeded(name);
 		if ( a!=null ) {
 			switch ( a.dict.type ) {
-				case ARG: chunks.add(new ArgRef(nodeContext,x.getText())); break;
-				case RET: chunks.add(new RetValueRef(rf.ruleCtx, x.getText())); break;
-				case LOCAL: chunks.add(new LocalRef(nodeContext,x.getText())); break;
-				case PREDEFINED_RULE: chunks.add(getRulePropertyRef(x));	break;
+				case ARG:
+					chunks.add(new ArgRef(nodeContext, name, escapedName));
+					break;
+				case RET:
+					chunks.add(new RetValueRef(rf.ruleCtx, name, escapedName));
+					break;
+				case LOCAL:
+					chunks.add(new LocalRef(nodeContext, name, escapedName));
+					break;
+				case PREDEFINED_RULE:
+					chunks.add(getRulePropertyRef(null, x));
+					break;
 				default:
 					break;
 			}
 		}
-		if ( node.resolver.resolvesToToken(x.getText(), node) ) {
-			chunks.add(new TokenRef(nodeContext,getTokenLabel(x.getText()))); // $label
+		if ( node.resolver.resolvesToToken(name, node) ) {
+			String tokenLabel = getTokenLabel(name);
+			chunks.add(new TokenRef(nodeContext, tokenLabel, target.escapeIfNeeded(tokenLabel))); // $label
 			return;
 		}
-		if ( node.resolver.resolvesToLabel(x.getText(), node) ) {
-			chunks.add(new LabelRef(nodeContext,getTokenLabel(x.getText()))); // $x for x=ID etc...
+		if ( node.resolver.resolvesToLabel(name, node) ) {
+			String tokenLabel = getTokenLabel(name);
+			chunks.add(new LabelRef(nodeContext, tokenLabel, target.escapeIfNeeded(tokenLabel))); // $x for x=ID etc...
 			return;
 		}
-		if ( node.resolver.resolvesToListLabel(x.getText(), node) ) {
-			chunks.add(new ListLabelRef(nodeContext,x.getText())); // $ids for ids+=ID etc...
+		if ( node.resolver.resolvesToListLabel(name, node) ) {
+			chunks.add(new ListLabelRef(nodeContext, name, escapedName)); // $ids for ids+=ID etc...
 			return;
 		}
-		Rule r = factory.getGrammar().getRule(x.getText());
+		Rule r = factory.getGrammar().getRule(name);
 		if ( r!=null ) {
-			chunks.add(new LabelRef(nodeContext,getRuleLabel(x.getText()))); // $r for r rule ref
+			String ruleLabel = getRuleLabel(name);
+			chunks.add(new LabelRef(nodeContext, ruleLabel, target.escapeIfNeeded(ruleLabel))); // $r for r rule ref
 		}
 	}
 
@@ -198,9 +213,9 @@ public class ActionTranslator implements ActionSplitterListener {
 			return;
 		}
 		switch ( a.dict.type ) {
-			case ARG: chunks.add(new ArgRef(nodeContext,y.getText())); break; // has to be current rule
+			case ARG: chunks.add(new ArgRef(nodeContext, y.getText(), target.escapeIfNeeded(y.getText()))); break; // has to be current rule
 			case RET:
-				chunks.add(new QRetValueRef(nodeContext, getRuleLabel(x.getText()), y.getText()));
+				chunks.add(new QRetValueRef(nodeContext, getRuleLabel(x.getText()), y.getText(), target.escapeIfNeeded(y.getText())));
 				break;
 			case PREDEFINED_RULE:
 				chunks.add(getRulePropertyRef(x, y));
@@ -217,7 +232,8 @@ public class ActionTranslator implements ActionSplitterListener {
 	public void setAttr(String expr, Token x, Token rhs) {
 		gen.g.tool.log("action-translator", "setAttr "+x+" "+rhs);
 		List<ActionChunk> rhsChunks = translateActionChunk(factory,rf,rhs.getText(),node);
-		SetAttr s = new SetAttr(nodeContext, x.getText(), rhsChunks);
+		String name = x.getText();
+		SetAttr s = new SetAttr(nodeContext, name, target.escapeIfNeeded(name), rhsChunks);
 		chunks.add(s);
 	}
 
@@ -225,7 +241,8 @@ public class ActionTranslator implements ActionSplitterListener {
 	public void nonLocalAttr(String expr, Token x, Token y) {
 		gen.g.tool.log("action-translator", "nonLocalAttr "+x+"::"+y);
 		Rule r = factory.getGrammar().getRule(x.getText());
-		chunks.add(new NonLocalAttrRef(nodeContext, x.getText(), y.getText(), r.index));
+		String name = y.getText();
+		chunks.add(new NonLocalAttrRef(nodeContext, x.getText(), name, target.escapeIfNeeded(name), r.index));
 	}
 
 	@Override
@@ -233,7 +250,8 @@ public class ActionTranslator implements ActionSplitterListener {
 		gen.g.tool.log("action-translator", "setNonLocalAttr "+x+"::"+y+"="+rhs);
 		Rule r = factory.getGrammar().getRule(x.getText());
 		List<ActionChunk> rhsChunks = translateActionChunk(factory,rf,rhs.getText(),node);
-		SetNonLocalAttr s = new SetNonLocalAttr(nodeContext, x.getText(), y.getText(), r.index, rhsChunks);
+		String name = y.getText();
+		SetNonLocalAttr s = new SetNonLocalAttr(nodeContext, x.getText(), name, target.escapeIfNeeded(name), r.index, rhsChunks);
 		chunks.add(s);
 	}
 
@@ -246,24 +264,7 @@ public class ActionTranslator implements ActionSplitterListener {
 		try {
 			Class<? extends TokenPropertyRef> c = tokenPropToModelMap.get(y.getText());
 			Constructor<? extends TokenPropertyRef> ctor = c.getConstructor(StructDecl.class, String.class);
-			TokenPropertyRef ref =
-				ctor.newInstance(nodeContext, getTokenLabel(x.getText()));
-			return ref;
-		}
-		catch (Exception e) {
-			factory.getGrammar().tool.errMgr.toolError(ErrorType.INTERNAL_ERROR, e);
-		}
-		return null;
-	}
-
-	// $text
-	RulePropertyRef getRulePropertyRef(Token prop) {
-		try {
-			Class<? extends RulePropertyRef> c = thisRulePropToModelMap.get(prop.getText());
-			Constructor<? extends RulePropertyRef> ctor = c.getConstructor(StructDecl.class, String.class);
-			RulePropertyRef ref =
-				ctor.newInstance(nodeContext, getRuleLabel(prop.getText()));
-			return ref;
+			return ctor.newInstance(nodeContext, getTokenLabel(x.getText()));
 		}
 		catch (Exception e) {
 			factory.getGrammar().tool.errMgr.toolError(ErrorType.INTERNAL_ERROR, e);
@@ -272,28 +273,24 @@ public class ActionTranslator implements ActionSplitterListener {
 	}
 
 	RulePropertyRef getRulePropertyRef(Token x, Token prop) {
-		Grammar g = factory.getGrammar();
 		try {
-			Class<? extends RulePropertyRef> c = rulePropToModelMap.get(prop.getText());
+			Class<? extends RulePropertyRef> c = (x != null ? rulePropToModelMap : thisRulePropToModelMap).get(prop.getText());
 			Constructor<? extends RulePropertyRef> ctor = c.getConstructor(StructDecl.class, String.class);
-			RulePropertyRef ref =
-				ctor.newInstance(nodeContext, getRuleLabel(x.getText()));
-			return ref;
+			return ctor.newInstance(nodeContext, getRuleLabel((x != null ? x : prop).getText()));
 		}
 		catch (Exception e) {
-			g.tool.errMgr.toolError(ErrorType.INTERNAL_ERROR, e, prop.getText());
+			factory.getGrammar().tool.errMgr.toolError(ErrorType.INTERNAL_ERROR, e, prop.getText());
 		}
 		return null;
 	}
 
 	public String getTokenLabel(String x) {
 		if ( node.resolver.resolvesToLabel(x, node) ) return x;
-		return factory.getGenerator().getTarget().getImplicitTokenLabel(x);
+		return target.getImplicitTokenLabel(x);
 	}
 
 	public String getRuleLabel(String x) {
 		if ( node.resolver.resolvesToLabel(x, node) ) return x;
-		return factory.getGenerator().getTarget().getImplicitRuleLabel(x);
+		return target.getImplicitRuleLabel(x);
 	}
-
 }
--- a/tool/src/org/antlr/v4/codegen/model/InvokeRule.java
+++ b/tool/src/org/antlr/v4/codegen/model/InvokeRule.java
@@ -24,8 +24,8 @@ import java.util.List;
 
 /** */
 public class InvokeRule extends RuleElement implements LabeledOp {
-	public final String originalName;
 	public final String name;
+	public final String escapedName;
 	public final OrderedHashSet<Decl> labels = new OrderedHashSet<Decl>(); // TODO: should need just 1
 	public final String ctxName;
 
@@ -41,8 +41,8 @@ public class InvokeRule extends RuleElement implements LabeledOp {
 		Target target = gen.getTarget();
 		String identifier = ast.getText();
 		Rule r = factory.getGrammar().getRule(identifier);
-		this.originalName = r.name;
-		this.name = r.escapedName;
+		this.name = r.name;
+		this.escapedName = gen.getTarget().escapeIfNeeded(name);
 		ctxName = target.getRuleFunctionContextStructName(r);
 
 		// TODO: move to factory
--- a/tool/src/org/antlr/v4/codegen/model/InvokeRule.java
+++ b/tool/src/org/antlr/v4/codegen/model/InvokeRule.java
@@ -9,12 +9,12 @@ package org.antlr.v4.codegen.model;
 import org.antlr.v4.codegen.ActionTranslator;
 import org.antlr.v4.codegen.CodeGenerator;
 import org.antlr.v4.codegen.ParserFactory;
+import org.antlr.v4.codegen.Target;
 import org.antlr.v4.codegen.model.chunk.ActionChunk;
 import org.antlr.v4.codegen.model.decl.Decl;
 import org.antlr.v4.codegen.model.decl.RuleContextDecl;
 import org.antlr.v4.codegen.model.decl.RuleContextListDecl;
 import org.antlr.v4.parse.ANTLRParser;
-import org.antlr.v4.runtime.atn.RuleTransition;
 import org.antlr.v4.runtime.misc.OrderedHashSet;
 import org.antlr.v4.tool.Rule;
 import org.antlr.v4.tool.ast.ActionAST;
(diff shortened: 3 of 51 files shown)


KvanTTT:
I didn't want to make a draft because of failed checks. But actually, it does not care for now.

KvanTTT:
> To get an updated, combined list, the easiest would be to use trxgrep and look for anything that ends in an underscore. But, that would be a merged list for all targets.

I realized it's ok to get a list even for all targets. I'll create a big test with all these words and run it on different runtimes. Failed runtimes will indicate correct targets of these words :)

kaby76:
Repeating what I said over in the closed Issue, here is the aggregated bad word list (with underscores) derived from symbols in grammars-v4.

[all_underscore.txt](https://github.com/antlr/antlr4/files/7802013/all_underscore.txt)

KvanTTT:
@parrt I think it makes sense to review and merge it now since there are a lot of fixes already. I'll add other reserved words provided by @kaby76 and fix potential bugs in the following requests (but now all tests are okay).

parrt:
> I decided to use `orginalName` for a name from grammar and `escapedName` for a name in runtime.

I always worry about making the smallest possible change, but I guess this makes sense to highlight the difference now.

Just so I'm clear, the list of reserved words will always be specific to a target right?   I don't think it makes sense to have a general list that is always escaped.

I will try to review this branch today

runtime-testsuite/resources/org/antlr/v4/test/runtime/descriptors/ParserExec/ReservedWordsEscaping.txt
```diff
@@ -0,0 +1,35 @@
...
+Parser
+
+[grammar]
+grammar G;
+
+root
+    : {0==0}? continue+ {<writeln("$text")>}
+    ;
+
+continue returns [int return]
```
parrt, line 14: I think we should add an argument here as well, per my other comment. Java definitely generates variables from grammar argument names.

tool/src/org/antlr/v4/codegen/model/InvokeRule.java
```diff
@@ -24,40 +24,43 @@
 
 /** */
 public class InvokeRule extends RuleElement implements LabeledOp {
-	public String name;
-	public OrderedHashSet<Decl> labels = new OrderedHashSet<Decl>(); // TODO: should need just 1
-	public String ctxName;
+	public final String originalName;
+	public final String name;
```
parrt, line 28: shouldn't this be the original name?  what is the difference? if it is the escaped name then we should probably call it that.

tool/src/org/antlr/v4/codegen/model/InvokeRule.java
```diff
@@ -24,40 +24,43 @@
...
 
-		this.name = ast.getText();
 		CodeGenerator gen = factory.getGenerator();
-		Rule r = factory.getGrammar().getRule(name);
-		ctxName = gen.getTarget().getRuleFunctionContextStructName(r);
+		Target target = gen.getTarget();
+		String identifier = ast.getText();
+		Rule r = factory.getGrammar().getRule(identifier);
+		this.originalName = r.name;
+		this.name = r.escapedName;
```
parrt, line 45: Yep, the `name` should either be the original or say it is escaped.   I think this should be consistent with code generation templates. Hmm... now that I think about it I actually do prefer `r.name` as the original so that it is as it was, without change, both here and in the code generation templates.   I think it's understood that `<r.name>` is the original name and this leaves many fewer changes when we look at this change later.

tool/src/org/antlr/v4/codegen/model/Lexer.java
```diff
@@ -7,18 +7,19 @@
...
-import java.util.LinkedHashMap;
-import java.util.Map;
+import java.util.*;
 
 public class Lexer extends Recognizer {
-	public Map<String,Integer> channels;
-	public LexerFile file;
-	public Collection<String> modes;
+	public final Map<String, Integer> channels;
+	public final Collection<String> channelNames;
```
parrt, line 19: was this field introduced to make code generation easier?

tool/src/org/antlr/v4/codegen/model/Lexer.java
```diff
@@ -28,7 +29,20 @@ public Lexer(OutputModelFactory factory, LexerFile file) {
 		this.file = file; // who contains us?
 
 		Grammar g = factory.getGrammar();
-		channels = new LinkedHashMap<String, Integer>(g.channelNameToValueMap);
+		Target target = factory.getGenerator().getTarget();
+
+		channels = new LinkedHashMap<>();
+		channelNames = new ArrayList<>();
```
parrt, line 35: Shouldn't this be escaped channel names or something?  seems inconsistent

tool/src/org/antlr/v4/codegen/model/RuleActionFunction.java
```diff
@@ -12,17 +12,19 @@
...
+	public final int ruleIndex;
 
 	/** Map actionIndex to Action */
 	@ModelElement public LinkedHashMap<Integer, Action> actions =
 		new LinkedHashMap<Integer, Action>();
 
 	public RuleActionFunction(OutputModelFactory factory, Rule r, String ctxType) {
 		super(factory);
-		name = r.name;
+		originalName = r.name;
```
parrt, line 26: yep, this definitely seems backwards...

tool/src/org/antlr/v4/codegen/target/CSharpTarget.java
```diff
@@ -8,22 +8,119 @@
...
+		return reservedWords;
+	}
+
+	@Override
+	protected String escapeWord(String word) {
+		return "@" + word;
+	}
+
+	@Override
+	protected boolean supportsWordEscaping() { return true; }
```
parrt, line 122: why not just have `escapeWord(s)` return `s` instead of having a separate method to test whether we need to escape?

tool/src/org/antlr/v4/codegen/target/CppTarget.java
```diff
@@ -37,34 +36,24 @@
 		"switch", "template", "this", "thread_local", "throw", "true",
 		"try", "typedef", "typeid", "typename", "union", "unsigned",
 		"using", "virtual", "void", "volatile", "wchar_t", "while",
-		"xor", "xor_eq"
-	};
+		"xor", "xor_eq",
 
-	/** Avoid grammar symbols in this set to prevent conflicts in gen'd code. */
-	protected final Set<String> badWords = new HashSet<String>();
+		"rule", "parserRule"
```
parrt, line 41: wow `parserRule` is a reserved word in C++?

tool/src/org/antlr/v4/codegen/target/DartTarget.java
```diff
@@ -38,10 +33,9 @@
 		"default", "get", "rethrow", "while",
 		"deferred", "hide", "return", "with",
 		"do", "if", "set", "yield",
-	};
 
-	/// Avoid grammar symbols in this set to prevent conflicts in gen'd code.
-	protected final Set<String> badWords = new HashSet<String>();
+		"rule", "parserRule"
```
parrt, line 37: I see that this was copied from the previous targets but I'm wondering what the logic is...

parrt:
looking really good!  thanks for your efforts here. I just have a number of tiny comments and then the one issue about naming.  I think we want to leave `<r.name>` in the code generation templates as the original name and keep it consistent so that field `name` in support code should be the original name.  We then have `name` and `escapedName` or whatever in the code generation templates, and a field if we need it that way.

KvanTTT:
> Just so I'm clear, the list of reserved words will always be specific to a target right? I don't think it makes sense to have a general list that is always escaped.

Generally yes, but I'm trying to figure out if common words exist (for instance, it looks like the word `rule` is common for all targets, but I haven't found out why it is so yet).

tool/src/org/antlr/v4/codegen/model/InvokeRule.java
KvanTTT: `originalName` is used in templates for expression with concatenation or with case changing: `<r.originalName>_sempred`, `<f.originalName; format="cap">`. Just `name` is used for plain identifiers: `<r.name>(...)`.

tool/src/org/antlr/v4/codegen/model/InvokeRule.java
KvanTTT: In this way, we are returning to the option when `for` is being transformed to `RULE_for_` (and this option causes other fixes, see https://github.com/antlr/antlr4/pull/3453). We've decided to escape only conflicting identifiers, i.e.: `for_`.

tool/src/org/antlr/v4/codegen/model/Lexer.java
KvanTTT: Rather more correct. Literal channel and mode names are being preserved, but identifier names are being escaped:

```java
	public static final int
		A=1, B=2;
	public static final int
		break_=2;
	public static final int
		for_=1;
	public static String[] channelNames = {
		"DEFAULT_TOKEN_CHANNEL", "HIDDEN", "break"
	};

	public static String[] modeNames = {
		"DEFAULT_MODE", "for"
	};
```

tool/src/org/antlr/v4/codegen/model/Lexer.java
KvanTTT: Yes, inconsistent. More correctly: `orginialChannelNames` or `runtimeChannels` but I wanted to introduce minimal code changes.

tool/src/org/antlr/v4/codegen/model/RuleActionFunction.java
KvanTTT: Actually, there is a logic here:

* If an entity is located in tool package (analyzer) then it has `name` for original name and `escapedName` for runtime name (if presented)
* If an entity is located in codegen package then it has `name` for runtime name and `originalName` for grammar name.

It's for minimizing code changes.

tool/src/org/antlr/v4/codegen/target/CSharpTarget.java
KvanTTT: Yes, now this method is redundant and should be deleted (but it's required for https://github.com/antlr/antlr4/pull/3453).

tool/src/org/antlr/v4/codegen/target/CppTarget.java
KvanTTT: I don't think so, this word is presented in all targets and I don't know why. I'm trying to figure it out right now.

tool/src/org/antlr/v4/codegen/target/DartTarget.java
KvanTTT: As well as I.

parrt:
> Generally yes, but I'm trying to figure out if common words exist (for instance, it looks like the word `rule` is common for all targets, but I haven't found out why it is so yet).

Yeah, that is really strange. There must be a quirk of code generation that I/we were trying to overcome.  I say we try it without those weird reserved words and see what happens.

KvanTTT:
> I think we want to leave <r.name> in the code generation templates as the original name and keep it consistent so that field name in support code should be the original name. We then have name and escapedName or whatever in the code generation templates, and a field if we need it that way.

Do you mean `escapedName` should be always escaped name independence of class package (tool or runtime)? `originalName` should be eliminated since `name` is always the original name. It's good for consistency and clarity but may cause more template changes that you'd like to get rid as I understand.

parrt:
> Do you mean `escapedName` should be always the escaped name independence of class package (tool or runtime)?

I would imagine that the simplest is that name is always the original name and `escapedName` is either the original name, or if necessary for that to get her target, and escaped name.  In other words, the code generation templates would always refer to name when they know for a fact there's no collision, such as in a string `"<r.name>"`. But, if there is a chance that the name could cause a problem in the generated code, we would always refer to it as `<r.escapedName>` even if the name is not actually escaped. Does that make sense?

> originalName should be eliminated since name is always the original name.

Yep!

tool/src/org/antlr/v4/codegen/model/InvokeRule.java
parrt: Well, We are assuming `RULE_<r.name>` here since we know that there will not be a collision.

tool/src/org/antlr/v4/codegen/model/Lexer.java
parrt: Without looking too closely, I was just wondering why we needed a new field when we must've been generating the list of channel names previously in the code generation.   I tend to avoid saving a field if I can computed on the fly quickly, to prevent having to keep things consistent between fields channels and channelNames, for example.

tool/src/org/antlr/v4/codegen/model/RuleActionFunction.java
parrt: Well, things get blurred when we're looking at the tool and the Java runtime together so I think we need to make these consistent. name should always be the original.  if we have the potential for an escaped value, that field should be so named.

tool/src/org/antlr/v4/codegen/target/CppTarget.java
parrt: hahaha, yeah, so weird

tool/src/org/antlr/v4/codegen/model/Lexer.java
KvanTTT: Yes, it's a good note. But we don't have a class for mode or channel that contains information about original and escaped names. Should I introduce them?

```
class Channel {
    int key;
    String name;
    String escapedName;
    ...
```

parrt: Oh I see.  we do need original and escaped but let's use the HashMap key set for the originals and create escapeChannelNames field for the escaped ones to keep it consistent :)

KvanTTT:
>  But, if there is a chance that the name could cause a problem in the generated code, we would always refer to it as <r.escapedName> even if the name is not actually escaped. Does that make sense?

Yes, it sounds reasonable and I like that. The only thing I was afraid of is massive template changes. Ok, I will try to implement it.

kaby76:
I'm also trying to envision how the changed code would work.

Let's suppose the grammar Foobar.g4 `grammar Foobar; boolean: 'true' | 'false';`, with pom.xml for the Antlr4 Maven Plugin that gives entry point "boolean". This grammar is valid in CSharp but invalid in Java.

* The grammar works for C#. I can generate a working C# parser, and there would be no compilation errors. The Antlr tool would work and the compilation of the generated files and existing code would work fine.
* This code will not work for Java. If I try to do this for the Java target (`java -jar C:/Users/Kenne/Downloads/antlr-4.9.3-complete.jar -Dlanguage=Java -encoding utf-8  Foobar.g4`), I cannot because the Antlr tool will emit `error(134): Foobar.g4:2:0: symbol boolean conflicts with generated code in target language or runtime`.

Up to now, if I wanted this "target agnostic", I'd have to change quite a bit. In the .g4 file, I would change `boolean` to `boolean_`; In the pom.xml, I would change `<entryPoint>boolean</entryPoint>` to `<entryPoint>boolean_</entryPoint>`. If I had written a listener or visitor, I would have to change that because the renamed symbol would result in new generated code declarations.

With this PR change, would I have to change anything in the driver, visitors or listeners I wrote in C#? Is the entry point in C# still `parser.boolean();`?

For Java, it would all be new code because the Antlr tool wouldn't have even generated any Java code prior to the PR. What would be the entry point be? `parser.boolean()` works fine for C# but not for Java, right? Are you planning on a string-named entry point, e.g., parser.EntryPoint("boolean") instead of "parser.boolean()"?

parrt:
Hi @kaby76.  current plan is to generate `boolean()` if it does not conflict for that target but `boolean_()`, or whatever the escape is, if it would conflict with the target.  That would mean that a maven plugin that wanted to call a rule would have to call the antlr escape function to get the appropriate function name.

KvanTTT:
@kaby76 good note.

> Up to now, if I wanted this "target agnostic", I'd have to change quite a bit. In the .g4 file, I would change boolean to boolean_; In the pom.xml, I would change <entryPoint>boolean</entryPoint> to <entryPoint>boolean_</entryPoint>.

It looks like the maven plugin relies on reflection. But I think it is possible to call `Target.escapeIfNeeded` from the ANTLR tool to get the correct method name for a certain target. That's why I don't think a big problem. Also, it's a more rare case compared to the number of conflicting identifiers and related issues.

> Are you planning on a string-named entry point, e.g., parser.EntryPoint("boolean") instead of "parser.boolean()"?

I don't think so, at least not now. By the way, a similar issue exists: https://github.com/antlr/antlr4/issues/1614
But as I've said above, it's not a big problem for now and it can be resolved using `Target.escapeIfNeeded`.

kaby76:
Yes, this will complicate things for trgen because the method to call (e.g., "parser.method_()") is per-target and per-version of the Antlr tool. The implementation for `Target.escapeIfNeeded(string) => string` is in the Java Antlr tool code (antlr4/tool/src/org/antlr/v4/codegen/Target.java), not in the runtime (C#, Python3, JavaScript, etc). And, there isn't a table generated by the Antlr tool to note the mapping ("boolean" => "boolean_"). Likely I will need to do call from C# to Java of escapeIfNeeded() in the Antlr tool jar. Ick.

KvanTTT:
In this case, I suggest using even a more simple algorithm: at first try to find the method `x`, if it's not found, then use `x_`.

KvanTTT:
Also, I have another idea, maybe more correct but it needs implementation. It's introducing a new parse rule option `{ root=true; }` that provides the following:

* Root (actually it can be several roots) is defined right in grammar, everyone shouldn't think about what rule should be used for parsing, just call `parser.parseRoot()` or similar.
* ANTLR will report `unused rule` warnings if at least one root is defined (I've described how it works in detail: 
https://github.com/antlr/antlr4/issues/1069#issuecomment-170295982).

Also, the new option is fully back portable.

ericvergnaud:
> Also, I have another idea, maybe more correct but it needs implementation. It's introducing a new parse rule option `{ root=true; }` (no ANTLR grammar changes are required) that provides the following:
> 
> * Root (actually it can be several roots) is defined right in grammar, everyone shouldn't think about what rule should be used for parsing, just call `parser.parseRoot()` or similar.
> * ANTLR will report `unused rule` warnings if at least one root is defined (I've described how it works in detail:
>   [Optionally report unused lexer/parser rules #1069 (comment)](https://github.com/antlr/antlr4/issues/1069#issuecomment-170295982)).
> 
> Almost in all use cases, a single root is used. But it's not necessary to use `root` option at all, the new option is fully back compatible.

Not sure I like this idea. Once a parser is generated, you might want to call a rule, if only for testing, even though you thought you'd never need to (so didn't declare it as root (=start)). And you don't want to regenerate everything just because of that.

KvanTTT:
> Not sure I like this idea. Once a parser is generated, you might want to call a rule, if only for testing, even though you thought you'd never need to (so didn't declare it as root (=start)). And you don't want to regenerate everything just because of that.

You will be able to call any rule as it was before. `Roots` is just an additional member that will contain rules marked with `root` option. Everything else will be the same. Also, `root` option affects generation warnings:

```g4
root { options root=true; } : x; // no warning since root=true, available from parser, roots contains this rule
no_root: 'else'; // warning since there is no reference and no option, but it's still available from parser
x: 'x'; // no warning since x is reachable, it's still available from parser
```

Without `root` option:

```g4
root : x; // no warning (option root is absent)
no_root: 'else'; // no warning 
x: 'x'; // no warning
```

I suggest such an option because in most cases something like `compilationUnit` is used.

kaby76:
I would skip on adding reports on unused rules. It seems beyond what a parser generator should do. Aside, Trash does this in two ways that complement each other:

* using a beautiful single XPath expression that finds a LHS symbol not used on the RHS of any rule;
* parsing the target source code and looking for a reference to a parser rule method. (I think I implemented this only for C#, but I was planning to do that for all targets, and using XPath expressions for the implementation.)

I'd also skip on the start rule spec. As Eric points out, a grammar can have multiple entry points (e.g., ISO C++ for the preprocessor and post-processed C++ code).

tool/src/org/antlr/v4/codegen/model/Lexer.java
parrt: Ok, let's go with consistency and more extensive changes here. thanks!

parrt:
Yup. I'm opposed to adding new option and limiting start rules.

kaby76:
As I noted [here](https://github.com/antlr/antlr4/pull/3453#issuecomment-1004073352) and [here](https://github.com/antlr/antlr4/pull/3453#issuecomment-1003770186), I am working out a reserved word list per target. After I got through the CSharp target, I was surprised that "await" wasn't a bad word. While it is ndeed an okay parser rule name to use, I then realized that the current tool already checks for "bad words". So I killed the running script that does the checking, and restarted it with a custom Antlr tool 4.9.3 that has minimal "bad word" checking. The script is attached. 
[try-symbol.txt](https://github.com/antlr/antlr4/files/7815790/try-symbol.txt). It will take some time. Note, I noticed that the "reserved-word lists" contain names that begin with underscore. These are skipped because the name cannot be a TOKEN_REF nor RULE_REF in a grammar. Not sure why they are in any of the "reserved-word lists".

KvanTTT:
> After I got through the CSharp target, I was surprised that "await" wasn't a bad word. 

Yes, this is because it was introduced not in the first version. `var` is also not a keyword in C#.

> As I noted here and here, I am working out a reserved word list per target.

Thank you for your work but wow I think it's enough to have global bad words list because I've already included them in a big test that I'm going to add in the following pull request. So, there is no need to split them by different runtimes.

KvanTTT:
Interesting: PHP runtime does require a bad list at least for keywords because the following code is valid (you try in a [sandbox](https://wtools.io/php-sandbox/b8Ee)):

```php
<?php

class Test {
  public $return = "return";
}

$test = new Test();
echo $test->return;
```

That's why there are almost no changes in PHP.stg at least yet.

But PHP has another problem because of case insensitiveness for some identifiers: https://github.com/antlr/antlr4/issues/3462

@parrt generally I've fixed all notes, and I'm going to update the reserved words list and fix potential other issues in the following PR since this request already has a lot of changes.

parrt:
> I think it's enough to have global bad words list

I think per target is what we want. No need to combine I don't think.

parrt:
> generally I've fixed all notes, and I'm going to update the reserved words list and fix potential other issues in the following PR since this request already has a lot of changes.

sounds good. I'll take a look shortly.  let's try to merge a solution to runtime collisions soon!

KvanTTT:
> I think per target is what we want. No need to combine I don't think.

I meant global list is convenient for test creating :) Of course, all targets have specified reserved word lists.

KvanTTT:
> I don't think it makes sense to have a general list that is always escaped.

If such a list exists, the error [`RESERVED_RULE_NAME`](https://github.com/antlr/antlr4/blob/d8e152592fb4dbdd6c9fda2c169031cca17dfc4f/tool/src/org/antlr/v4/tool/ErrorType.java#L901) should be reported on the generation step, not on runtime. Currently, we have the following words in this list:

```
EOF
MORE
SKIP
```

But I'm afraid it's not the full list and it will be filled up after further investigation and words provided by @kaby76

tool/src/org/antlr/v4/codegen/ActionTranslator.java
```diff
@@ -272,28 +273,24 @@ RulePropertyRef getRulePropertyRef(Token prop) {
...
+			Class<? extends RulePropertyRef> c = (x != null ? rulePropToModelMap : thisRulePropToModelMap).get(prop.getText());
 			Constructor<? extends RulePropertyRef> ctor = c.getConstructor(StructDecl.class, String.class);
-			RulePropertyRef ref =
-				ctor.newInstance(nodeContext, getRuleLabel(x.getText()));
-			return ref;
+			return ctor.newInstance(nodeContext, getRuleLabel((x != null ? x : prop).getText()));
 		}
 		catch (Exception e) {
-			g.tool.errMgr.toolError(ErrorType.INTERNAL_ERROR, e, prop.getText());
+			factory.getGrammar().tool.errMgr.toolError(ErrorType.INTERNAL_ERROR, e, prop.getText());
```
parrt, line 282: why factory.getGrammar() here but just target below line 289?

parrt: sorry i'm an idiot. diff values.

tool/src/org/antlr/v4/codegen/model/chunk/NamedActionChunk.java
```diff
@@ -0,0 +1,14 @@
+package org.antlr.v4.codegen.model.chunk;
+
+import org.antlr.v4.codegen.model.decl.StructDecl;
+
+public abstract class NamedActionChunk extends ActionChunk {
```
parrt, line 5: I'm not sure NamedActionChunk name is the right one.  Looks to me like it is some kind of symbol reference chunk.

parrt:
Fantastic work @KvanTTT.  beautiful. I just have that one nit about the naming of the superclass chunk... what you think?

tool/src/org/antlr/v4/codegen/model/chunk/NamedActionChunk.java
KvanTTT: I'm not sure about this entity naming, probably a more deep understanding is needed. Now it's superclass for `LabelRef`, `LocalRef`, `NonLocalAttrRef`, `RetValueRef`, `SetAttr`, `TokenRef`.

Ok, maybe you are right. I'm renaming it to `SymbolReferenceChunck`, is it ok?

parrt: Sure maybe shorter like `SymbolRefChunk` to be similar.

KvanTTT:
Also, dart tests are failed. But it looks like a problem on CI:

```
Error: Could not resolve the package 'antlr4' in 'package:antlr4/antlr4.dart'.
```

It works fine locally.

KvanTTT:
> > As I noted here and here, I am working out a reserved word list per target.
> 
> Thank you for your work but wow I think it's enough to have global bad words list because I've already included them in a big test that I'm going to add in the following pull request. So, there is no need to split them by different runtimes.

Ok, reserved word lists per target are also useful because with them I can update reserved words in ANTLR, probably remove actually not reserved words.

kaby76:
Here are "bad words" for C#, Java, Dart, JavaScript. Working out C++, Go, Python3, PHP.

[final-list.javascript.txt](https://github.com/antlr/antlr4/files/7817642/final-list.javascript.txt)
[final-list.csharp.txt](https://github.com/antlr/antlr4/files/7817643/final-list.csharp.txt)
[final-list.dart.txt](https://github.com/antlr/antlr4/files/7817644/final-list.dart.txt)
[final-list.java.txt](https://github.com/antlr/antlr4/files/7817645/final-list.java.txt)

Again, this uses a custom Antlr 4.9.3 tool with minimal bad word list, build of the parser, and a parse in order to test from start to end what is a bad word.

kaby76:
Attached are the last two "bad words", for Cpp and Go.

[final-list.cpp.txt](https://github.com/antlr/antlr4/files/7833996/final-list.cpp.txt)
[final-list.go.txt](https://github.com/antlr/antlr4/files/7834021/final-list.go.txt)

I did a check to see if this PR fixes [Issue #1070](https://github.com/antlr/antlr4/issues/1070) that it claims to fix. Indeed, there are many "bad words" that are not listed in the tables at https://github.com/antlr/antlr4/blob/7d2a37e68eacbaedade04903be802d9fc0e78ad1/tool/src/org/antlr/v4/codegen/target/GoTarget.java#L27 , and are not mangled on the generated files.

Checking the simple grammar `grammar Foobar; x : Accept; Accept: 'xxx';` against the latest source for the Antlr tool, this works fine for CSharp, but it does not for Go: the "go build" fails. There is no name mangling for `Accept`. 

This change does not fix "symbol conflicts": I will still have to change a "target agnostic" grammar for Go that declares a lexer rule for "Accept".

parrt:
Ok, @KvanTTT: care to update the word lists?

KvanTTT:
Sure, I'll update the word lists but a bit later. I've encountered yet another symbol conflicts problem and I'm thinking how to resolve them in a better way.

Also, I think there is another way of bad words extracting. They can be extracted analytically based on runtime code, not only empirically based on incorrect generated code. For instance, we have bad word `terminal` in Dart runtime because there is conflict with default listener method [`enterTerminal`](https://github.com/antlr/antlr4/blob/master/tool/resources/org/antlr/v4/tool/templates/codegen/Dart/Dart.stg#L116-L130): 

```
  /// The default implementation does nothing.
  @override
  void enterEveryRule(ParserRuleContext ctx) {}

  /// The default implementation does nothing.
  @override
  void exitEveryRule(ParserRuleContext ctx) {}

  /// The default implementation does nothing.
  @override
  void visitTerminal(TerminalNode node) {}

  /// The default implementation does nothing.
  @override
  void visitErrorNode(ErrorNode node) {}
```

But other problem words such as `everyRule`, `errorNode` can be extracted as well. This approach works with other runtimes as well.

kaby76:
To compute the "bad word" list analytically, one would have to look at the .stg files (since the name that is generated is some string in the .stg file (e.g. [Enter<lname; format="cap">](https://github.com/antlr/antlr4/blob/2be5a44ba10760303fc8600b2816fce8e74fdaf5/tool/resources/org/antlr/v4/tool/templates/codegen/CSharp/CSharp.stg#L83)), along with computed strings by the Tool), the Antlr runtimes (all public methods, fields, etc.), and any other libraries that one commonly uses.  For the Cpp target, we include stddef.h, which defines NULL. Any preprocessor symbol could turn the generated code on its head, especially in the first table in the generated source code for the parser that enumerates the lexer grammar symbols. One can't anticipate all the symbol conflicts at tool invocation time because the generated code will exist in an environment that one cannot predict.

KvanTTT:
Probably we should combine both approaches to extract the most general bad words list.

kaby76:
Sure. An update to the script is here. 
[try-symbol.txt](https://github.com/antlr/antlr4/files/7835332/try-symbol.txt)

But this illustrates why I've been an advocate of parameterizing templates, including the attributes and values themselves. This is exactly why I wrote trgen to allow the CI tester in grammars-v4 to specify the templates rather than just rely on hardwired templates, but doesn't go nearly far enough. Of course, it goes against "strict model-view separation".

KvanTTT:
@kaby76 take attention following words are reserved in ANTLR, it's not necessary to add them to the bad words list:

* catch
* finally
* fragment
* grammar
* import
* options
* throws

Also, for the following words ANTLR reports `RESERVED_RULE_NAME` or similar errors, they also shouldn't be added:

* EOF
* MORE
* SKIP

KvanTTT:
> Here are "bad words" for C#, Java, Dart, JavaScript. Working out C++, Go, Python3, PHP.

Hi! Do you have updated for Python and PHP?

kaby76:
> Hi! Do you have updates for Python and PHP?

Sorry Ivan, I hadn't prioritized working on this. I can work out the PHP keywords the next couple of days. Then, I'll check what was done for Python. (I've been working out a machine learning scraper for grammars from documents in Trash.)

KvanTTT:
Ok, no worry, no rush. I'm just checking the status.

kaby76:
In determining the reserved words in the PHP target, I tested the tool and target against the current list of reserved words in [PHPTarget.java](https://github.com/antlr/antlr4/blob/f5f777d5af3a8e847465787e6112872065440014/tool/src/org/antlr/v4/codegen/target/PHPTarget.java#L18) _with a special-built tool that removes all reserved word checks for PHP. In other words, I emptied out the list, built and ran the tool against the latest "dev" branch code to observe any problems. Note, I left in the double underscore symbols--even though one cannot use them in a grammar.

It turns out almost all the so-called "reserved words" do not cause any problem whatsoever. Attached is a grammar that I used to check the list and the generated PHP code. 
[Test.txt](https://github.com/antlr/antlr4/files/8146766/Test.txt) 
[TestParser.txt](https://github.com/antlr/antlr4/files/8146775/TestParser.txt)
[TestListener.txt](https://github.com/antlr/antlr4/files/8146776/TestListener.txt)
[TestVisitor.txt](https://github.com/antlr/antlr4/files/8146777/TestVisitor.txt)

I am using is PHP 7.4.3 (cli) (built: Nov 25 2021 23:16:22) ( NTS ). I have not tested this against PHP 8. There is no stated minimum requirements in the online documentation (), but Composer for antlr/antlr4-php-runtime states it is for ["PHP 7 and 8.0 runtime for ANTLR 4"](https://packagist.org/packages/antlr/antlr4-php-runtime). Really, properly released software should test the software against a stated minimum and maximum version, and tested in the CI build.

I will continue the experimental check using a version of grammars-v4 modified to remove the trailing underscores in symbol names. (I can do this quickly using Trash trrename.)

parrt:
ok, please let us know if we can simplify the list of naughty words. wow PHP is very permissive I see.

parrt:
Ok, please let me know here https://github.com/antlr/antlr4/pull/3460  if you would like to make a PR to reduced things; maybe coordinate with @KvanTTT

kaby76:
I performed the additional tests for symbol conflicts for the PHP target. Here are the results and discussion.

# Reserved words for PHP

	CLASS
	addContextToParseTree
	addParseListener
	catch
	consume
	createErrorNode
	createTerminalNode
	dumpDFA
	enterOuterAlt
	enterRecursionRule
	enterRule
	exitRule
	finally
	getATNWithBypassAlts
	getBuildParseTree
	getContext
	getCurrentRuleName
	getCurrentToken
	getDFAStrings
	getErrorHandler
	getExpectedTokens
	getExpectedTokensWithinCurrentRule
	getInputStream
	getInvokingContext
	getNumberOfSyntaxErrors
	getParseListeners
	getPrecedence
	getRuleIndex
	getRuleInvocationStack
	getSourceName
	getTokenFactory
	getTokenStream
	inContext
	isExpectedToken
	isMatchedEOF
	isTrace
	match
	matchWildcard
	notifyErrorListeners
	parserRule
	precpred
	pushNewRecursionContext
	removeParseListener
	removeParseListeners
	reset
	setBuildParseTree
	setContext
	setErrorHandler
	setInputStream
	setTokenFactory
	setTokenStream
	setTrace
	triggerEnterRuleEvent
	triggerExitRuleEvent
	unrollRecursionContexts

# Discussion

1) PHP doesn't seem to tell me in one swoop all duplicated definitions. This was a problem in testing, forcing me to edit ONE non-terminal (aka "parser rule") at a time, build, and test. PHP seems to output one error, then quit. So, if there are multiple problems with the grammar, finding and correcting all problems is time consuming.
2) There is only one terminal (aka, "lexer rule") that results in a symbol conflict: CLASS. There may be others.
3) Most, but not all, methods in the [parser class](https://github.com/antlr/antlr-php-runtime/blob/12f51dcb932510481286fd14d59823ed242eda4e/src/Parser.php#L28) result in a symbol conflict. E.g., if the rule "match:'a';" is defined, then it will result in a symbol conflict. What this means is that this list is fragile: what is a "bad word" will depend on the runtime. If someone modifies the parser class with a new method, that method name will need to be tested and possibly entered in the "bad word" list. I haven't researched why "context" is a valid parser rule name when there is a method in the parser class by that name as well. There are other names like this.

Attached is a grammar file with the additional conflicts (i.e., the commented out rules).

[Test.txt](https://github.com/antlr/antlr4/files/8172281/Test.txt)

I would prefer not to make the changes to the code. There are other PRs I have to complete, and I have a lot of work to do on [Trash](https://github.com/kaby76/Domemtech.Trash). Thanks.

PR #3453, opened 2022-01-02
Escape reserved words during grammar generation (@parrt version)

### LINKED ISSUE #1070: Have ANTLR4 prevent conflict with user rule names by behind-the-scenes renaming of its own variables

KvanTTT:
fixes #1070

Always escape `x` to `x_`, `X` to `X_`, `XContext` to `X_Context` (@parrt version, see https://github.com/antlr/antlr4/issues/1070#issuecomment-1003596172). It does not relate to languages with reserved words escaping (C#, Swift).

Deprecate `USE_OF_BAD_WORD`

### DIFF
--- a/runtime-testsuite/resources/org/antlr/v4/test/runtime/descriptors/LexerExec/ReservedWordsEscaping.txt
+++ b/runtime-testsuite/resources/org/antlr/v4/test/runtime/descriptors/LexerExec/ReservedWordsEscaping.txt
@@ -1,0 +1,23 @@
+[notes]
+https://github.com/antlr/antlr4/issues/1070
+
+[type]
+Lexer
+
+[grammar]
+lexer grammar L;
+
+channels { break }
+
+A: 'a' -> mode(for);
+
+mode for;
+B: 'b' -> channel(break);
+
+[input]
+ab
+
+[output]
+[@0,0:0='a',<1>,1:0]
+[@1,1:1='b',<2>,channel=2,1:1]
+[@2,2:1='<EOF>',<-1>,1:2]
--- a/runtime-testsuite/resources/org/antlr/v4/test/runtime/descriptors/ParserExec/ReservedWordsEscaping.txt
+++ b/runtime-testsuite/resources/org/antlr/v4/test/runtime/descriptors/ParserExec/ReservedWordsEscaping.txt
@@ -1,0 +1,36 @@
+[notes]
+https://github.com/antlr/antlr4/issues/1070
+
+[type]
+Parser
+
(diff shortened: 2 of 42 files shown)


KvanTTT:
@kaby76 or @studentmain could you provide updated lists of reserved words (both keywords and conflicting symbols) for all runtimes while I'm completing the fixes?

kaby76:
> @kaby76 or @studentmain could you provide updated lists of reserved words (both keywords and conflicting symbols) for all runtimes while I'm completing the fixes?

As we have an undocumented requirement in grammars-v4 that an Antlr g4 should be target agnostic, the "bad word list" applied to all targets, not a specific target. And there was no differentiation between keyword vs a conflicting symbol. The initial version was culled from the Antlr tool sources, discussed [here](https://github.com/antlr/grammars-v4/issues/2153). But, you already are modifying that code.

To get a combined list, the easiest would be to use trxgrep and look for anything that ends in an underscore. But, that would be a merged list for all targets. I would have to write a grammar against each bad word, and test it against each of the targets, easily doable but I can't come up with that quickly.

KvanTTT:
> There is no differentiation between keyword vs a conflicting symbol --all bad words are off limits regardless

Differentiation is worth only for languages that support keywords escaping (C#, Swift). But if this info is unknown, a word is treated as a conflicting symbol.

> To get an updated, combined list, the easiest would be to use trxgrep and look for anything that ends in an underscore. But, that would be a merged list for all targets. I would have to write a grammar against each bad word, and test it against each of the targets, easily doable but I can't come up with that quickly.

It would be great. No rush, especially since I haven't finished either.

KvanTTT:
I'm closing it in favor of https://github.com/antlr/antlr4/pull/3451 Let's continue there.

kaby76:
Here is the aggregated list of "bad symbols" (with an underscore at the end).

[all_underscore.txt](https://github.com/antlr/antlr4/files/7801985/all_underscore.txt)

It was computed using the following script on a clean git clone of grammars-v4.

[get-all-unames.txt](https://github.com/antlr/antlr4/files/7801989/get-all-unames.txt)

PR #3455, opened 2022-01-02
[Go] Fix (*BitSet).equals

jcking:
Currently code assumes slices have the same length for equivalence. This is incorrect, as its possible for one BitSet to have trailing zeros and another to have the best fit. So we should only be comparing set bits and ignoring trailing zeros.


jcking:
@parrt

PR #3087, opened 2021-02-17
Add Super property

terwey:
Add a super property to the Base*Visitor generated code allowing to only override the required Visitors vs implementing each one in its entirety.

Example code:

```
	type fooVisitor struct {
		*parser.BaseAnalyzerVisitor
	}

	is := antlr.NewInputStream("4+6")
	lexer := parser.NewAnalyzerLexer(is)
	stream := antlr.NewCommonTokenStream(lexer, antlr.TokenDefaultChannel)
	p := parser.NewAnalyzerParser(stream)

	fooVisitor := & fooVisitor{}
	fooVisitor.BaseAnalyzerVisitor = &parser.BaseAnalyzerVisitor{}
	
	// this is new
	fooVisitor.BaseAnalyzerVisitor.SetSuper(fooVisitor)
	fooVisitor.Visit(p.Prog())
```

NB: in draft until signed

PR #3456, opened 2022-01-03
Go visitor

dragonfax:
Here is my take on a solution to the Go Visitor implementation. The existing implementation doesn't work, but its only because the pattern used by most languages isn't natively supported by Go.

### The Problem

The current visitor solution for antlr in all of its supported languages, uses virtual methods. 

Go unfortunately doesn't have virtual methods. Its a language that doesn't truly include Objects or Inheritence. It has simple data "structs", and instead provides a means of "composing" structs. But when you compose a struct out of other structs, it never really becomes those other structs. It just "contains" them. So the children have no inate access to the parent struct.

### The Solution

You can still do abstract classes in Go, the problem comes when the base class has implementation (non abstract methods) that must call these methods (implemented in the subclass). Without virtual methods, this isn't possible. 

A common way to solve this in Go is to have a struct be self referential. When you construct the composable abstract struct (super class), pass it a reference to the outer parent struct (subclass). So the super class can make method calls to the sub classes (would-be virtual) methods.

@terwey touched on this solution at the beginning of the year.
* https://github.com/antlr/antlr4/pull/3087

This lets us keep the same Visitor design that Antlr uses with all the other languages it supports. The same experience cross-language boundaries.

### Concerns

Recursion doesn't become an issue here, as the only time the base class calls into sub class, is for dealing with the next layer down of nodes in the tree (aka the children of the current node), or when its dealing with a leaf node. So if a programmer does make a mistake, the deepest the stack will run is the depth of the parse tree and no farther.

Go's garbage collector has no issues with circular references, as its a Mark&Sweep design.

Here we also remove those interfaces around each context, as they weren't adding anything to the implementation of visitors, and in fact just acted as a barrier to use the contexts themselves, requiring a cast (what called a "type assertion" in Go), every time.

### Examples

There is an example repo to show what a project using this feature in Go looks like. 
* https://github.com/dragonfax/antlr_example_visitor_go

The generated parser sources are checked into the repo for convenience.

I find the calculator example a little played out so this example reads a java source file and outputs a go struct of the fields.

There is a java version that parallels the design of the go version, to help those without a lot of Go experience understand that example.
* https://github.com/dragonfax/antlr_example_visitor_java

### Tests

I'm still working on groking the build and test architecture of this project so that I can fix the related tests. 

### Followup

There is another branch to apply on top of this one that adds Go Generics support. But that feature isn't officially in the language until the February release.

### Correspondence

@jcking @sheik


dragonfax:
I originally tried something similar to @sheik's latest solution. Using Go's duck-typing, thus removing the large Visitor interface completed. But admitedly, it leads to a completely different design to visitors than Antlr uses for all its other languages. Rather, it seems like a good idea to keep the general design patterns homogenous, to aid adoption and support throughout the library.

dragonfax:
Looks like removing those interfaces is way more of a task than I expected. So I've reverted that commit to the PR and the tests pass. Without it, my changes just boil down to the same thing that Terwey was doing. But its still a good solution to the problem.

PR #3459, opened 2022-01-04
Remove unused variable from generated code

jcking:
These variables are unused and can cause static analyzers to complain, specifically Dart when it is opinionated.

### DIFF
--- a/tool/resources/org/antlr/v4/tool/templates/codegen/Dart/Dart.stg
+++ b/tool/resources/org/antlr/v4/tool/templates/codegen/Dart/Dart.stg
@@ -855,7 +855,6 @@ class <lexer.name> extends <superClass; null="Lexer"> {
 SerializedATN(model) ::= <<
 <if(rest(model.segments))>
 <! requires segmented representation !>
-static const int _serializedATNSegments = <length(model.segments)>;
 <model.segments:{segment|static final String _serializedATNSegment<i0> =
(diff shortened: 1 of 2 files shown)


jcking:
@parrt

parrt:
Wow so `_serializedATNSegments` is not used?  That seems odd. @lingyv-li any ideas?

KvanTTT:
Also, it does not affect API since it's private.

parrt:
yes I can't see any use of the value so let's get rid of it.

KvanTTT:
Git blame says that line was changed 9 years ago :) https://github.com/jcking/antlr4/blame/ca2bfedb2c1c92f6370a33b4e385a553cecdff54/tool/resources/org/antlr/v4/tool/templates/codegen/Java/Java.stg#L970

KvanTTT:
And looks like was that member was not used even then: https://github.com/jcking/antlr4/commit/c592e416379b33355d1daf7da82f22c5ea53b1d5

parrt:
haha.

ericvergnaud:
Can we make sure that testLargeLexer is still being run ?
If it isn’t then it’s unsafe to drop segmentation.
If it is, then we need to understand why the Java spec doesn’t yell

parrt:
It is testing large lexer it appears; there are 40 tests from intellij for java TestLexerExec and 40 from:

```
$ mvn -Dtest=java.TestLexerExec test
...
```

Issue #972, opened 2015-08-05
Allow Generated Classes to be Package Private


jdconrad:
I would like to be able to specify visibility of the generated class files (including the RuleContexts that are static inner classes of the generated Parser) as a command line option when using the ANTLR tool.  Is this something that could be added to a future version of ANTLR?

hanjos:
+1 to this.

There have been previous issues about this (#324, #326 and #336, all closed), but nothing official seems to have come of it, as far as I can see. You could try @jchambers' pull requests.

hanjos:
**Update:** My current solution is to use [maven-replacer-plugin](https://code.google.com/p/maven-replacer-plugin/) to remove the public modifier from the generated classes. Worked well enough:

```xml
<plugin>
  <groupId>com.google.code.maven-replacer-plugin</groupId>
  <artifactId>maven-replacer-plugin</artifactId>
  <version>1.4.1</version>
  <executions>
    <execution>
      <phase>process-sources</phase>
      <goals>
        <goal>replace</goal>
      </goals>
    </execution>
  </executions>
  <configuration>
    <includes>
      <include>target/generated-sources/antlr4/**/*.java</include>
    </includes>
    <variableTokenValueMap>
      public class=class,public interface=interface
    </variableTokenValueMap>
  </configuration>
</plugin>
```

The version I'm using (1.4.1) is in Maven Central, and a few people have exported the project to GitHub, so there should be no danger of sudden broken builds or lost source code.

That being said, this is still a hack :smile:

sjamesr:
As far as I can tell, this should be a relatively straightforward change to the Java code generation template, plus a way of passing the option through to the template. Would there be any interest in a patch here?

eshu:
+1 to this issue

PR #3461, opened 2022-01-04
[C++] Mark generated classes as final

jcking:
Mark generated classes final when possible. This allows the compiler to potentially apply optimizations as the class hierarchy will be known within the compilation unit. This change also removes unnecessary `virtual` keyword usage, as `override` implies `virtual`.

### DIFF
--- a/tool/resources/org/antlr/v4/tool/templates/codegen/Cpp/Cpp.stg
+++ b/tool/resources/org/antlr/v4/tool/templates/codegen/Cpp/Cpp.stg
@@ -43,7 +43,7 @@ cppTypeInitMap ::= [
 LexerHeader(lexer, atn, actionFuncs, sempredFuncs, superClass = {antlr4::Lexer}) ::= <<
 <namedActions.context>
 
-class <file.exportMacro> <lexer.name> : public <superClass> {
+class <file.exportMacro> <lexer.name> final : public <superClass> {
 public:
 <if (lexer.tokens)>
   enum {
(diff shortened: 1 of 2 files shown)


jcking:
@mike-lischke

mike-lischke (commented):
No, not a good idea. It surely makes sense to allow deriving your own class from a generated parser (for example), if a base class is not enough, for certain reasons.

I admit that won't happen often, but is nonetheless a valid scenario. And the benefit doesn't outweigh the loss.

jcking:
How about as a grammar option then?

```
options {
final = true;
}
```

Or something similar?

KvanTTT:
If the option is implemented, other runtimes should support this option as well, not only C++. Also, see the similar issue but about methods: https://github.com/antlr/antlr4/issues/972

mike-lischke:
A grammar option requires changes in the ANTLR4 tool, which is not so welcome.

mike-lischke:
@parrt @jcking I think we should close this PR and open a new one if Justin has good arguments yet for his changes (though I'm still not fond of make generated classes always final, but an option is probably acceptable).

This PR includes now > 700 files, which are impossible to review (and don't belong to the patch anyway).

KvanTTT:
> This PR includes now > 700 files, which are impossible to review (and don't belong to the patch anyway).

This is because of the target branch. It should be changed to `dev`.

parrt:
well, first let's take a look after resetting it to use the development branch, which I not able to do directly. But, yes, let's make sure we have a good reason to do everything. and all changes represent a risk as well as a benefit.

parrt:
Hi Guys. I reset this to compare to dev.  is it still relevant? @mike-lischke @jcking

parrt:
Yeah, I think I'll close as it is my intention to allow overriding even rule parse methods. thanks guys...

PR #3474, opened 2022-01-06
GitHub action php

ericvergnaud:


KvanTTT:
Please squash the same commits like `Update run-tests-php.cmd` before merging because they don't contains useful info for other users and pollute git history.

parrt:
I usually squash but I screwed up yesterday and made a whole bunch of updates to the master, because I thought it would only take one or two commits.  then I got lost in bashing my way to victory and realized I could not squash the commits because others might have forked the repository etc.. oops.

parrt:
Anyway, congrats on getting this stuff to run @ericvergnaud !!!   I could not figure out why it wouldn't install Python yesterday. ugh

PR #3478, opened 2022-01-06
[C++] Optimizations and cleanups and const correctness, oh my

jcking:
Various changes related to optimizations, cleanup, and const correctness fixes. The optimizations are related to `std::shared_ptr` and getting rid of unnecessary copying when possible as we as some prerequisite work for future improvements. The cleanup is marking things as final or removing odd whitespace. The const correctness is self explanatory.

Additionally this makes LL1Analyzer final and removes its virtual methods. Its not possible for generated code to override the LL1Analyzer and having virtual in the hot path is not free, unlike Java. In C++ virtual methods effectively guarantee CPU cache misses.

Lastly it moves a few methods in the hot path to be inline, specifically Recognizer::set/getState.

### DIFF
--- a/runtime/Cpp/runtime/src/Recognizer.cpp
+++ b/runtime/Cpp/runtime/src/Recognizer.cpp
@@ -149,14 +149,6 @@ bool Recognizer::precpred(RuleContext * /*localctx*/, int /*precedence*/) {
 void Recognizer::action(RuleContext * /*localctx*/, size_t /*ruleIndex*/, size_t /*actionIndex*/) {
 }
 
-size_t Recognizer::getState() const {
-  return _stateNumber;
-}
-
-void Recognizer::setState(size_t atnState) {
-  _stateNumber = atnState;
-}
-
 void Recognizer::InitializeInstanceFields() {
   _stateNumber = ATNState::INVALID_STATE_NUMBER;
   _interpreter = nullptr;
--- a/runtime/Cpp/runtime/src/Recognizer.h
+++ b/runtime/Cpp/runtime/src/Recognizer.h
@@ -113,7 +113,7 @@ namespace antlr4 {
 
     virtual void action(RuleContext *localctx, size_t ruleIndex, size_t actionIndex);
 
-    virtual size_t getState() const ;
+    size_t getState() const { return _stateNumber; }
 
     // Get the ATN used by the recognizer for prediction.
     virtual const atn::ATN& getATN() const = 0;
@@ -126,7 +126,7 @@ namespace antlr4 {
     ///  invoking rules. Combine this and we have complete ATN
     ///  configuration information.
     /// </summary>
-    void setState(size_t atnState);
+    void setState(size_t atnState) { _stateNumber = atnState; }
 
     virtual IntStream* getInputStream() = 0;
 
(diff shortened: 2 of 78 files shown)


jcking:
@mike-lischke

mike-lischke:
Passing a shared_ptr as const ref is not a good idea. I had a [discussion on Stackoverflow](https://stackoverflow.com/questions/37610494/passing-const-shared-ptrt-versus-just-shared-ptrt-as-parameter) 5 years ago, exactly because I was thinking about doing that. You are killing the shared_ptr promise to ensure a pointer is valid as long as at least one place in code holds the shared_ptr.

If you really want to improve that (what I had in mind since I wrote that code) is to pass through raw pointers and make sure there's "one source of truth" that owns the pointer (using a unique_ptr then).

jcking:
> Passing a shared_ptr as const ref is not a good idea. I had a [discussion on Stackoverflow](https://stackoverflow.com/questions/37610494/passing-const-shared-ptrt-versus-just-shared-ptrt-as-parameter) 5 years ago, exactly because I was thinking about doing that. You are killing the shared_ptr promise to ensure a pointer is valid as long as at least one place in code holds the shared_ptr.
> 
> If you really want to improve that (what I had in mind since I wrote that code) is to pass through raw pointers and make sure there's "one source of truth" that owns the pointer (using a unique_ptr then).

I think I switched most to be pass by value, which avoids incrementing and decrementing the reference count when paired with `std::move`, increasing throughput. The other changes where I made it const reference was to copying when possible with return values, avoids incrementing/decrementing reference counts.

mike-lischke:
The std::move semantic doesn't apply here, however. It's about ensuring pointer validity and that should not be compromised to get a potential performance gain (which isn't even relevant here).

Note: my comment relates only to changes like this one: `ATNConfig(Ref<ATNConfig> const& c, Ref<SemanticContext> semanticContext);`. In this case you are not passing-by-value (which was the case before), but by reference. And this makes it possible to invalidate the `Ref` in deeper call levels, without the upper level knowing.

As I wrote above: passing a raw point around would be the preferred solution, but requires measure to ensure there's a single point of ownership.

mike-lischke:
I just realised, it's the other way around: you removed passing by reference. This is totally fine then!

jcking:
> I just realised, it's the other way around: you removed passing by reference. This is totally fine then!
> 
> However, I wonder why you didn't apply that to all cases. For example: `ATNConfig(Ref<ATNConfig> const& c, Ref<SemanticContext> semanticContext);` you removed the `const&` from the second parameter. Why not from the first too?

The implementation of the constructor does not actually save a reference to ATNConfig, it only accesses members. Realistically it should probably just accept `const ATNConfig &c` but that was a larger change that I was willing to do for this specific bit.

mike-lischke (approved):
@parrt This patch is ready to merge.

parrt:
something is weird here with the unit test. Let me rerun them.

jcking:
Rebased off of master. Let's see what happens.

jcking:
All passed now, was due to being on an older base of master.

@parrt

parrt:
Thanks again for all your help!

Issue #3481, opened 2022-01-08
[Dart] minor adjustments to support stricter dart analyzer configurations.


modulovalue:
* It would be great if the following rules could also be excluded here:

https://github.com/antlr/antlr4/blob/09e917ea8b0939c835c38189031c93840f3c41cc/tool/resources/org/antlr/v4/tool/templates/codegen/Dart/Dart.stg#L186

```
// ignore_for_file: unused_import, unused_local_variable, prefer_single_quotes, file_names, directives_ordering, prefer_final_locals, avoid_catching_errors, unnecessary_parenthesis, prefer_const_declarations
```

file_names: Dart prefers snake_case over PascalCase.
directives_ordering: generated directives at the top of the file are not ordered.
prefer_final_locals: mutable local values should be marked as final.
avoid_catching_errors: the errors within antlr should extend the Exception type.
unnecessary_parenthesis: self explanatory.
prefer_const_declarations: self explanatory.

* The following line should not have dynamic as its type, but use var to allow the compiler to infer the type. The dart analyzer doesn't support this type of 'dynamic' use when some stricter checks are enabled.

https://github.com/antlr/antlr4/blob/09e917ea8b0939c835c38189031c93840f3c41cc/tool/resources/org/antlr/v4/tool/templates/codegen/Dart/Dart.stg#L321

```
var _localctx = ...
```

modulovalue:
Dart added support for disabling all lints at once:

```dart
 // ignore_for_file: type=lint
```

This should now be preferred over enumerating all to-be-ignored lints explicitly.

Spiralis:
In my flutter/dart setup, I was not even able to use the generated dart files, until I renamed them to snake_case. I think antlr3 had an option to make filenames snake-case, IIRC?

PR #3482, opened 2022-01-08
minor adjustments to support stricter dart analyzer configurations

modulovalue:
See https://github.com/antlr/antlr4/issues/3481


modulovalue:
I see now why the dynamic was there :(

The nullsafety migration is not done, using dynamic is a good bandaid, but not a solution.

That would have to happen first. So I'll close this.

PR #3483, opened 2022-01-08
Preserve line separators for input runtime tests data

### LINKED ISSUE #1098: Incorrect tokens line number for files with '\r' newline.

KvanTTT:


KvanTTT:
@parrt please review

runtime-testsuite/test/org/antlr/v4/test/runtime/BaseRuntimeTest.java
```diff
@@ -426,35 +420,39 @@ public static ErrorQueue antlrOnString(String workdir,
...
 
 		List<Pair<String, String>> pairs = new ArrayList<>();
-		String[] lines = dtext.split("\r?\n");
-		for (String line : lines) {
-			if ( line.startsWith("[") && line.length()>2 &&
-				 sections.contains(line.substring(1, line.length() - 1)) ) {
-				if ( currentField!=null ) {
-					pairs.add(new Pair<>(currentField,currentValue));
+
+		Matcher matcher = linePattern.matcher(dtext);
```
parrt, line 428: hiya. not sure i find the complex regex more readable than my simple section detector.

parrt: Reader has to go find and then understand that hairball haha

runtime-testsuite/test/org/antlr/v4/test/runtime/BaseRuntimeTest.java
```diff
@@ -42,18 +44,10 @@
...
-		"Java",
-		"Node",
-		"PHP",
-		"Python2", "Python3",
-		"Swift"
-	};
+	final static Set<String> sections = new HashSet<>(Arrays.asList(
+		"notes", "type", "grammar", "slaveGrammar", "start", "input", "output", "errors", "flags", "skip"
+	));
+	final static Pattern linePattern = Pattern.compile("([^\\r\\n]*)(\r?\n)");
```
parrt, line 50: to simplify how about a-zA-Z instead of the `^` based one?

runtime-testsuite/test/org/antlr/v4/test/runtime/BaseRuntimeTest.java
```diff
@@ -426,35 +420,39 @@ public static ErrorQueue antlrOnString(String workdir,
...
+				newSection = sections.contains(sectionName);
+			}
+
+			if (newSection) {
+				if (currentField!=null) {
+					pairs.add(new Pair<>(currentField, currentValue.toString()));
 				}
-				currentField = line.substring(1, line.length() - 1);
-				currentValue = null;
+				preserveSeparator = sectionName.equals("input");
```
parrt, line 445: i see; you needed to upgrade code to handle test files with various line terminators.

runtime-testsuite/test/org/antlr/v4/test/runtime/BaseRuntimeTest.java
KvanTTT: It's not correct, because we have to read files line by line separated by line separators. File content contains different words and chars.

parrt: oh so you see why i don't like regex. i quickly looked and saw `[...]` as matching the section names. haha.  why not split on `\r?\n` then?

KvanTTT: Because in some tests line separators matter. Specifically, in tests in this PR: `LineSeparator_LF.txt eol=lf, LineSeparator_CRLF.txt`. Also, in general, input changing is not good.

parrt: Yeah,I was wondering about that... What about `.*?\r?\n` for the regex?

BTW, do we assume that the same line Terminator is used throughout the entire test descriptor file? I think we should.

KvanTTT: > Yeah,I was wondering about that... What about .*?\r?\n for the regex?

Yes, it's possible to use `(.*?)(\r?\n)` (with groups) but I don't see a big difference. Also, I try to get rid of greedy quantifiers if possible.

> BTW, do we assume that the same line Terminator is used throughout the entire test descriptor file? I think we should.

Line terminator may differ even within one section (it's relevant for input). For other sections line terminator does not matter (and it's being replaced by `\n` because ANTLR uses `\n` for output).

parrt: Ah, yes, you're right about input/output. Hm..ok, this is close enough. let's go for it.

PR #3487, opened 2022-01-11
[C++] Implement dynamic BitSet

jcking:
Currently `antlrcpp::BitSet` just inherits from `std::bitset<2048>` which more than likely has a size of 256 bytes. This makes popcount unlikely to be very efficient on most platforms as the entire 256 bytes has to be inspected. This pull request switches to a dynamic bitset implementation that stores `sizeof(uintptr_t) * 8` bits inline and uses the heap for anything larger. Thus storing bits 0 through 63 is free.


chakpongchung:
is there any reason we dont want to use dynamic bitset from boost? @jcking

jcking:
Mandatory external dependencies are banned in the ANTLRv4 runtime.

chakpongchung:
> Mandatory external dependencies are banned in the ANTLRv4 runtime.
boost is used in so many c++ projects anyway. I see std::any is used in 4.9.3+ which bumps the requirement to c++17 already. Why not just include boost to simplify the code rather than reinventing the wheel? @jcking @parrt

jcking:
Requiring C++17 is a far cry from requiring boost. At that rate why not
require abseil, folly, or any of those. While boost is used by a lot of
projects, it is absent from many of the largest C++ code bases. boost is a
hammer, we just need a nail. Dynamic bitsets are also fairly easy to
implement.

runtime/Cpp/runtime/src/Recognizer.h
```diff
@@ -17,35 +18,38 @@ namespace antlr4 {
...
-    virtual dfa::Vocabulary const& getVocabulary() const = 0;
+    virtual const Vocabulary& getVocabulary() const = 0;
 
     /// <summary>
     /// Get a map from token names to token types.
     /// <p/>
     /// Used for XPath and tree pattern compilation.
     /// </summary>
-    virtual std::map<std::string_view, size_t> getTokenTypeMap();
+    const std::unordered_map<std::string_view, size_t>& getTokenTypeMap() const;
```
hzeller, line 40: I'd probably have all these map types typedefs in the public section of the class, so that it is easier to change and avoids repetition of the same otherwise somewhat lengthy type.

runtime/Cpp/runtime/src/VocabularyImpl.h
```diff
@@ -0,0 +1,37 @@
...
+                   antlrcpp::Span<const std::string_view> displayNames);
+
+    std::string_view getLiteralName(size_t tokenType) const override;
+
+    std::string_view getSymbolicName(size_t tokenType) const override;
+
+    std::string getDisplayName(size_t tokenType) const override;
+
+  private:
+    antlrcpp::Span<const std::string_view> _literalNames;
```
hzeller, line 32: Since they are only set in the constructor, we can make them `const`
(and with that, we also prevent accidental non-initialization)

runtime/Cpp/runtime/src/VocabularyImpl.h
```diff
@@ -0,0 +1,37 @@
...
+
+    VocabularyImpl(antlrcpp::Span<const std::string_view> literalNames,
+                   antlrcpp::Span<const std::string_view> symbolicNames)
+        : VocabularyImpl(literalNames, symbolicNames, {}) {}
+
+    VocabularyImpl(antlrcpp::Span<const std::string_view> literalNames,
+                   antlrcpp::Span<const std::string_view> symbolicNames,
+                   antlrcpp::Span<const std::string_view> displayNames);
+
+    std::string_view getLiteralName(size_t tokenType) const override;
```
hzeller, line 25: Since our class is final now, using `final` instead of `override` would give the reader more information.

runtime/Cpp/runtime/src/atn/ATNConfigSet.cpp
```diff
@@ -208,7 +208,7 @@ std::string ATNConfigSet::toString() const {
     ss << ",uniqueAlt = " << uniqueAlt;
   }
 
-  if (conflictingAlts.size() > 0) {
+  if (conflictingAlts.count() > 0) {
```
hzeller, line 211: Is this a container that has empty() ? Then this is typically the cheapest implementation for that question.

`if (!conflictingAlts.empty()) { ... }`

runtime/Cpp/runtime/src/support/InlinedVector.h
```diff
@@ -0,0 +1,346 @@
...
+    InlinedVector(InlinedVector<T, N, Allocator> &&other) {
+      move_from(std::forward<InlinedVector<T, N, Allocator>>(other));
+    }
+
+    ~InlinedVector() {
+      destroy();
+    }
+
+    InlinedVector<T, N, Allocator>& operator=(const InlinedVector<T, N, Allocator> &other) {
+      if (this != std::addressof(other)) {
```
hzeller, line 70: To reduce the cyclomatic complexity and improve readability, you could return early after the aliasing check:

```c++
if (this == std::addressof(other)) return *this;
if (capacity()....
```

hzeller (commented):
Nice!
A few comments while reading through the PR.

Issue #1863, opened 2017-05-09
Add a more understandable message than "Serialized ATN data element .... element ... out of range 0..65535"


jbarotin:
Hi,

I tried to build a big grammar file with antlr4. I've  got the following message when I execute the ANTLR  code that generate java source : 
```
Exception in thread "main" java.lang.UnsupportedOperationException: Serialized ATN data element ```
101246 element 11 out of range 0..65535
        at org.antlr.v4.runtime.atn.ATNSerializer.serialize(ATNSerializer.java:361)
        at org.antlr.v4.runtime.atn.ATNSerializer.getSerialized(ATNSerializer.java:601)
        at org.antlr.v4.codegen.model.SerializedATN.<init>(SerializedATN.java:22)
        at org.antlr.v4.codegen.model.Recognizer.<init>(Recognizer.java:64)
        at org.antlr.v4.codegen.model.Lexer.<init>(Lexer.java:27)
        at org.antlr.v4.codegen.OutputModelController.lexer(OutputModelController.java:151)
        at org.antlr.v4.codegen.OutputModelController.buildLexerOutputModel(OutputModelController.java:104)
        at org.antlr.v4.codegen.CodeGenerator.generateLexer(CodeGenerator.java:119)
        at org.antlr.v4.codegen.CodeGenPipeline.process(CodeGenPipeline.java:54)
        at org.antlr.v4.Tool.processNonCombinedGrammar(Tool.java:404)
        at org.antlr.v4.Tool.process(Tool.java:354)
        at org.antlr.v4.Tool.processGrammarsOnCommandLine(Tool.java:321)
        at org.antlr.v4.Tool.main(Tool.java:168)
```

From what I understand, my grammar file is too long.

Is that possible to get a message to explain that the limitation of the library is reached ?

For your test, here it is my grammar file ( a bit cryptic cause I anonymize it) :
https://gist.github.com/jbarotin/7f7caf419b9038c2a548969265788392

ericvergnaud:
Hi,
By the looks of it, you are confusing structure and content. No structure or language grammar has thousands of keywords.  Instead they have patterns to consume all those words or sentences generically. You should look at determining those patterns. Antlr will produce an AST which you can then traverse to interpret the exact words.
Eric

Envoyé de mon iPhone

jbarotin:
Thanks for your answer, you are exactly describing the way I'm implementing my grammar right now. 

[...]

PR #3488, opened 2022-01-11
Normalize line separators for .txt files (fix Windows tests)

KvanTTT:


KvanTTT:
I still don't understand how to force `\n` line separators for Windows instead of `\r\n`. Git doesn't accept the change of separators. All tests expect `\n` line separator.

ericvergnaud:
Have you tried: git config --global core.autocrlf false
See https://docs.github.com/en/get-started/getting-started-with-git/configuring-git-to-handle-line-endings

KvanTTT:
Yes, I've tried with no effect. Ok, I'll restore line separators and replate line separators tests with handwritten tests.

parrt:
Oh sorry. didn't see this yet, @KvanTTT (Just emailed you).

KvanTTT:
Yes, I'm working on it.

KvanTTT:
@parrt the fix is ready.

runtime-testsuite/test/org/antlr/v4/test/runtime/ExtraTests.java
```diff
@@ -0,0 +1,53 @@
+package org.antlr.v4.test.runtime;
+
+import java.util.ArrayList;
+import java.util.List;
+
+public class ExtraTests {
+	public static RuntimeTestDescriptor[] getRuntimeTestDescriptors(String targetName) {
```
KvanTTT, line 7: These tests are useful. We can write here test for generated and very large grammars. For instance, [LargeLexer.txt](https://github.com/antlr/antlr4/blob/master/runtime-testsuite/resources/org/antlr/v4/test/runtime/descriptors/LexerExec/LargeLexer.txt) can be moved here.

parrt: Why not include LargeLexer etc... in main section?

parrt: Can you remind me what the purpose is for a new file ExtraTests.java ?

parrt: Is it only way to distinguish \r\n from \n tests?

KvanTTT: > Why not include LargeLexer etc... in main section?

Because it contains a lot of duplicated lines. It can be replaced by a short and clear test with several lines (also, it's possible to vary the number of tokens and other parameters).

> Is it only way to distinguish \r\n from \n tests?

For now yes. But I'm working on https://github.com/antlr/antlr4/issues/1863 and other related bugs that also require big tests that ideally should be generated.

KvanTTT: LargeLexer can be rewritten by the following way: https://github.com/antlr/antlr4/pull/3493/files#diff-aa2cc9084de70a064d9035752f274ac7e6551b512db05f5665cc64d92bdd0d2dR57-R83

parrt: Hmm.. I think the whole point of the descriptor files is to avoid building tests in Java but it seems like we are now creating tests in Java that manually create descriptor objects.

parrt: Certainly I can see that creating large test grammars in line in the descriptor files is weird. I toyed with the idea of creating a special grammar that looked like <filename.g4> instead of the actual text.  Seems like that would solve our problem right?

KvanTTT: > Hmm.. I think the whole point of the descriptor files is to avoid building tests in Java but it seems like we are now creating tests in Java that manually create descriptor objects.

All rules have exceptions. If it's convenient, a handwritten test can be used.

> I toyed with the idea of creating a special grammar that looked like <filename.g4> instead of the actual text. Seems like that would solve our problem right? This would also work for input/output files, many of which would have different line endings

I'm failed with saving a file with a certain line ending instead of a system line ending, git does not allow it. That's why I restored your code that normalizes line endings (`\n`). Also, I don't see a big difference between the descriptor file with inlined big grammar and the descriptor file with ref to a file with big grammar. And an external tool is required for generating such big grammars (sometimes with big input and output). It looks more complicated.

runtime-testsuite/test/org/antlr/v4/test/runtime/ExtraTests.java
```diff
@@ -0,0 +1,81 @@
...
+		result.output = "[@0,0:0='1',<1>,1:0]\n" +
+				"[@1,1:2='\\r\\n',<2>,1:1]\n" +
+				"[@2,3:3='2',<1>,2:0]\n" +
+				"[@3,4:5='\\r\\n',<2>,2:1]\n" +
+				"[@4,6:6='3',<1>,3:0]\n" +
+				"[@5,7:6='<EOF>',<-1>,3:1]\n";
+		return result;
+	}
+
+	static RuntimeTestDescriptor getLargeLexerDescriptor(String targetName) {
```
KvanTTT, line 55: Ok, I've moved `LargeLexer` in the current PR for the demo.

parrt: Ok. Yeah, I'm still thinking we are making a big departure from the point of descriptor files to solve this \r\n thing. Seems we're off the road to shove in a test case for a tiny fraction of the number of tests

KvanTTT: I have yet another generated test [`getAtnStatesSizeMoreThan65535Descriptor`](https://github.com/antlr/antlr4/pull/3493/files#diff-aa2cc9084de70a064d9035752f274ac7e6551b512db05f5665cc64d92bdd0d2dR84) in another PR. It's too excess to use yet another big file with uniform data.

runtime-testsuite/test/org/antlr/v4/test/runtime/ExtraTests.java
parrt: I'll try to think about this today.  One first thought. We don't need the regex anymore if we use \n. split works great and is much easier to understand.

runtime-testsuite/test/org/antlr/v4/test/runtime/ExtraTests.java
parrt: I see. yes, starting to make sense to me for generated examples.

runtime-testsuite/test/org/antlr/v4/test/runtime/ExtraTests.java
KvanTTT: Ok, I'll remove regex. Definitely, now it does not make sense since we have other tests for line endings checking.

KvanTTT: Done. Please don't squash my commits since they are logically different.

Also, in the future, we can use use `!fixup` commits for easier changes review. Before the final PR merge, all `!fixup` are being merged to original commits for clear history. I mean `!fixup Change something` is being merged to `Change something`.

parrt: hi. the fixup thing might be hard for me to remember and complicate my life ;)

parrt:
Ok, looking good now in my mind except that we've tried to group tests now by area like lexing, parsing.  Instead of adding new TestExtra, shouldn't we add the new tests to TestLexer or whatever?

runtime-testsuite/test/org/antlr/v4/test/runtime/ExtraTests.java
KvanTTT: Ok. I just always use interactive rebase in IDEA and never have troubles :)

KvanTTT:
Ok, maybe it makes sense. I just didn't want to mix descriptor and handwritten tests but probably it does not matter.

parrt:
> Ok, maybe it makes sense. I just didn't want to mix descriptor and handwritten tests but probably it does not matter.

yeah, probably better to mix gen vs files to keep test topics/areas the same. This way we can also avoid adding new TestX.java file

parrt:
Seems github is hosed. This times out in actions: `Download action repository 'stCarolas/setup-maven@v4' (SHA:1d56b37995622db66cce1214d81014b09807fb5a)`

KvanTTT:
> yeah, probably better to mix gen vs files to keep test topics/areas the same. This way we can also avoid adding new TestX.java file

Indeed, much fewer changes.

parrt:
@ericvergnaud does the antlr data center look ok from your side? Any network issues? thanks!

runtime-testsuite/test/org/antlr/v4/test/runtime/BaseRuntimeTest.java
```diff
@@ -367,6 +366,13 @@ public static ErrorQueue antlrOnString(String workdir,
 				System.err.println("Can't read descriptor file "+fname);
 			}
 		}
+
```
parrt, line 369: this makes me kinda nervous but better than altering all target test files I guess.

runtime-testsuite/test/org/antlr/v4/test/runtime/ExtraTests.java
```diff
@@ -0,0 +1,68 @@
+package org.antlr.v4.test.runtime;
+
+public class ExtraTests {
```
parrt, line 3: Sorry to be fussy but "Extra" isn't very specific. Maybe "GeneratedTests" or even "GeneratedLexerDescriptors"?

parrt: I'll merge then think/tweak of name

parrt:
Thanks for the hard work, @KvanTTT !

runtime-testsuite/test/org/antlr/v4/test/runtime/ExtraTests.java
KvanTTT: Yes, but I haven't found a better name. Generated is also not very correct since it contains line separator tests. Maybe handwritten or something like this.

ericvergnaud:
> @ericvergnaud does the antlr data center look ok from your side? Any network issues? thanks!

No apparent issue here, some builds are successful.

PR #3492, opened 2022-01-14
[JavaScript] Implement packed BitSet

jcking:
Currently JavaScript uses an Object for `BitSet`. This change switches to using an array of numbers, similar to Go and Dart. The approach is similar to Dart. Additionally I cleaned up package.json, switched from `jest` to `jasmine`, and got rid of yarn as its not needed.


ericvergnaud:
so basically you're proposing 4 PRs in one without any preliminary discussion ?
Please break this down and let's discuss

Issue #3338, opened 2021-11-03
Serialized ATN data element 810567 element 11 out of range 0..65535


ccrfrog:
I have a `.g4` file in which one of the syntax rules has more than 80,000 alternatives,
when I run `antlr4 -Dlanguage=Go -visitor`
below exception occured

```
Exception in thread "main" java.lang.UnsupportedOperationException: Serialized ATN data element 810567 element 11 out of range 0..65535
        at org.antlr.v4.runtime.atn.ATNSerializer.serialize(ATNSerializer.java:361)
        at org.antlr.v4.runtime.atn.ATNSerializer.getSerialized(ATNSerializer.java:601)
        at org.antlr.v4.Tool.generateInterpreterData(Tool.java:745)
        at org.antlr.v4.Tool.processNonCombinedGrammar(Tool.java:400)
        at org.antlr.v4.Tool.process(Tool.java:361)
        at org.antlr.v4.Tool.processGrammarsOnCommandLine(Tool.java:328)
        at org.antlr.v4.Tool.main(Tool.java:172)
```

KvanTTT:
Out of curiosity: where do you need such a big number of alternatives?

parrt:
Fixed by https://github.com/antlr/antlr4/pull/3591

Issue #2732, opened 2020-01-22
UnsupportedOperationException while generating code for large grammars.


mimuogi:
When generating code for large grammars we are getting this exception log. The code generation targets C#.

Exception in thread "main" java.lang.UnsupportedOperationException: Serialized ATN data element 85257 element 11 out of range 0..65535
  	at org.antlr.v4.runtime.atn.ATNSerializer.serialize(ATNSerializer.java:362)
  	at org.antlr.v4.runtime.atn.ATNSerializer.getSerialized(ATNSerializer.java:601)
  	at org.antlr.v4.Tool.generateInterpreterData(Tool.java:745)
  	at org.antlr.v4.Tool.processNonCombinedGrammar(Tool.java:400)
  	at org.antlr.v4.Tool.process(Tool.java:361)
  	at org.antlr.v4.Tool.processGrammarsOnCommandLine(Tool.java:328)
  	at org.antlr.v4.Tool.main(Tool.java:172)
  
  
  UserGrammar.g4

KvanTTT:
What runtime have you tried? [Official](https://github.com/antlr/antlr4/blob/master/doc/csharp-target.md) (this) or [optimized](https://github.com/tunnelvisionlabs/antlr4cs) by @sharwell?

Also, could you attach your grammar where such an exception occurs?

mimuogi:
> What runtime have you tried? Official (this) or optimized by @sharwell?
> Also, could you attach your grammar where such an exception occurs?

I tried the [official](https://github.com/antlr/antlr4/blob/master/doc/csharp-target.md) runtime. For privacy reasons I can't share the grammar at the issue and I had contact you through your email. Thank you for your help.

JenniferPruitt:
I noticed that there is a similar problem at https://github.com/antlr/antlr4/issues/840. Perhaps we can refer to [this issue](https://github.com/antlr/antlr4/issues/840) to find more context about the bug.

parrt:
Fixed by https://github.com/antlr/antlr4/pull/3591

Issue #840, opened 2015-03-19
Getting this error: Exception in thread "main" java.lang.UnsupportedOperationException: Serialized ATN data element out of range


agiovacchini:
I'm using last version pulled out from git (4.5.1) and built under OSx Yosemite, tryed both with jdk6 and jdk7 but with my file, which is: https://gist.github.com/agiovacchini/52cc6dc9210a65dfc659#file-example-g4 I get this error:

Exception in thread "main" java.lang.UnsupportedOperationException: Serialized ATN data element out of range
    at org.antlr.v4.runtime.atn.ATNSerializer.serialize(ATNSerializer.java:384)
    at org.antlr.v4.runtime.atn.ATNSerializer.getSerialized(ATNSerializer.java:563)
    at org.antlr.v4.codegen.model.SerializedATN.<init>(SerializedATN.java:46)
    at org.antlr.v4.codegen.model.Recognizer.<init>(Recognizer.java:87)
    at org.antlr.v4.codegen.model.Lexer.<init>(Lexer.java:51)
    at org.antlr.v4.codegen.OutputModelController.lexer(OutputModelController.java:176)
    at org.antlr.v4.codegen.OutputModelController.buildLexerOutputModel(OutputModelController.java:129)
    at org.antlr.v4.codegen.CodeGenerator.generateLexer(CodeGenerator.java:142)
    at org.antlr.v4.codegen.CodeGenPipeline.process(CodeGenPipeline.java:73)
    at org.antlr.v4.Tool.processNonCombinedGrammar(Tool.java:428)
    at org.antlr.v4.Tool.process(Tool.java:378)
    at org.antlr.v4.Tool.processGrammarsOnCommandLine(Tool.java:345)
    at org.antlr.v4.Tool.main(Tool.java:192)

The launch script is like this:
# !/bin/zsh

export CLASSPATH=".:./antlr4/dist/antlr4-4.5.1-complete.jar"
alias antlr4='java -jar ./antlr4/dist/antlr4-4.5.1-complete.jar'
alias grun='java org.antlr.v4.runtime.misc.TestRig'
antlr4 example.g4

parrt:
likely classpath issue. remove any prior from class path

agiovacchini:
I've already tryed with this, but same result:

#!/bin/zsh
export CLASSPATH="./antlr4/dist/antlr4-4.5.1-complete.jar"
[...]

PR #3493, opened 2022-01-15
Get rid of ATN serialization restrictions (ATN states size can be > 65535, up to 2^31-1), remove excess serialization and allocations

### LINKED ISSUE #1863: Add a more understandable message than "Serialized ATN data element .... element ... out of range 0..65535"

KvanTTT:
It's the limit of serialization, not of ANTLR tool itself. It can be and should be fixed. I've fixed it by using int serialization with varying bytes count that depends on int value:

```java
public void writeCompactUInt32(int value) {
	if (value < 0b1000_0000_0000_0000) {
		writeUInt16(value);
	} else {
		writeUInt16((value & 0b0111_1111_1111_1111) | (1 << 15));
		writeUInt16(value >>> 15);
	}
}
```

All changes are back-compatible excluding quite large lexers or parsers (with more than 32768 states). But actually, it does not matter as I've said before because of [version check on runtime](https://github.com/antlr/antlr4/blob/master/runtime/Java/src/org/antlr/v4/runtime/RuntimeMetaData.java#L144-L167).

If it's okay for Java, I'll fix other runtimes as well.

Fixes #1863, #2732, #3338

The PR depends on #3488

### DIFF
--- a/runtime/Java/src/org/antlr/v4/runtime/atn/ATNSerializer.java
+++ b/runtime/Java/src/org/antlr/v4/runtime/atn/ATNSerializer.java
@@ -23,16 +23,13 @@ import java.util.Map;
 import java.util.UUID;
 
 public class ATNSerializer {
-	public ATN atn;
-	private List<String> tokenNames;
-
-	private interface CodePointSerializer {
-		void serializeCodePoint(IntegerList data, int cp);
-	}
+	public final ATN atn;
+	private final List<String> tokenNames;
 
 	public ATNSerializer(ATN atn) {
 		assert atn.grammarType != null;
 		this.atn = atn;
+		this.tokenNames = null;
 	}
 
 	public ATNSerializer(ATN atn, List<String> tokenNames) {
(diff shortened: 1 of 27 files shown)


runtime/Java/src/org/antlr/v4/runtime/atn/ATNSerializer.java
```diff
@@ -98,20 +97,15 @@ public IntegerList serialize() {
...
-			data.add(stateType);
+			writer.writeCompactUInt32(stateType);
 
-			if (s.ruleIndex == -1) {
-				data.add(Character.MAX_VALUE);
-			}
-			else {
-				data.add(s.ruleIndex);
-			}
+			writer.writeUInt16(s.ruleIndex == -1 ? Character.MAX_VALUE : s.ruleIndex);
```
KvanTTT, line 102: Unfortunately, it doesn't allow more than 65534 tokens. It looks like some applications require more than 65534 tokens, see https://github.com/antlr/antlr4/issues/840#issuecomment-583880796:

> Sorry to raise this question again! I am trying to build a large dictionary of about 200,000 words (which use simple regular expressions to cover for edge cases. Unfortunately, I am running into the mentioned issue as well. From what I understand, this is due to the reason that the number of internal ATN states is limited to MAX_VALUE = '\uFFFF'. Is there a recommended way of using the ANTLR lexer as a tagger for large dictionaries.

But with the [new encoding format](https://github.com/antlr/antlr4/discussions/3494) it will be possible.

KvanTTT: Ok, it looks like there is a way to support such cases with saving backward compatibility. I'll try to implement it (but anyway it's quite dirty).

KvanTTT:
Ok, I managed to save back-compatibility (up to 2^14-1=16383 that is also quite big) using the following encoding scheme:

We only need positive integers and `-1` (0xFFFF) for null.

All tests for other runtime works except a couple of tests with big lexers. But now ANTLR supports up to 2^31-1 integers and doesn't have any restrictions related to serialization.

Also, I've optimized both the tool and Java runtime, I've removed excess serialization and allocations. If the code is good, I'll fix other runtimes.

KvanTTT:
@parrt since #3488 is merged, this request is ready for intermediate review (Java only)

KvanTTT:
I'm closing in favor of https://github.com/antlr/antlr4/pull/3505

PR #3507, opened 2022-01-24
[JavaScript] Implement packed BitSet

jcking:
Implements a packed BitSet, similar to Go and Dart runtimes. It is more memory efficient as it avoids using an Object as a hash map.


ericvergnaud:
Hi,
not sure memory is a bigger concern than speed, the current biggest concern with JS and Python.
Can you provide performance comparisons ?

jcking:
For the test `zeroThroughOneTwentySeven`, the new implementation is ~4.6 ms while the old implementation is ~5.7 ms. For the test `equality` the new implementation is ~0.1 ms while the old implementation is ~0.7 ms. For the test `bitwiseOr` the perf is similar. I would expect the new implementation to beat the old one in most cases or perform similarly.

ericvergnaud:
Atomic improvements are great if they don't come at the cost of deteriorating the broader use case.
So I'm interested in overall performance i.e. when used in a parsing scenario.

PR #3511, opened 2022-01-27
Fix #3510

### LINKED ISSUE #3510: Channel(HIDDEN) != channel(HIDDEN)!
Welcome to another "Oh geezus!"

"-> Channel(HIDDEN)" appears to work Java (or at least, it seems so). But, for CSharp it does not. The Antlr 4.9.3 tool generates a lexer with a reference to "_channel", which is private for CSharp. The interesting this to note is that "-> channel(HIDDEN)" works for both Java and CSharp, but there is no code generated that references "channel". It seems to be handled in a completely different manner, likely in the generated tables.

The code that handles uppercase Channel should probably be ripped out.

	// Template generated code from trgen 0.14.1
	grammar Arith;
	file_ : expression (SEMI expression)* EOF;
	expression : expression POW expression | expression (TIMES | DIV) expression | expression (PLUS | MINUS) expression | LPAREN expression RPAREN | (PLUS | MINUS)* atom ;
	atom : scientific | variable ;
	scientific : SCIENTIFIC_NUMBER ;
	variable : VARIABLE ;
	VARIABLE : VALID_ID_START VALID_ID_CHAR* ;
	SCIENTIFIC_NUMBER : NUMBER (E SIGN? UNSIGNED_INTEGER)? ;
	LPAREN : '(' ;
	RPAREN : ')' ;
	PLUS : '+' ;
	MINUS : '-' ;
	TIMES : '*' ;
	DIV : '/' ;
	GT : '>' ;
	LT : '<' ;
	EQ : '=' ;
	POINT : '.' ;
	POW : '^' ;
	SEMI : ';' ;
	WS : [ \r\n\t] + -> channel(HIDDEN) ;
	COMMENT : '//' ~[\n\r]* -> Channel(HIDDEN);
	fragment VALID_ID_START : ('a' .. 'z') | ('A' .. 'Z') | '_' ;
	fragment VALID_ID_CHAR : VALID_ID_START | ('0' .. '9') ;
	fragment NUMBER : ('0' .. '9') + ('.' ('0' .. '9') +)? ;
	fragment UNSIGNED_INTEGER : ('0' .. '9')+ ;
	fragment E : 'E' | 'e' ;
	fragment SIGN : ('+' | '-') ;

kaby76:
This is a fix for #3510. The problem is that "-> Channel(HIDDEN)" in the lexer is output to the generated lexer as "_channel = HIDDEN". This cannot work because the Antlr CSharp runtime sets the accessibility of the various fields ("_channel", "_type", "_mode") to private, and provides instead get and set properties. Since it already is using the method PushMode, and I don't want to change the runtime, this change uses property set instead.

### DIFF
--- a/tool/resources/org/antlr/v4/tool/templates/codegen/CSharp/CSharp.stg
+++ b/tool/resources/org/antlr/v4/tool/templates/codegen/CSharp/CSharp.stg
@@ -762,9 +762,9 @@ LexerSkipCommand()  ::= "Skip();"
 LexerMoreCommand()  ::= "More();"
 LexerPopModeCommand() ::= "PopMode();"
 
-LexerTypeCommand(arg, grammar)      ::= "_type = <tokenType.(arg)>;"
-LexerChannelCommand(arg, grammar)   ::= "_channel = <channelName.(arg)>;"
-LexerModeCommand(arg, grammar)      ::= "_mode = <modeName.(arg)>;"
+LexerTypeCommand(arg, grammar)      ::= "Type = <tokenType.(arg)>;"
+LexerChannelCommand(arg, grammar)   ::= "Channel = <channelName.(arg)>;"
+LexerModeCommand(arg, grammar)      ::= "CurrentMode = <modeName.(arg)>;"
 LexerPushModeCommand(arg, grammar)  ::= "PushMode(<modeName.(arg)>);"
 
 ActionText(t) ::= "<t.text>"


ericvergnaud:
Hi @kaby76 , thanks for this.
Would be great to add a test where the grammar uses those instructions.
I'm not sure C# would be the only one to fail

KvanTTT:
Yes, ideally every change should be covered by a test. It makes sense to add a runtime test that uses [all lexer commands](https://github.com/antlr/antlr4/blob/master/doc/lexer-rules.md#lexer-commands).

KvanTTT:
Could you please change the target branch from master to dev? Also, I think it's a quite important fix that should be included in the core ASAP (ANTLR 4.10).

kaby76:
> Could you please change the target branch from `master` to `dev`? Also, I think it's a quite important fix that should be included in the core ASAP (ANTLR 4.10).

How did people "fetch upstream" the updates from antlr4/ with conflicts to a forked repo? I looks like commits where moved around after I created this branch and I have to re-fork the entire repo from scratch.

KvanTTT:
No, I asked just to change the branch `master` -> `dev` in the settings of the merge request. Ok, nevermind, just create a new PR to dev.

KvanTTT:
https://docs.github.com/en/pull-requests/collaborating-with-pull-requests/proposing-changes-to-your-work-with-pull-requests/changing-the-base-branch-of-a-pull-request

kaby76:
No, that's not the problem. I couldn't never update my copy of the repo using "fetch upstream" unless I manually merged all the commits you guys moved around. I have several branches that I created after you folks move commits. I'm not going to do that. I need to start afresh.

PR #3513, opened 2022-01-29
Optimize ATN serialization data

KvanTTT:
Using plain chars for serialized ATN data instead of escaped where it's possible. Fixed the bug with the max size of ATN segment introduced in https://github.com/antlr/antlr4/pull/3438.

I've compared the size of source files for [`largeLexer`](https://github.com/antlr/antlr4/blob/master/runtime-testsuite/test/org/antlr/v4/test/runtime/GeneratedLexerDescriptors.java#L42) test:

| Language   | Before (KB) | After (KB) | Comment                             |
| ---------- | ------ | ----- | ----------------------------------- |
| Java       | 1331   | 902   |                                     |
| C#         | 2713   | 1997  |                                     |
| Dart       | 2582   | 1468  | + Fixed max ATN segment               |
| JavaScript | 2148   | 1702  |                                     |
| Go         | -      | -     | No change because of plain ushort16 |
| PHP        | 2287   | 971   | + Fixed max ATN segment               |
| Python     | 1711   | 1102  |                                     |
| Cpp        | 3158   | 1212  | + Fixed max ATN segment               |
| Swift      | 4900  | 1400  | JSON -> Binary serialization  (6900 -> 3400 for compilied files)  |

Also, I've fixed binary serialization in Swift and replaced JSON serialization.

Please do not squash since all commits are atomic.

### DIFF
--- a/tool/src/org/antlr/v4/codegen/Target.java
+++ b/tool/src/org/antlr/v4/codegen/Target.java
@@ -26,36 +26,52 @@ import org.stringtemplate.v4.STGroupFile;
 import org.stringtemplate.v4.StringRenderer;
 import org.stringtemplate.v4.misc.STMessage;
 
+import java.util.HashMap;
+import java.util.Map;
 import java.util.Set;
 
 /** */
 public abstract class Target {
-	/** For pure strings of Java 16-bit Unicode char, how can we display
-	 *  it in the target language as a literal.  Useful for dumping
+	protected final CodeGenerator gen;
+	private STGroup templates;
+	private boolean isPreviousOctal = false;
+
+	protected static final Map<Character, String> defaultCharValueEscape;
+	static {
+		// https://docs.oracle.com/javase/tutorial/java/data/characters.html
+		HashMap<Character, String> map = new HashMap<>();
+		addEscapedChar(map, '\t', 't');
+		addEscapedChar(map, '\b', 'b');
+		addEscapedChar(map, '\n', 'n');
+		addEscapedChar(map, '\r', 'r');
+		addEscapedChar(map, '\f', 'f');
+		addEscapedChar(map, '\'');
+		addEscapedChar(map, '\"');
+		addEscapedChar(map, '\\');
+		defaultCharValueEscape = map;
+	}
+
+	protected Target(CodeGenerator gen) {
+		this.gen = gen;
+	}
+
+	/** For pure strings of Unicode char, how can we display
+	 *  it in the target language as a literal. Useful for dumping
 	 *  predicates and such that may refer to chars that need to be escaped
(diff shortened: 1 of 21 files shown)


KvanTTT:
I don't exactly why CI reports not found `deserializeFromJson` for Swift, I have green tests on my machine. Could it be related to some caching?

KvanTTT:
@parrt please take a look at this optimization, it does not break anything.

parrt:
looks like there's some very good stuff in here. My brain is stuck on the other thing at the moment so let me finish up there.

parrt:
> Moreover, I can't find any substring `deserializeFromJson` in the entire project.

I see:

Screen Shot 2022-01-29 at 3 16 41 PM

parrt:
Let's resolve #3515 quickly before we resolve this one: https://github.com/antlr/antlr4/issues/3515

KvanTTT:
> I see:

Yes, but it's in the master branch, tests are being failed for the branch of the pull request.

> Let's resolve #3515 quickly before we resolve this one: #3515

It looks like you mixed up something because you pointed out the same issues.

parrt:
Sorry. Let's resolve #3515 quickly before we resolve this PR.

parrt:
Copying comment over here... "Actually, concerning this PR, why not use the binary string in the parser file like others? Now Swift build is much more complicated, isn't it?"

tool/src/org/antlr/v4/codegen/Target.java
```diff
@@ -329,21 +343,70 @@ public String encodeIntAsCharEscape(int v) {
...
+			case "JavaScript":
+			case "Python2":
+			case "Python3":
+				return String.format("\\u%04x", v);
+			case "CSharp":
+				return String.format("\\x%X", v);
+			case "Dart":
+			case "PHP":
+			case "Swift":
+				return String.format("\\u{%X}", v);
```
KvanTTT, line 408: I suggest adding `StandardTarget` eventually (we were discussing the question some time ago) to minimize the risk of misprinting in string literal:

```java
class StandardTarget {
    public final static String Java = "Java";
    ...
}
```

or add constants to each target:

```java
class JavaTarget {
    public final static String Key = "Java";
}
```

Those constants are used over the repository in many places.

parrt: If the Constants are used a lot, I can see creating such a mechanism. For now, let's focus on just the changes directly in the serialization.

parrt:
Wow. I love it!  You removed a whole bunch of code, I made things much easier to understand. Well done!!!  Not only that, the generated code is significantly smaller🕺. Merging... all tests are passing

PR #3514, opened 2022-01-29
Minor cleanup and add function for serialization stats

parrt:


runtime/Java/src/org/antlr/v4/runtime/misc/IntegerList.java
```diff
@@ -309,4 +311,29 @@ private int charArraySize() {
 		}
 		return result;
 	}
+
+	public void writeSerializedATNIntegerHistogram(String filename) {
```
KvanTTT, line 315: I don't think `IntegerList` is the appropriate place for this utility method. I suggest moving it to utilities (ideally out of runtime).

parrt: oh right! gees. thanks for catching. I had that in minder earlier but got confused at the end...haha

PR #3517, opened 2022-01-30
set java versions for tool

ericvergnaud:


parrt:
Hi. I think these are inherited from the master pom, which sets v11, right?

parrt:
I think it inherits 11 from parent...closing.

PR #3518, opened 2022-01-30
free serialized atn memory in java

ericvergnaud:


ericvergnaud:
@parrt according to my knowledge this would help free up memory after ATN initialization
the same should be achievable in C#
in C++, Python and JavaScript, I thing _serializedATN = null should suffice
not sure about PHP, Swift, Dart or Go

runtime/Java/src/org/antlr/v4/runtime/atn/ATNDeserializer.java
```diff
@@ -160,6 +160,7 @@ static protected boolean isFeatureSupported(UUID feature, UUID actualUuid) {
 	@SuppressWarnings("deprecation")
 	public ATN deserialize(char[] data) {
 		data = data.clone();
+		System.gc(); // presumably helps free up deserialized ATN
```
KvanTTT, line 163: It's not recommended to call `gc` manually, it may decrease performance. JVM better detects the best moment when gc should be called.

KvanTTT (commented):
Does it help to get rid of putting strings in the string pool?

KvanTTT:
> in C++, Python and JavaScript, I thing _serializedATN = null should suffice

I don't recommend doing that, it's an antipattern and this field is completely useless in this case. Instead, I suggest removing `_serializedATN` field and initializing ATN via a method.

parrt:
I am not overly concerned with freeing up 40k of memory at the moment, so if it's okay, I'm going to focus on cleaning up the ATN serialization, compressing it, etc... I can imagine that 40k could make a difference on a mobile phone but I'm wondering how many gymnastics I'd like in the generated code.

ericvergnaud:
I don't disagree, but then why are we trying to improve the ATN serialization when it works perfectly ?

parrt:
> I don't disagree, but then why are we trying to improve the ATN serialization when it works perfectly ?

To handle cases where ATN state numbers (values inside serialized string) go beyond 65535.

KvanTTT:
I think Java runtime code can be fixed as well as other runtimes code: just initialize all constant string/charr arrays within a method and get rid of the static field.

ericvergnaud:
That won’t free up the memory used by the constant coming from the class file 

Envoyé de mon iPhone

KvanTTT:
If it helps to get rid of placing constant string into string pool it's great.

parrt:
I imagine all string literals go into constant pool.

KvanTTT:
In this case I don't understand what advantages these changes have. We can just load data via method call instead of static field that is not being collect by gc.

ericvergnaud:
The advantage is that class containing the constant gets unloaded

PR #3519, opened 2022-01-30
Follow maven guidelines for maven-plugin

ericvergnaud:


parrt:
oh yeah!  does this get rid of that error during build? I literally did not know what to put for the scope to hush the warning.

ericvergnaud:
Yes it does

PR #3520, opened 2022-01-30
Fix .interp file parsing test for the Java runtime.

parrt:
Also includes separating the generation of the .interp file from writing it out so that we can use both independently.

Still working on the Python part.

### DIFF
--- a/runtime-testsuite/test/org/antlr/v4/test/runtime/BaseRuntimeTestSupport.java
+++ b/runtime-testsuite/test/org/antlr/v4/test/runtime/BaseRuntimeTestSupport.java
@@ -225,6 +225,7 @@ public abstract class BaseRuntimeTestSupport implements RuntimeTestSupport {
 
 		return atn;
 	}
+
 	protected void semanticProcess(Grammar g) {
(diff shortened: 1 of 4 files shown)


parrt:
All right I'm going to merge this because a not sure what to do about the Python runtime.  I have a number of  unit tests for the Java API token stream rewriter... maybe this needs to be ported to Python instead of using these submitted by @youkaichao ?

ericvergnaud:
Each runtime may already have its own test suite, and some already have i.e. Python, JavaScript...
As per API testing, I think it's a good idea to make those tests visible

parrt:
Ah. Yes, Swift has them. JavaScript too. Go does but not in separate dir. has `testing_assert_test.go` etc... Dart has one. Csharp tests 2 bug fixes.

parrt:
Given that, I'll merge so we can move forward.

KvanTTT:
Does it make sense to delete `*.interp` files after your changes?

parrt:
OH! Yes, it does make sense. shoot. i'll do and just push to upstream master

parrt:
Done.

PR #3522, opened 2022-01-31
[C++] Optimize LL1Analyzer

jcking:


jcking:
This optimizes LL1Analyzer by removing extra heap allocations, removing usage of shared_ptr where possible, and reducing stack usage by reducing the number of arguments passed on the stack recursively.

@mike-lischke

jcking:
I also got rid of the `finally` usage because that's really only useful for exceptions and there is no code catching exceptions from LL1Analyzer, so a throw will kill th state anyway. I also used LL1AnalyzerImpl to avoid changing the LL1Analyzer contract, but I'm happy to modify LL1Analyzer directly if you prefer.

runtime/Cpp/runtime/src/atn/LL1Analyzer.cpp
```diff
@@ -31,16 +166,11 @@ std::vector<misc::IntervalSet> LL1Analyzer::getDecisionLookahead(ATNState *s) co
...
-    antlrcpp::BitSet callRuleStack;
-    LOOK(s->transitions[alt]->target, nullptr, PredictionContext::EMPTY,
-          look[alt], lookBusy, callRuleStack, seeThruPreds, false);
-
+    LL1AnalyzerImpl impl(_atn, look[alt], false, false);
+    impl.LOOK(s->transitions[alt]->target, nullptr, PredictionContext::EMPTY);
     // Wipe out lookahead for this alternative if we found nothing
     // or we had a predicate when we !seeThruPreds
-    if (look[alt].size() == 0 || look[alt].contains(HIT_PRED)) {
+    if (look[alt].size() == 0 || look[alt].contains(Token::INVALID_TYPE)) {
```
mike-lischke, line 173: This probably doesn't change the value which is used to indicate the tested situation, but it changes semantic. Without looking up what HIT_PRED is, I can guess it means a predicate was hit. An invalid token type is something completely different. So why changing that at all?

mike-lischke (changes requested):
That's tricky to review, as a large block was moved around and I cannot compare both. Also I don't see a real benefit of moving a part of the implementation into an inner class, but it cleans things up a bit, so I'm willing to approve, but pls check my comment yet.

runtime/Cpp/runtime/src/atn/LL1Analyzer.cpp
jcking: I had switched it to make compilation work as HIT_PRED was defined in LL1Analyzer. I should have just qualified with as `LL1Analyzer::HIT_PRED`.

Switched it back, was being lazy.

mike-lischke (approved):
@parrt This patch can be merged.

PR #3523, opened 2022-02-01
[C++] Fix missing virtual destructors

jcking:
Fix broken compilation due to missing virtual destructors on `LexerActionExecutor` and `PredictionContext`. I removed the former previously thinking `std::enable_shared_from_this` was virtual itself, but apparently its not. `PredictionContext` was never virtual and it should have been.


jcking:
@parrt Quick fix.

PR #3528, opened 2022-02-05
Remove not used tokens from grammars, deprecate ANTLR 3 errors

KvanTTT:
`private`, `protected`, `public` tokens are not used anywhere at all as well as some imaginary tokens from the parser grammar.

Preparation before the rest fixes related to #1070 (to allow `private`, `protected`, `public` identifiers).

ANTLR 3 is not actual for a long time, it does not make sense to preserve compatibility with ANTLR 3 grammar in ANTLR 4 grammar. Also, ANTLR 3 errors are not covered by tests.

Please do not squash.


tool/src/org/antlr/v4/parse/ANTLRParser.g
```diff
@@ -60,36 +60,20 @@ options {
...
     CLOSURE;
     POSITIVE_CLOSURE;
     RANGE;
     SET;
-    CHAR_RANGE;
     EPSILON;
     ALT;
-    ALTLIST;
     ID;
-    ARG;
```
parrt, line 78: Wow. We were incredibly lazy to leave this grammar the way it was ha ha! thanks!

parrt:
Amazing cleanup work. Thank you!

PR #3533, opened 2022-02-08
Update go-target.md

luisespino:
Because the dev branch is now the default, `@master` must be added to the end of the go get, so that the deserialize ATN does not give a version error, while version 4.10 comes out.


parrt:
I've actually switched Master back to being the default so I will close! The old `go get` should work to get the latest stable.

parrt:
(Can you give it a try to get the latest without the `@master`?)

luisespino:
Hi, it works fine with the original go get. Thanks!

parrt:
hooray!

PR #3537, opened 2022-02-11
Fix bail error strategy nil interface panic

mattwiller:
I had experienced a nil interface conversion panic in the bail error strategy, which appeared simple to fix.  Please let me know if a test is needed, but I thought this trivial change might be okay without.


parrt:
Because we changed from master too dev, things are a little bit nutty so I'm going to manually pull this in by cherrypicking your commits into dev.

parrt:
Actually unfortunately of already changed how we do the contributor license agreement.  I'm going to make a parallel change as I can't bring in a new contributor until after I cut the 4.10 release.

PR #3538, opened 2022-02-12
keep getSerializedATNSegmentLimit at max int

parrt:
I generated code with `getSerializedATNSegmentLimit()` set to 80 (current dev value) and max int. Here is diff with C++ target:

Screen Shot 2022-02-12 at 1 02 24 PM

On the left we see small segments but it's unnecessary and wasteful compared to the big static array on right. This PR makes it look like right side.  Only Java really needs this.  Targets are free to ignore `getSerializedATNSegmentLimit()` as JavaScript does.

@KvanTTT @marcospassos @jcking @ericvergnaud @mike-lischke Any concerns?


KvanTTT:
It's ok, but I've already fixed it in https://github.com/antlr/antlr4/pull/3513

parrt:
Ok, well, maybe take the new comments i have here?

JavaScript does:

```
SerializedATN(model) ::= <<
<! only one segment, can be inlined !>

const serializedATN = ["<model.serialized; wrap={",<\n>    "}>"].join("");

>>
```

And C++ can be simplified to the equivalent if someone wants to put the effort in.

mike-lischke:
Looks fine to me.

KvanTTT:
> Ok, well, maybe take the new comments i have here?

Let's merge your branch and I'll rebase on master and drop my commit.

> And C++ can be simplified to the equivalent if someone wants to put the effort in.

Yes, I've already done it as well. Also for Dart and PHP.

PR #3544, opened 2022-02-16
[C++] Improve multithreaded performance, fix TSAN error, and fix profiling ATN simulator setup

jcking:
Various small improvements.

- Improve multithreaded performance by using `insert` instead of `find` then `insert`. This reduces the amount of time spent in critical sections. 
- Fix extremely rare race condition that I introduced as I unlocked then dereferenced an iterator.
- Fix parser profiling setup. Currently calling `setProfile` more than once would crash in debug builds.


jcking:
@mike-lischke

mike-lischke (approved):
@parrt This C++ patch can be merged.

parrt:
awesome patch. thanks!

PR #3545, opened 2022-02-17
[C++] Cleanup ATNDeserializer and remove related deprecated methods from ATNSimulator

jcking:
Simple cleanup of `ATNDeserializer` that removes virtual methods, since they cannot really be overridden anyway, moves private methods to the source file, and removes related deprecated methods from ATNSimulator since nobody should be using those.

### DIFF
--- a/runtime/Cpp/runtime/src/atn/ATNDeserializer.cpp
+++ b/runtime/Cpp/runtime/src/atn/ATNDeserializer.cpp
@@ -61,59 +61,125 @@ using namespace antlrcpp;
 
 namespace {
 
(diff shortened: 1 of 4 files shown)


jcking:
@mike-lischke

mike-lischke (approved):
@jcking I wish you would not shuffle around so much code. That removes the pieces from their surrounding context and is difficult to review.  If I didn't miss anything then I think this patch is ok, however.

@parrt This patch is ready for merge.

parrt:
two of the tests fail but I think it has something to do with the continuous integration server.

PR #3546, opened 2022-02-20
Increase ATN states size limit, simplify ATN serialization

### LINKED ISSUE #1863: Add a more understandable message than "Serialized ATN data element .... element ... out of range 0..65535"

### LINKED ISSUE #2732: UnsupportedOperationException while generating code for large grammars.

### LINKED ISSUE #3338: Serialized ATN data element 810567 element 11 out of range 0..65535

KvanTTT:
I'm returning to the increasing of ATN states size and I've also simplified serialization (related to Unicode encoding).

ATN states size can be > 65535, up to 2^31-1:

* fixes #1863
* fixes #2732
* fixes #3338

Take a look at C# code: it contains small changes because it already has `Read` and `Write` methods (as well as other runtimes except for Java).

If Java and C# are ok, I'll complete other runtimes.

Writing, Reading methods have comprehensive tests: `testATNDataWriterReaderCompact`, `testATNDataWriterReaderRaw`.

### DIFF
--- a/runtime-testsuite/test/org/antlr/v4/test/runtime/BaseRuntimeTest.java
+++ b/runtime-testsuite/test/org/antlr/v4/test/runtime/BaseRuntimeTest.java
@@ -366,9 +366,10 @@ public abstract class BaseRuntimeTest {
 		}
 
 		if (group.equals("LexerExec")) {
-			descriptors.add(GeneratedLexerDescriptors.getLineSeparatorLfTest(targetName));
-			descriptors.add(GeneratedLexerDescriptors.getLineSeparatorCrLfTest(targetName));
+			descriptors.add(GeneratedLexerDescriptors.getLineSeparatorLfDescriptor(targetName));
+			descriptors.add(GeneratedLexerDescriptors.getLineSeparatorCrLfDescriptor(targetName));
 			descriptors.add(GeneratedLexerDescriptors.getLargeLexerDescriptor(targetName));
+			descriptors.add(GeneratedLexerDescriptors.getAtnStatesSizeMoreThan65535Descriptor(targetName));
 		}
 
 		return descriptors.toArray(new RuntimeTestDescriptor[0]);
--- a/runtime-testsuite/test/org/antlr/v4/test/runtime/GeneratedLexerDescriptors.java
+++ b/runtime-testsuite/test/org/antlr/v4/test/runtime/GeneratedLexerDescriptors.java
@@ -1,7 +1,9 @@
 package org.antlr.v4.test.runtime;
 
+import java.util.Collections;
+
 public class GeneratedLexerDescriptors {
-	static RuntimeTestDescriptor getLineSeparatorLfTest(String targetName) {
+	static RuntimeTestDescriptor getLineSeparatorLfDescriptor(String targetName) {
 		UniversalRuntimeTestDescriptor result = new UniversalRuntimeTestDescriptor();
 		result.name = "LineSeparatorLf";
 		result.targetName = targetName;
@@ -20,7 +22,7 @@ public class GeneratedLexerDescriptors {
 		return result;
 	}
 
-	static RuntimeTestDescriptor getLineSeparatorCrLfTest(String targetName) {
+	static RuntimeTestDescriptor getLineSeparatorCrLfDescriptor(String targetName) {
 		UniversalRuntimeTestDescriptor result = new UniversalRuntimeTestDescriptor();
 		result.name = "LineSeparatorCrLf";
 		result.targetName = targetName;
@@ -63,6 +65,52 @@ public class GeneratedLexerDescriptors {
 		result.input = "KW400";
 		result.output = "[@0,0:4='KW400',<402>,1:0]\n" +
 				"[@1,5:4='<EOF>',<-1>,1:5]\n";
+		return result;
+	}
+
+	static RuntimeTestDescriptor getAtnStatesSizeMoreThan65535Descriptor(String targetName) {
+		UniversalRuntimeTestDescriptor result = new UniversalRuntimeTestDescriptor();
+		result.name = "AtnStatesSizeMoreThan65535";
+		result.notes = "Regression for https://github.com/antlr/antlr4/issues/1863";
+		result.targetName = targetName;
+		result.testType = "Lexer";
+
+		final int tokensCount = 1024;
+		final String suffix = String.join("", Collections.nCopies(70, "_"));
+
+		String grammarName = "L";
+		StringBuilder grammar = new StringBuilder();
+		grammar.append("lexer grammar ").append(grammarName).append(";\n");
+		grammar.append('\n');
+		StringBuilder input = new StringBuilder();
+		StringBuilder output = new StringBuilder();
+		int startOffset;
+		int stopOffset = -2;
+		for (int i = 0; i < tokensCount; i++) {
+			String value = "T_" + i + suffix;
+			grammar.append(value).append(": '").append(value).append("';\n");
+			input.append(value).append('\n');
+
+			startOffset = stopOffset + 2;
+			stopOffset += value.length() + 1;
+
+			output.append("[@").append(i).append(',').append(startOffset).append(':').append(stopOffset)
+					.append("='").append(value).append("',<").append(i + 1).append(">,").append(i + 1)
+					.append(":0]\n");
+		}
+
+		grammar.append("\n");
+		grammar.append("WS: [ \\t\\r\\n]+ -> skip;\n");
+
+		startOffset = stopOffset + 2;
+		stopOffset = startOffset - 1;
+		output.append("[@").append(tokensCount).append(',').append(startOffset).append(':').append(stopOffset)
+				.append("='<EOF>',<-1>,").append(tokensCount + 1).append(":0]\n");
+
+		result.grammar = grammar.toString();
+		result.grammarName = grammarName;
+		result.input = input.toString();
+		result.output = output.toString();
 		return result;
 	}
 }
--- a/runtime/CSharp/src/Atn/ATNDeserializer.cs
+++ b/runtime/CSharp/src/Atn/ATNDeserializer.cs
@@ -45,8 +45,7 @@ namespace Antlr4.Runtime.Atn
 			ReadRules (atn);
 			ReadModes (atn);
 			IList<IntervalSet> sets = new List<IntervalSet>();
-			ReadSets (atn, sets, ReadInt);
-	        ReadSets (atn, sets, ReadInt32);
+			ReadSets (sets);
 	        ReadEdges (atn, sets);
 			ReadDecisions (atn);
 			ReadLexerActions (atn);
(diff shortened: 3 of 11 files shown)


parrt:
Can you describe what needs to change to handle 32-bit ints using 16-bit unicode?  I assume this is easy to fix for non-Java. Just use int not short.  What is effect on Java ATN code size?

KvanTTT:
> Can you describe what needs to change to handle 32-bit ints using 16-bit unicode?

We used a different encoding for 16-bit and 32-bit values: https://github.com/antlr/antlr4/blob/db8a48394a6e30a4bc0554f1eaee1ffbee311938/runtime/Java/src/org/antlr/v4/runtime/atn/ATNSerializer.java#L166-L201 Now encoding is universal and even more compact.

> Basically using high bits to encode right?

Yes, two high bits are used for service purposes.

>  Also wouldn't this hurt the common case to use upper bit(s)?

All big Unicode values are encoded using not more than 2 integers in tests as it was before, I've checked: https://github.com/antlr/antlr4/pull/3546/files#diff-45a1efdccdd2d1ab783a98c0d1dc54c09914461d0c4bb0237d9dbe17bd4168dcR39

3 bytes are used for very big values (> 2^30) and for negative numbers except -1 (negative numbers aren't used for serialization at all):

> Hmm...again I'm worried about a very rare cases causes a big code change. Not that this approach is "wrong" but always gotta worry about breaking stuff.

All values within the range `Integer.MIN_VALUE..Integer.MAX_VALUE` are covered by tests https://github.com/antlr/antlr4/pull/3546/files#diff-3ac04ac7612676174d4634b5bea3e2ef612af5a06f6887090e36450067be8d6fR163-R207:

```java
@Test public void testATNDataWriterReaderCompact() {
	IntegerList integerList = new IntegerList();
	ATNDataWriter writer = new ATNDataWriter(integerList, "Java");
	assertEquals(1, writer.write(0));
	assertEquals(1, writer.write(-1));
	assertEquals(1, writer.write(42));
	assertEquals(2, writer.write(1 << 14));
	assertEquals(2, writer.write(0xFFFF));
	assertEquals(3, writer.write(Integer.MAX_VALUE));
	assertEquals(3, writer.write(Integer.MIN_VALUE));
	assertEquals(13, integerList.size());
	char[] charArray = Utils.toCharArray(integerList);
	ATNDataReader reader = new ATNDataReader(charArray);
	assertEquals(0, reader.read());
	assertEquals(-1, reader.read());
	assertEquals(42, reader.read());
	assertEquals(1 << 14, reader.read());
	assertEquals(0xFFFF, reader.read());
	assertEquals(Integer.MAX_VALUE, reader.read());
	assertEquals(Integer.MIN_VALUE, reader.read());
}
@Test public void testATNDataWriterReaderRaw() {
	IntegerList integerList = new IntegerList();
	ATNDataWriter writer = new ATNDataWriter(integerList, "Java");
	writer.writeInt32(0);
	writer.writeInt32(-1);
	writer.writeInt32(42);
	writer.writeInt32(1 << 14);
	writer.writeInt32(0xFFFF);
	writer.writeInt32(Integer.MAX_VALUE);
	writer.writeInt32(Integer.MIN_VALUE);
	assertEquals(7 * 2, integerList.size());
	char[] charArray = Utils.toCharArray(integerList);
	ATNDataReader reader = new ATNDataReader(charArray);
	assertEquals(0, reader.readInt32());
	assertEquals(-1, reader.readInt32());
	assertEquals(42, reader.readInt32());
	assertEquals(1 << 14, reader.readInt32());
	assertEquals(0xFFFF, reader.readInt32());
	assertEquals(Integer.MAX_VALUE, reader.readInt32());
	assertEquals(Integer.MIN_VALUE, reader.readInt32());
}
```

parrt:
Hmm... I think @ericvergnaud had a concern about double encoding; first we encode to get larger integers and then we have to use modified UTF-8. in principle that's okay but it makes me nervous.  I think Eric's concern was about all of the decoding on a phone.

Also, 14 bits only gives us 16384 max, which I think is hitting the size of the lexer for SQL grammars. I don't remember where we put the numbers I had but isn't it going to hurt grammar size for such large but 16-bit conforming grammars?

 I see that we allow 32 bit values and sets but I don't see how our current code allows 32-bit single edge transitions in the ATN. do you have any idea if that is the case?

KvanTTT:
> Hmm... I think @ericvergnaud had a concern about double encoding; first we encode to get larger integers and then we have to use modified UTF-8. in principle that's okay but it makes me nervous. I think Eric's concern was about all of the decoding on a phone.

In my changes, there is no need to encode small numbers at first and big numbers at second, natural order is preserved. It should not break decoding on a phone.

Also, it looks like such Unicode encoding was introduced by @bhamiltoncx in https://github.com/antlr/antlr4/commit/fd4246cf3f90f0e4ee961b8254a8c2220baf0c16

> Also, 14 bits only gives us 16384 max, which I think is hitting the size of the lexer for SQL grammars. I don't remember where we put the numbers I had but isn't it going to hurt grammar size for such large but 16-bit conforming grammars?

It does not hit the size of the lexer for SQL grammars because the max value is `13163` there: https://github.com/antlr/antlr4/pull/3505#issuecomment-1030716062 Anyway, SQL lexer is very big, other grammars are smaller.

> I see that we allow 32 bit values and sets but I don't see how our current code allows 32-bit single edge transitions in the ATN. do you have any idea if that is the case?

I've added the generated test [getAtnStatesSizeMoreThan65535Descriptor](https://github.com/antlr/antlr4/pull/3546/files#diff-859a1117c3f13396162959e824858c01d179a61f3280ab64fd68c9c9490b2729R71) that fails on the previous version. Actually, max is 31-bit transitions because of int is signed type.

parrt:
thanks for the link back to the empirical tests I did for the SQL grammar.

> Anyway, SQL lexer is very big, other grammars are almost always smaller

If this is true, seems like we are doing some major surgery to support a tiny subset of the population. on the other hand, it sounds like there are a number of people that are finding this useful; you passed some links to issues.

we could minimize the size of this change by limiting it to only those targets that need to use strings to encode integer lists, right?  certainly that his java. can C# do static integer arrays properly? if so, then we don't need this for C# either.  we should definitely not do this for any targets such as C++ that can directly encode static arrays.  does JavaScript use strings?  Dang, yep, looks like JavaScript uses strings as well:

```
const serializedATN = ["<model.serialized; wrap={",<\n>    "}>"].join("");
```

 it looks like Java, C#, JS, Dart, Php, Python2/3, Swift.  So, everybody except Go and C++? Dang. we should definitely get away from strings for any language that allows it.  certainly this should be avoided for Python as it can simply do `atn = [1,2,3,4...]` right?

parrt:
Ok, I've spent a lot of time thinking about this today and went back over the arguments from before.  basically I'd like to stick with my original suggestion:

> In other words, we begin the process by serializing an ATN into a list of integers. Then, we figure out the maximum value and see if everything fits in 16 bits. If so, we leave everything as is, otherwise we convert ALL int values to two \uXXXX chars rather than a single char.

It seems to me that we should simply get an IntegerList out of `serialize()` with the version number first and then the word size (16 or 32) with all words 16 or 32, no bit twiddling.  The version number plus the word size indicates the encoding for words.  (We can add a method to IntegerList to compute the maximum value.)  In the future, it's possible I'd consider the optimization you have here, but I'd rather not do bit twiddling across all targets in one big move. The targets can do whatever they want with this information, generating strings or simply creating static short or int lists like C++ and Go. any target that is allowed to generate static integers should do so.

Sounds like we still need to convert the Java ATNDeserializer to use a "reader", as you have suggested, so that it can read either 16 or 32 depending on what kind of data we have stored.  Others seem to use something similar but for example Python directly refers to readInt vs readInt32.  We'll need a generic "read" that pulls 16 or 32 bits depending on the value following the version number.

KvanTTT:
How to represent `-1` value in 16-bit and 32-bit modes, `0xFFFF` and `0xFFFFFFFF`? I am afraid the current serialization works incorrectly if positive value `0xFFFF` conflicts with the special value reserved for `-1`.

parrt:
Yup, sounds like we need to limit to 0xFFFF - 1 for valid 16 bit values.

You're right shifting by 16 is twiddling :) but simple enough I think.

KvanTTT:
> Yup, sounds like we need to limit to 0xFFFF - 1 for valid 16 bit values.

It's quite a strange solution, especially considering ANTLR is also a binary parser and `0xFFFF` is a completely correct token type. I've created an issue for this: https://github.com/antlr/antlr4/issues/3555

> It seems to me that we should simply get an IntegerList out of serialize() with the version number first and then the word size (16 or 32) with all words 16 or 32, no bit twiddling. The version number plus the word size indicates the encoding for words. (We can add a method to IntegerList to compute the maximum value.)

This solution has the following disadvantages and it's not so simple:

* First two integers of ATN always should take 1 element instead of 2 (to preserve backward compatibility for version field). It complicates conversion to char arrays.
* It's unclear how to store `-1` value in 16-bit and 32-bit modes, `0xFFFF` and `0xFFFFFFFF`? It is ugly and actually incorrect because of the abovementioned issue with the `FFFF` token type. Also, Java runtime requires -2 shifting.
* It generates bigger ATNs. Plain Int arrays take twice more memory.
* It complicates the serialization format because the second field with int size (16 or 32) should be filled after completion of the whole serialization step. Also, functionality with detection of max value should be integrated into `IntegerList`.

Ok, let's postpone fixing mentioned bugs because I don't think implementing such encoding is a very good idea. Actually, I think using `IntegerList` for serialization is not very good at all (specifically, restricting integers within 0..0xFFFF range), `ByteList` is preferable and it's a standard way in most binary serializers. I've described binary encoding using byte array in https://github.com/antlr/antlr4/discussions/3494

parrt:
Ok, sounds good. Valid points you raise but we can discuss in future.  Thanks for all your efforts.

PR #3547, opened 2022-02-21
Fix for #3510

kaby76:
This is a fix for #3510. The problem is that "-> Channel(HIDDEN)" in the lexer is output to the generated lexer as "_channel = HIDDEN". This cannot work because the Antlr CSharp runtime sets the accessibility of the various fields ("_channel", "_type", "_mode") to private, and provides instead get and set properties. Since it already is using the method PushMode, and I don't want to change the runtime, this change uses property set instead.

### DIFF
--- a/tool/resources/org/antlr/v4/tool/templates/codegen/CSharp/CSharp.stg
+++ b/tool/resources/org/antlr/v4/tool/templates/codegen/CSharp/CSharp.stg
@@ -762,9 +762,9 @@ LexerSkipCommand()  ::= "Skip();"
 LexerMoreCommand()  ::= "More();"
 LexerPopModeCommand() ::= "PopMode();"
 
-LexerTypeCommand(arg, grammar)      ::= "_type = <tokenType.(arg)>;"
-LexerChannelCommand(arg, grammar)   ::= "_channel = <channelName.(arg)>;"
-LexerModeCommand(arg, grammar)      ::= "_mode = <modeName.(arg)>;"
+LexerTypeCommand(arg, grammar)      ::= "Type = <tokenType.(arg)>;"
+LexerChannelCommand(arg, grammar)   ::= "Channel = <channelName.(arg)>;"
+LexerModeCommand(arg, grammar)      ::= "CurrentMode = <modeName.(arg)>;"
 LexerPushModeCommand(arg, grammar)  ::= "PushMode(<modeName.(arg)>);"
 
 ActionText(t) ::= "<t.text>"


runtime/CSharp/tests/issue-3510/ChannelTest.g4
```diff
@@ -0,0 +1,7 @@
+grammar ChannelTest;
+
+prog : INT+ EOF;
+INT : [0-9]+;
+AAA : [a-z]+ -> channel(HIDDEN); // Note--lowercase "channel"
+BBB : [A-Z]+ -> Channel(HIDDEN); // Note--uppercase "Channel"
+WS : [ \t\r\n] -> skip;
```
KvanTTT, line 7: It should be written using a descriptor (https://github.com/antlr/antlr4/tree/dev/runtime-testsuite/resources/org/antlr/v4/test/runtime/descriptors/LexerExec) to cover all runtimes but not C# only.

Also, please add tests on other commands if you can (see [lexer-commands](https://github.com/antlr/antlr4/blob/dev/doc/lexer-rules.md#lexer-commands) description):

* skip
* more
* popMode
* pushMode
* type
* channel

runtime/CSharp/tests/issue-3510/ChannelTest.g4
```diff
@@ -0,0 +1,7 @@
+grammar ChannelTest;
+
+prog : INT+ EOF;
+INT : [0-9]+;
+AAA : [a-z]+ -> channel(HIDDEN); // Note--lowercase "channel"
+BBB : [A-Z]+ -> Channel(HIDDEN); // Note--uppercase "Channel"
```
KvanTTT, line 6: Not sure it makes sense to use both options for `channel` here. Case insensitivity should be check in another test.

KvanTTT (changes requested):

runtime/CSharp/tests/issue-3510/ChannelTest.g4
kaby76: OK, was wondering what those were for.

runtime/CSharp/tests/issue-3510/ErrorListener.cs
```diff
@@ -0,0 +1,18 @@
+using Antlr4.Runtime;
```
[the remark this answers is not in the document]
kaby76: Agree. Will move this over to a descriptor.

runtime/CSharp/tests/issue-3510/ChannelTest.g4
KvanTTT: BTW, we have similar tests but in `tool-testsuite`: https://github.com/antlr/antlr4/blob/dev/tool-testsuite/test/org/antlr/v4/test/tool/TestLexerActions.java But actually they should be moved to `runtime-testsuite`.

PR #3548, opened 2022-02-22
Fix for #3441 and #3443 -- expose Reset() and String() methods.

kaby76:
This is a fix for the missing API method String() in the [Token](https://github.com/antlr/antlr4/blob/2be5a44ba10760303fc8600b2816fce8e74fdaf5/runtime/Go/antlr/token.go#L21), and Reset() in the [Lexer](https://github.com/antlr/antlr4/blob/2be5a44ba10760303fc8600b2816fce8e74fdaf5/runtime/Go/antlr/lexer.go#L18).

Go requires any exported data or functions to be capitalized. These functions are required for this code to print out the token stream:

        j := 0
        for {
            t := lexer.NextToken()
            fmt.Print(j)
            fmt.Print(" ")
            // missing fmt.Println(t.String())
            fmt.Println(t.GetText())
            if t.GetTokenType() == antlr.TokenEOF {
                break
            }
            j = j + 1
        }
        // missing lexer.Reset()

IToken.ToString() in C# and Token.toString() in Java are available because whatever implementation is used, it is derived from `object`. Thus, in C#, you can do `t.ToString()` or Java, `t.toString()`. Yet, `t.String()` in Go causes a compilation error.

[Reset() in C#](https://github.com/antlr/antlr4/blob/2be5a44ba10760303fc8600b2816fce8e74fdaf5/runtime/CSharp/src/Lexer.cs#L112) and [reset() in Java](https://github.com/antlr/antlr4/blob/2be5a44ba10760303fc8600b2816fce8e74fdaf5/runtime/Java/src/org/antlr/v4/runtime/Lexer.java#L88) are explicitly listed `public` in the interface. Thus, in C#, you can do `lexer.Reset();`, and Java, `lexer.reset();`. Yet,  `lexer.Reset()` or `lexer.reset()` in Go results in a compilation error.

### DIFF
--- a/runtime/Go/antlr/testing_api_test.go
+++ b/runtime/Go/antlr/testing_api_test.go
@@ -1,0 +1,35 @@
+// Run via "go test"
+
+package antlr
+
+import (
+	"testing"
+)
+
+func next(t *testing.T, lexer *LexerB, want string) {
+    var token = lexer.NextToken()
+    var got = token.String()
+    if got != want {
+        t.Errorf("got %q, wanted %q", got, want)
+    }
+}
+
+func TestString(t *testing.T){
+    str := NewInputStream("a b c 1 2 3")
+    lexer := NewLexerB(str)
+    next(t, lexer, "[@-1,0:0='a',<1>,1:0]")
+    next(t, lexer, "[@-1,1:1=' ',<7>,1:1]")
+    next(t, lexer, "[@-1,2:2='b',<1>,1:2]")
+    next(t, lexer, "[@-1,3:3=' ',<7>,1:3]")
+    next(t, lexer, "[@-1,4:4='c',<1>,1:4]")
+    next(t, lexer, "[@-1,5:5=' ',<7>,1:5]")
+    next(t, lexer, "[@-1,6:6='1',<2>,1:6]")
+    next(t, lexer, "[@-1,7:7=' ',<7>,1:7]")
+    next(t, lexer, "[@-1,8:8='2',<2>,1:8]")
+    next(t, lexer, "[@-1,9:9=' ',<7>,1:9]")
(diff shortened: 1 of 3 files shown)


runtime/Go/antlr/testing_api_test.go
```diff
@@ -0,0 +1,35 @@
...
+    next(t, lexer, "[@-1,3:3=' ',<7>,1:3]")
+    next(t, lexer, "[@-1,4:4='c',<1>,1:4]")
+    next(t, lexer, "[@-1,5:5=' ',<7>,1:5]")
+    next(t, lexer, "[@-1,6:6='1',<2>,1:6]")
+    next(t, lexer, "[@-1,7:7=' ',<7>,1:7]")
+    next(t, lexer, "[@-1,8:8='2',<2>,1:8]")
+    next(t, lexer, "[@-1,9:9=' ',<7>,1:9]")
+    next(t, lexer, "[@-1,10:10='3',<2>,1:10]")
+    next(t, lexer, "[@-1,11:10='<EOF>',<-1>,1:11]")
+}
```
KvanTTT, line 32: Should it be a test on `Reset` method as well?

kaby76: Yeah, looks like I missed that. Thanks.

kaby76:
Waiting for https://github.com/antlr/antlr4/pull/3558

jimidle:
Guys - I have fixed this in the /v4 code, where it should be - I used this code and added the new test file (though really, those test files are useless and I need to actually add some tests. The build tests are adequate for now, but there is supposed to be a test suite in the actual go runtime itself . What is there right now is just a few half-hearted tests that don't really help anything.

Anyway, you can close this PR now as I have used @kaby76 Ken's code and put it in the /v4 code. That PR will come in shortly and close the issues that this PR was correctly addressing.

PR #3556, opened 2022-02-23
ATN serialization improvements

KvanTTT:
* Move decode method from ATNSerializer to ATNDeserializerHelper
* Get rid of excess ATN serialization (for interpreter data and generated file)

### DIFF
--- a/tool/src/org/antlr/v4/Tool.java
+++ b/tool/src/org/antlr/v4/Tool.java
@@ -390,27 +390,19 @@ public class Tool {
 
 		if ( generate_ATN_dot ) generateATNs(g);
 
-		if (gencode && g.tool.getNumErrors()==0 ) {
-			String interpFile = generateInterpreterData(g);
-			try (Writer fw = getOutputFileWriter(g, g.name + ".interp")) {
-				fw.write(interpFile);
-			}
-			catch (IOException ioe) {
-				errMgr.toolError(ErrorType.CANNOT_WRITE_FILE, ioe);
-			}
-		}
+		IntegerList atnData = gencode && g.tool.getNumErrors()==0
+			? generateInterpreterData(g)
+			: null;
 
 		// PERFORM GRAMMAR ANALYSIS ON ATN: BUILD DECISION DFAs
 		AnalysisPipeline anal = new AnalysisPipeline(g);
 		anal.process();
 
-		//if ( generate_DFA_dot ) generateDFAs(g);
-
 		if ( g.tool.getNumErrors()>prevErrors ) return;
 
 		// GENERATE CODE
 		if ( gencode ) {
-			CodeGenPipeline gen = new CodeGenPipeline(g, codeGenerator);
+			CodeGenPipeline gen = new CodeGenPipeline(g, codeGenerator, atnData);
 			gen.process();
 		}
 	}
@@ -698,27 +690,27 @@ public class Tool {
 		}
 	}
 
-	public static String generateInterpreterData(Grammar g) {
+	public IntegerList generateInterpreterData(Grammar g) {
 		StringBuilder content = new StringBuilder();
 
 		content.append("token literal names:\n");
 		String[] names = g.getTokenLiteralNames();
 		for (String name : names) {
-			content.append(name + "\n");
+			content.append(name).append("\n");
 		}
 		content.append("\n");
 
 		content.append("token symbolic names:\n");
 		names = g.getTokenSymbolicNames();
 		for (String name : names) {
-			content.append(name + "\n");
+			content.append(name).append("\n");
 		}
 		content.append("\n");
 
 		content.append("rule names:\n");
 		names = g.getRuleNames();
 		for (String name : names) {
-			content.append(name + "\n");
+			content.append(name).append("\n");
 		}
 		content.append("\n");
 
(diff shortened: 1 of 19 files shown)


parrt:
Hi.  can you remind me what the primary purposes of shuffling code? as usual I don't like creating new files unless we have to.  could you help me understand the excess serialization going on? thanks!

KvanTTT:
I've moved `decode` out of the serializer because of the following reasons:

* It's used only for tests. There is no need to keep it in the ANTLR core package
* Actually it's for deserialization, not for serialization

parrt:
can you remind me what the excess serialization stuff you're referring to? thanks.

KvanTTT:
> Not sure we need separate class/file for a function. Why not just put decode() into TestATNSerialization.java?

Because it's already quite a big class and `decode` method is also not small. I just would like to split tests and helper code.

> can you remind me what the excess serialization stuff you're referring to?

I've encountered it during debugging. I've noticed the ATN is being serialized (call of `ATNSerializer.serialize`) both for `.interp` file and for serialized data for lexer/parser.

parrt:
> Because it's already quite a big class and `decode` method is also not small. I just would like to split tests and helper code.

Yes, big is bad but on the other hand so it is making a class with one method and introducing a new file.  I think I would prefer just dumping the decode method into the test class.

Hmm... looks like some of the targets have implemented decode() like C++ (ATNSerializer.h):

```
    virtual std::string decode(const std::wstring& data);
```

> I've noticed the ATN is being serialized (call of ATNSerializer.serialize) both for .interp file and for serialized data for lexer/parser.

Ah. Hmm...let me pull in your branch and use a good difference tool instead of the website see  what the changes really look like.

tool/src/org/antlr/v4/Tool.java
```diff
@@ -727,26 +719,31 @@ public static String generateInterpreterData(Grammar g) {
...
 			content.append("mode names:\n");
 			for (String mode : ((LexerGrammar)g).modes.keySet()) {
-				content.append(mode + "\n");
+				content.append(mode).append("\n");
 			}
 		}
 		content.append("\n");
 
-		IntegerList serializedATN = ATNSerializer.getSerialized(g.atn, g.getLanguage());
-		// Uncomment if you'd like to write out histogram info on the numbers of
```
parrt, line 742: seems like we should leave this in

tool/src/org/antlr/v4/codegen/BlankOutputModelFactory.java
```diff
@@ -29,7 +30,7 @@
 	public ParserFile parserFile(String fileName) { return null; }
 
 	@Override
-	public Parser parser(ParserFile file) { return null; }
```
parrt, line 32: at first glance it looks strange that lexer() function below does not also need an ATN.

tool/src/org/antlr/v4/codegen/model/Lexer.java
```diff
@@ -21,12 +22,11 @@
 	public final Collection<String> modes;
 	public final Collection<String> escapedModeNames;
 
-	@ModelElement public LinkedHashMap<Rule, RuleActionFunction> actionFuncs =
-		new LinkedHashMap<Rule, RuleActionFunction>();
+	@ModelElement public LinkedHashMap<Rule, RuleActionFunction> actionFuncs = new LinkedHashMap<>();
 
-	public Lexer(OutputModelFactory factory, LexerFile file) {
-		super(factory);
-		this.file = file; // who contains us?
```
parrt, line 29: seems like that's a useful comment

tool/src/org/antlr/v4/codegen/model/Parser.java
```diff
@@ -7,17 +7,18 @@
...
 public class Parser extends Recognizer {
-	public ParserFile file;
+	public final ParserFile file;
 
-	@ModelElement public List<RuleFunction> funcs = new ArrayList<RuleFunction>();
+	@ModelElement public final List<RuleFunction> funcs = new ArrayList<RuleFunction>();
 
-	public Parser(OutputModelFactory factory, ParserFile file) {
-		super(factory);
-		this.file = file; // who contains us?
```
parrt, line 21: same as lexer; a useful comment

tool/src/org/antlr/v4/codegen/model/Recognizer.java
```diff
@@ -63,13 +65,14 @@ public Recognizer(OutputModelFactory factory) {
 
 		ruleNames = g.rules.keySet();
 		rules = g.rules.values();
-		atn = new SerializedATN(factory, g.atn);
```
parrt, line 66: I'm not a fan of flipping all of these IF to `?:`.  I don't think they are is easy to read.

parrt: you're doing all sorts of code revisions that are not really needed but are costing my time to review.

KvanTTT: Sorry.

parrt: not a problem. You are a valuable contributor and a hardcore tech guy!!  I just think maybe we need to synchronize better :) your instincts seem very good but I have to fit that into my limited focus time.

tool/src/org/antlr/v4/Tool.java
```diff
@@ -390,27 +390,19 @@ public void processNonCombinedGrammar(Grammar g, boolean gencode) {
...
-		if (gencode && g.tool.getNumErrors()==0 ) {
-			String interpFile = generateInterpreterData(g);
-			try (Writer fw = getOutputFileWriter(g, g.name + ".interp")) {
-				fw.write(interpFile);
-			}
-			catch (IOException ioe) {
-				errMgr.toolError(ErrorType.CANNOT_WRITE_FILE, ioe);
-			}
-		}
+		IntegerList atnData = gencode && g.tool.getNumErrors()==0
```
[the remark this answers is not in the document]
KvanTTT: Unfortunately, it's not so simple. Code generator uses `atnData` deeply internally during generation. StringTemplate library uses the `serialized` field from [`SerializedATN`](https://github.com/antlr/antlr4/blob/dev/tool/src/org/antlr/v4/codegen/model/SerializedATN.java).

image

Stack trace:

image

`atnData` for interp file is being calculated in [Tool.java](https://github.com/antlr/antlr4/blob/master/tool/src/org/antlr/v4/Tool.java#L400). It should be passed to the code generator somehow.

tool/src/org/antlr/v4/codegen/BlankOutputModelFactory.java
KvanTTT: Yes, because lexer works a bit differently and gets atnData from somewhere else. I'll take a look if it's possible to unify the parameter.

tool/src/org/antlr/v4/codegen/model/Lexer.java
KvanTTT: Maybe renamed `file` to `containingFile` instead of using such a comment?

tool/src/org/antlr/v4/codegen/BlankOutputModelFactory.java
KvanTTT: It depends on this code: https://github.com/antlr/antlr4/blob/dev/tool/src/org/antlr/v4/codegen/OutputModelController.java#L140-L152 There is no call to `lexer` method from `OutputModelController` but call to constructor directly.

parrt:
I will close this as it is superseded by the big PR I just pulled in.

Issue #3452, opened 2022-01-02
Go target generates non-unique tables for parser and lexer


kaby76:
This is a problem with grammars such as [grammars-v4/csharp/ (the C# language)](https://github.com/antlr/grammars-v4/tree/master/csharp), or grammars-v4/cpp, where there are two more parser grammars within one directory, i.e., for preprocessing.

The Antlr tool generates for Go tables parserATN, literalNames, symbolicNames, ruleNames for a parser in scope of the package. Generated this way, the Go compiler complain that you are trying to define these tables twice. To work around this problem, the Antlr tool must be called with a different package name using `-package` on the command line. And, since Go restricts sources with a package name to have one package name in one directory, and that the package name equal the name of the enclosing directory, the files have to be placed in separate dirctories using the `-o` and `-lib` options on the Antlr tool call.

It would make my life far, far easier if the Antlr tool would just generate a unique name for the table (e.g., append the name of the grammar) and offer getters. All the other targets (CSharp, Java, Dart, JavaScript, Python3) have classes supported by the language, so it's unique only to the Go target.

For additional details, see https://github.com/antlr/grammars-v4/issues/2452#issuecomment-1003720415

Issue #3446, opened 2021-12-30
Go runtime and Antlr tool should use/generate nested pointer to structs, not nested structs


kaby76:
This is a problem I discovered while porting [grammars-v4/sql/plsql to Go](https://github.com/antlr/grammars-v4/issues/2452#issuecomment-1002544972). It appears that the nested struct of a parser base class cannot be initialized. This is a big problem because it means that any initialization of a "base class" must be performed in the driver code after creating the parser object. This is completely different to how every other target works. In implementing an OOP date structure in Go, the base classes referenced in the generated parser and lexer must be pointers.

Issue #3264, opened 2021-09-02
Golang target @ member variables are global


dkfitzpatrick:
In the Java language target, variables declared in the @lexer::members section get declared within the lexer class.   In the golang target the variables get declared as global variables.    For example:

// TestLexer.g4
lexer grammar TestLexer;

@lexer::members {
   var depth int
}

OPAR   : '(' { depth++ } ;
CPAR   : ')' { depth-- } ;

It is not clear if this is by design or not (but it would seem not).    I am not familiar with the Antlr internals, but it seems the change is only related to the Go.stg template file.    From this:

type <lexer.name> struct {
	*<if(superClass)><superClass><else>antlr.BaseLexer<endif>
	channelNames []string
	modeNames []string
	// TODO: EOF string
}

To this (moving the template section for @lexer::members):

type <lexer.name> struct {
	*<if(superClass)><superClass><else>antlr.BaseLexer<endif>
	channelNames []string
	modeNames []string
       <if(namedActions.members)>
       <namedActions.members>
       <endif>
	// TODO: EOF string
}

And access the values via the receiver (aka l.depth).

My application was requiring recursive substitution - hence the ability to have lexer-instance specific values is important
(same issues apply to generated parsers).

Any pointers to operator error, or work-arounds highly appreciated.

anqurvanillapy:
Thanks to Go's structural interfaces, you could extend your lexer in this way:

```go
// MyLexerState is the 
type MyLexerState struct {
    depth int
}

func (s *MyLexerState) Open() {
    s.depth++
}

func (s *MyLexerState) Close() int {
    d := s.depth
    s.depth--
    return d
}

// StatefulLexer embeds the lexer state.
type StatefulLexer struct {
    *parser.MyLexer
    MyLexerState
}
```

And in your grammar file, do some structural typing tricks to call the interface:

```go
```

anqurvanillapy:
Well, here's a dirtier trick, but it works now:

* Notice that we could add any new methods to the lexer, from the `@lexer::header`, and the field `Interpreter` is an interface, so let's extend its implementation

```g4
@lexer::header{

// MyLayoutInterpreter extends the original interpreter with a user-defined state, here is `LayoutState`.
type MyLayoutInterpreter struct {
  antlr.ILexerATNSimulator
  LayoutState
}

func NewMyLayoutLexer(input antlr.CharStream) *MyLexer {
  l := NewMyLexer(input)
  // Re-assign the interpreter.
[...]

PR #3558, opened 2022-02-23
[Go] Fix for issue 3557 -- get "go test" working again

kaby76:
# What does this PR fix?

* #3547 (Go target)

# Discussion

This is a fix for the Go runtime so that "go test" works. The problem is that some of the runtime tests are within the runtime itself, and I don't think they are tested in GitHub Actions.

The change is essentially a couple of test files in the runtime. One of the files didn't explicitly state the "LexerA.g4" grammar, so I inserted a comment so people don't need to guess WTF it is.

# Why is this PR important?

It's blocking testing for another PR I want to make.


kaby76:
Prereq for https://github.com/antlr/antlr4/pull/3548

runtime/Go/antlr/tokenstream_rewriter_test.go
```diff
@@ -10,6 +10,16 @@ import (
...
 
+
+/* Assume the following grammar for this test.
+
+lexer grammar LexerA;
+A : 'a';
+B : 'b';
+C : 'c';
+
+*/
```
KvanTTT, line 21: Where did you take it from? Probably it should be extracted to a separate file.

KvanTTT (approved):

runtime/Go/antlr/tokenstream_rewriter_test.go
kaby76: Of course it should be in a file. But, so should the grammar in the comment [here](https://github.com/antlr/antlr4/blob/f5f777d5af3a8e847465787e6112872065440014/runtime/Go/antlr/testing_lexer_b_test.go#L12).

But, even if we decide to leave it here, and pull out the grammars into files, and generate the lexers for testing, the generated code will not build. That's because we have the lovely [global variable problem](https://github.com/antlr/antlr4/issues/3452). And, rather than deal with the problem in an intelligent manner, the output from the tool is "cleverly" hand-edited.

Even if we fix #3452, API tests should not even go here. How can one test the visibility of the API [from within the package[(https://github.com/antlr/antlr4/blob/f5f777d5af3a8e847465787e6112872065440014/runtime/Go/antlr/testing_lexer_b_test.go#L5)? It's all just wrong. The tests should not be here. It needs to go into the [runtime-testsuite](https://github.com/antlr/antlr4/tree/dev/runtime-testsuite) directory.

KvanTTT: > So, who knows if this is right. The "experts" are so much smarter than I. I only have a ~1/2 century of programming that I have done, but it's nothing.

Everyone can mistake despite the big experience as well as suggest a good solution. And it's okay.

kaby76: It's normal to make mistakes. We are all human. It's wrong to not take ownership.  This has been here for a long time, enough time to reflect and try to fix.

I think it best to add the code as is. #3452 / #3264 needs to be addressed post 4.10. I have a fix for this, but it needs to be integrated with a fix with #3446, which I also have.

parrt:
thanks!

parrt:
@kaby76  let me know if you need anything else :)

jimidle:
Guys - I don't think we should go down this path for Go testing. I think it needs me to dedicate some serious time to write Go runtime tests in Go idiomatic form. I am going to do some serious rummaging around in the go runtime in the next few weeks, and when I have got the performance where I think it should be, then I will look at stuff like this. 

Talking of performance, when I get some changes in to the `dev` branch, perhaps you will have time to run the comparison benchmark you have @kaby76 ? That seems some serious effort you put in there. I would like to get Go performance to be top of the tree. I know what is wrong, but some of the go stuff will require quite a bit of, shall we say, "correcting" to get rid of useless memory allocations and the like.

`go test./...` does work from the /v4 code, but there are valuable tests in there right now. 

I suggest that we close this. BTW - I am just going through all PRs and issue that pertain to Go at the moment so I can clean them up/rework them/ask for them to be closed. Start with a clean slate so to speak.

kaby76:
@jimidle I'll be looking at performance for Go soon. My plan is to: clean up and extend the test driver generator used in testing grammars-v4; add "individual" parsing back into selected targets and grammars, because "grouped" parsing (aka "warm-up") doesn't help for targets like PHP but the grammar "works" otherwise); add in parse tree comparison for all tests across all targets; finally, set up tests for performance across targets for selected grammars. It'll likely be a few weeks before performance tests are written.

KvanTTT:
>  I think it needs me to dedicate some serious time to write Go runtime tests in Go idiomatic form.

I'm not sure it's correct. We have several runtimes and ideally tests should be compatible with all of them. It's not idiomatic form but universal.

parrt:
We definitely want to keep the general tests but you're free to make whatever extra tests you want in your runtime directories.

jimidle:
I should have explained better. The kind of tests that would be here are
low level things specifically to go. You are 100% correct that we should
not try to recreate the higher level test bed here. Nor would I wish to.

These are low level unit and regression tests, and perhaps more
importantly, performance/memory tests that keep history and point out any
swings for the worst after changes.

The other thing that goes here is Example tests, which are auto included in
the Go documentation.

PR #3560, opened 2022-02-26
Provide Javascript port of TokenStreamRewriter

### LINKED ISSUE #1767: TokenStreamRewriter for javascript?
Hi.
I already use antlr for a few years with java target. Now I need to use with javascript target. I followed the tutorial and everything went well. Now I missed a class: TokenStreamRewriter. Does it have a different name? I searched on google and found nothing with this topic.

eirikur-grid:
# What
Add a JavaScript port of the TokenStreamRewriter that includes most, but not all, of the methods from the Java implementation.

# Why
So that JavaScript developers can benefit from having it to implement all sorts of neat things (refactorings, translation etc.) 

# Scope
## Included
 - `insertAfter`
 - `insertBefore`
 - `replace`
 - `delete`
 - `getText`
 - `getProgram`
 - `getTokenStream`

## Omitted
The following methods were omitted because 
a) I couldn't find any tests for them and 
b) their utility is not immediately obvious to me.  

 - `rollback`
 - `deleteProgram`
 - `getLastRewriteTokenIndex`

If someone needs them then we can obviously add support for them in a subsequent PR.

# How
By
 - Porting **most** of the `TokenStreamRewriter` class
 - Porting tests and supporting fixtures (there are a couple of toy grammars used by the tests)
 - Adding more tests to get close to 100% coverage

Credit to @sharwell for having ported the TokenStreamRewriter and the corresponding unit tests to TypeScript (see https://github.com/tunnelvisionlabs/antlr4ts). That has made creating a JavaScript port relatively straight-forward.

## Drive-bys
 - For added convenience, I added a Makefile within the spec folder to automate the code generation from the toy grammars.
 - I added a dev dependency for [c8](https://github.com/bcoe/c8) and a run target in package.json that generates a code coverage report. This I used to find uncovered sections in the `TokenStreamRewriter`.
 - I added an explicit dev dependency on eslint (seems it was transitive before) and upgraded it to the latest version. I then tweaked the parser options in order for it to recognize class static fields. See this [SO thread](https://stackoverflow.com/questions/60046847/eslint-does-not-allow-static-class-properties) for more context.
 - I made a few additions to `.gitignore` to exclude autogenerated files that don't need to be placed under source control.
 - I added an `npm run lint` target

Closes #1767

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -83,6 +83,9 @@ nbactions*.xml
 /gen4/
 /tool/playground/
 tmp/
+**/generatedCode/*.interp
+**/generatedCode/*.tokens
+**/generatedCode/*.bak
 
 # Configurable build files
 bilder.py
@@ -106,6 +109,9 @@ runtime/PHP
 
 # Swift binaries
 .build/
+
+# Code coverage reports
+coverage/
 
 # Cpp generated build files
 runtime/Cpp/CMakeCache.txt
--- a/runtime/JavaScript/spec/rewriter/Makefile
+++ b/runtime/JavaScript/spec/rewriter/Makefile
@@ -1,0 +1,16 @@
+ANTLR_VERSION = 4.12.0
+ANTLR_JAR = .antlr/antlr-$(ANTLR_VERSION)-complete.jar
+
+.antlr:
+	mkdir .antlr
+
+$(ANTLR_JAR): .antlr
+	curl https://www.antlr.org/download/antlr-$(ANTLR_VERSION)-complete.jar -o $(ANTLR_JAR)
+
+abc: abc.g4 $(ANTLR_JAR)
+	java -jar $(ANTLR_JAR) -Dlanguage=JavaScript -no-listener abc.g4 -o generatedCode/
+	sed -i.bak "s/import antlr4 from 'antlr4'/import antlr4 from '..\/..\/..\/src\/antlr4\/index.node.js'/" generatedCode/abc.js
+
+calc: calc.g4 $(ANTLR_JAR)
+	java -jar $(ANTLR_JAR) -Dlanguage=JavaScript -no-listener calc.g4 -o generatedCode/
+	sed -i.bak "s/import antlr4 from 'antlr4'/import antlr4 from '..\/..\/..\/src\/antlr4\/index.node.js'/" generatedCode/calc.js
--- a/runtime/JavaScript/spec/rewriter/generatedCode/abc.js
+++ b/runtime/JavaScript/spec/rewriter/generatedCode/abc.js
@@ -1,0 +1,37 @@
+// Generated from abc.g4 by ANTLR 4.12.0
+// jshint ignore: start
+import antlr4 from '../../../src/antlr4/index.node.js';
+
+
+const serializedATN = [4,0,3,13,6,-1,2,0,7,0,2,1,7,1,2,2,7,2,1,0,1,0,1,1,
+1,1,1,2,1,2,0,0,3,1,1,3,2,5,3,1,0,0,12,0,1,1,0,0,0,0,3,1,0,0,0,0,5,1,0,0,
+0,1,7,1,0,0,0,3,9,1,0,0,0,5,11,1,0,0,0,7,8,5,97,0,0,8,2,1,0,0,0,9,10,5,98,
+0,0,10,4,1,0,0,0,11,12,5,99,0,0,12,6,1,0,0,0,1,0,0];
+
+
+const atn = new antlr4.atn.ATNDeserializer().deserialize(serializedATN);
+
+const decisionsToDFA = atn.decisionToState.map( (ds, index) => new antlr4.dfa.DFA(ds, index) );
+
+export default class abc extends antlr4.Lexer {
+
+    static grammarFileName = "abc.g4";
+    static channelNames = [ "DEFAULT_TOKEN_CHANNEL", "HIDDEN" ];
+	static modeNames = [ "DEFAULT_MODE" ];
+	static literalNames = [ null, "'a'", "'b'", "'c'" ];
+	static symbolicNames = [ null, "A", "B", "C" ];
+	static ruleNames = [ "A", "B", "C" ];
+
+    constructor(input) {
+        super(input)
+        this._interp = new antlr4.atn.LexerATNSimulator(this, atn, decisionsToDFA, new antlr4.atn.PredictionContextCache());
+    }
+}
+
+abc.EOF = antlr4.Token.EOF;
+abc.A = 1;
+abc.B = 2;
+abc.C = 3;
+
+
+
--- a/runtime/JavaScript/spec/rewriter/generatedCode/calc.js
+++ b/runtime/JavaScript/spec/rewriter/generatedCode/calc.js
@@ -1,0 +1,50 @@
+// Generated from calc.g4 by ANTLR 4.12.0
+// jshint ignore: start
+import antlr4 from '../../../src/antlr4/index.node.js';
+
+
+const serializedATN = [4,0,7,38,6,-1,2,0,7,0,2,1,7,1,2,2,7,2,2,3,7,3,2,4,
+7,4,2,5,7,5,2,6,7,6,1,0,4,0,17,8,0,11,0,12,0,18,1,1,4,1,22,8,1,11,1,12,1,
+23,1,2,1,2,1,3,1,3,1,4,1,4,1,5,1,5,1,6,4,6,35,8,6,11,6,12,6,36,0,0,7,1,1,
+3,2,5,3,7,4,9,5,11,6,13,7,1,0,0,40,0,1,1,0,0,0,0,3,1,0,0,0,0,5,1,0,0,0,0,
+7,1,0,0,0,0,9,1,0,0,0,0,11,1,0,0,0,0,13,1,0,0,0,1,16,1,0,0,0,3,21,1,0,0,
+0,5,25,1,0,0,0,7,27,1,0,0,0,9,29,1,0,0,0,11,31,1,0,0,0,13,34,1,0,0,0,15,
+17,2,97,122,0,16,15,1,0,0,0,17,18,1,0,0,0,18,16,1,0,0,0,18,19,1,0,0,0,19,
+2,1,0,0,0,20,22,2,48,57,0,21,20,1,0,0,0,22,23,1,0,0,0,23,21,1,0,0,0,23,24,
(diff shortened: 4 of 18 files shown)


runtime/JavaScript/src/antlr4/BufferedTokenStream.js
```diff
@@ -83,6 +83,10 @@ export default class BufferedTokenStream extends TokenStream {
 		this.index = this.adjustSeekIndex(index);
 	}
 
+	get size() {
+		return this.tokens.length;
+	}
```
eirikur-grid, line 88: This attribute is used by the `TokenStreamRewriter`. Seems that it was missing so I added it. In the Java run-time, this attribute is defined on the `IntStream` interface and subsequently implemented in `BufferedTokenStream.java`.

runtime/JavaScript/src/antlr4/BufferedTokenStream.js
```diff
@@ -148,7 +152,7 @@ export default class BufferedTokenStream extends TokenStream {
 		return n;
 	}
 
-// Get all tokens from start..stop inclusively///
+	// Get all tokens from start..stop inclusively///
```
eirikur-grid, line 155: Automatic code formatting kicked in on this file.

runtime/JavaScript/src/antlr4/TokenStreamRewriter.js
```diff
@@ -0,0 +1,340 @@
+import Token from "./Token.js";
+import Interval from "./misc/Interval.js";
+
+export default class TokenStreamRewriter {
+    // eslint-disable-next-line no-undef
+    static DEFAULT_PROGRAM_NAME = "default";
+
+    /**
+     * @param {import("./CommonTokenStream")} tokens 
+     */
```
eirikur-grid, line 10: I now realize that the JSDoc declarations should probably be improved for this class. I'll put some effort into adding more of them.

runtime/JavaScript/spec/TokenStreamRewriterSpec.js
```diff
@@ -0,0 +1,733 @@
...
+ */
+function getRewriter(lexerClass, input) {
+    const chars = new antlr4.InputStream(input);
+    const lexer = new lexerClass(chars);
+    const tokens = new antlr4.CommonTokenStream(lexer);
+    tokens.fill();
+    return new antlr4.TokenStreamRewriter(tokens);
+}
+
+// TODO: change test names to BDD statements
```
eirikur-grid, line 18: I can tweak the test names. When I started porting these tests, jest was still being used as a test runner. They now feel a bit out of place once situated in the describe/it context of BDD tests.

ericvergnaud:
@eirikur-grid sorry if it's taken us so long to look at this
would you be able to rebase from this PR https://github.com/antlr/antlr4/pull/4027
and provide an associated d.ts file

eirikur-grid:
This week is going to be quite hectic for me. I'll take a look shortly after Christmas.

eirikur-grid:
@ericvergnaud I've rebased and added the type definitions file.

.gitignore
```diff
@@ -83,6 +83,9 @@ nbactions*.xml
 /gen4/
 /tool/playground/
 tmp/
+**/generatedCode/*.interp
```
ericvergnaud, line 86: I guess **/generatedCode/* would be better ?

runtime/JavaScript/spec/Makefile
```diff
@@ -0,0 +1,15 @@
...
+
+.antlr:
+	mkdir .antlr
+
+.antlr/antlr-4.11.1-complete.jar: .antlr
+	curl https://www.antlr.org/download/antlr-4.11.1-complete.jar -o $(ANTLR_JAR)
+
+abc: abc.g4 .antlr/antlr-4.11.1-complete.jar
+	java -jar $(ANTLR_JAR) -Dlanguage=JavaScript -no-listener abc.g4 -o generatedCode/
+	sed -i.bak "s/import antlr4 from 'antlr4'/import antlr4 from '..\/..\/src\/antlr4\/index.js'/" generatedCode/abc.js
```
ericvergnaud, line 11: you'll need to update this to point to index.node.js

ericvergnaud:
@eirikur-grid we're about to release, and this is a good time to finalize things.
Could you:
 - squash your commits and gpg-sign off the squashed commit (this replaces the contributors.txt strategy)
 - move your changes in index.js to index.node.js and index.web.js
 - add an export in index.d.ts
 - move your spec/* files under spec/rewriter/* to clarify their purpose
 - not mandatory but a .ts test would also be welcome (to validate the ts api)
 If you can do that rapidly enough, this will make it into the next release

eirikur-grid:
> @eirikur-grid we're about to release, and this is a good time to finalize things. Could you:
> 
>     * squash your commits and gpg-sign off the squashed commit (this replaces the contributors.txt strategy)
> 
>     * move your changes in index.js to index.node.js and index.web.js
> 
>     * add an export in index.d.ts
> 
>     * move your spec/* files under spec/rewriter/* to clarify their purpose
> 
>     * not mandatory but a .ts test would also be welcome (to validate the ts api)
> 
> 
> If you can do that rapidly enough, this PR will make it into the next release

You caught me at a very busy time, hence the late reply. I should be able to get around to this before the end of this week.

ericvergnaud:
HeyToo bad, version 4.12 just went out Please rebase and we’ll ship this in the next releaseEricEnvoyé de mon iPhoneLe 19 févr. 2023 à 23:13, Eiríkur Fannar Torfason ***@***.***> a écrit :﻿

@eirikur-grid we're about to release, and this is a good time to finalize things. Could you:
* squash your commits and gpg-sign off the squashed commit (this replaces the contributors.txt strategy)

* move your changes in index.js to index.node.js and index.web.js

* add an export in index.d.ts

* move your spec/* files under spec/rewriter/* to clarify their purpose

* not mandatory but a .ts test would also be welcome (to validate the ts api)

If you can do that rapidly enough, this PR will make it into the next release

You caught me at a very busy time, hence the late reply. I should be able to get around to this before the end of this week.

.gitignore
eirikur-grid: A single leading asterisk will not match the slash character. The double asterisk will and is therefore needed to match `generatedCode` in nested directories.

The reason for listing three different file extensions is that I don't want to exclude the entire `generatedCode` directory, as it contains .js files for the toy grammars and we need those for the tests.

runtime/JavaScript/spec/rewriter/Makefile
```diff
@@ -0,0 +1,15 @@
+ANTLR_JAR = .antlr/antlr-4.12.0-complete.jar
+
+.antlr:
+	mkdir .antlr
+
+.antlr/antlr-4.12.0-complete.jar: .antlr
```
ericvergnaud, line 6: please avoid hard coded versions, use variables instead to ease the release process

historical-contributors-agreement.txt
```diff
@@ -326,5 +326,6 @@ YYYY/MM/DD, github id, Full name, email
 2021/12/16, Ketler13, Oleksandr Martyshchenko, oleksandr.martyshchenko@gmail.com
 2021/12/25, Tinker1024, Tinker1024, tinker@huawei.com
 2021/12/31, Biswa96, Biswapriyo Nath, nathbappai@gmail.com
+2022/02/26, eirikur-grid, Eiríkur Fannar Torfason, eirikur [dot] torfason [at] gmail [dot] com
```
ericvergnaud, line 329: this is no longer the correct model, please use GitHub DCO in your next commit

runtime/JavaScript/spec/rewriter/Makefile
eirikur-grid: Good point, fixed.

historical-contributors-agreement.txt
eirikur-grid: Done

runtime/JavaScript/spec/rewriter/Makefile
```diff
@@ -0,0 +1,16 @@
+ANTLR_VERSION = 4.12.0
```
ericvergnaud, line 1: I think we have a small problem here
The purpose if the CI is not to test against the latest official version, rather against the version being tested.
So the makefile should use the tool from the local maven repo, not from the antlr web site.

runtime/JavaScript/spec/rewriter/generatedCode/abc.js
```diff
@@ -0,0 +1,37 @@
+// Generated from abc.g4 by ANTLR 4.12.0
```
ericvergnaud, line 1: this needs to be regenerated during tests or it will fail with a version mismatch

runtime/JavaScript/spec/rewriter/generatedCode/calc.js
```diff
@@ -0,0 +1,50 @@
+// Generated from calc.g4 by ANTLR 4.12.0
```
ericvergnaud, line 1: same here

runtime/JavaScript/spec/rewriter/Makefile
eirikur-grid: OK, how do I do that? Are there any examples within the repo that I can follow? I can see that the Java tests are capable of generating lexers/parsers from a grammar at run-time but what about other languages?

When I started working on this, I found Python tests using generated code (runtime/Python2/tests/mocks/TestLexer.py) that has been commited to the repo, so I assumed this was OK.

eirikur-grid: I think my contribution ends here. Over a year ago, I found myself needing a JavaScript version of the TokenStreamRewriter. That is no longer the case, so I have no vested interest in putting more effort into this.

ericvergnaud: All generic grammar tests generate parsers at CI time, but they are run from java.

Every GitHub action runs the "Build ANTLR with Maven" step, which builds the tool and registers it in the local maven repo, which should be found at ~/m2/repository/org/antlr/antlr4/${version}/antlr4-${version}-complete.jar.

You are correct in observing that many target-specific tests do not follow the expected policy, but let's not take things worse :-)

ericvergnaud: Sorry to hear that, and apologies for taking so long to consider this PR. I hope your interest will grow again soon. This is useful work for the community.

ericvergnaud:
@parrt ready to merge, I suggest we deal with the versioning issue at a global level since it affects all targets

parrt:
Re "derived from antlr4ts", license appears to be "The ANTLR Project" so we're good there, right?  https://github.com/tunnelvisionlabs/antlr4ts/blob/master/LICENSE

PR #3561, opened 2022-02-27
Convert Python2/3 to use int arrays not strings for ATN encodings

parrt:
Also make target indicate int vs string.

@ericvergnaud @KvanTTT how does this look?  Simplifies our world a lot.  Any interest in making the analogous changes in other targets that allow big int arrays?

### DIFF
--- a/runtime/Python2/src/antlr4/atn/ATNDeserializer.py
+++ b/runtime/Python2/src/antlr4/atn/ATNDeserializer.py
@@ -21,8 +21,12 @@ class ATNDeserializer (object):
         self.stateFactories = None
         self.actionFactories = None
 
-    def deserialize(self, data):
+    def deserialize_str(self, data):
         self.data = [ ord(c) for c in data ]
+        return self.deserialize(data)
+
+    def deserialize(self, data):
+        self.data = data
         self.pos = 0
         self.checkVersion()
         atn = self.readATN()
--- a/runtime/Python2/src/antlr4/atn/ATNDeserializer.py
+++ b/runtime/Python2/src/antlr4/atn/ATNDeserializer.py
@@ -21,10 +21,6 @@ class ATNDeserializer (object):
         self.stateFactories = None
         self.actionFactories = None
 
-    def deserialize_str(self, data):
-        self.data = [ ord(c) for c in data ]
-        return self.deserialize(data)
-
     def deserialize(self, data):
         self.data = data
         self.pos = 0
--- a/tool/src/org/antlr/v4/codegen/Target.java
+++ b/tool/src/org/antlr/v4/codegen/Target.java
@@ -343,8 +343,7 @@ public abstract class Target {
 			throw new IllegalArgumentException(String.format("Cannot encode the specified value: %d", v));
 		}
 
-		String language = getLanguage();
-		if (language.equals("Go") || language.equals("Cpp")) {
+		if ( ATNSerializedAsInts() ) {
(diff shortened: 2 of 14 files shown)


tool/src/org/antlr/v4/codegen/Target.java
```diff
@@ -689,6 +689,10 @@ public boolean supportsOverloadedMethods() {
 		return true;
 	}
 
+	public boolean ATNSerializedAsInts() {
```
KvanTTT, line 692: I recommend using the following naming for methods: `isATNSerializedAsInts()`. At least the leading letter should be in lowercase.

tool/src/org/antlr/v4/codegen/Target.java
```diff
@@ -689,6 +689,10 @@ public boolean supportsOverloadedMethods() {
...
+		return false;
```
KvanTTT, line 693: Probably it makes sense to return `true` by default because early you said `Ints` are more preferable than `Strings`.

tool/src/org/antlr/v4/codegen/target/CppTarget.java
```diff
@@ -86,6 +86,10 @@ protected boolean shouldUseUnicodeEscapeForCodePointInDoubleQuotedString(int cod
 		}
 	}
 
+	public boolean ATNSerializedAsInts() {
```
KvanTTT, line 89: `@Override` annotation is missing.

runtime/Python2/src/antlr4/atn/ATNDeserializer.py
```diff
@@ -21,8 +21,12 @@ def __init__(self, options = None):
         self.stateFactories = None
         self.actionFactories = None
 
-    def deserialize(self, data):
+    def deserialize_str(self, data):
         self.data = [ ord(c) for c in data ]
+        return self.deserialize(data)
+
+    def deserialize(self, data):
+        self.data = data
```
KvanTTT, line 29: Do we need both of these methods? I suggest keeping only the single `deserialize`.

tool/src/org/antlr/v4/codegen/Target.java
parrt: good suggestion. yes, I had trouble coming up with a name that didn't violate rules or look weird. I'll change.

tool/src/org/antlr/v4/codegen/Target.java
parrt: sounds good; I will reverse it

tool/src/org/antlr/v4/codegen/target/CppTarget.java
parrt: done

runtime/Python2/src/antlr4/atn/ATNDeserializer.py
parrt: probably not but I was worried that something else might use it.  I will delete it and see what happens.

KvanTTT (approved):

parrt:
@marcospassos Could you add this method to PhpTarget.java?

```java
	@Override
		return false;
	}
```

Thanks!

parrt:
Unless you'd like to simplify the target and gen arrays of int instead of strings like this PR does for Python. up to you!!

ericvergnaud:
Less code, love it!

PR #3564, opened 2022-03-02
Some bugfixes

KvanTTT:


tool/src/org/antlr/v4/automata/ParserATNFactory.java
```diff
@@ -472,24 +472,26 @@ public Handle elemList(List<Handle> els) {
...
 				epsilon(el.right, els.get(i+1).left);
 			}
 		}
 		Handle first = els.get(0);
-		Handle last = els.get(n -1);
-		if ( first==null || last==null ) {
-			g.tool.errMgr.toolError(ErrorType.INTERNAL_ERROR, "element list has first|last == null");
-		}
-		return new Handle(first.left, last.right);
+		Handle last = els.get(n - 1);
```
ericvergnaud, line 493: That's a significant change in behavior, that might break users code

KvanTTT: I've just removed `first==null || last==null` check, `null` elements are being processed later. How it could affect user code? All tests are passing.

ericvergnaud: It no longer throws a tool error, that's a significant change.
Not sure how to reproduce the throwing scenario, but it would be better to have a test for that (which throws with the existing code) and check that it still does with the  new code.
We never know what developers do with development tools, they might be sending some random grammars and expecting an error if the grammar isn't valid...

ericvergnaud: I'm not saying this is important, but not sure the first message can't be improved ? How about putting the undefined rules in quotes ?

KvanTTT: Maybe it can, but I didn't change it. This is the existing error that is being thrown in https://github.com/antlr/antlr4/blob/dev/tool/src/org/antlr/v4/automata/ParserATNFactory.java#L297 Also, it looks like we don't use quotes for other similar errors, see for example https://github.com/antlr/antlr4/blob/e19c0abea36409dbd1b89035a37f8145b517406d/tool/src/org/antlr/v4/tool/ErrorType.java#L204

parrt:
Hi.  I think I will close this because as I mentioned I'm not a big fan of removing IFs in favor of `?:` and I can't risk changing the ATN factory.  basically, I think I'm unlikely to accept changes to the code that are purely for the purpose of code cleanup.  I'm really in maintenance mode.

KvanTTT:
Ok, I can rollback code style changes. But I don't see significant changes of ATN factory, I've fixed NRE that should not be thrown, it's quite important especially for plugin.

parrt:
Hi. please roll back to avoid style changes so that I can see the actual changes. :)  then I can take a quickly look :)

KvanTTT:
It's already done :) https://github.com/antlr/antlr4/pull/3632

PR #3565, opened 2022-03-04
[C++] Switch to camel case for generated static data variable names

jcking:
I realized that generated names are difficult to read without `_` separating the parser/lexer name and the variable name. In example, `XPathLexerstaticData` looks awfully like `XPathLexerStaticData`. While `XPathLexer_staticData` is distinctly different from `XPathLexerStaticData`. While this deviates from the style overall, it is significantly more readable for generated code and makes it clear that `XPathLexer_staticData` is related to `XPathLexer`.

### DIFF
--- a/runtime/Cpp/runtime/src/tree/xpath/XPathLexer.cpp
+++ b/runtime/Cpp/runtime/src/tree/xpath/XPathLexer.cpp
@@ -37,11 +37,11 @@ struct XPathLexerStaticData final {
   std::unique_ptr<antlr4::atn::ATN> atn;
 };
 
-std::once_flag XPathLexer_onceFlag;
-XPathLexerStaticData *XPathLexerstaticData = nullptr;
+std::once_flag xpathLexerOnceFlag;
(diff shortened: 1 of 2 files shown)


jcking:
@mike-lischke Was looking at generated code and realized without `_` it is difficult to discern what is going on.

mike-lischke:
Why not give `XPathLexer_staticData` a completely different name? Also with the underscore they seem to be related and that raises the question why there are two of them.

jcking:
I think I was able to get it to use `xpathLexerStaticData`, which is more readable and consistent. Waiting on tests to confirm.

mike-lischke (approved):
Yes, better and now even conforms to variable naming style :-)

@parrt This patch is ready for merge.

mike-lischke:
Ah, @jcking, just a minor thing, but can you also edit the PR title to reflect what was actually done. Might come in handy in later searches.

jcking:
Done. Changed title.

PR #3567, opened 2022-03-05
use ints not chars for C#

parrt:


parrt:
Woot! Another victory. no more string ATNs for C#!  @KvanTTT @ericvergnaud

tool/src/org/antlr/v4/codegen/target/CSharpTarget.java
```diff
@@ -173,7 +173,7 @@ private void reportError(STMessage msg) {
 
 	@Override
 	public boolean isATNSerializedAsInts() {
-		return false;
+		return true;
```
KvanTTT, line 176: I guess this method should be completely removed since abstract method has the same implementation.

runtime/CSharp/src/Tree/Xpath/XPathLexer.cs
```diff
@@ -46,7 +43,7 @@ public const int
 	};
 
 	public static readonly string[] ruleNames = {
-		"Anywhere", "Root", "Wildcard", "Bang", "ID", "NameChar", "NameStartChar",
+		"Anywhere", "Root", "Wildcard", "Bang", "ID", "NameChar", "NameStartChar",
```
KvanTTT, line 46: Excess whitespace at the end on the line (or changed line break?).

runtime/CSharp/src/Tree/Xpath/XPathLexer.cs
```diff
@@ -64,7 +61,7 @@ public XPathLexer(ICharStream input, TextWriter output, TextWriter errorOutput)
 		null, null, null, "'//'", "'/'", "'*'", "'!'"
 	};
 	private static readonly string[] _SymbolicNames = {
-		null, "TokenRef", "RuleRef", "Anywhere", "Root", "Wildcard", "Bang", "ID",
+		null, "TokenRef", "RuleRef", "Anywhere", "Root", "Wildcard", "Bang", "ID",
```
KvanTTT, line 64: Excess whitespace.

runtime/CSharp/src/Tree/Xpath/XPathLexer.cs
```diff
@@ -101,61 +98,34 @@ public override IVocabulary Vocabulary
 	}
 	private void ID_action(RuleContext _localctx, int actionIndex) {
 		switch (actionIndex) {
-		case 0:
+		case 0:
```
KvanTTT, line 101: Why excess whitespaces appeared in the PR?

runtime/CSharp/src/Tree/Xpath/XPathLexer.cs
parrt: weird.  this is generated code though so...

tool/src/org/antlr/v4/codegen/target/CSharpTarget.java
parrt: I will take care of this once I get rid of more bring atn stuff.

runtime/CSharp/src/Tree/Xpath/XPathLexer.cs
parrt: generated code so I've no idea

PR #3568, opened 2022-03-06
JavaScript: Change string to int array for serialized ATN

parrt:
@KvanTTT @ericvergnaud @jcking And now JavaScript uses an int array not a string! woot!  Soon we will worry less about large ATN.

### DIFF
--- a/tool/src/org/antlr/v4/codegen/target/JavaScriptTarget.java
+++ b/tool/src/org/antlr/v4/codegen/target/JavaScriptTarget.java
@@ -94,6 +94,6 @@ public class JavaScriptTarget extends Target {
 
 	@Override
 	public boolean isATNSerializedAsInts() {
-		return false;
+		return true;
 	}
 }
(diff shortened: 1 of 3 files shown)


tool/src/org/antlr/v4/codegen/target/JavaScriptTarget.java
```diff
@@ -94,6 +94,6 @@ public boolean supportsOverloadedMethods() {
 
 	@Override
 	public boolean isATNSerializedAsInts() {
```
KvanTTT, line 96: The method should be removed since the basic implementation is the same.

ericvergnaud: I'd suggest making the overridden method purely abstract. The base class shouldn't make any assumption about the derived classes.

KvanTTT: Not sure since we only need strings for Java runtime. Other runtimes can use ints for encoding.

parrt: When I/we can shift others to int-only then I guess we can remove method. good idea.

PR #3569, opened 2022-03-07
perf: Add default implementation for Visit in ParseTreeVisitor. 

chenquan:
Reference: https://github.com/antlr/antlr4/blob/ad29539cd2e94b2599e0281515f6cbb420d29f38/runtime/Java/src/org/antlr/v4/runtime/tree/AbstractParseTreeVisitor.java#L18


chenquan:
@parrt PTAL

PR #3570, opened 2022-03-07
[C++] Remove unused LexerATNSimulator::match_calls

jcking:
The variable appears to be unused and atomic operations are not exactly free.

### DIFF
--- a/runtime/CSharp/src/Atn/LexerATNSimulator.cs
+++ b/runtime/CSharp/src/Atn/LexerATNSimulator.cs
@@ -45,8 +45,6 @@ namespace Antlr4.Runtime.Atn
 
 		readonly SimState prevAccept = new SimState();
 
-		public static int match_calls = 0;
-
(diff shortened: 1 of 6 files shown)


mike-lischke:
I'm not sure, but I believe this is for debugging. @parrt Do you remember what that `match_calls` is for?

parrt:
I can only think that this is for debugging. I notice that it is not in the job a target. @ericvergnaud  any idea what `match_calls` is for other than debugging?  I see it in these targets

Screen Shot 2022-03-08 at 4 19 12 PM

@jcking care to fix the other targets in parallel?  if not I can easily make a separate PR

mike-lischke (approved):

ericvergnaud:
I believe it is for debugging, particularly useful across targets to check if the algorithm behaves the same for a given grammar and input

parrt:
ah! Ok, well, I hope we are beyond most issue with the ATN!  :)

PR #3572, opened 2022-03-07
[C++] Force generated static data type name to titlecase

jcking:


jcking:
@parrt Assuming tests pass.

parrt:
@mike-lischke seems ok to me; it's an `_` added to prevent collisions on a var name.

mike-lischke:
Hard to say if that is ok. Leading underscore denotes private variables. Are these all private? I doubt it, otherwise they had been prefixed with an underscore already.

Also without explanation I see no reason why these names should be changed at all.

jcking:
Okay, different approach that keeps the style. I forced the type name to be capitlized, which should avoid the conflict.

The issue is that some grammars use lowecase names. This causes the static data type name and variable name to be the same, causing compilation to fail. Forcing the type name to be capitlized should address this edge case.

parrt:
That works for me what I will wait to hear from @mike-lischke

mike-lischke (approved):
Yes, that's better. Thanks.

@parrt ready for merge

PR #3573, opened 2022-03-08
Don't decode utf16 in ATN deserialization for Go; it's just a bunch of uint16

parrt:


parrt:
@KvanTTT @jcking does this look right?

KvanTTT:
It looks like this is already fixed by https://github.com/antlr/antlr4/pull/3571

jcking:
I'm happy having it in a separate pill request. Let's keep it in ***@***.***

PR #3574, opened 2022-03-08
[Go] Cleanup and fix ATN deserialization verification

jcking:
Updates `ATNDeserializationOptions`, makes some things private, fixes ATN deserialization, and enables verification by default.

Currently ATN deserialization is incorrect, but it is not clear what the incorrect behavior would surface as.


jcking:
@parrt

parrt:
That does look more Go idiomatic. All tests are passing.

PR #3575, opened 2022-03-10
Use int array not string for ATN in Swift

parrt:


parrt:
And another int array!! yay! @KvanTTT @jcking @ericvergnaud

parrt:
woohoo! Only php and dart left I think?

PR #3576, opened 2022-03-10
[C++] Fix generated Lexer static data constructor

jcking:
I missed one from my previous pull request.


jcking:
@parrt Missed one. :(

PR #3583, opened 2022-03-14
[Java] Preserve serialized ATN version 3 compatibility

jcking:
Java is a bit special in that it is the only language where it is normal for pre-compiled code to be distributed. This makes it difficult to hard cut to a new serialized ATN version. There are a bunch of libraries which break unexpectedly, including some Android libraries. This patch preserves backwards compatibility for ATN serialized version 3, by simply skipping the UUID and assuming all features were supported.

### DIFF
--- a/runtime/Java/src/org/antlr/v4/runtime/atn/ATNDeserializer.java
+++ b/runtime/Java/src/org/antlr/v4/runtime/atn/ATNDeserializer.java
@@ -20,10 +20,10 @@ import java.util.Locale;
  * @author Sam Harwell
  */
 public class ATNDeserializer {
-	public static final int SERIALIZED_VERSION;
-	static {
-		SERIALIZED_VERSION = 4;
-	}
+
+  static final int LEGACY_SERIALIZED_VERSION = 3;
+
+	public static final int SERIALIZED_VERSION = 4;
 
 	interface UnicodeDeserializer {
 		// Wrapper for readInt() or readInt32()
@@ -90,6 +90,12 @@ public class ATNDeserializer {
 
 		int p = 0;
 		int version = toInt(data[p++]);
+    if (version == LEGACY_SERIALIZED_VERSION) {
+      // Preserve backwards compatibility for version 3. We simply skip over the UUID and assume all
+      // features were supported.
+      p += 8;
+      version = SERIALIZED_VERSION;
+    }
 		if (version != SERIALIZED_VERSION) {
 			String reason = String.format(Locale.getDefault(), "Could not deserialize ATN with version %d (expected %d).", version, SERIALIZED_VERSION);
 			throw new UnsupportedOperationException(new InvalidClassException(ATN.class.getName(), reason));


runtime/Java/src/org/antlr/v4/runtime/atn/ATNDeserializer.java
```diff
@@ -90,6 +90,12 @@ public ATN deserialize(char[] data) {
 
 		int p = 0;
 		int version = toInt(data[p++]);
+    if (version == LEGACY_SERIALIZED_VERSION) {
+      // Preserve backwards compatibility for version 3. We simply skip over the UUID and assume all
+      // features were supported.
+      p += 8;
+      version = SERIALIZED_VERSION;
+    }
```
KvanTTT, line 98: Use tabs instead of spaces (ANTLR uses tabs formatting).

KvanTTT (commented):
If it's applied to Java it should be applied to other targets as well.

Also, I don't think it's a usefull change because it complicates the code. Moreover, it won't work if [32-bit serialization](https://github.com/antlr/antlr4/discussions/3580) is implemented.

KvanTTT:
> Java is a bit special in that it is the only language where it is normal for pre-compiled code to be distributed. 

Java is not a such single language. At least C# language is similar.

runtime/Java/src/org/antlr/v4/runtime/atn/ATNDeserializer.java
```diff
@@ -20,10 +20,10 @@
  * @author Sam Harwell
  */
 public class ATNDeserializer {
-	public static final int SERIALIZED_VERSION;
-	static {
-		SERIALIZED_VERSION = 4;
-	}
+
+  static final int LEGACY_SERIALIZED_VERSION = 3;
```
KvanTTT, line 24: It's unclear what exacly `LEGACY_SERIALIZED_VERSION` is. What if `SERIALIZED_VERSION` will be incremented?

runtime/Java/src/org/antlr/v4/runtime/atn/ATNDeserializer.java
parrt: Yeah,I am abandoning backward compatibility except for the java target because we have special internal problem to resolve. I'll deal with this when I can look at the 32 bit implementation.

parrt:
> Java is not a such single language. At least C# language is similar.

well, I think we've converted that to use a list of integers as well. so in this regard it is now different. Only java does the string serialization.

runtime/Java/src/org/antlr/v4/runtime/atn/ATNDeserializer.java
parrt: This is a loose patch to solve an internal problem we have.

runtime/Java/src/org/antlr/v4/runtime/atn/ATNDeserializer.java
parrt: Yeah, I guess I will do a bit of copy and paste programming for java if I get the 32-bit deserialization going.  thanks for pointing this out

KvanTTT:
> well, I think we've converted that to use a list of integers as well. so in this regard it is now different. Only java does the string serialization.

But it does not relate to the topic.

jcking:
@parrt I think we no longer need this, we should revert it as it can cause confusion for older versions of ANTLRv4. It causes a random exception instead of a graceful exception that is actionable.

parrt:
@jcking I have reverted and pushed back to dev.

PR #3584, opened 2022-03-14
[C++] Avoid using dynamic_cast where possible by using hand rolled RTTI

jcking:
`dynamic_cast` in hot code paths is much slower than using hand rolled RTTI using an enum. CPU profiling shows that `dynamic_cast` is one of the single most expensive calls. This change switches to using a hand rolled RTTI-like implementation using an enum.


QuantamHD:
@parrt I'm going to upload my draft PR for posterity, and include the performance numbers.

QuantamHD:
I'm waiting on Google to approve the DCO in this repo, but here's my version https://github.com/antlr/antlr4/compare/dev...QuantamHD:surelog_perf?expand=1

It results in about a 25% performance uplift in walltime, and 11% reduction in profile cycles.

Before Changes
image


After Changes
image

After the changes strcmp and dynamic_cast are nowhere near the top of the profile. I think @jcking's changes are great. I made mine more targeted by implementing specific is* functions for hot paths whereas Justin's is far more general. Either way removing RTTI in hot paths is a must.

runtime/Cpp/runtime/src/atn/SingletonPredictionContext.cpp
```diff
@@ -45,11 +49,11 @@ bool SingletonPredictionContext::operator == (const PredictionContext &o) const
     return true;
   }
 
-  const SingletonPredictionContext *other = dynamic_cast<const SingletonPredictionContext*>(&o);
-  if (other == nullptr) {
+  if (o.getContextType() != PredictionContextType::EMPTY && o.getContextType() != PredictionContextType::SINGLETON) {
```
mike-lischke, line 52: Why do we need the extra test for ::EMPTY here?

mike-lischke (commented):
Looks good, just a small question for you @jcking .

runtime/Cpp/runtime/src/atn/PredictionContext.cpp
```diff
@@ -389,11 +389,11 @@ std::string PredictionContext::toDOTString(const Ref<const PredictionContext> &c
   });
 
   for (const auto &current : nodes) {
-    if (is<const SingletonPredictionContext>(current)) {
+    if (current->getContextType() == PredictionContextType::SINGLETON || current->getContextType() == PredictionContextType::EMPTY) {
```
hzeller, line 392: Since we're calling the virtual `getContextType()` multiple times in this block (here and below, in total 3 times), it might be worthwhile calling it only once, remembering the result and do the cheap comparisons without the virtual method call overhead multiple times. Picking this as an example, there are also the other places that might need to call virtual functions multiple times.

```c++
  for (const auto &current : nodes) {
     const PredictionContextType contextType = current->getContextType();
      if (contextType == PredictionContextType::SINGLETON || contextType == PredictionContextType::EMPTY) {
            //...
            if (contextType == PredictionContextType::EMPTY) {
              // ...
            }
       }
  }
```

runtime/Cpp/runtime/src/atn/ATNStateType.h
```diff
@@ -0,0 +1,36 @@
...
+    TOKEN_START = 6,
+    RULE_STOP = 7,
+    BLOCK_END = 8,
+    STAR_LOOP_BACK = 9,
+    STAR_LOOP_ENTRY = 10,
+    PLUS_LOOP_BACK = 11,
+    LOOP_END = 12,
+  };
+
+  std::string atnStateTypeName(ATNStateType atnStateType);
```
hzeller, line 33: Not necessarily now (not sure what the current baseline requirements are, is antlr already compiling with c++17?), but longer term, using `std::string_view` here is probably a good idea.

runtime/Cpp/runtime/src/atn/SingletonPredictionContext.h
```diff
@@ -25,6 +25,8 @@ namespace atn {
 
     static Ref<const SingletonPredictionContext> create(Ref<const PredictionContext> parent, size_t returnState);
 
+    PredictionContextType getContextType() const override;
```
hzeller, line 28: For all classes that are not extended anymore, it would probably be good to use `final` instead of `override`.
Not only is this a good documentation for the reader of the code, but the compiler can also use that for devirtualization if it otherwise can know which class it is dealing with.

jcking: Unfortunately SingletonPredictionContext is inherited by EmptyPredictionContext, so it cannot be final. However, whether or not we need PredictionContextType::EMPTY I will have to look.

runtime/Cpp/runtime/src/atn/ATNStateType.h
jcking: We could use string_view, but then we cannot ever dynamically allocate. For example, if we wanted to do "INVALID(%d)".

runtime/Cpp/runtime/src/atn/PredictionContext.cpp
jcking: Fixed.

runtime/Cpp/runtime/src/atn/SingletonPredictionContext.cpp
jcking: Fixed, we do not now. I removed EmptyPredictionContext as its not useful and can be implemented purely in terms of SIngletonPredictionContext. It also avoids people instantiating EmptyPredictionContext, since some code uses pointer equality to the PredictionContext::EMPTY.

runtime/Cpp/runtime/src/atn/SingletonPredictionContext.h
jcking: I was able to remove EmptyPredictionContext and make SingletonPredictionContext final. However it won't matter much, as most invocations use the base class rather than the derived. So the compiler cannot do much.

hzeller: I usually find it most useful as documentation for the reader of the code, actually.

runtime/Cpp/runtime/src/atn/ATNStateType.h
hzeller: Given that `ATNStateType` is an enum class it would be very hard to get an invalid value in there. So I'd prefer performance over new string allocations to be prepared for an unlikely force-typecast any time.

runtime/Cpp/runtime/src/atn/ATNStateType.cpp
```diff
@@ -0,0 +1,32 @@
...
+      return "BLOCK_END";
+    case ATNStateType::STAR_LOOP_BACK:
+      return "STAR_LOOP_BACK";
+    case ATNStateType::STAR_LOOP_ENTRY:
+      return "STAR_LOOP_ENTRY";
+    case ATNStateType::PLUS_LOOP_BACK:
+      return "PLUS_LOOP_BACK";
+    case ATNStateType::LOOP_END:
+      return "LOOP_END";
+    default:
```
hzeller, line 29: I suggest not having a default case. That way, the compiler automatically will warn when there is a new enum value that is not handled. In general whenever there are default cases in switch/case statements that are supposed to handle all cases of an enum, we should consider removing them to have the compiler help us check for correctness.

So I suggest to move the `return "Invalid";` to the end, outside the switch/case.

runtime/Cpp/runtime/src/atn/TransitionType.cpp
```diff
@@ -0,0 +1,28 @@
...
+      return "ACTION";
+    case TransitionType::SET:
+      return "SET";
+    case TransitionType::NOT_SET:
+      return "NOT_SET";
+    case TransitionType::WILDCARD:
+      return "WILDCARD";
+    case TransitionType::PRECEDENCE:
+      return "PRECEDENCE";
+    default:
```
hzeller, line 25: Similar suggestion here with the default case: Don't have a default case here, but move the `return "INVALID";` to the end of the function.

runtime/Cpp/runtime/src/atn/ATNStateType.h
jcking: atnStateTypeName and transitionTypeName are only used by toString methods, which are not called anywhere in the ANTLR runtime except for debugging, which is off by default. So performance is not an issue in this case, unless the user is calling toString in which cause I would suggest not unless debugging because a bunch of them are not performant.

mike-lischke (approved):
From my standpoint the patch is fine, but @hzeller has a few good points here, which might be worth to be considered.

runtime/Cpp/runtime/src/atn/TransitionType.cpp
jcking: Ugh, didn't public comment. I moved it to the end, but used UNKNOWN. ATNStateType uses INVALID.

runtime/Cpp/runtime/src/atn/ATNStateType.cpp
jcking: Same as other, fixed.

jcking:
> From my standpoint the patch is fine, but @hzeller has a few good points here, which might be worth to be considered.

Addressed the switch cases. @parrt we should be good assuming tests for C++ are successful.

parrt:
Ok, great! I'll address the php failing issue later.

Issue #3577, opened 2022-03-10
[C++] Mingw. Link errors in Release configuration.


ArsenyVelichko:
I tried to build my C++ application together with the corresponding ANTLR-runtime library under mingw 9.0 and encountered the following link errors in the Release configuration:
```
C:/Qt/Tools/mingw900_64/bin/../lib/gcc/x86_64-w64-mingw32/11.2.0/../../../../x86_64-w64-mingw32/bin/ld.exe: 
CMakeFiles/RailDSL.dir/antlr/MetaGrammar/MetaGrammarLexer.cpp.obj:MetaGrammarLexer.cpp:(.text+0x714):
undefined reference to `antlr4::atn::LexerATNSimulator::SimState::~SimState()'
C:/Qt/Tools/mingw900_64/bin/../lib/gcc/x86_64-w64-mingw32/11.2.0/../../../../x86_64-w64-mingw32/bin/ld.exe:
CMakeFiles/RailDSL.dir/antlr/MetaGrammar/MetaGrammarLexer.cpp.obj:MetaGrammarLexer.cpp:
(.text$_ZN6antlr43atn17LexerATNSimulatorD0Ev[_ZN6antlr43atn17LexerATNSimulatorD0Ev]+0x1c): 
undefined reference to `antlr4::atn::LexerATNSimulator::SimState::~SimState()'
C:/Qt/Tools/mingw900_64/bin/../lib/gcc/x86_64-w64-mingw32/11.2.0/../../../../x86_64-w64-mingw32/bin/ld.exe: 
CMakeFiles/RailDSL.dir/antlr/MetaGrammar/MetaGrammarParser.cpp.obj:MetaGrammarParser.cpp:(.text+0x3e56):
undefined reference to `antlr4::atn::LexerATNSimulator::SimState::~SimState()'
C:/Qt/Tools/mingw900_64/bin/../lib/gcc/x86_64-w64-mingw32/11.2.0/../../../../x86_64-w64-mingw32/bin/ld.exe: 
CMakeFiles/RailDSL.dir/antlr/MetaGrammar/MetaGrammarParser.cpp.obj:MetaGrammarParser.cpp:(.text+0x3ed6):
undefined reference to `antlr4::atn::LexerATNSimulator::SimState::~SimState()'
```
Referring to the LexerATNSimulator.h header file, I noticed that the ```ANTLR4CPP_PUBLIC``` macro was missing from the nested class.
[...]

PR #3588, opened 2022-03-16
[C++] Add ANTLR4CPP_PUBLIC attributes to various symbols

jcking:
As the name suggests, add `ANTLR4CPP_PUBLIC` to various symbols. I additionally cleaned up `SimState` while I was there, as it cannot be virtual since it is used by value as a member. So using a struct is more simple.

Addresses #3577


jcking:
@mike-lischke

PR #3593, opened 2022-03-22
Use linux DCO not our old contributors certificate of origin

parrt:
See the discussion here: https://github.com/antlr/antlr4/discussions/3589


parrt:
Note that I have changed the PR template so that it notifies people that the Linux distribution certificate of origin is in effect.  No need to actually sign something.

parrt:
oops. had in dev not a branch

PR #3594, opened 2022-03-22
[C++] Refactor and optimize SemanticContext, fix bug in filterPrecedencePredicates

jcking:
Cleanup `SemanticContext` and remove use of `dynamic_cast` related to it. Additionally this fixes 2 flaws in `filterPrecedencePredicates` that @parrt and I found.

- While filtering precedence predicates Java removes the precedence predicates from the set while C++ does not. This change fixes this by filtering out precedence predicates during iteration and adding the min or max after.
- Equality of and/or is undefined currently, as `std::unordered_set` makes no guarantees on iteration order. So two identical sets could iterate differently, resulting in two separate and/or semantic contexts being different. This change fixes this by preserving insertion order, making it well defined.


jcking:
@mike-lischke Continuation of last pull request. This one is just cleanup and optimizations.

runtime/Cpp/runtime/src/atn/SemanticContext.cpp
```diff
@@ -51,43 +51,79 @@ namespace {
...
+          precedencePredicate = std::static_pointer_cast<const SemanticContext::PrecedencePredicate>(std::move(semanticContext));
+        }
+      } else {
+        auto [existing, inserted] = operandSet.insert(semanticContext.get());
+        if (inserted) {
+          operandList.push_back(std::move(semanticContext));
+        }
+      }
+    }
+  }
```
mike-lischke, line 72: You must have an `insertSemanticContext` function already, but you added another one. Why's that?

jcking: This one accepts a rvalue reference `Ref<const SemanticContext>`, while the other takes a lvalue reference. This avoids an extra reference/unreference in `shared_ptr` under certain circumstances.

mike-lischke: I'm afraid of code duplication, but ok.

jcking:
Oh, I hit request review at the same time you approved. lol Ignore that.

@parrt

PR #3595, opened 2022-03-22
[C++] Fix bugs in SemanticContext

jcking:
This fixes 2 flaws in filterPrecedencePredicates that @parrt and I found in the C++ implementation.

- While filtering precedence predicates Java removes the precedence predicates from the set while C++ does not. This change fixes this by filtering out precedence predicates during iteration and adding the min or max after.
- Equality of and/or is undefined currently, as std::unordered_set makes no guarantees on iteration order. So two identical sets could iterate differently, resulting in two separate and/or semantic contexts being different when they should be equal. This change fixes this by preserving insertion order, making it well defined.

### DIFF
--- a/runtime/Cpp/runtime/src/atn/SemanticContext.cpp
+++ b/runtime/Cpp/runtime/src/atn/SemanticContext.cpp
@@ -3,7 +3,11 @@
  * can be found in the LICENSE.txt file in the project root.
  */
 
+#include <functional>
+#include <unordered_set>
+
 #include "misc/MurmurHash.h"
+#include "support/Casts.h"
 #include "support/CPPUtils.h"
 #include "support/Arrays.h"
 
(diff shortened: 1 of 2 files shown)


jcking:
@parrt See if you can get a grammar that tests the fixed behavior so we know what other runtimes are affected.

@mike-lischke

runtime/Cpp/runtime/src/atn/SemanticContext.cpp
```diff
@@ -13,6 +17,41 @@ using namespace antlr4;
...
+  };
+
+  template <typename Comparer>
+  void insertSemanticContext(const Ref<const SemanticContext> &semanticContext,
+                             std::unordered_set<const SemanticContext*, SemanticContextHasher, SemanticContextComparer> &operandSet,
+                             std::vector<Ref<const SemanticContext>> &operandList,
+                             Ref<const SemanticContext::PrecedencePredicate> &precedencePredicate) {
+    if (semanticContext != nullptr) {
+      if (is<const SemanticContext::PrecedencePredicate>(semanticContext)) {
+        if (precedencePredicate == nullptr || (Comparer{})(downCast<const SemanticContext::PrecedencePredicate*>(semanticContext.get())->precedence, precedencePredicate->precedence)) {
```
mike-lischke, line 41: Two typecasts? Can you down cast to the Comparer directly? Also it seems odd that you use a C-style cast here, which we are usually try to avoid.

runtime/Cpp/runtime/src/atn/SemanticContext.cpp
```diff
@@ -208,35 +241,31 @@ std::string SemanticContext::AND::toString() const {
...
+      insertSemanticContext<std::greater<int>>(operand, operands, opnds, precedencePredicate);
     }
   } else {
-    operands.insert(b);
+    insertSemanticContext<std::greater<int>>(b, operands, opnds, precedencePredicate);
   }
 
-  std::vector<Ref<const PrecedencePredicate>> precedencePredicates = filterPrecedencePredicates(operands);
-  if (!precedencePredicates.empty()) {
-    // interested in the transition with the highest precedence
```
mike-lischke, line 231: Please keep this comment. There's not so much description in the code, so every line counts :-)

runtime/Cpp/runtime/src/atn/SemanticContext.cpp
jcking: This was template stuff. I switched it so the comparer is passed via a parameter and no more C style looking cast, its just a operator() call now.

jcking: its comparing the precedence value, int.

runtime/Cpp/runtime/src/atn/SemanticContext.cpp
jcking: Added back.

mike-lischke (approved):

jcking:
@parrt Good to go.

PR #3596, opened 2022-03-23
[Java] Used LinkedHashSet to preserve insertion order in SemanticContext

jcking:
Currently SemanticContext equality and hashing assume a well defined iteration order for `HashSet`. `HashSet` does not actually guarantee this, so we need to switch to `LinkedHashSet` which guarantees insertion order.


jcking:
@parrt

parrt:
I think I'm going to need time to look at this in more detail after 4.10.  should be okay but you never know...

PR #3597, opened 2022-03-25
[Go] Do not export Array2DHashSet which is an implementation detail

jcking:
`Array2DHashSet` is an implementation detail that is not directly exposed by any other interfaces or structures. It should be considered an implementation detail and not exported.


jcking:
@parrt

PR #3599, opened 2022-03-27
Prep 4.10

parrt:


parrt:
Hiya. I’d really like to use the next three days to get the release out of 4.10. I want to incorporate it into the plug-in for IntelliJ which has a number of bugs that I want to fix and get out the door. There’s plenty of time to fix more stuff in the future but let me know what’s critical to get in this release. @ericvergnaud pointed out that backward compatibility could be important for some targets, particularly those without a build system like a JavaScript and python. JavaScript in particular is vulnerable because there is a massive chain of dependencies that occur. He has incorporated backward compatibility for the previous version of the ATN. I’m open to backward compatibility for other targets if you think it’s important, otherwise I’m simply going to declare that people must regenerate source from the grammars.

parrt:
github actions will be in maintenance soon today they claim...but is dead now already. haha.

parrt:
I'm mostly ready to go!!! waiting for CI to finish.

parrt:
Ok, I'm moving all of my prep work into dev folks. we are really close. just got to get final testing going.

PR #3602, opened 2022-03-28
[Swift] use stdlib (single pass)

Dante-Broggi:
The first commit requires a minimum Swift version of Swift 5

### DIFF
--- a/runtime/Swift/Sources/Antlr4/misc/BitSet.swift
+++ b/runtime/Swift/Sources/Antlr4/misc/BitSet.swift
@@ -633,7 +633,7 @@ public class BitSet: Hashable, CustomStringConvertible {
 
         while true {
             if word != 0 {
-                let bit = (u * BitSet.BITS_PER_WORD) + BitSet.numberOfTrailingZeros(word)
+                let bit = (u * BitSet.BITS_PER_WORD) + word.trailingZeroBitCount
                 return bit
             }
             u += 1
@@ -642,10 +642,6 @@ public class BitSet: Hashable, CustomStringConvertible {
             }
             word = words[u]
         }
-    }
-
-    public static func numberOfTrailingZeros(_ i: Int64) -> Int {
-        return i.trailingZeroBitCount
     }
 
     /// 
@@ -674,7 +670,7 @@ public class BitSet: Hashable, CustomStringConvertible {
 
         while true {
             if word != 0 {
-                return (u * BitSet.BITS_PER_WORD) + BitSet.numberOfTrailingZeros(word)
+                return (u * BitSet.BITS_PER_WORD) + word.trailingZeroBitCount
             }
             u += 1
             if u == wordsInUse {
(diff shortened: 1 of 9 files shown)


runtime/Swift/Sources/Antlr4/misc/BitSet.swift
```diff
@@ -645,41 +645,7 @@ public class BitSet: Hashable, CustomStringConvertible {
...
-            n = n - 4
-            x = y
-        }
-        y = x << 2
-        if y != 0 {
-            n = n - 2
-            x = y
-        }
-        return Int(n - ((x << 1) >>> 31))
+        return i.trailingZeroBitCount
```
KvanTTT, line 648: Is containing function required at all? `numberOfTrailingZeros` can be replaced with `trailingZeroBitCount`

runtime/Swift/Sources/Antlr4/misc/BitSet.swift
```diff
@@ -814,35 +780,7 @@ public class BitSet: Hashable, CustomStringConvertible {
...
-            x <<= 4
-        }
-        if x >>> 30 == 0 {
-            n += 2
-            x <<= 2
-        }
-        n -= x >>> 31
-
-        return Int(n)
+        return i.leadingZeroBitCount
```
KvanTTT, line 783: `numberOfLeadingZeros` -> `leadingZeroBitCount` for all usings.

runtime/Swift/Sources/Antlr4/misc/BitSet.swift
```diff
@@ -903,16 +841,7 @@ public class BitSet: Hashable, CustomStringConvertible {
...
-        // HD, Figure 5-14
-        i = i - ((i >>> 1) & 0x5555555555555555)
-        i = (i & 0x3333333333333333) + ((i >>> 2) & 0x3333333333333333)
-        i = (i + (i >>> 4)) & 0x0f0f0f0f0f0f0f0f
-        i = i + (i >>> 8)
-        i = i + (i >>> 16)
-        i = i + (i >>> 32)
-
-        return Int(i) & 0x7f
+        return i.nonzeroBitCount
```
KvanTTT, line 844: `bitCount` -> `nonzeroBitCount` for all usings.

runtime/Swift/Sources/Antlr4/misc/BitSet.swift
Dante-Broggi: I have not yet changed any public API (which these functions are) but as you are requesting it, I will inline these and remove them.

KvanTTT: Anyway we've changed a lot of public API in 4.10 and removed outdated function. Awaiting @parrt for the final decision.

parrt:
hi. yeah, maybe we can revisit after 4.10? changed the ATNDeserialization a lot!

runtime/Swift/Sources/Antlr4/misc/BitSet.swift
parrt: What duplications are we talking about? Sorry I'm trying to review lots of different things.

KvanTTT: Nevermind, @Dante-Broggi has already removed useless public API functions.

PR #3603, opened 2022-03-28
[Go] Fix symbol collision in generated lexers and parsers

jcking:
If a lowercase grammar name is used, the functions will collide. This change ensures they do not by enforcing capitalization when necessary.


jcking:
@parrt

PR #3608, opened 2022-03-29
[C++] Refactor PredictionContext and yet more performance improvements

jcking:
Refactor PredictionContext, remove some `dynamic_cast`, and avoid copying `std::shared_ptr` when possible

### DIFF
--- a/runtime/Cpp/runtime/src/atn/ATNSimulator.cpp
+++ b/runtime/Cpp/runtime/src/atn/ATNSimulator.cpp
@@ -23,12 +23,11 @@ void ATNSimulator::clearDFA() {
   throw UnsupportedOperationException("This ATN simulator does not support clearing the DFA.");
 }
 
-PredictionContextCache& ATNSimulator::getSharedContextCache() {
+PredictionContextCache& ATNSimulator::getSharedContextCache() const {
   return _sharedContextCache;
 }
 
-Ref<const PredictionContext> ATNSimulator::getCachedContext(Ref<const PredictionContext> const& context) {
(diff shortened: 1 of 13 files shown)


hzeller:
Now that #3609 is merged, this needs a rebase...

jcking:
> Now that #3609 is merged, this needs a rebase...

Yep, done.

jcking:
@parrt This one is good.

jcking:
@parrt Meant the other one. Let's leave this until I rebase.

runtime/Cpp/runtime/src/atn/ATNSimulator.cpp
```diff
@@ -23,12 +23,11 @@ void ATNSimulator::clearDFA() {
   throw UnsupportedOperationException("This ATN simulator does not support clearing the DFA.");
 }
 
-PredictionContextCache& ATNSimulator::getSharedContextCache() {
+PredictionContextCache& ATNSimulator::getSharedContextCache() const {
```
hzeller, line 26: Usually it is a good idea to _not_ make a method const if it returns a non-const reference, i.e. allows the caller to change the internal state of the object by accessing the writable reference to an internal member ( https://google.github.io/styleguide/cppguide.html#Use_of_const ).

Probably a conscious choice, but just want to caution.

jcking: In this case, the PredictionContextCache is not actually owned by the ATNSimulator, its given a reference to it during construction.

PR #3609, opened 2022-03-29
[C++] Devirtualize hand rolled RTTI for performance

jcking:


runtime/Cpp/runtime/src/atn/PredictionContext.h
```diff
@@ -65,13 +65,10 @@ namespace atn {
...
     const size_t cachedHashCode;
 
-  protected:
-    explicit PredictionContext(size_t cachedHashCode);
-
   public:
     virtual ~PredictionContext() = default;
 
-    virtual PredictionContextType getContextType() const = 0;
+    PredictionContextType getContextType() const;
```
hzeller, line 71: The implementation seems to be missing, resulting in a linking error.

hzeller:
We migh talso need to tackle the `SemanticContext` hierarchy.

hzeller:
> We migh talso need to tackle the `SemanticContext` hierarchy.

ah, I misread, it is included.

runtime/Cpp/runtime/src/atn/ATNState.h
```diff
@@ -114,14 +114,20 @@ namespace atn {
...
+  protected:
+    explicit ATNState(ATNStateType stateType) : _stateType(stateType) {}
 
   private:
     /// Used to cache lookahead during parsing, not used during construction.
 
     misc::IntervalSet _nextTokenWithinRule;
     std::atomic<bool> _nextTokenUpdated { false };
 
+    ATNStateType _stateType;
```
hzeller, line 129: These are only set once in the constructor, so can be made `const`.
The advantage is that then the compiler will also make sure e.g. that there is no default constructor or assignment operators, so we make sure that there is no accidental expensive copies generated (so, for instance, ` ATNState() = delete;` would not be needed, as there can be no default constructor generated as there is at least one const member)

This is true for all the changes added in this PR, that all can be made const: `ATNStateType _stateType`, `LexerActionType actionType`, `PredictionContextType _contextType`, `SemanticContextType _contextType`, `TransitionType _transitionType` and `ParseTreeType _treeType`

runtime/Cpp/runtime/src/atn/PredictionContext.h
jcking: Woops. Fixed.

runtime/Cpp/runtime/src/atn/ATNState.h
jcking: PredictionContextType cannot be, because ArrayPredictionContext needs to be moveable in another pull request. The rest are now const though.

hzeller:
Nice. Testing with Surelog showed a ~1.5% improvement (which means the improvment is likely much better than that, as this is measuring end2end of Surelog's regression tests of which parsing is only a part).

parrt:
thanks guys!!!

PR #3610, opened 2022-03-30
Avoid dynamic_cast<> if we already know the type

hzeller:
In that case, downCast() is sufficient.


runtime/Cpp/runtime/src/ParserRuleContext.h
```diff
@@ -1,11 +1,12 @@
-﻿/* Copyright (c) 2012-2017 The ANTLR Project. All rights reserved.
+/* Copyright (c) 2012-2017 The ANTLR Project. All rights reserved.
```
hzeller, line 1: FYI: This is the superfluous and Byte Order Mark, that was removed while editing. This is UTF-8, not UTF-16.

PR #3612, opened 2022-03-30
[C++] Add T::is for type hierarchy checks and remove some dynamic_cast

jcking:
We removed a lot of usage of RTTI in favor of a hand rolled RTTI, however a downside is it is more error prone to check for non base classes. For example, checking if an `ATNState` is an instance of `DecisionState`. This pull request adds a `is` static method to each class in the hierarchy to check whether it is an instance of it by inspecting the appropriate enum states.

It additionally gets rid of some more `dynamic_cast`.


runtime/Cpp/runtime/src/atn/DecisionState.h
```diff
@@ -12,6 +12,18 @@ namespace atn {
 
   class ANTLR4CPP_PUBLIC DecisionState : public ATNState {
   public:
+    static bool is(const ATNState &atnState) {
+      const auto stateType = atnState.getStateType();
+      return stateType == ATNStateType::BLOCK_START ||
```
hzeller, line 17: mmh, so with these many terms comparing with the proposed bitset might be cheaper.

hzeller (commented):
Nice, this will certainly improve readability (and performance for the places with removed dynamic casts)

Now, we can make the `getXYZType()` functions in the base class protected again, as we only expect the static functions using it.

runtime/Cpp/runtime/src/atn/DecisionState.h
jcking: The set of enums are technically 2 contiguous spans. So we can shift this to 4 comparisions.

runtime/Cpp/runtime/src/atn/ATNDeserializer.cpp
```diff
@@ -578,52 +578,52 @@ void ATNDeserializer::verifyATN(const ATN &atn) const {
...
+      checkCondition(StarLoopEntryState::is(state->transitions[0]->target));
     }
 
-    if (is<LoopEndState*>(state)) {
+    if (LoopEndState::is(state)) {
       checkCondition((downCast<LoopEndState*>(state))->loopBackState != nullptr);
     }
 
-    if (is<RuleStartState*>(state)) {
+    if (RuleStartState::is(state)) {
```
hzeller, line 610: This one doesn't seem to be defined.

runtime/Cpp/runtime/src/atn/DecisionState.h
hzeller: but wouldn't a bit-set be easier to read anyway ?

runtime/Cpp/runtime/src/atn/ATNDeserializer.cpp
jcking: Missed that one. Done.

runtime/Cpp/runtime/src/atn/DecisionState.h
jcking: We also have requirements on some of the type enums being the same as the serialization values. So using a bitset won't really work unless we implement yet another mapping. Keeping the base classes out of the enum is preferred. Now DecisionState::is and BlockStartState::is are smaller, as it takes advantage of some of the enums being continuous.

runtime/Cpp/runtime/src/atn/SetTransition.cpp
```diff
@@ -11,8 +11,8 @@
 using namespace antlr4;
 using namespace antlr4::atn;
 
-SetTransition::SetTransition(TransitionType transitionType, ATNState *target, misc::IntervalSet aSet)
-  : Transition(transitionType, target), set(aSet.isEmpty() ? misc::IntervalSet::of(Token::INVALID_TYPE) : std::move(aSet)) {
+SetTransition::SetTransition(ATNState *target, misc::IntervalSet aSet)
```
hzeller, line 14: The `NotSetTransition` constructor still calls the constructor  with the `TransitionType` parameter which is now gone.

jcking: That was an unrelated change that I thought I undid, apparently I missed that bit. Fixed.

jcking:
@parrt Assuming the MacOS runns pick this up eventually and pass, this one is good.

jcking:
@parrt Looks like macOS passed, woot.

PR #3613, opened 2022-03-31
[C++] Avoid copying statically generated serialized ATNs

jcking:
Generated parsers and lexers have a serialized ATN embeded with them as a static array. This data is read only and loaded as part of the executable or library. Currently we copy this into an `std::vector`. Previously C++ serialized ATNs as multiple segments, but this is no longer the case. Now there is no reason for us to copy the serialized ATN. This pull request introduces `SerializedATN` and `SerializedATNView` which are analygous to `std::string` and `std::string_view`.


runtime/Cpp/runtime/src/tree/xpath/XPathLexer.cpp
```diff
@@ -61,7 +61,7 @@ void xpathLexerInitialize() {
       "STRING"
     }
   );
-  static const int32_t serializedATNSegment0[] = {
+  static const int32_t serializedATNSegment[] = {
```
hzeller, line 64: These could even be `static constexpr`. Probably sizeof() below will be smart enough to already know this at compile-time even with `const`, but with `constexpr` it doesn't have an excuse not to know.

hzeller (commented):
Could the view just be a `std::span<int32_t>` ?

Without knowing much detail  - the whole initialization and explicit use of mutex looks potentially error prone, and it looks like they are used for a global single instance ?
Could some of these static initializations done with a singleton way of initializing things ?

```
static TheThing CreateTheThing() {...}

const TheThing &GetGlobalThing() {
   static const TheThing static_thing = CreateTheThing();
   return static_thing;
}
```

The code generated will make sure that the thing-creation only happens onece.

jcking:
> Could the view just be a `std::span<int32_t>` ?
> 
> Without knowing much detail - the whole initialization and explicit use of mutex looks potentially error prone, and it looks like they are used for a global single instance ? Could some of these static initializations done with a singleton way of initializing things ?
> 
> ```
> static TheThing CreateTheThing() {...}
> 
> const TheThing &GetGlobalThing() {
>    static const TheThing static_thing = CreateTheThing();
>    return static_thing;
> }
> ```
> 
> The code generated will make sure that the thing-creation only happens onece.

Its std::once_flag, not mutex.

std::span is C++20, ANTLR is currently at C++17 with no intention to go to C++20 anytime soon. So we are replicating std::span basically, but trimmed down.

jcking:
@parrt Relatively isolated improvement. Should be good.

PR #3620, opened 2022-04-03
No longer necessary too generate sets or single atom transit that are bigger than 16bits.

parrt:


runtime/Java/src/org/antlr/v4/runtime/atn/CodePointTransitions.java
```diff
@@ -6,40 +6,25 @@
...
  *
- * To keep the serialized ATN size small, we only inline atom and
- * range transitions for Unicode code points <= U+FFFF.
- *
- * Whenever we encounter a Unicode code point > U+FFFF, we represent that
- * as a set transition (even if it is logically an atom or a range).
+ * Previously, we distinguished between atom and range transitions for
+ * Unicode code points <= U+FFFF and those above. We used a set
+ * transition for a Unicode code point > U+FFFF. Now that we can serialize
+ * 32-bit int/chars in the ATN serialization, this is no longer necessary.
```
KvanTTT, line 16: I don't think it's a useful comment. Git keeps history well.

parrt: Much easier to read that comment then it is to know that I'm supposed to dig through a GitHub log sequence ;)

PR #3623, opened 2022-04-04
[Swift] An interval ought to be a value

Dante-Broggi:
Interval was a pointer to 2 Ints
it ought to be just 2 Ints, which is smaller and more semantically correct,
with no need for a cache.

However, this breaks AnyObject conformance and changes type metadata but people shouldn't be relying on those for an Interval.


KvanTTT:
Interesting. I'm wondering how it affects performance.

PR #3624, opened 2022-04-04
Allow to switch off building C++ tests, needed if used as sub-project

hzeller:
The new option to cmake ANTLR_BUILD_CPP_TESTS is default
on (so the behavior is as before), but it provides a way to
switch off if not needed.

The C++ tests pull in an external dependency (googletests),
which might conflict if ANTLR is used as a subproject in
another cmake project.

Signed-off-by: Henner Zeller <h.zeller@acm.org>


hzeller:
Needed for 4.10

parrt:
Sounds good to me. Let's see what @mike-lischke thinks.

mike-lischke (approved):

parrt:
Thanks!

PR #3627, opened 2022-04-05
[C++] Implement configurable PredictionContextMergeCache

jcking:


jcking:
Assuming tests pass, this should be good. @parrt 

I removed the weight feature from the LRU and shifted the clear counter to the parser atn simulator so that clear actually clears.

PR #3630, opened 2022-04-06
[C++] Free ATNConfig lookup set in readonly ATNConfigSet

jcking:
Currently the underlying storage lives on as only `clear` is used. To actually free the underlying storage we have to perform a swap. The original intent was to free the storage when the config becomes readonly as its going to live on for a long time in the DFA cache.


jcking:
@parrt Assuming tests pass, this should be good. Avoids keeping the underlying storage of the set around after making it readonly.

PR #3632, opened 2022-04-08
Fix NPE for undefined label, fix antlr#2788

KvanTTT:
The fixed version of https://github.com/antlr/antlr4/pull/3564

### DIFF
--- a/tool-testsuite/test/org/antlr/v4/test/tool/TestSymbolIssues.java
+++ b/tool-testsuite/test/org/antlr/v4/test/tool/TestSymbolIssues.java
@@ -542,4 +542,22 @@ public class TestSymbolIssues extends BaseJavaToolTest {
 
 		testErrors(test, false);
 	}
+
+	// ISSUE: https://github.com/antlr/antlr4/issues/2788
+	@Test public void testUndefinedLabel() {
+		String[] test = {
+				"grammar Test;" +
+				"root\n" +
+				"    : root a\n" +
+				"    | b [error]\n" +
+				"    ;\n" +
+				"\n" +
+				"a: 'a';\n" +
+				"b: 'b';",
+
+				"error(" + ErrorType.INTERNAL_ERROR.code + "): Test.g4:2:30: internal error: Rule error undefined \n"
+		};
+
+		testErrors(test, false);
+	}
 }
--- a/tool/src/org/antlr/v4/automata/ParserATNFactory.java
+++ b/tool/src/org/antlr/v4/automata/ParserATNFactory.java
@@ -472,12 +472,21 @@ public class ParserATNFactory implements ATNFactory {
             if ( el.left.getNumberOfTransitions()==1 ) tr = el.left.transition(0);
             boolean isRuleTrans = tr instanceof RuleTransition;
             if ( el.left.getStateType() == ATNState.BASIC &&
+				el.right != null &&
 				el.right.getStateType()== ATNState.BASIC &&
-				tr!=null && (isRuleTrans && ((RuleTransition)tr).followState == el.right || tr.target == el.right) )
-			{
+				tr!=null && (isRuleTrans && ((RuleTransition)tr).followState == el.right || tr.target == el.right) ) {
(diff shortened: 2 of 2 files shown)


tool/src/org/antlr/v4/automata/ParserATNFactory.java
```diff
@@ -472,24 +472,28 @@ public Handle elemList(List<Handle> els) {
...
 			}
 		}
 		Handle first = els.get(0);
-		Handle last = els.get(n -1);
-		if ( first==null || last==null ) {
-			g.tool.errMgr.toolError(ErrorType.INTERNAL_ERROR, "element list has first|last == null");
-		}
-		return new Handle(first.left, last.right);
+		Handle last = els.get(n - 1);
+		return new Handle(first != null ? first.left : null, last != null ? last.right : null);
```
parrt, line 496: Seems dangerous to remove an error message

tool/src/org/antlr/v4/automata/ParserATNFactory.java
```diff
@@ -472,24 +472,28 @@ public Handle elemList(List<Handle> els) {
...
+				el.right != null &&
 				el.right.getStateType()== ATNState.BASIC &&
-				tr!=null && (isRuleTrans && ((RuleTransition)tr).followState == el.right || tr.target == el.right) )
-			{
+				tr!=null && (isRuleTrans && ((RuleTransition)tr).followState == el.right || tr.target == el.right) ) {
 				// we can avoid epsilon edge to next el
-				if ( isRuleTrans ) ((RuleTransition)tr).followState = els.get(i+1).left;
-                else tr.target = els.get(i+1).left;
+				Handle handle = null;
+				if (i + 1 < els.size())
```
parrt, line 480: can you wrap if blocks in {...} even for a single statement please? thanks!

tool/src/org/antlr/v4/automata/ParserATNFactory.java
parrt, line 496: also I don't like all of these nested `?:`.  they are hard to read

KvanTTT: If I restore the check, two errors will occur:

```
error(20): Test.g4:2:30: internal error: Rule error undefined 
error(20):  internal error: element list has first|last == null 
```

The message of internal error `element list has first|last == null` is duplicated, unclear for the end-user, and does not help to resolve an issue. The first message is much more informative.

KvanTTT: Done, I've forced pushed the commit.

KvanTTT:
`Error: There is not enough space on the disk.` on GitHub CI 😞

parrt:
looks great. thanks!!!

parrt:
> [`Error: There is not enough space on the disk.` on GitHub CI](https://github.com/antlr/antlr4/runs/5890009729?check_suite_focus=true) 😞

booo!!!

tool-testsuite/test/org/antlr/v4/test/tool/TestSymbolIssues.java
```diff
@@ -542,4 +542,22 @@ public void testLabelsForTokensWithMixedTypesLRWithoutLabels() {
...
+				"    ;\n" +
+				"\n" +
+				"a: 'a';\n" +
+				"b: 'b';",
+
+				"error(" + ErrorType.INTERNAL_ERROR.code + "): Test.g4:2:30: internal error: Rule error undefined \n"
+		};
+
+		testErrors(test, false);
+	}
```
mike-lischke, line 562: I know this is already more than 2 years old, but I'm currently on porting this test and I wonder how it can even work. Neither the old ANTLRParser.g4, nor the new ANTLRv4Parser.g4 grammar allow an arg action block after a rule ref (with or w/o element options). Consequently I get a syntax error for that grammar code. Yet the test seems to indicate the grammar is valid and only errors out because a non-existing rule is referenced. Can someone please explain, how that's possible?

PR #3635, opened 2022-04-08
[C++] Remove more dynamic_cast usage

jcking:
<!--


jcking:
@parrt This is good as well. Last minute removal of more dynamic cast.

PR #3637, opened 2022-04-09
add license prefix

ericvergnaud:


KvanTTT:
Maybe remove such headers from all files? Why is such duplication required? Also, we are constantly forgetting to update the year in all such headers, there are a lot of files with 2017 year.

parrt:
I think all files need the copyright protection a lawyer told me some years ago.

KvanTTT:
Maybe at least to remove end year? It's tedious to update it every time.

parrt:
Agreed that it is tedious. Unfortunately, I don't know what the law says about whether we need a date on there...

PR #3639, opened 2022-04-09
attempt to fix php build on circle-ci

ericvergnaud:


KvanTTT:
Not actual for now, should be closed.

PR #3655, opened 2022-04-14
Update publishing instruction for Dart

lingyv-li:
Dart `pub` tool got integrated in the `dart` CLI now. Updating the instruction.


lingyv-li:
Fixes #3654

Issue #3651, opened 2022-04-13
[Cpp Runtime] MSVC: Release 4.10 fails to compile. Missing definition for antlrcpp::s2ws


amykyta3:
Runtime source does not provide any definition for `antlrcpp::s2ws`:

```
antlr4-cpp-runtime\ANTLRFileStream.cpp(17): error C2653: 'antlrcpp': is not a class or namespace name
antlr4-cpp-runtime\ANTLRFileStream.cpp(17): error C3861: 's2ws': identifier not found
```
From my attempt to upgrade to 4.10: https://github.com/SystemRDL/systemrdl-compiler/runs/6001839774

Grepping the contents of runtime shows this is missing:
```
$ grep -r s2ws antlr4-cpp-runtime-4.10-source/
antlr4-cpp-runtime-4.10-source/runtime/src/ANTLRFileStream.cpp:  std::ifstream stream(antlrcpp::s2ws(fileName), std::ios::binary);
antlr4-cpp-runtime-4.10-source/demo/Windows/antlr4-cpp-demo/main.cpp:  std::wstring s = antlrcpp::s2ws(tree->toStringTree(&parser)) + L"\n";
```
as does searching the contents of the antlr repository: https://github.com/antlr/antlr4/search?q=s2ws

amykyta3:
Possibly related: #3467

PR #3656, opened 2022-04-14
[C++] Remove reference to antlrcpp:s2ws

### LINKED ISSUE #3467: [Cpp Runtime] MSVC: ANTLRFileStream uses removed "s2ws"
Hello,

master branch of runtime cpp fails to compile with MSVC on the following line:

image

thanks for your support,

jcking:
Fix #3467 and #3651.


jcking:
For 4.11 we can just use `<filesystem>`.

PR #3658, opened 2022-04-14
Export runtime version in Python3 runtime

xrmx:
To make it easier for runtime user to choose a generated parser
compatible for the runtime.
Runtime 4.10 broke compatibility with parser compiled with 4.7.2 that
were working till 4.9.X and so a way to have the version is handy if one
want to support multiple runtime in the same codebase.

Signed-off-by: Riccardo Magliocchetti <riccardo.magliocchetti@gmail.com>


KvanTTT:
It makes sense for an end user. But it should be fixed for all runtimes, not only Python.

xrmx:
I can add the same for Python2 and javascript easily, i can try to hack something for the rest of the runtimes but I'll take a bit.

lbisceglia:
I just noticed this PR, I was working on an independent change to accomplish something very similar for the Go runtime.
PR: https://github.com/antlr/antlr4/pull/3665

lbisceglia:
Since you've moved the version to a different file, you'll have to update this as well: https://github.com/antlr/antlr4/blob/master/scripts/files-to-update.txt

lbisceglia:
Created a PR for Golang support: https://github.com/antlr/antlr4/pull/3671

lbisceglia:
Created a PR for Golang support: https://github.com/antlr/antlr4/pull/3671

parrt:
I think a lot of the runtimes already have the version information in the runtime metadata. E.g., https://github.com/antlr/antlr4/blob/master/runtime/Java/src/org/antlr/v4/runtime/RuntimeMetaData.java#L70

PR #3669, opened 2022-04-19
Update Swift Package to support static library

HellGhost:
Add static library distribution in SPM


parrt:
Closing https://github.com/antlr/antlr4/pull/3743 in favor of this one.  Should we move to Swift 5.5 or 5.7?

PR #3672, opened 2022-04-21
Preven bndtools from erroneously adding "org.antlr.v4.gui" to the Import-Package header

badfish69:
Since a recent update of the version of maven-bundle-plugin used when building the java runtime, the resulting jar file fails to load in OSGI servers due to an unsatisfied package dependency:
`Unable to resolve org.antlr.antlr4-runtime/4.10.1: missing requirement [org.antlr.antlr4-runtime/4.10.1] osgi.wiring.package; filter:="(osgi.wiring.package=org.antlr.v4.gui)"`

The cause of this is an erroneous reference to `org.antlr.v4.gui` in the Import-Package header in MANIFEST.MF.

It appears that newer versions of bndtools are a little to good at discovering class usage, and are detecting the static string reference in the `Class.forName("org.antlr.v4.gui.TestRig")` call in TestRig.java.

This commit obscures the class name a bit to prevent bndtools from finding it at compile-time.

### DIFF
--- a/runtime/Java/src/org/antlr/v4/runtime/misc/TestRig.java
+++ b/runtime/Java/src/org/antlr/v4/runtime/misc/TestRig.java
@@ -16,9 +16,14 @@ import java.lang.reflect.Method;
  */
 @Deprecated
 public class TestRig {
+	/*
+	 * Ofuscate the TestRig class name a bit as bndtools (via maven-bundle-plugin) is too smart for its
+	 * own good and erroneously adds it to the Import-Package header when packaging
+	 */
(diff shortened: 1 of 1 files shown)


rovarga:
Ah, nice hack. Wouldn't it be better to have the import as optional (as in #3675 )?

badfish69:
@rovarga Your fix is probably better.
Alternatively, you could turn the Class.forName() analysis off entirely with a
`<_noclassforname>true</_noclassforname>` directive
I guess it depends on whether there is ever likely to be a need to resolve the deprecated TestRig class at runtime in an OSGI container?

parrt:
Closed in favor of the other

PR #3692, opened 2022-05-02
3691 - fix hardcoded dependency for OSGI manifest

jherkel:
Fix OSGI manifest

Signed-off-by: Jakub Herkel <j.herkel@gmail.com>


jherkel:
fixed with https://github.com/antlr/antlr4/pull/3675

PR #3694, opened 2022-05-05
[C++] Introduce synchronization and container abstractions

jcking:
In most cases the synchronization and container implementations provided by the C++ standard library are sufficient. However, there are some environments where users may wish to substitute other implementations. One of those is Abseil. By default, the implementations provided by the C++ standard library are used and should be preferred.


jcking:
@hzeller

jcking:
> Wasn't there also a `std::call_once` in `ATNDeserializationOptions.cpp` and in `XPathLexer.cpp` ?

Whoops, thanks. Forgot about those.

GoToCoding:
Sorry for writing to finished PR.

Are you considering to fully move to abseil types instead of std?
Because this duplicating defines look little strange if you want to use abseil flatmap in more and more places.
[Of course that's just my opinion]

PR #3697, opened 2022-05-06
Improve ExternalProject cmake file

Krzmbrzl:
Two small improvements to the CMake file used to setup the cpp runtime
as an externa project in cmake.

The first commit just silences a warning in newer cmake versions, whereas the latter
ensures that a user does not have to take care of setting up the correct include directories
themselves. Linking to either `antl4_static`  or `antlr4_shared` will now automatically
populate the include directories of the target these are linked to.


Krzmbrzl:
Ah whoops - on a fresh run, this will now error because the include dir does not exist yet during configure time.

See also https://stackoverflow.com/questions/45516209/cmake-how-to-use-interface-include-directories-with-externalproject

I'll fix this and force-push the last commit to include the fix.

PR #3702, opened 2022-05-07
[WIP] New script to download antlr jar(s) and run tool on grammars

parrt:
Tired of java classpath? haha @mike-lischke @jcking @KvanTTT @ericvergnaud 

Will also augment to find java VM and install as part of this.

Comments welcome.


kaby76:
I keep seeing people tripping up on Java, just [this a few days ago](https://github.com/kaby76/Antlr4BuildTasks/issues/31#issuecomment-1120230409). For [Antlr4BuildTasks](https://github.com/kaby76/Antlr4BuildTasks), I do the same thing as you--test $path for a java--but then if not found and Windows x64, use a JRE I include in the package, otherwise crap out. I'm now writing code to just download a specific SDK from OpenJDK because I cannot rely on Java "installed" correctly for the tool.

parrt:
Hi @kaby76 yeah, installing the Java VM just a run a tool is a burden, particularly now that Oracle makes it painful with click wrap license etc.  I'm going to be using this python code, which seems to work great:

https://github.com/jyksnw/install-jdk

You can specify what version and whether you want JRE or JDK etc...

The idea is for me most people will have python already installed on their system and so they can do a `pip install someantlr4tool` and get something that seems like an executable that will do the needful.

On the other hand, making an executable package that they can just double click on to install with everything sounds pretty good too.

But all works great so far. I ran out yesterday in fact when trying to test something on the plug-in:

```
$ python ~/antlr/code/antlr4/scripts/antlr4.py *.g4
```

parrt:
The script now has setup.py so running `antlr4` works after you `pip install .` and installs jre 11 if needed.

kaby76:
Fantastic. Less work answering questions from folks who mess up a Java install. I just did the same over for Antlr4BuildTasks, although there are different downloads for different OSes?

parrt:
that python lib is supposed to handle windows, linux, mac. fingers crossed. will try soon on other systems. :)

parrt:
To my surprise, this works on windows 10.  It would NOT, however, get the `entry_points` so I could run just `antlr4`.  Wasted an hour or two.   Anybody else understand the wonders of windows?

parrt:
Linux ubuntu from AWS. It has python 3.10 but no pip. wtf?!  Had to run:

```
sudo apt update
sudo apt install python3-pip
```

Then from my scripts/antlr4 dir in antlr:

```
pip install .
```

Hmm..didn't work.  Trying

```
python setup.py install
```

Looks better. I can now run tool!

```bash
ubuntu@ip-172-30-0-187:~/antlr4$ antlr4
latest 4.10.1
Downloading antlr4-4.10.1-complete.jar
ANTLR tool needs Java to run; install Java JRE 11 yes/no (default yes)? 
Installed Java in /home/ubuntu/.jre/jdk-11.0.15+10-jre; remove that dir to uninstall
ANTLR Parser Generator  Version 4.10.1
 -o ___              specify output directory where all output is generated
 -lib ___            specify location of grammars, tokens files
 -atn                generate rule augmented transition network diagrams
 -encoding ___       specify grammar file encoding; e.g., euc-jp
 -message-format ___ specify output style for messages in antlr, gnu, vs2005
 -long-messages      show exception details when available for errors and warnings
 -listener           generate parse tree listener (default)
 -no-listener        don't generate parse tree listener
 -visitor            generate parse tree visitor
 -no-visitor         don't generate parse tree visitor (default)
 -package ___        specify a package/namespace for the generated code
 -depend             generate file dependencies
 -D<option>=value    set/override a grammar-level option
 -Werror             treat warnings as errors
 -XdbgST             launch StringTemplate visualizer on generated code
 -XdbgSTWait         wait for STViz to close before continuing
 -Xforce-atn         use the ATN simulator for all predictions
 -Xlog               dump lots of logging info to antlr-timestamp.log
 -Xexact-output-dir  all output goes into -o dir regardless of paths/package
```

I'm calling that a victory.  @hzeller might be interested in this.

parrt:
Next, I'm thinking a profile run that doesn't require building gen'd code (works like intellij plugin).  It needs a grammar and input, that's it. Generates lookahead profile info. @mike-lischke and @hzeller might like.

ericvergnaud:
How about a lambda function on AWS that takes the grammars as input and returns the generated files as output ?

Envoyé de mon iPad

mike-lischke:
> Next, I'm thinking a profile run that doesn't require building gen'd code (works like intellij plugin). It needs a grammar and input, that's it. Generates lookahead profile info. @mike-lischke and @hzeller might like.

I use a similar approach for the grammar debugger in my extension. It uses the serialized interpreter data. I guess you can use that too for the profiler.

parrt:
@ericvergnaud I've been thinking a great deal about servers and a collaborative website for grammars etc... For mine with a lot of experience warned me that you can run up bills of $100,000 before you can shut off the processes because people can DDOS your site and cause it to consume resources. He warned me that I would have to spend a huge amount of time thinking about security. Another solution is simply to rent an instance at Amazon or google that is limited to I certain amount of disk space. Then the disk and CPU are limited two fixed costs and we only have to worry about network traffic spiking and costing money.  I can easily put a server on my one gigabit fiber line in my house on a separate private Netwerk but my provider won't give me a static IP.

parrt:
I got a good idea from @hzeller.  He suggested building a website where you can put your grammar and sample input to get profiling information and all sorts of trees or whatever. Then, you can create a link to give to somebody else that will reproduce exactly what you were looking at.  A cool way to send bug reports for example.  This is something that python tutor does. Click on this link and it will show you bit of python I was in the middle of executing:

https://pythontutor.com/visualize.html#code=print%28%22Hi%20guys!%22%29&cumulative=false&curInstr=1&heapPrimitives=nevernest&mode=display&origin=opt-frontend.js&py=3&rawInputLstJSON=%5B%5D&textReferences=false

Naturally that won't work for huge grammar or input which means we would need to store them on the disk somewhere. I was originally just thinking of an S3 bucket but then my friend warned to me about security and money.

KvanTTT:
Yes, security is important because actions in grammar potentially allows RCE (Remote Code Execution) and other vulnerabilities. But I think it's the solvable problem. There are a lot of online compilers/interpreters that allow code compilation on cloud.

KvanTTT:
I've been thinking about the idea of "cloud parser generator" for a long time. But I suggest moving disscusion to the separated topic: https://github.com/antlr/antlr4/discussions/3717

kaby76:
I'm trying to coordinate the download areas for my tools (which are complete app generators with a build file for any OS, any target) with that here. As far as I can tell, the JRE is placed in c:/users/name/.jre for Windows, ~/.jre elsewhere; and the Anrtlr complete jar is place c:/users/name/.m2 for Windows, ~/.m2 elsewhere.

Why ".m2"? Why not ".m3", ".m4", or something related Antlr, e.g., ".antlrtemps"? There's nothing in the source to say what the significance of "m2" is. Thanks.

parrt:
Yeah, I just used whatever that install package did for the JRE; you're right it would be good to coordinate.  Yeah, the .m2 is an artifact of the maven thing. That directory is the root of the repository cache.

parrt:
I think I'm going to add antlr4-prof script today, that takes a grammar and an input file than dumps out profiling information either in csv or formatted text.

parrt:
Right now everything is in the scripts subdirectories of the main repository... It feels kind of weird to have a sub tool that has a different versioning sequence. Should this stuff be pulled out into a different repository like antlr4-tools?

ericvergnaud:
IT would also be an opportunity to collect data on common pitfalls

Envoyé de mon iPad

parrt:
Yes a good idea. I've been keeping logs of everything coming in right now including the errors.

PR #3707, opened 2022-05-09
Cpp: Disable warnings for external project

Krzmbrzl:
When adding ANTLR as an external project, the downstream users will
(generally) not be interested in compiler warnings emitted for ANTLR
itself, as they won't be able to do anything about them anyway (in their
project).

Thus, this commit adds a new cmake option DISABLE_WARNINGS that can be
used to suppress compiler warnings. This flag is then subsequently set
whenever ANTLR is used as an ExternalProject.


parrt:
It seems kind of dangerous to disable warnings. I don't understand the build file enough to know and what circumstances they would be turned off.

Krzmbrzl:
It's not dangerous at all. The warnings are only turned off if someone uses ANTLR as an external project - aka as a 3rdparty dependency. In these cases having warnings for ANTLR base code does not gain anything as the respective developers won't be able to maintain the ANTLR code itself. Therefore any warning emitted there is not really actionable anyway. 

Warnings will still be shown when compiling ANTLR for ANTLR's sake - aka when working on ANTLR itself. Thus, actual ANTLR developers will still continue to see warnings as usual.

PR #3709, opened 2022-05-10
New SLL_OR_LL prediction mode

KvanTTT:
## Motivation

Many resources suggest using two-stage parsing strategy (SLL at first and LL if SLL has been failed) for fast parsing ([Improving the performance of an ANTLR parser](https://tomassetti.me/improving-the-performance-of-an-antlr-parser/) by @ftomassetti). Moreover, even the book `The Definitive ANTLR 4 Reference by Terence Parr, main author of ANTLR` and article `Adaptive LL(*) Parsing: The Power of Dynamic Analysis` by @parrt say two-stage parsing is optimal for most use cases (if input doesn't contain errors but it's common case).

The current solution looks the following way:

```java
// this code comes from The Definitive ANTLR 4 Reference by Terence Parr, main author of ANTLR
// try with simpler/faster SLL(*)
parser.getInterpreter().setPredictionMode(PredictionMode.SLL);
// we don't want error messages or recovery during first try
parser.removeErrorListeners();
parser.setErrorHandler(new BailErrorStrategy());
try {
  parser.startRule();
  // if we get here, there was no syntax error and SLL(*) was enough;
  // there is no need to try full LL(*)
}
catch (ParseCancellationException ex) {
  // thrown by BailErrorStrategy
  tokens.reset();
  // rewind input stream
  parser.reset();
  // back to standard listeners/handlers
  parser.addErrorListener(ConsoleErrorListener.INSTANCE);
  parser.setErrorHandler(new DefaultErrorStrategy());
  // full now with full LL(*)
  parser.getInterpreter().setPredictionMode(PredictionMode.LL);
  parser.startRule();
}
```

But it's too bulky and has a lot of unwanted side effects (clearing error listeners and changing error strategy). Frankly, it's too lazy to write the abovementioned code every time. Also it might be unnecessary details for ANTLR users.

## Solution

I suggest introducing the new mode `SLL_OR_LL` that allows shrinking the code to only two lines:

```java
parser.getInterpreter().setPredictionMode(PredictionMode.SLL_OR_LL);
parser.startRule();
```

The switching is being performed internally, error listeners and error handler are untouched.

If idea is ok and is going to be accepted, I'll fix other runtimes.


tool/resources/org/antlr/v4/tool/templates/codegen/Java/Java.stg
```diff
@@ -411,14 +411,19 @@ RuleFunction(currentRule,args,code,locals,ruleCtx,altLabelCtxs,namedActions,fina
...
+			_errHandler.reportError(this, re);
+			_errHandler.recover(this, re);
+		}
 	}
 	<endif>
 	finally {
 		<finallyAction>
 		exitRule();
 	}
+	if (shouldReparseOrThrow(_localctx.exception)) {
```
parrt, line 424: You are usually very concerned with class file size; won't this erase all your previous gains? Not saying it's wrong but... seems like something to consider.

KvanTTT: For me user convenience has more priority than generated/compiled size. Also it's not big increasing of genereted code (+5 small lines for every rule). If it's possible to decrease the code size than it should be decreased if it doesn't break other gains.

BTW, implemented architecture will allow SLL recover on real syntax error if such algorithm is implemented (I remember an issue about diambigution of real and errors because of SLL weakness).

KvanTTT: I updated Java.stg, now it's even less than 5 small lines :)

parrt:
Potentially very interesting. I'll eval as soon as I can relearn algorithm details.

parrt:
What's the benefit divided by size of change here?  I.e., how much faster is it?

parrt:
Oh, so this change doesn't improve performance really?  I'm looking for the motivation :)

KvanTTT:
The main benefit is user experience and correctness. I'm almost sure there are a lot of users that don't know about two-stage strategy. With the new mode, it becomes very easy to improve parser performance without fuss with exceptions, error listeners and handlers (it's performance improvement but indirect :) ).

parrt:
The benefit of ease-of-use only matters if as you say it "improves parser performance".  Do we have any idea of how much benefit we get? I just don't remember from when I did testing in the old days.  They would have to be strong performance improvements to justify a widespread change across all targets I would think.

KvanTTT:
By the way, I've also introduced two missing recognition exceptions `MissingTokenException` and `UnwantedTokenException`, it looks like every parsing error should have own exception.

Ok, let's postpone it until best times. Maybe one day it would be possible to recover SLL more efficiently.

PR #3722, opened 2022-05-18
Ensure that only the tokens needed are fetched

TristonianJones:
Update the Go and Java runtimes to match the C++ runtime for syncing to only the tokens needed
to read the text from the supplied interval. 

The `fill()` methods search for EOF as a signal to conclude the token syncing; however, when there
is an errant input, EOF may be missing and this can result in `fill()` re-reading the entire token stream
for each error encountered in the input. As such `fill()` has been replaced with `sync()` calls which
provide equivalent functionality.

Signed-off-by: Tristan Swadell <tswadell@google.com>

### DIFF
--- a/runtime/Go/antlr/common_token_stream.go
+++ b/runtime/Go/antlr/common_token_stream.go
@@ -331,10 +331,12 @@ func (c *CommonTokenStream) GetTextFromRuleContext(interval RuleContext) string 
 
 func (c *CommonTokenStream) GetTextFromInterval(interval *Interval) string {
 	c.lazyInit()
-	c.Fill()
 
 	if interval == nil {
+		c.Fill()
 		interval = NewInterval(0, len(c.tokens)-1)
+	} else {
+		c.Sync(interval.Stop)
 	}
 
 	start := interval.Start
--- a/runtime/Go/antlr/common_token_stream_test.go
+++ b/runtime/Go/antlr/common_token_stream_test.go
@@ -152,3 +152,27 @@ func TestCommonTokenStreamCannotConsumeEOF(t *testing.T) {
 	assert.Equal(1, tokens.Size())
 	assert.Panics(tokens.Consume)
 }
+
+func TestCommonTokenStreamGetTextFromInterval(t *testing.T) {
+	assert := assertNew(t)
+	lexEngine := &commonTokenStreamTestLexer{
+		tokens: []Token{
+			newTestCommonToken(1, " ", LexerHidden),                    // 0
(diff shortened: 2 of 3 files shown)


runtime/Go/antlr/common_token_stream.go
```diff
@@ -331,10 +331,12 @@ func (c *CommonTokenStream) GetTextFromRuleContext(interval RuleContext) string
 
 func (c *CommonTokenStream) GetTextFromInterval(interval *Interval) string {
 	c.lazyInit()
-	c.Fill()
 
 	if interval == nil {
+		c.Fill()
 		interval = NewInterval(0, len(c.tokens)-1)
+	} else {
+		c.Sync(interval.Stop)
```
jcking, line 339: We need to double check whether this is supposed to be `interval.Stop` or `interval.Stop - 1`. `Sync` is documented as being an index and `interval.Stop` is documented as being exclusive. However there is some oddities on whether this may or may not be true.

jcking: Okay, intervals in Go are seriously broken. We are not going to fix it in this pull request, so if this works it works.

TristonianJones: When I do `interval.Stop-1` it would return empty text in the case where the token is a single character, so I believe the code is correct only because the `Sync` call is compensating by adding another character to fetch under the covers. The same behavior exists in Java as well.

parrt:
Seems harmless enough and if it's mimicking C++ seems OK.  Is the continuous integration system down @ericvergnaud ?

runtime/Go/antlr/common_token_stream.go
parrt: Hmm... Everything was inclusive if I remember correctly...yeah, we should take a look at this in more detail.

KvanTTT:
There are some problems with self-hosted CI again. Any chance to switch to cloud CI again?

parrt:
@ericvergnaud can you kick the CI?  Seems like github actions, at least self hosted, are unreliable. I guess that means that if we bought some fancy server in the cloud it we have the same problem right?

KvanTTT:
They are already fine. Is it possible to restart the build?

KvanTTT:
>  I guess that means that if we bought some fancy server in the cloud it we have the same problem right?

Yes, I guess it's more reliable option.

parrt:
oh and now it says all fine. that was fast. maybe too fast. haha

PR #3726, opened 2022-05-27
Fixing bug identified by https://github.com/antlr/antlr4/discussions/…

parrt:
Fix per https://github.com/antlr/antlr4/discussions/3712   We didn't see a bug but code is wrong.


parrt:
Just ran https://github.com/antlr/performance and number of live instances and live object heap size for java are about the same with this change so didn't blow anything up.

Issue #2568, opened 2019-06-06
Add @SuppressWarnings and @Generated annotations to all generated classes and interfaces


Vampire:
To prevent some code quality tools to report issues, it would be nice to have the `@SuppressWarnings` annotation and also a `@Generated` annotation on all generated classes.

Currently the `@SuppressWarnings` is only present on the lexer and the parser, but for example not on the listener, base listener, visitor and base visitor.

For `@Generated` it might be a good idea to have this optional, as in latest Java releases this annotation was moved out of the JRE into an own library.
If the `@Generated` annotation is added, the single-line comment about the file being generated could eventually be left out.

amrith92:
I think `@SuppressWarnings` is already being added, however, `@Generated` is not. 👍 to add support for this.

Vampire:
> I think @SuppressWarnings is already being added

As I said, only on the lexer and parser, not on the others like listener, base listener, visitor or base vistor.

Regarding generated annotation there is also #2512

dhalperi:
:+1: to this.

Would be happy to send a PR if it would be reviewed -- seems like the maintainers have not had much time to respond to changes, especially to Java code, recently.

ericvergnaud:
Hi,

I’ll happily review @Generated
I’m more reluctant to review @SuppressWarnings for 2 reasons:
 - it would be a better fix to not generate code that raises warnings
 - which warnings are raised depends on the jdk version and compiler, and unused warnings generate warnings…

Eric

magicwerk:
I created PR #3732 which allows to specify the desired annotation using command line option -generated.
Example:
-generated javax.annotation.processing.Generated

This should also fix #2512.

Regarding the discussion about @SuppressWarnings:
The created Lexer and Parser file already contain @SuppressWarnings, so the information in this issue is probably outdated.

Vampire:
> The created Lexer and Parser file already contain @SuppressWarnings, so the information in this issue is probably outdated.

Why do you think so?
I already said it in the original comment and in the response to the first comment stating the same like you.
Yes, it is on those two, but missing on the others.
So which part is outdated?

Vampire:
Btw. both `@Generated` annotations (`javax.annotation.Generated` and `javax.annotation.processing.Generated`)  are suboptimal as both have only `SOURCE` retention and some tools like JaCoCo need at least `CLASS` retention to see the annotation as they do not work on the source code. So indeed an option might be the best choice.

Issue #2512, opened 2019-03-16
javax.annotation.processing.Generated


dfa1:
It could be useful to (optionally) annotate Java sources with `javax.annotation.processing.Generated` [1]:

> The Generated annotation is used to mark source code that has been generated. It can also be used to differentiate user written code from generated code in a single file.

Some tools, notably sonar [2], are able to automatically detect such annotation and skip raising of issues.

I don't know if it is possible  but I could help in testing, documentation or even a PR (with some guidance).

[1] https://docs.oracle.com/javase/9/docs/api/javax/annotation/processing/Generated.html
[2] https://jira.sonarsource.com/browse/SONARJAVA-71

ericvergnaud:
Hi
It would definitely make sense (and is easy to achieve) but would unfortunately break compatibility of generated sources with Java 7 and 8, which we still support.
Éric 

Envoyé de mon iPhone

dfa1:
Hi, 

I completely agree, as I mentioned it should be optional (default off) in order to support users on Java 8 and 7.

ChristianLutz:
Hello, 
For java 7 and 8 there is [`javax.annotation.Generated`](https://docs.oracle.com/javase/8/docs/api/javax/annotation/Generated.html) available.

ericvergnaud:
Interesting.
Then may I suggest you submit a PR?
It should be quite trivial to add this to the java.stg template

ChristianLutz:
Hello,
is there any way to distinguish within the template between the java target version? Because for 7 and 8 we must use `javax.annotation.Generated` and for > java 9 `javax.annotation.processing.Generated`.

ericvergnaud:
From what I can see it is also available in Java 9
https://docs.oracle.com/javase/9/docs/api/javax/annotation/Generated.html <https://docs.oracle.com/javase/9/docs/api/javax/annotation/Generated.html>

Did you give it a try?

Vampire:
The whole `java.xml.ws.annotation` module which contains `javax.annotation.Generated` was deprecated in Java 9 and removed in Java 11.
You can still use `javax.annotation.Generated` in Java 9+ if you depend on the module `java.xml.ws.annotation` which is not the best idea, or by adding a dependency on `javax.annotation:javax.annotation-api` wich also provides the `javax.annotation.Generated` annotation.
But it would be better, if you could select which annotation to use.
[...]

PR #3732, opened 2022-05-30
Annotate Java files as being generated

magicwerk:
Signed-off-by: magicwerk <magicwerk@gmail.com>

### DIFF
--- a/tool/resources/org/antlr/v4/tool/templates/codegen/Java/Java.stg
+++ b/tool/resources/org/antlr/v4/tool/templates/codegen/Java/Java.stg
@@ -48,6 +48,9 @@ ParserFile(file, parser, namedActions, contextSuperClass) ::= <<
 package <file.genPackage>;
 <endif>
 <namedActions.header>
+<if(file.generatedImport)>
+import <file.generatedImport>;
+<endif>
 import org.antlr.v4.runtime.atn.*;
 import org.antlr.v4.runtime.dfa.DFA;
 import org.antlr.v4.runtime.*;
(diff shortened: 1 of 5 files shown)


parrt:
hi. I don't think it's a good idea to add an option to the entire tool just to get this tiny feature for one out of 10 targets.  What problem are you solving here may I ask?

magicwerk:
The problem I try to solve is referenced:
- Add @SuppressWarnings and @Generated annotations to all generated classes and interfaces #2568
There is also another related issue:
- javax.annotation.processing.Generated #2512 
In both issues, feedback was positive that a PR would be welcomed and reviewed.
So this PR tries to solve theses issues.

parrt:
Thanks very much for the detailed response. Could you modify this so all it does is add @generated("ANTLR") in Java.stg? Can I don't think I'm ready for options or any other warning suppression at this time. thanks!

magicwerk:
Unfortunately things are not that easy...
The original annotation @javax.annotation.Generated is not longer part of Java 11.
You can use @javax.annotation.processing.Generated instead or bring back the old javax stuff with the artifact javax.annotation:javax.annotation-api.
As Antlr targets a wide ranch of Java versions, a command line option would definitely make more sense.

parrt:
hm...ok, yeah, I don't want to add an option just for this. As it doesn't make sense to add something that will only work with certain versions then I guess I'll close. sorry!

Vampire:
@parrt could you maybe reconsider this?
It would be super helpful.
Especially as both mentioned `Generated` annotations have `SOURCE` level retention and if you for example want JaCoCo to be able to exclude files, they need to have a `Generated` annotation with at least `CLASS` retention as it does not work with the source code.
So being able to pick an annotation to apply, or maybe even multiple to also add warning suppression, would really be helpful.

Issue #1839, opened 2017-04-25
[Feature Request] Target for Rust Programming Language


ikarienator:
I propose we have a target for Rust Programming Language since:

1. Rust has finally delivered stable release last year (currently version 1.16).
2. Rust has attracted more and more engineers recent years.

Rust has become a mature programming language with the constant effort from Mozilla. The I haven't found any good parser generators in the Rust world yet. Hopefully ANTLR can be one.

mike-lischke:
This will only happen if someone from the commuity creates such a target and sends a pull request to the main ANTLR repo.

neatnerd:
I would be willing to contribute, is somebody working on this subject already, so I dont do double work?

KvanTTT:
You can also create a new topic in [ANTLR google group](https://groups.google.com/forum/#!forum/antlr-discussion)

spease:
This would also be useful for a work project I'm working on.

keppel2:
This would be great. A minimal implementation that omits some features would be fine (the Go implementation doesn't have a default visitor implementation for example). I just want to walk the tree myself.

metadave:
I'd like to give this a shot unless someone else is already working on it.

KvanTTT:
Yes, I think it's a great idea!

xipheCom:
Would be awesome! Where can I watch the progress?

rrevenantt:
@metadave I have done quite a lot of work on it recently. You can see it in https://github.com/rrevenantt/antlr4rust 
A decent bit of things are already working. If anybody wants to help i would gladly appreciate it.

BTW How complete target implementation should be to be merged in?

mike-lischke:
@rrevenantt I recommend to use a clone of the antlr repo instead of an own repo, with an own structure. This way you can later create a pull request to get your code merged in and you can directly use the scripts, folders and test infrastructure of the main repo. Keep in mind that your target has to run the entire test suite that's executed for each of the targets.

metadave:
@rrevenantt excellent! if that's the case, then perhaps I'll pick a different (new) target to work on, as I've wanted to learn all the internals of Antlr4 for quite some time.

rrevenantt:
@mike-lischke Thanks. I was thinking about using git-subtree, so i have created it as a separate repo. Anyway I can easily merge them later. Nevertheless I am already using test infrastructure of the main repo, and quite some tests from test suit are already passing.

[...]

PR #3736, opened 2022-06-03
Runtime tests refactoring

### LINKED ISSUE #3367: Refactor runtime testsuite
There is a lot of duplicated code in runtime test classes that can be extracted to a common abstract class. This issue can be assigned this issue to me.

KvanTTT:
### Motivation

* Make it easy to integrate new targets (Kotlin, Rust) to the ANTLR infrastructure. Initially I was going to integrate Kotlin target and I encountered a lot of code duplication and unclear architecture. I decided to rewrite it at first.
* Prepare the platform for application like [Online grammar editor based on ANTLR](https://github.com/antlr/antlr4/discussions/3717). It will be possible to reuse existing code for compiling and executing target code on different runtimes
* Deduplicate and simplify code for more convenient maintaining

fixes #3367 (partially)

@parrt I think it's ready for review. It's -2500 lines of code but functionality is the same! I touched only test files.

### Description

* Introduce universal workflow the used for all runtimes (`BaseRuntimeTestSupport`).
  It contains the following stages:
  * Generate
  * Compile
  * Execute

  After completion of every stage, corresponding state is being created. Every state contains exception (if a error occurred) and reference to the previous state. Java execution state is special because it contains java objects that are used in some tests. It's not necessary to run the workflow until the last stages, some test require only generation or compilation.
* Clarify API of `BaseRuntimeTestSupport`. Now almost all methods have basic implementation
* Clarity error messages
* Extract `stg` templates and other resource to separate files and simplify them
* Clear CI build and run scripts
* Extract `FileUtils`, `Generator`, clarify other util classes
* Improve performance:
    * Use cache for resources (do not load resource twice)
    * Extract static code parts

## Example

Now `BaseCSharpTest` looks the following way:

```cs
/*
 * Copyright (c) 2012-2017 The ANTLR Project. All rights reserved.
 */
package org.antlr.v4.test.runtime.csharp;

import org.antlr.v4.test.runtime.*;
import org.antlr.v4.test.runtime.states.CompiledState;
import org.antlr.v4.test.runtime.states.GeneratedState;
import org.stringtemplate.v4.ST;

import java.nio.file.Paths;

import static org.antlr.v4.test.runtime.FileUtils.mkdir;
import static org.antlr.v4.test.runtime.FileUtils.writeFile;

public class BaseCSharpTest extends BaseRuntimeTestSupport {
	@Override
	public String getLanguage() { return "CSharp"; }

	@Override
	public String getTitleName() { return "C#"; }

	@Override
	public String getExtension() { return "cs"; }

	@Override
	public String getRuntimeToolName() { return "dotnet"; }

	@Override
	public String getExecFileName() { return getTestFileName() + ".dll"; }

	private final static String testProjectFileName = "Antlr4.Test.csproj";
	private final static String cSharpAntlrRuntimeDllName =
			Paths.get(getCachePath("CSharp"), "Antlr4.Runtime.Standard.dll").toString();

	private final static String cSharpTestProjectContent;

	static {
		ST projectTemplate = new ST(RuntimeTestUtils.getTextFromResource("org/antlr/v4/test/runtime/helpers/Antlr4.Test.csproj.stg"));
		projectTemplate.add("runtimeLibraryPath", cSharpAntlrRuntimeDllName);
		cSharpTestProjectContent = projectTemplate.render();
	}

	@Override
	protected void initRuntime() throws Exception {
		String cachePath = getCachePath();
		mkdir(cachePath);

		String[] args = new String[]{
				"dotnet",
				"build",
				Paths.get(getRuntimePath(), "src", "Antlr4.csproj").toString(),
				"-c",
				"Release",
				"-o",
				cachePath
		};

		runCommand(args, cachePath, "build " + getTitleName() + " ANTLR runtime");
	}

	@Override
	public CompiledState compile(RunOptions runOptions, GeneratedState generatedState) {
		Exception exception = null;
		try {
			writeFile(getTempDirPath(), testProjectFileName, cSharpTestProjectContent);
			runCommand(new String[]{"dotnet", "build", testProjectFileName, "-c", "Release"}, getTempDirPath(),
					"build C# test binary");
		} catch (Exception e) {
			exception = e;
		}
		return new CompiledState(generatedState, exception);
	}
}
```

and its template:

```stg
using System;
using Antlr4.Runtime;
using Antlr4.Runtime.Tree;
using System.Text;

public class Test {
	public static void Main(string[] args) {
		Console.OutputEncoding = Encoding.UTF8;
		Console.InputEncoding = Encoding.UTF8;
		var input = CharStreams.fromPath(args[0]);
		var lex = new <lexerName>(input);
		var tokens = new CommonTokenStream(lex);
<if(parserName)>
		var parser = new <parserName>(tokens);
<if(debug)>
		parser.AddErrorListener(new DiagnosticErrorListener());
<endif>
		parser.BuildParseTree = true;
		var tree = parser.<parserStartRuleName>();
		ParseTreeWalker.Default.Walk(new TreeShapeListener(), tree);
<else>
		tokens.Fill();
		foreach (object t in tokens.GetTokens())
			Console.Out.WriteLine(t);
<if(showDFA)>
		Console.Out.Write(lex.Interpreter.GetDFA(Lexer.DEFAULT_MODE).ToLexerString());
<endif>
<endif>
	}
}

<if(parserName)>
class TreeShapeListener : IParseTreeListener {
	public void VisitTerminal(ITerminalNode node) { }
	public void VisitErrorNode(IErrorNode node) { }
	public void ExitEveryRule(ParserRuleContext ctx) { }

	public void EnterEveryRule(ParserRuleContext ctx) {
		for (int i = 0; i \< ctx.ChildCount; i++) {
			IParseTree parent = ctx.GetChild(i).Parent;
			if (!(parent is IRuleNode) || ((IRuleNode)parent).RuleContext != ctx) {
				throw new Exception("Invalid parse tree shape detected.");
			}
		}
	}
}
<endif>
```

Compare it with the previous version: https://github.com/antlr/antlr4/blob/dev/runtime-testsuite/test/org/antlr/v4/test/runtime/csharp/BaseCSharpTest.java


parrt:
That is really impressive. We've been needing a refactoring of the code for a long time so thank you! The only scary part is the size of the change but I saw nothing worrisome as I went through it. Perhaps others could take a quick peek like @ericvergnaud

ericvergnaud:
Will do over the weekend

KvanTTT:
By the way it's not all changes. I know how Java performance can be increased and I've encountered some multithreading issues and useless locks. Some side effects also should be eliminated. But I decided to postpone these changes to the next time since the current request is already too big.

parrt:
@KvanTTT by "Java performance" do you mean of the testing or normal parse runtime?

KvanTTT:
> by "Java performance" do you mean of the testing or normal parse runtime?

I mean testing. We have method [`execClass`](https://github.com/antlr/antlr4/blob/master/runtime-testsuite/test/org/antlr/v4/test/runtime/java/BaseJavaTest.java#L349) and currently it's being executed out of process because `TEST_IN_SAME_PROCESS` is always false. But in-process testing is implemented incorrectly, it doesn't not work in multithreading mode since global output stream is used (`System.setOut(new PrintStream(stdoutOut));`). It should be either removed of fixed. I suggest fixing since it improves performance quite good.

KvanTTT:
Also, I think [locking on generator](https://github.com/antlr/antlr4/blob/master/runtime-testsuite/test/org/antlr/v4/test/runtime/BaseRuntimeTest.java#L306) also should be removed. Currently I don't understand why ANTLR tool is not thread safe and actually it looks like it's thread-safe since all tests works fine without locking in multi-threaded mode and I haven't found global state that may affect multithreading.

parrt:
> global output stream is used (System.setOut(new PrintStream(stdoutOut));). 

I guess we would have to change all of the templates that generate output? I think we have template refs like `<print(...)>` or something.

parrt:
> actually it looks like it's thread-safe since

The code does not seem to use any globals so I can't remember why this was locked... Running the Java tests is really slow for sure... Let me try it myself

KvanTTT:
> I guess we would have to change all of the templates that generate output? I think we have template refs like <print(...)> or something.

Yes, we have to redirect output to another stream, not standard. Actually, I've already done that in local commit but haven't publish it yet to get rid of a lot of changes.

KvanTTT:
> Running the Java tests is really slow for sure...

They are quite fast especially compare to C++ :) Anyway they can be improved more to make them even faster.

parrt:
3min -> 2min 30s for the runtime test when I turn off locking. (no parallelism requested)

parrt:
With `mvn -Dparallel=classes -DthreadCount=4 -Dtest=java.** test` it drops to 1m 30s!  Assuming no lock. I think you can go ahead and add that to this PR if you like.  Seems to work.  I see no obvious data collisions.

parrt:
BTW, I'm unlikely to integrate new targets given the hassles we already have for targets.

KvanTTT:
> BTW, I'm unlikely to integrate new targets given the hassles we already have for targets.

What hassles do you mean?

parrt:
Keeping many different Target language is up-to-date all by itself is painful when I don't know how to use most of them haha. Also learning how to publish each one of these is a hassle. I literally cannot published the Dart Target at the moment. Think about the trouble that we go through to get the continuous integration servers working. Target developers come and go but I'm left maintaining thanks.

KvanTTT:
So, it means it doesn't make sense to work on Kotlin target, isn't it?

parrt:
> So, it means it doesn't make sense to work on Kotlin target, isn't it?

Kotlin would be the easiest to target to integrate as it could simply reuse the Java runtime; I'm assuming that the target would really just be a new template file that generated source in Kotlin.

But, as usual, I have to ask the question about motivation. I've not heard people asking for it but that doesn't mean the market is not there.  I haven't heard anybody asking for rust either but if the market appears then I can imagine supporting those targets.  Kotlin is really cool but the market seems to be limited to android development for the most part and I'm not sure how many of those people are parsing.

KvanTTT:
> Kotlin would be the easiest to target to integrate as it could simply reuse the Java runtime; I'm assuming that the target would really just be a new template file that generated source in Kotlin.

Not sure about that, Kotlin has other targets except of JVM (JavaScript, Native). The entire runtime should be ported to Kotlin.

> I haven't heard anybody asking for rust either but if the market appears then I can imagine supporting those targets.

Not sure about objectivity but we have the issue about Rust runtime support with almost 100 votes: https://github.com/antlr/antlr4/issues/1839

KvanTTT:
Hi. I suggest merging since all tests are okay (except of problematic Swift and C++).

parrt:
@ericvergnaud any time to take a quick peek?

ericvergnaud:
No, I’m facing blocking internet issues, so focusing on resolving them

parrt:
This is likely to be completely correct or super close so I'm merging.  We can always tweak.

parrt:
Actually mac CI is failing (not tests) again...i'll pull and test.

KvanTTT:
Thanks! I have another bunch of tests improvements. Will make pull request ASAP.

parrt:
dang. I tried several times myself but w/o freeze.  Is it for a specific target that it freezes?  Can YourKit identify the lock?

KvanTTT:
>  I tried several times myself but w/o freeze.

It happens not every time. Multithreading issues are hard to detect and debug because they are not reproducable. I've tried on JavaScript runtime.

> Can YourKit identify the lock?

I haven't tried yet, but I'll try. Not necessary YourKit but maybe another tool.

KvanTTT:
Interesting, it looks like I've found one of potential issues and even IntelliJ warns about that!

image

parrt:
Wow.  I wonder why it's wait()'ing there. Line 199:

```
SingletonPredictionContext.create(ctx, ((RuleTransition)t).followState.stateNumber);
```

Doesn't lead to any obvious locking.  Does it say what it's locking on?

Hmm...what's this in PredictionContext?

```
public static int globalNodeCount = 0;
```

That looks like trouble.

KvanTTT:
> Wow. I wonder why it's wait()'ing there. Line 199:

Most likely it's because of class loading deadlock, other threads wait on initializer.

> Doesn't lead to any obvious locking. Does it say what it's locking on?

I'd say it might occur if two or more thread are trying to initialize `EmptyPredictionContext` and another instance of  `PredictionContext` simultaniously. It's rare case but it's possible. Also, there are questions on Stackoverflow: https://stackoverflow.com/questions/48755164/referencing-subclass-in-static-variable, https://stackoverflow.com/questions/48009445/what-does-class-loading-deadlock-mean-here

Anyway, it could be easily fixed by moving static fields to its declarations (move `EMPTY` to `EmptyPredictionContext`).

> That looks like trouble.

It's trouble but not significant since identifiers are only used inside `toDOTString` and it's not critical. Anyway it can be easily fixed by using `AtomicInteger` instead of plain `int`.

KvanTTT:
BTW potential deadlocks is a critical problem. In fact ANTLR Java runtime can not be considered as thread-safe because of them.

parrt:
> BTW potential deadlocks is a serious problem. In fact ANTLR Java runtime can not be considered as thread-safe because of them.

Hmm...I thought it was because of something in ANTLR tool or is that a separate lock disucssion for parallel testing.  We've never seen a Java runtime deadlock that I'm aware of.  Do we just have to worry about EmptyPredictionContext?

KvanTTT:
> We've never seen a Java runtime deadlock that I'm aware of.

But it does not fully exclude the bug. It occurs rarely.

> Do we just have to worry about EmptyPredictionContext?

Yes, but it's located in `runtime` dir, not `tool` and not `tests`.

KvanTTT:
Anyway, I have it fixed in another branch.

PR #3739, opened 2022-06-04
runtime/cpp: fix version detection in FindANTLR.cmake

Xiretza:
Minor versions can have multiple digits, such as in 4.10.1.


parrt:
Hi. Emerged some thing that solve this one already. Thank you.  https://github.com/antlr/antlr4/pull/3706

PR #3740, opened 2022-06-05
[Go] Add Iterative tree walker in go

rishabh-arya95:
Adding the iterative tree walker implementation for GO, inspired by the java iterative walker
https://github.com/antlr/antlr4/pull/1231


parrt:
Now that I have merged the update to the Go Target made by @jimidle maybe you guys can take another look to ensure it is compatible.

jimidle:
I will take a look at this after 4.11.0 is released. Right now, I do not know if anyone has attempted this, but it is probably best done with generics and current tree stuff does not use generics... yet.

jimidle:
@parrt Would you please assign this to me, or label it with Go.

jimidle:
This code now needs to be moved in to the /v4 version of the go runtime. I have not tried this and there are no tests, but as it is a new file, it seems safe enough. We can merge this one and I can copy it to the v4 version.

parrt:
@jimidle so I should merge this one and then you will copy it to v4?

jimidle:
Yes. Let’s do that. The OP hasn’t resubmitted. It’s a small addition, so I
can just do it.

parrt:
done @jimidle !

PR #3742, opened 2022-06-05
Update License Files to Help Autodetection

### LINKED ISSUE #1952: GitHub license detection is broken
Pull request #1628 broke GitHub's automatic license detection for the repository. The third party license notices should be moved to a separate file, **THIRD-PARTY-NOTICES.txt**, rather than included inline in **LICENSE.txt**. See tunnelvisionlabs/antlr4cs#184 for an example.

msymons:
* Adjust the text of of the BSD license so that it is now recognised as  BSD-3-Clause license  by GitHub auto-detection.  
* Move MIT license from LICENSE.txt to THIRD-PARTY-NOTICES.txt & attribute correctly,  This fixes #1952
* Update copyright year in LICENSE.txt to include current year 2022
* Update license name in project POM to `BSD-3-Clause` (Official SPDX short identifier for the license).  This helps ensure that downstream SCA tooling can correctly identify the license.
* Update URLs in project POM to use https, where https is used.

### DIFF
--- a/LICENSE.txt
+++ b/LICENSE.txt
@@ -1,52 +1,28 @@
-[The "BSD 3-clause license"]
-Copyright (c) 2012-2017 The ANTLR Project. All rights reserved.
+Copyright (c) 2012-2022 The ANTLR Project. All rights reserved.
 
 Redistribution and use in source and binary forms, with or without
 modification, are permitted provided that the following conditions
 are met:
 
- 1. Redistributions of source code must retain the above copyright
-    notice, this list of conditions and the following disclaimer.
- 2. Redistributions in binary form must reproduce the above copyright
-    notice, this list of conditions and the following disclaimer in the
-    documentation and/or other materials provided with the distribution.
- 3. Neither the name of the copyright holder nor the names of its contributors
-    may be used to endorse or promote products derived from this software
-    without specific prior written permission.
+1. Redistributions of source code must retain the above copyright
+notice, this list of conditions and the following disclaimer.
 
-THIS SOFTWARE IS PROVIDED BY THE AUTHOR ``AS IS'' AND ANY EXPRESS OR
-IMPLIED WARRANTIES, INCLUDING, BUT NOT LIMITED TO, THE IMPLIED WARRANTIES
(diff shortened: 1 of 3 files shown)


msymons:
Failed DCO, not realizing the it requires use of PRIMARY email address, not just the one that I am happy to have public.

msymons:
@parrt,  apart from the DCO problem, is this a PR that looks OK?  If it does then I will resubmit it so that DCO passes.

LICENSE.txt
```diff
@@ -1,52 +1,28 @@
-[The "BSD 3-clause license"]
-Copyright (c) 2012-2017 The ANTLR Project. All rights reserved.
+Copyright (c) 2012-2022 The ANTLR Project. All rights reserved.
```
KvanTTT, line 1: Is it possible to update the end year for all files in the repository? (2017 -> 2022)

msymons: I am sure it is possible to do that although I would not be the right person to do that...  being a non-dev and not having an IDE or other tooling to help update files in bulk.  I know that copyright rules can differ between jurisdictions, but I would think that the correct way to update copyright on each file is as it is worked on.  In other words,  do not claim 2022 copyright on a file last updated in (say) 2019.

I updated the date on the license file because it applies to the project as a whole.  More importantly,  it just seems crazy to be making changes to the license file and and yet neglect to take the opportunity to tweak the date.

parrt:
> Failed DCO, not realizing the it requires use of PRIMARY email address, not just the one that I am happy to have public.

hi. i thought it was just the email used with github. Sorry for the hassle.

msymons:
> > Failed DCO, not realizing the it requires use of PRIMARY email address, not just the one that I am happy to have public.

> hi. i thought it was just the email used with github. Sorry for the hassle.

The email  address that I used is one of my backup GitHub email addresses...  but not the address that is flagged as "Primary" on my account.  The DCO error said that this primary address is what it was looking for.  So, personally,  perhaps we should blame the DCO documentation =) 

Do note that the 3 commits in the PR are all "verified".

I can re-create the PR if need be...  but did want to check that all was OK with the actual changes made.

parrt:
Looks good. Thanks, @msymons !

PR #3743, opened 2022-06-08
[Swift] let the Swift Package Manager choose between static or dynamic linking 

100mango:
When building an executable target using Swift Package Manager,  We should depend on a static Antlr runtime framework to make a portable executable binary.

Or we will have to deal with the linker problem like:

```
dyld[36790]: Library not loaded: @rpath/libAntlr4.dylib
  Referenced from: /Users/100mango/Desktop/RossetaDartDemo/Rosseta
  Reason: tried: '/usr/lib/swift/libAntlr4.dylib' (no such file),
 '/Users/100mango/Desktop/RossetaDartDemo/../lib/libAntlr4.dylib' (no such file),
 '/usr/lib/swift/libAntlr4.dylib' (no such file), 
'/Users/100mango/Desktop/RossetaDartDemo/../lib/libAntlr4.dylib' (no such file),
 '/usr/local/lib/libAntlr4.dylib' (no such file), 
'/usr/lib/libAntlr4.dylib' (no such file)
```

### DIFF
--- a/Package.swift
+++ b/Package.swift
@@ -8,7 +8,7 @@ let package = Package(
     products: [
         .library(
             name: "Antlr4",
-            type: .dynamic,
+            type: .static,
             targets: ["Antlr4"]),
     ],
     targets: [
--- a/Package.swift
+++ b/Package.swift
@@ -8,7 +8,6 @@ let package = Package(
     products: [
         .library(
             name: "Antlr4",
-            type: .static,
             targets: ["Antlr4"]),
     ],
     targets: [


parrt:
hi. How certain are you of this change? Well this affect people that are expecting dynamic?

100mango:
@parrt Good Point👍  A better approach is to leave this parameter unspecified (done)。

According to Apple Docs:
" Leave this parameter unspecified to let the Swift Package Manager choose between static or dynamic linking (recommended)."

KvanTTT:
Hi. Not sure it should be removed. Have you tried runtime tests with removed option? They don't work for me if the option is removed because `Antlr4.lib` and `Antlr4.dll` are absent (I try Windows). But it looks actual for Mac/Linux as well.

HellGhost:
https://github.com/antlr/antlr4/pull/3669

parrt:
Hi gang. Can you help me understand the status here? Do we need this and #3669 or does #3669 supersede this one? thanks!

serhiybobyr-tomtom:
> Hi gang. Can you help me understand the status here? Do we need this and #3669 or does #3669 supersede this one? thanks!

By default the current changes are enough and are recommended. [#3669](https://github.com/antlr/antlr4/pull/3669)  supersedes this one. Beside the recommended way it also specifies 2 more products. So as a result `swift build` will produce static and dynamic libraries the default one will only produce object files which then you can choose to what used it for. BTW the other PR bumps swift version from 5.3 straight to 5.6 which is not related to this at all and raises a question why not 5.5 or 5.7.

parrt:
Closed in favor of https://github.com/antlr/antlr4/pull/3669. thanks!

PR #3750, opened 2022-06-17
Runtime tests refactoring 2 (concurrent tests, no duplication in runtime tests)

KvanTTT:
* All tests now can be run concurrent
* Use [dynamic tests](https://junit.org/junit5/docs/current/user-guide/#writing-tests-dynamic-tests), remove a lot of duplicated code
* Update jUnit: 4.13.2 -> [5.8.2](https://junit.org/junit5/)
* Simplify CI run scripts, set up parallelization via config (the same config for maven and IDE)
* Unify output for all runtimes
* `RuntimeTestDescriptor` now is immutable
* Remov outdated and not used files (.stg)

@parrt it also can be reviewed. It's almost -5K lines with the same or improved functionality.


parrt:
Should I still review given deadlock we're seeing or wait a bit?

runtime-testsuite/resources/org/antlr/v4/test/runtime/descriptors/CompositeParsers/BringInLiteralsFromDelegate.txt
```diff
@@ -18,6 +18,4 @@ s
 =a
 
 [output]
-"""S.a
-"""
-
+"""S.a"""
```
parrt, line 21: I'm nervous about changing the output. You must have changed the semantics of this.  Sometimes the \n is part of the output.

runtime-testsuite/test/org/antlr/v4/test/runtime/RuntimeTestDescriptor.java
```diff
@@ -1,65 +1,84 @@
-/*
```
parrt, line 1: removing copyright?

runtime-testsuite/test/org/antlr/v4/test/runtime/cpp/CppRuntimeTests.java
```diff
@@ -0,0 +1,11 @@
+package org.antlr.v4.test.runtime.cpp;
+
+import org.antlr.v4.test.runtime.RuntimeTests;
+import org.antlr.v4.test.runtime.RuntimeRunner;
+
+public class CppRuntimeTests extends RuntimeTests {
```
parrt, line 6: Not sure we want to remove all these test entry points. It was very easy to select and run individual tests even if having the duplicated code around was "bad".

KvanTTT: Actually the previous tests were also being generated and it was not very convenient to select individual tests. You had to generate them at first. With the current system both groups and individual tests are being generated automatically:

image

With the dynamic tests there are no dummy `testOne` methods.

runtime-testsuite/test/org/antlr/v4/test/runtime/RuntimeTestDescriptor.java
KvanTTT: It's not intentially. I can restore if it matters.

runtime-testsuite/resources/org/antlr/v4/test/runtime/descriptors/CompositeParsers/BringInLiteralsFromDelegate.txt
KvanTTT: In this case we have action `a : '=' 'a' {<write("\"S.a\"")>};` that definitely should not print new line. That's why I think it's correct change. Also, there is inconsistent code across runtimes: some of the them prints new lines, some of them are not. The code that fixes new lines stuff is place in separated commit: https://github.com/antlr/antlr4/pull/3750/commits/08f5ae7098ad084e9abfa60f3b74d14e656e2118

KvanTTT:
> Should I still review given deadlock we're seeing or wait a bit?

I'll fix deadlock in another request. It doesn't break anything since lock on generator is not removed in the current request.

runtime-testsuite/test/org/antlr/v4/test/runtime/cpp/CppRuntimeTests.java
parrt: heh, that does look a lot better! And also removes code which is fantastic!

runtime-testsuite/test/org/antlr/v4/test/runtime/RuntimeTestDescriptor.java
parrt: Might as well keep the copyright on everything.

runtime-testsuite/resources/org/antlr/v4/test/runtime/descriptors/CompositeParsers/BringInLiteralsFromDelegate.txt
parrt: From a quick scan it looks like you have always added a new line across runtimes? Then just stripped?

KvanTTT: Trailing new line was always being added to the output: https://github.com/antlr/antlr4/blob/dev/runtime-testsuite/test/org/antlr/v4/test/runtime/TestOutputReading.java#L35 I rewrote the behavior and now all new lines are being handled accurately: https://github.com/KvanTTT/antlr4/blob/runtime-tests-refactoring-2/runtime-testsuite/test/org/antlr/v4/test/runtime/StreamReader.java#L40 (it only removes `\r` chars to unify output across platforms).

parrt:
Just checking @KvanTTT: am i able to run each test and group of tests individually?  I.e., TestSets.foo() and TestSets for a specific target?

KvanTTT:
> Just checking @KvanTTT: am i able to run each test and group of tests individually? I.e., TestSets.foo() and TestSets for a specific target?

Yes, but you have to run tests at first to let IDEA to display groups and tests. It's not necessary to wait for the completion. After tests are generated, you can choose either individual test or entire group to run. You can try it on my branch before merging.

parrt:
I see so i "run all" once then I can select from that list to individually run tests.  Hmm...yeah, I guess that is a decent trade to remove all the redundant classes.  Does it still work though from maven with: 

```
$ mvn -Dtest=java.*Lexer* test
```

runtime-testsuite/resources/org/antlr/v4/test/runtime/descriptors/CompositeParsers/BringInLiteralsFromDelegate.txt
parrt: ok. cool.

KvanTTT:
> Does it still work though from maven with:

I haven't found a solution, but I suspect it should be possible since IDEA can run groups.

parrt:
Hmm...ok. let me pull in and check it out.

parrt:
Hmm...can't get intellij to run the tests. what do i run?

parrt:
Ah. You forgot to filter the files....mac's put that crazy file in there.

Screen Shot 2022-06-18 at 2 58 44 PM

parrt:
Ok, cool. It works after I delete that .DS_Store file.  I'm able to run individual groups.

parrt:
@KvanTTT Can you modify the https://github.com/antlr/antlr4/blob/master/doc/antlr-project-testing.md to suit changes?

KvanTTT:
Sure. Should I change generation of tests (3rd to 4th)?

parrt:
> Sure. Should I change generation of tests (3rd to 4th)?

haha sure why not!?

parrt:
Can you filter for dirs only or ignore .DS_Store too?

KvanTTT:
> Can you filter for dirs only or ignore .DS_Store too?

Ok, I'll add `.DS_Store` to `.gitignore`. But it looks like it relevant only for Mac: https://en.wikipedia.org/wiki/.DS_Store

parrt:
It's not in the repo, the mac can just add to the disk and your code simply finds all entries and assumes dir and valid group of tests I think.

KvanTTT:
Interesting, it's already in .gitignore: https://github.com/antlr/antlr4/blob/a8e048ad9d4d1325e030ba394216877d24f16611/.gitignore#L17 Ok, got it. It looks like all dirs that start with dot `.` should be ignored.

parrt:
Dang. Hope that wasn't me ;)  It must be an IDE auto-adding.

KvanTTT:
Should work fine now.

parrt:
Great job, @KvanTTT !!!

PR #3752, opened 2022-06-19
Parallel lock free testing, remove potential deadlocks, cache static data, go to descriptor via test

### LINKED ISSUE #3370: Parallel runtime tests execution

### LINKED ISSUE #3756: Remove potential deadlocks in Java
`Referencing subclass from superclass initializer might lead to class loading deadlock` here:

* https://github.com/antlr/antlr4/blob/dev/runtime/Java/src/org/antlr/v4/runtime/atn/SemanticContext.java#L35
* https://github.com/antlr/antlr4/blob/dev/runtime/Java/src/org/antlr/v4/runtime/atn/PredictionContext.java#L28
* https://github.com/antlr/antlr4/blob/dev/runtime/Java/src/org/antlr/v4/runtime/RuleContext.java#L69

And spread changes to other runtimes.

### LINKED ISSUE #3757: Cache static data in ANTLR tool for performance (don't load every time from disk)

KvanTTT:
@parrt I've fixed Java and C# runtimes. If it's ok, I'll unify other runtimes as well (where multithreading is actual)

Now Java runtime tests are superfast since they are inprocess (30s instead of 1.5 minutes)!

Also, now it's very easy to navigate to test data via tests in IntelliJ! See the video:

https://user-images.githubusercontent.com/1150330/174623809-59da7c66-af67-419d-83c8-bdfa69e2026e.mp4

* Parallel runtime tests execution, fixes #3370
* Remove potential deadlocks in Java, fixes #3756
* Cache static data in ANTLR tool for performance, fixes #3757


runtime-testsuite/resources/descriptors/ParserExec/ListLabelForClosureContext.txt
```diff
@@ -7,6 +7,7 @@ Parser
 
 [grammar]
 grammar T;
+@members {<setOutStream()>}
```
parrt, line 10: i wonder if this could simply be added to output template for all tests?

runtime/Java/src/org/antlr/v4/runtime/atn/PredictionContext.java
```diff
@@ -67,19 +63,19 @@ protected PredictionContext(int cachedHashCode) {
 	}
 
 	/** Convert a {@link RuleContext} tree to a {@link PredictionContext} graph.
-	 *  Return {@link #EMPTY} if {@code outerContext} is empty or null.
+	 *  Return {@link EmptyPredictionContext#Instance} if {@code outerContext} is empty or null.
 	 */
 	public static PredictionContext fromRuleContext(ATN atn, RuleContext outerContext) {
-		if ( outerContext==null ) outerContext = RuleContext.EMPTY;
+		if ( outerContext==null ) outerContext = ParserRuleContext.EMPTY;
```
parrt, line 69: Are we sure this is always ParserRuleContext and not LexerRuleContext?

parrt: haha. there is no LexerRuleContext.  trying read this again after 10 years

runtime-testsuite/resources/descriptors/ParserExec/ListLabelForClosureContext.txt
KvanTTT: Actually it's actual only for Java runtime. It allows setting up custom output stream during function call via reflection: https://github.com/antlr/antlr4/pull/3752/files#diff-ac01582a52fe387cfa1e210df27a9f0b85a2c4470f71cb9e9701d26574455d29R153

```stg
setOutStream() ::= <<
private java.io.PrintStream outStream = System.out;

public void setOutStream(java.io.PrintStream outStream) {
	this.outStream = outStream;
}
>>
writeln(s) ::= <<outStream.println(<s>);>>
write(s) ::= <<outStream.print(<s>);>>
writeList(s) ::= <<outStream.println(<s; separator="+">);>>
```

For all other runtimes `setOutStream` is just empty, output is always being printed to console instead of custom stream.

Another way is to add support of custom output stream to runtime, but it requires changes in runtime and I'm not sure it's valuable for now.

runtime/Java/src/org/antlr/v4/runtime/atn/PredictionContext.java
parrt: actually, RuleContext might be what we need as an empty for it is general for lexer and parser I think. Why did you convert to ParserRuleContext?

runtime/Java/src/org/antlr/v4/runtime/atn/PredictionContext.java
```diff
@@ -67,19 +63,19 @@ protected PredictionContext(int cachedHashCode) {
...
 
 		// if we are in RuleContext of start rule, s, then PredictionContext
 		// is EMPTY. Nobody called us. (if we are empty, return empty)
-		if ( outerContext.parent==null || outerContext==RuleContext.EMPTY ) {
-			return PredictionContext.EMPTY;
+		if ( outerContext.parent==null || outerContext==ParserRuleContext.EMPTY ) {
```
parrt, line 73: same here. maybe shouldn't be ParserRuleContext?

runtime/Java/src/org/antlr/v4/runtime/atn/PredictionContext.java
KvanTTT: I didn't change the type, it always was `ParserRuleContext`: https://github.com/antlr/antlr4/pull/3752/commits/4b6918a52e7959935ee128b05275a6fad01cbf41#diff-20c6552f826df3e37bfcf31155a2e4360d2b4fea4fbb4ad6d27832946ff7718fL69 Also, there is no class named `LexerRuleContext` in ANTLR runtime library.

runtime/Java/src/org/antlr/v4/runtime/atn/PredictionContext.java
KvanTTT: Possible supertypes of: `RuleContext`:

image

Moreover, all concrete classes are supertypes of `ParseRuleContext`, not `RuleContext`. I don't understand meaning of `RuleContext`, it looks redundant.

tool/src/org/antlr/v4/Tool.java
```diff
@@ -139,10 +139,6 @@ public Option(String fieldName, String name, OptionArgType argType, String descr
 	protected boolean haveOutputDir = false;
 	protected boolean return_dont_exit = false;
 
-    // The internal options are for my use on the command line during dev
-    public static boolean internalOption_PrintGrammarTree = false;
```
parrt, line 143: i think i use this don't I (might leave in for debuggin)

runtime-testsuite/resources/descriptors/ParserExec/ListLabelForClosureContext.txt
parrt: Maybe we use a subclass of Parser / Lexer (for java runtime testing only) that had the setOutStream() defined?

runtime/Java/src/org/antlr/v4/runtime/atn/PredictionContext.java
parrt: ah!! cool.

runtime/Java/src/org/antlr/v4/runtime/atn/PredictionContext.java
parrt: I think lexer uses that. Parser tracks things to build parse tree.

parrt:
A huge amount of work here. thank you!

tool/src/org/antlr/v4/Tool.java
KvanTTT: I tried to minimize number of static fields since they could cause multithreading issues, it doesn't look safe. Also, it looks like temporary code that is used only during testing. I suggest either using temporary code without comitting or creating permanent option without static field.

KvanTTT:
> A huge amount of work here. thank you!

Thank you for your reviews and PR accepting :) I suggest changing other runtimes (moving `EMPTY` instances) to unify them with Java runtime (and probably to prevent potential deadlocks). I've complete it for Java and C#. It looks actual for C++, Swift., maybe Go.

parrt:
Are we sure the others have deadlock issues?  Not sure I have the energy to update. Maybe we can ping those target folks.

runtime-testsuite/resources/descriptors/ParserExec/ListLabelForClosureContext.txt
KvanTTT: Good idea, I'll try it.

KvanTTT:
> Are we sure the others have deadlock issues? Not sure I have the energy to update. Maybe we can ping those target folks.

Not sure (because they can use another initialization mechanizm) but not fully excluded (C# is very similar to Java). But I also like the idea of synchonized API across runtimes. I can do that.

parrt:
I had a random thought: Is any user code relying on EMPTY? I don't think so from a quick check but we want to be careful we don't break any user code as a result of this.

KvanTTT:
> I had a random thought: Is any user code relying on EMPTY? I don't think so from a quick check but we want to be careful we don't break any user code as a result of this.

Yes, we should be careful, but removing deadlocks is much more important than insignificant changing of API. And since we change Java code, code of other runtimes can be changed as well.

KvanTTT:
I've almost completed everything, but I also want to update documentation.

KvanTTT:
I think now it's ready to review and merge.

tool/src/org/antlr/v4/codegen/Target.java
```diff
@@ -92,14 +88,22 @@ public String getVersion() {
 	}
 
 	public STGroup getTemplates() {
+		String language = getLanguage();
+		STGroup templates = languageTemplates.get(language);
+
 		if (templates == null) {
```
parrt, line 94: We might have a race here. If one thread gets templates==null it will enter the IF. Another thread could also get null and enter the IF before the synchronization occurs. Maybe we should move the synchronization under the whole function or surround function body with `synchronized (languageTemplates)`?

tool/src/org/antlr/v4/tool/ErrorManager.java
```diff
@@ -216,25 +219,37 @@ public void emit(ErrorType etype, ANTLRMessage msg) {
...
-        }
-        else if ( url==null ) {
-            rawError("no such message format file "+fileName+" retrying with default ANTLR format");
-            setFormat("antlr"); // recurse on this rule, trying the default message format
-            return;
-        }
-        format = new STGroupFile(url, "UTF-8", '<', '>');
-        format.load();
+		STGroupFile loadedFormat = loadedFormats.get(formatName);
+		if (loadedFormat == null) {
```
parrt, line 223: Maybe the situations aren't races but it will still try to load multiple times technically if multiple threads reach the IF at the same time. Is this worth fixing? Since it's not critical code we might want to simply synchronize the method or synchronize on `loadedFormats`.

parrt:
In the middle of the review but wanted to comment that this is very well structured...each commit does a logical unit of work. thanks!  And amazing piece of refactoring overall 💯

runtime-testsuite/test/org/antlr/v4/test/runtime/RuntimeTests.java
```diff
@@ -120,10 +118,20 @@ private static void test(RuntimeTestDescriptor descriptor, RuntimeRunner runner)
 
 		FileUtils.mkdir(runner.getTempDirPath());
 
-		String sourceName = "org/antlr/v4/test/runtime/templates/" + targetName + ".test.stg";
-		String template = RuntimeTestUtils.getTextFromResource(sourceName);
-		STGroup targetTemplates = new STGroupString(sourceName, template, '<', '>');
-		targetTemplates.registerRenderer(String.class, rendered);
+		STGroup targetTemplates = cachedTargetTemplates.get(targetName);
+		if (targetTemplates == null) {
```
parrt: Just marking this as a similar situation... Not a big deal but multiple loads could occur in a multi threaded environment.

doc/antlr-project-testing.md
```diff
@@ -87,6 +87,8 @@ And the result of testing the entire subdirectory:
 
 All test are run in parallel both via maven and via IDE.
 
+In IntelliJ, it's very easy to go to source by right-click on any test and pressing `Jump to source` (F4).
```
parrt, line 90: "right-clicking"

runtime-testsuite/resources/org/antlr/v4/test/runtime/helpers/RuntimeTestParser.java
```diff
@@ -0,0 +1,14 @@
...
+import org.antlr.v4.runtime.TokenStream;
+
+public abstract class RuntimeTestParser extends Parser {
+	protected java.io.PrintStream outStream = System.out;
+
+	public RuntimeTestParser(TokenStream input) {
+		super(input);
+	}
+
+	public void setOutStream(java.io.PrintStream outStream) {
```
parrt, line 11: hooray! This saved a lot.

runtime-testsuite/test/org/antlr/v4/test/runtime/FileUtils.java
```diff
@@ -60,6 +61,18 @@ public static boolean moveFile(File sourceFir, String destDir, String fileName)
 		return file.renameTo(newFile);
 	}
 
+	public static void replaceInFile(Path sourcePath, String target, String replacement) throws IOException {
+		replaceInFile(sourcePath, sourcePath, target, replacement);
+	}
+
+	public static void replaceInFile(Path sourcePath, Path destPath, String target, String replacement) throws IOException {
+		String content = new String(Files.readAllBytes(sourcePath), StandardCharsets.UTF_8);
```
parrt, line 69: Is it guaranteed that all of the files we are reading will be UTF-8?  Do some of our test files read Japanese or Chinese characters for example?

doc/antlr-project-testing.md
```diff
@@ -87,6 +87,8 @@ And the result of testing the entire subdirectory:
```
parrt, line 87: Somethings not right in the version of the file you updated I think. The master branch has a bunch of extra information... Maybe I screwed up and did not pull this into the dev branch previously?  Here is master: https://github.com/antlr/antlr4/blob/master/doc/antlr-project-testing.md

I found it because I was looking for the part where maven still contest individual targets like:

```
$ mvn -Dtest=cpp.** test
```

tool/src/org/antlr/v4/codegen/Target.java
KvanTTT: I don't think it's a race here. In the code below `synhronized` is used:

```java
if (templates == null) {
	synchronized (languageTemplates) {
		templates = languageTemplates.get(language);
		if (templates == null) {
```

I use double-check for the best performance, locking is used only one time. It's a commonly used pattern: https://en.wikipedia.org/wiki/Double-checked_locking

Plain reading of the field is thread-safe.

KvanTTT: Ok, it looks like it's better to use `ConcurrentHashMap` in any case of multithreading: https://stackoverflow.com/q/33350489/1046374

tool/src/org/antlr/v4/tool/ErrorManager.java
KvanTTT: Multiple loading is not a big problem, and it doesn't occur in the code because lock (`synchonized`) is written in the code below. But `HashMap` should be replaced to `ConcurrentHashMap` if it's used in multi-threaded code.

KvanTTT: > Since it's not critical code we might want to simply synchronize the method or synchronize on loadedFormats.

runtime-testsuite/test/org/antlr/v4/test/runtime/FileUtils.java
KvanTTT: I haven't encountered the problems in out tests.

tool/src/org/antlr/v4/tool/ErrorManager.java
KvanTTT: Double-checking locking makes sure that the protected section is being calculated once and allows getting rid of excess locks.

parrt: The pattern does look familiar...yep. thanks. I missed the double check inside.. thanks.

tool/src/org/antlr/v4/codegen/Target.java
parrt: cool.

KvanTTT: It looks like the performance hit is insignificant: https://stackoverflow.com/a/12008833/1046374

Since `synchonized` on method looks clearer maybe it makes sense to use it everywhere.

doc/antlr-project-testing.md
KvanTTT: It was replaced by `mvn -Dtest=JavaRuntimeTests test` in https://github.com/antlr/antlr4/commit/f5843abaf5e7e2d763941417b04698e4b5d86b84#diff-8a1a1aeb8d87968c5aaee5b9a2c38e954b31e729767509607d65b6c87be49dcbL119 But `mvn -Dtest=java.** test` also works. I can restore the command with two asterisks if you want.

KvanTTT:
> In the middle of the review but wanted to comment that this is very well structured...each commit does a logical unit of work. thanks! And amazing piece of refactoring overall 💯

I'd like to thank my @JetBrains team. They taught me to keep structured git history and write very conscious code.

runtime-testsuite/resources/org/antlr/v4/test/runtime/helpers/RuntimeTestParser.java
KvanTTT: Definitely!

tool/src/org/antlr/v4/codegen/Target.java
parrt: yeah, much simpler to do on method level.

doc/antlr-project-testing.md
KvanTTT: Done.

parrt:
An amazing PR if I ever saw one thanks!!!  Merging...

PR #3754, opened 2022-06-21
[C++] Fix building shared library in mingw toolchain

Biswa96:
This fixes building shared library by defining proper symbols.
Previously, the symbols ANTLR4CPP_EXPORTS and ANTLR4CPP_STATIC
were defined for MSVC toolchain only. So, mingw toolchain was
excluded in that condition. Hence, the issue is solved by
moving the compile definitions in a general `if(WIN32)` check
which is independent of compiler toolchain.

Signed-off-by: Biswapriyo Nath <nathbappai@gmail.com>

### DIFF
--- a/runtime/Cpp/runtime/CMakeLists.txt
+++ b/runtime/Cpp/runtime/CMakeLists.txt
@@ -87,10 +87,14 @@ set(extra_share_compile_flags "")
 set(extra_static_compile_flags "")
 set(static_lib_suffix "")
 
-if(CMAKE_CXX_COMPILER_ID MATCHES "MSVC")
-  set(static_lib_suffix "-static")
+# For any toolchain targeting Win32 platform whether mingw or msvc
+if(WIN32)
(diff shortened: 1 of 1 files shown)


mike-lischke (approved):
Seems odd we check for Win 32 bit, given that most people are now on 64bit.

Biswa96:
The name "WIN32" does not mean 32 bit only. "WIN32" is true for any Microsoft Windows OS, irrespective of CPU architecture. See more https://cmake.org/cmake/help/latest/variable/WIN32.html

mike-lischke:
WIN32 means "Windows 32 bit". It doesn't matter if that is still available on all Windows versions (which is purely for compatibility reasons). But anyway...

parrt:
Have we reached a consensus on the way they should be?

mike-lischke:
I don't know CMake. Someone else must review this patch.

PR #3767, opened 2022-06-28
Swift runtime tests run on Windows!

KvanTTT:
* Include Swift runtime tests to CircliCI config
* Use crossplatform semaphore instead of unsafe raw `pthread_mutex_t` #3755
* Fir Swift runtime warnings
* Fir Swift build directory
* Unify Swift runtime with others
* Remove outdated files
* Simplify and clarity CircleCI scripts
* NET: 3.1 -> 6.0
* Minor fixes

@parrt I think it makes sense to disable MacOS agent since it's not reliable and it still doesn't work. I fixed Swift runtime tests and now they work on Linux (CircleCI). Also, Swift runtime tests work on Windows but only from IntelliJ (I haven't found out why they don't work from maven).


parrt:
Yeah, I'm not sure that github has gotten their Mac support to work very well but @ericvergnaud might have more information since he knows the set up.

KvanTTT:
Now it's ready for review. DCO is failed because I have one commit from another machine, but all CircleCI tests are green, including tests for Swift.

parrt:
Ok, I will likely get to this tomorrow as my brain is stuck elsewhere. This will give @ericvergnaud I chance to comment on Mac tests.

KvanTTT:
Sure, no rush. Anyway, I haven't touched Mac pipelines. It can be removed later.

.circleci/scripts/install-linux-csharp.sh
```diff
@@ -10,5 +8,4 @@ sudo apt-get update; \
   sudo apt-get update && \
   sudo apt-get install -y dotnet-sdk-3.1
 export PATH=$PATH:~/.dotnet
-echo "done installing .Net SDK"
```
parrt, line 13: Probably OK, but it looks like this is missing a new line on the end, which annoys UNIX.

parrt:
Looks absolutely amazing!  Wow Swift on circleci :)

PR #3775, opened 2022-07-05
Consolidate runtime tests

KvanTTT:
[TestVisitors.java](https://github.com/antlr/antlr4/blob/dev/runtime-testsuite/test/org/antlr/v4/test/runtime/java/api/TestVisitors.java) tests are actual for all runtimes not only for Java. But it uses quite runtime specific code. I suggest adding new sections `codeDeclaration` and `codeCall` to descriptor file that look the following way:

```
[type]
Parser

[grammar]
grammar VisitorBasic;

s
:	'A' EOF
;

A : 'A';

[start]
s

[input]
A

[codeDeclaration]
[Java]
static class VisitorBasicStringBaseVisitor extends VisitorBasicBaseVisitor<String> {
    @Override
    public String visitTerminal(TerminalNode node) {
        return node.getSymbol().toString() + "\n";
    }

    @Override
    protected String defaultResult() {
        return "";
    }

    @Override
    protected String aggregateResult(String aggregate, String nextResult) {
        return aggregate + nextResult;
    }
}

[CSharp]
class VisitorBasicStringBaseVisitor : VisitorBasicBaseVisitor<String>
{
	public override string VisitTerminal(ITerminalNode node)
	{
		return node.Symbol + "\n";
	}

	protected override string AggregateResult(string aggregate, string nextResult)
	{
		return aggregate + nextResult;
	}
}

...

[codeCall]
[Java]
outStream.print(new VisitorBasicStringBaseVisitor().visit(tree));

[CSharp]
Console.Write(new VisitorBasicStringBaseVisitor().Visit(tree));

...

[output]
[@0,0:0='A',<1>,1:0]
[@1,1:0='<EOF>',<-1>,1:1]
```

Moreover, with the new test infrastructure, it's possible to test code even without grammar:

```
[type]
Runtime

[codeDeclaration]
[Java]
static void surrogateRangeIntegerToCharArray() {
	IntegerList l = new IntegerList();
	// Java allows dangling surrogates, so (currently) we do
	// as well. We could change this if desired.
	l.add(0xDC00);
	char[] expected = new char[] { 0xDC00 };
	assertArrayEquals(expected, l.toCharArray());
}

[codeCall]
[Java]
surrogateRangeIntegerToCharArray();
```

@parrt what do you think about this?


parrt:
Hmm... I like the idea of consolidating the visitor tests but I worry about putting multiple languages in one descriptor file. Maybe there's a new way to think about this? Let me finish up the json stuff before I think about this :)  You'll note that I have moved the unit tests to the appropriate descriptors area :)

KvanTTT:
> I like the idea of consolidating the visitor tests

Visitors are only one example, but actually there are more runtime tests that should be consolidated if it's possible and reasonable.

> but I worry about putting multiple languages in one descriptor file. Maybe there's a new way to think about this?

Yes, I also thought about that and came up with the following considerations:

* **Separated files**
  * Pros: Support of highlighting and syntax checking
  * Cons: A lot of files related to one descriptor (1 descriptor + up to 10 files for every runtime)
* **Single file**
  * Pros: Everything in one descriptor, it's easier to manage code and get rid of missed tests
  * Cons: Lack of highlighting and syntax checking support

But we can support both schemes depending on circumstances. At first, try to read a section with code from a descriptor file. If such section is not found, search for runtime files by mask: `*.txt.<ext>` (for instance, `test.txt.cs` for C#).

If code size is small, then a single descriptor is more preferable than multiple files. Otherwise, if code size is quite big, then use multiple runtime files that are being searched by mask.

PR #3776, opened 2022-07-06
Don't make class SementicContext::Empty final.

hzeller:
Don't make class SementicContext::Empty final.
    
The class does not implement the pure virtual methods so can
not be considered 'final'. Clang 10 and 11 complain about this.
    
Context: https://github.com/chipsalliance/Surelog/issues/3081


hzeller:
CC @KvanTTT

hzeller:
The MacOSX CI issues seem to be unrelated (It fails attempting to  [download something](https://github.com/antlr/antlr4/runs/7218419721?check_suite_focus=true#step:1:27))

PR #3777, opened 2022-07-06
Fix CMake syntax for variable expansion

hs-apotell:
When using variables to compare (like in if clause) the variable
shouldn't be quoted. More details can be found at the link below:

https://cmake.org/cmake/help/latest/command/if.html#variable-expansion


hs-apotell:
Is there a reason (apart from not finding a volunteer) Windows CI for runtime Cpp was excluded? Was it attempted and have known issues?

ericvergnaud:
it just needs someone to give it a go

Issue #3783, opened 2022-07-12
[Infra] Check builds for runtime


hs-apotell:
@parrt I am attempting to get the CI builds working for the different runtime targets. I noticed that number of runtime builds are disabled for one reason or other.

Also, the workflow is using a self-hosted runner. Does it have to work with self-hosted machines? Looks like lot of the installation for needed software is happening as part of the workflow which is already installed on Github hosted machines. Builds will be a lot faster if GitHub hosted machines are used if things like VS isn't required to install on every build. If self-hosted is a requirement, I am going to need access to be able to kick off builds. At the moment, the builds are waiting forever because there are no self-hosted machines available to pick up the pending jobs.

I have already upgraded all the actions in the the Windows workflow to the latest and also got a few runtime tests working but all on Github hosted machines.

ericvergnaud:
The reason we're using self hosted is because at the time we built this, we could not get free hosting on circle-ci for macOS and windows

Despite the looks of it, the software is not installed on every build since the installer skips the install if it's already installed.

KvanTTT:
> The reason we're using self hosted is because at the time we built this, we could not get free hosting on circle-ci for macOS and windows

But self hosted CI works unstable. For instance, the latest build of dev failed: https://github.com/antlr/antlr4/runs/7222752630?check_suite_focus=true On the other hand, it looks like circle-ci works well and stable for the latest time.

@hs-apotell great! Please suggest a pull request to review changes.

ericvergnaud:
It was also working well with travis-ci until they pulled the plug.
I'd suggest not jumping on conclusions here. 
It's true that from time to time the self-hosted CI encounters network issues.
But if we consume too much free time from circle-ci, then we'll have no CI at all.

hs-apotell:
I can try creating a PR to see if the same setup would work for self-hosted. I have only verified my changes on GitHub hosted machines.

The longest build is about 40 minutes (cpp) which is well under limit of 6 hours. AFAIK for public repositories there is no maximum allotted time each month. Plus, if you are paying to host it, you might as well pay for time on GitHub and let others take care of keeping the machines up to date. Optionally, you can always have both and let them run in parallel.

hs-apotell:
> Please suggest a pull request to review changes.

#3785

[...]

PR #3785, opened 2022-07-12
Issue #3783: Fix CI check builds for windows platform

hs-apotell:

### DIFF
--- a/.github/scripts-windows/run-tests-cpp.cmd
+++ b/.github/scripts-windows/run-tests-cpp.cmd
@@ -1,6 +1,3 @@
-set ChocolateyInstall=C:\tools\chocolatey
-C:\ProgramData\chocolatey\bin\cinst.exe visualstudio2022-workload-vctools -y
-
 cd runtime-testsuite
 mvn -Dtest=cpp.** test
 cd ..
--- a/.github/scripts-windows/run-tests-python2.cmd
+++ b/.github/scripts-windows/run-tests-python2.cmd
@@ -1,3 +1,3 @@
 cd runtime-testsuite
-mvn -Dantlr-python2-exec="C:\Python27\python.exe" -Dtest=python2.** test
+mvn -Dantlr-python2-exec="%PYTHON_HOME%\python.exe" -Dtest=python2.** test
 cd ..
--- a/.github/scripts-windows/run-tests-python3.cmd
+++ b/.github/scripts-windows/run-tests-python3.cmd
@@ -1,3 +1,3 @@
 cd runtime-testsuite
-mvn -Dantlr-python3-exec="C:\Python310\python.exe" -Dtest=python3.** test
+mvn -Dantlr-python3-exec="%PYTHON_HOME%\python.exe" -Dtest=python3.** test
 cd ..
--- a/.github/workflows/windows.yml
+++ b/.github/workflows/windows.yml
@@ -7,69 +7,178 @@ on:
     branches: [ dev ]
 
 jobs:
-  build:
-    runs-on: [self-hosted, windows, x64]
+  cpp-builds:
+    name: "windows-cpp (${{ matrix.compiler }})"
+    runs-on: windows-latest
     strategy:
       fail-fast: false
       matrix:
-        TARGET: [java, python2, python3, javascript, csharp, dart, go, php]
+        compiler: [ cl, clang ]
+
+    steps:
+    - name: Setup Clang
+      if: matrix.compiler == 'clang'
+      uses: egor-tensin/setup-clang@v1
+      with:
+        version: 13
+        platform: x64
+        cygwin: 0
+
+    - name: Check out code
+      uses: actions/checkout@v2
+
+    - name: Build
+      shell: cmd
+      run: |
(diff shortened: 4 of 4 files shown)


.github/scripts-windows/run-tests-cpp.cmd
```diff
@@ -1,5 +1,5 @@
-set ChocolateyInstall=C:\tools\chocolatey
-C:\ProgramData\chocolatey\bin\cinst.exe visualstudio2022-workload-vctools -y
+REM set ChocolateyInstall=C:\tools\chocolatey
+REM C:\ProgramData\chocolatey\bin\cinst.exe visualstudio2022-workload-vctools -y
```
KvanTTT, line 3: Just remove these lines since there is no need for them.

.github/workflows/windows.yml
```diff
@@ -2,74 +2,115 @@ name: Windows
...
+        TARGET: [
+          cpp,
+          csharp,
+          dart,
+          go,
+          java,
+          javascript,
+          python2,
+          python3,
+          php,
```
KvanTTT, line 25: Have you also tried to set up Swift? In theory, it should work, see https://github.com/antlr/antlr4/pull/3767

KvanTTT (commented):
It looks great, thanks! Special thanks for setting up C++ on Windows!

.github/workflows/windows.yml
```diff
@@ -2,74 +2,115 @@ name: Windows
 
 on:
   push:
-    branches: [ dev ]
+    branches: [ dev, ci ]
```
KvanTTT, line 5: Is `ci` branch temporary?

.github/workflows/windows.yml
hs-apotell: Sure.

.github/workflows/windows.yml
hs-apotell: Yes, it's my forked repo branch to test the changes. I needed this since I couldn't get the forked repo to build my changes because of self-hosted setting.

hs-apotell:
My personal grudge with the current setup is that all logs are lost. When builds fail, all the information I have is that the build failed. The errors (and warnings) aren't propagated back to where the user has access to them. At the moment, there are numerous warnings in C++ build that needs to be resolved. But nobody is seeing them and so nobody is fixing them.

hs-apotell:
I have added two new workflows to use the native build system (cmake) to build cpp runtime so all the errors and warnings are shown. The configuration builds with both the cl and the clang (not clang-cl) compiler.

@KvanTTT Resolved all the above comments. And, no, I haven't tried the swift builds. I haven't not educated in this language so there are failures I wouldn't know what to do about them. I don't want to block this PR because of that potential failure. If required, I will follow up with another PR with swift specific change. This PR was focused primarily on getting the Windows + cpp builds to work and be verified on the CI.

.github/scripts-windows/run-tests-python3.cmd
```diff
@@ -1,3 +1,3 @@
 cd runtime-testsuite
-mvn -Dantlr-python3-exec="C:\Python310\python.exe" -Dtest=python3.** test
+mvn -Dantlr-python3-exec="%PYTHON_HOME%\python.exe" -Dtest=python3.** test
```
ericvergnaud, line 2: I disapprove this change.
This runs in an always on docker instance with both Python2 and Python3 installed
The python version at PYTHON_HOME is unknown

.github/scripts-windows/run-tests-python2.cmd
```diff
@@ -1,3 +1,3 @@
 cd runtime-testsuite
-mvn -Dantlr-python2-exec="C:\Python27\python.exe" -Dtest=python2.** test
+mvn -Dantlr-python2-exec="%PYTHON_HOME%\python.exe" -Dtest=python2.** test
```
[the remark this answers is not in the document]
ericvergnaud: We don't need this to be portable, currently it only runs on 4 docker instances on a physical box in my office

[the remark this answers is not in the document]
hs-apotell: A hardcoded path won't wok on Github cloud though. Python is not installed on that path.

Just may be - We might be on the different page here. Please follow comments on #3783 if you haven't yet.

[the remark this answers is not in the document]
KvanTTT: Is it possible to use just different environment variables `%PYTHON2_HOME%` and `%PYTHON3_HOME%` to make it workable everywhere?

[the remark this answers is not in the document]
hs-apotell: Sure. @ericvergnaud Would that resolve your concern?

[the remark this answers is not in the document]
ericvergnaud: It might, but we don't need to solve problems we don't have. The existing scripts work just fine, they're just not portable, which is not something we need at this point.
Sorry for the tone, it's great that you've gotten the cpp build to run on Windows. I suggest you restrict this PR to just that.

parrt:
Maybe an attempt to run on github servers will solve all of this or will this same problem occur?

hs-apotell:
Abandoned in favor of #3792

PR #3787, opened 2022-07-15
[C++] Remove libuuid dependency 

Technius:
This PR removes the libuuid dependency from the C++ runtime, since the libuuid headers are not referenced anywhere. Note that libuuid is only used as a dependency on Linux.

Context: I'm working on a C++ project that uses Nix to manage dependencies and was trying to use the CMake ANTLR config file (as generated by `-DANTLR_INSTALL=on`), but I was getting linker errors in my project as 1) libuuid is not checked for in the exported CMake config file; and 2) an unnecessary `-luuid` was added to the antlr-runtime target's interface link libraries.


parrt:
Makes me a bit nervous... @mike-lischke or @jcking might have an opinion.

mike-lischke:
Won't removing the lib cause linker errors for others because of missing symbols? How can that work without the linked uuid lib?

Technius:
As far as I can tell, there is no `#include <uuid/uuid.h>` in the runtime files or any generated C++ files, so I don't believe that a symbol referencing libuuid will be contained in the antlr C++ runtime object files. Perhaps I'm missing something? In any case, I tried the following experiment: I ran a fresh `ubuntu:latest` docker image, installed `libantlr4-runtime4.9`, and then ran `nm -D $(find /usr/lib -name libantlr4-runtime.so.4.9)`. The symbol list did not contain anything beginning with `uuid_`.

parrt:
Oh!  @mike-lischke Isn't this related to something we removed recently in the ATN serialization? We *used* to have a UUID in there but it has been removed from all target.

mike-lischke:
Yes, that's what I thought too. I did a quick search and cannot find any reference to uuid. So it looks like the lib reference is just an artifact and can be removed. IIRC @KvanTTT wrote the patch that removed the uuid stuff, right?

KvanTTT:
Probably this: https://github.com/antlr/antlr4/commit/54dbececd02474201139c421e5a27d5d7c722bba#diff-5d9ed1fbfe4d52ab8f90ed8d2ad3e1b0c9f9f606cb1c00da23be067cde69ca9a

parrt:
Thanks everyone

jcking:
You can remove any libuuid references. I must have missed one. C++ no
longer had any external dependencies, aside from th standard library of
course.

PR #3792, opened 2022-07-16
Issue #3783: CI Check Builds

hs-apotell:
Issue #3783: CI Check Builds

Adding workflow for Github hosted CI check builds. Supports all three
supported platforms with all runtime languages except swift.

Also, includes cpp native builds across all supported platforms i.e.
independent on the test run builds.

Every configuration build generates an artifact (archive of the entire
repository) and is uploaded as a result of the build (regardless of
success or failure).

### DIFF
--- a/.github/workflows/hosted.yml
+++ b/.github/workflows/hosted.yml
@@ -1,0 +1,290 @@
+name: antlr4
+
+on:
+  push:
+    branches: [ master, dev, hostedci ]
+  pull_request:
+    branches: [ master, dev ]
+
+jobs:
+  cpp-builds:
+    runs-on: ${{ matrix.os }}
+
+    strategy:
+      fail-fast: false
+      matrix:
+        os: [
+          macos-latest,
+          ubuntu-latest,
+          windows-latest
+        ]
+        compiler: [ clang, gcc ]
+        exclude:
+          - os: windows-latest
+            compiler: gcc
+        include:
+          - os: windows-latest
+            compiler: cl
+
+    steps:
+    - name: Install dependencies (Ubuntu)
+      if: startswith(matrix.os, 'ubuntu')
+      run: |
+        sudo apt-get update -qq
+        sudo apt install -y ninja-build
+
+    - name: Install dependencies (MacOS)
+      if: startswith(matrix.os, 'macos')
+      run: brew install ninja
+
+    - name: Setup Clang
+      if: (matrix.compiler == 'clang') && !startswith(matrix.os, 'macos')
+      uses: egor-tensin/setup-clang@v1
+      with:
+        version: 13
+        platform: x64
+        cygwin: 0
+
+    - name: Check out code
+      uses: actions/checkout@v2
+
+    - name: Use ccache
+      if: startswith(matrix.os, 'macos') || startswith(matrix.os, 'ubuntu')
+      uses: hendrikmuhs/ccache-action@v1
+      with:
+        key: ${{ matrix.os }}-${{ matrix.compiler }}
+
+    - name: Configure shell (Ubuntu)
+      if: startswith(matrix.os, 'ubuntu')
+      run: echo 'PATH=/usr/lib/ccache:'"$PATH" >> $GITHUB_ENV
+
+    - name: Configure shell (MacOS)
+      if: startswith(matrix.os, 'macos')
+      run: echo "PATH=$(brew --prefix)/opt/ccache/libexec:$PATH" >> $GITHUB_ENV
+
+    - name: Build (Windows)
+      if: startswith(matrix.os, 'windows')
+      shell: cmd
+      run: |
+        call "C:\Program Files\Microsoft Visual Studio\2022\Enterprise\VC\Auxiliary\Build\vcvars64.bat"
+
+        if "${{ matrix.compiler }}" EQU "cl" (
+          echo 'CC=cl' >> $GITHUB_ENV
+          echo 'CXX=cl' >> $GITHUB_ENV
+        ) else (
+          echo 'CC=clang' >> $GITHUB_ENV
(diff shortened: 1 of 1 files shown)


ericvergnaud:
Looks great: doesn't break self-hosted 
What's in the way for also having swift ?

ericvergnaud:
And let's find out how many free builds we get

parrt:
@hs-apotell Amazing. Pulling in!  I should merge that file into dev as well right so PRs and our dev gets tested?

parrt:
Oh, it's in dev now. sorry. thanks!

.github/workflows/hosted.yml
```diff
@@ -0,0 +1,290 @@
+name: antlr4
+
+on:
+  push:
+    branches: [ master, dev, hostedci ]
```
KvanTTT, line 5: Is `hostedci` branch required? It looks temporary.

.github/workflows/hosted.yml
```diff
@@ -0,0 +1,290 @@
...
+        cmake --build out/Debug --parallel
+
+        cmake -G Ninja -DCMAKE_BUILD_TYPE=Release -DANTLR_BUILD_CPP_TESTS=OFF -S . -B out/Release
+        cmake --build out/Release --parallel
+
+    - name: Prepare artifacts
+      if: always()
+      run: |
+        cd ${{ github.workspace }}/..
+        tar czfp antlr_${{ matrix.os }}_${{ matrix.compiler }}.tgz antlr4
```
KvanTTT, line 132: Is it possible to use more standard `zip` format instead of `tgz`?

.github/workflows/hosted.yml
```diff
@@ -0,0 +1,290 @@
...
+        extensions: mbstring
+        tools: composer
+
+    - name: Check out code
+      uses: actions/checkout@v2
+
+    - name: Checkout antlr PHP runtime
+      uses: actions/checkout@v2
+      with:
+        repository: antlr/antlr-php-runtime
```
KvanTTT, line 244: Probably the following condition:

```
if: matrix.target == 'php'
```

is missing. PHP runtime is only actual for PHP tests. It looks like the step is being executed for all runtimes: https://github.com/antlr/antlr4/runs/7371687695?check_suite_focus=true

KvanTTT (commented):
Is it also possible to include test for tool itself? (not runtime) See https://app.circleci.com/pipelines/github/antlr/antlr4/1885/workflows/cb672c7b-206e-4390-b823-0a3f6f6454da/jobs/20238 for instance.

KvanTTT:
Am I understand correctly that now CircleCI is not required at all since GitHub supports all OS.

hs-apotell:
I would let the current change bake for about a week and follow up later with others (swift and tool).

I would like to fix all the compile warnings in native build and also update the builder and test runner to post the logs. If you can assist with the latter (it's Java code) that would be great.

.github/workflows/hosted.yml
hs-apotell: I used this branch for local testing in my fork without having to modify dev branch. It plays no role in upstream repo.

.github/workflows/hosted.yml
hs-apotell: I am not sure if the default VM image for all platforms have zip installed or not. Will have to investigate.

BTW, _standard_ is very much platform dependent. On Windows, zip but on Posix it would be tar/tgz.

.github/workflows/hosted.yml
hs-apotell: Fixed in  #3797

parrt:
Hi @hs-apotell it's getting confusing with all of the hosted versus Remote testing... It looks like it's more reliable hosted than remote. Could you check the actions and see if you concur?  @ericvergnaud are you OK if we turn off the remote CI?

Something is screwy because some of the targets are now failing but it's hard to tell looking at all of the actions what's going on so I wanted to simplify.  Seems we get a lot of timeouts from maven :(

parrt:
(PS: i just tweaked to use 1.19 Go)

parrt:
Hi. E.g., the php run just failed again due to timeout:

https://github.com/antlr/antlr4/runs/7821060959?check_suite_focus=true

```
Run stCarolas/setup-maven@v4.4
  with:
    maven-version: 3.5.4
  env:
    JAVA_HOME: /Users/runner/hostedtoolcache/Java_Zulu_jdk/11.0.16-8/x64
downloading https://archive.apache.org/dist/maven/maven-3/3.5.4/binaries/apache-maven-3.5.4-bin.tar.gz
Error: connect ETIMEDOUT 138.[2](https://github.com/antlr/antlr4/runs/7821060959?check_suite_focus=true#step:5:2)01.1[3](https://github.com/antlr/antlr4/runs/7821060959?check_suite_focus=true#step:5:3)1.13[4](https://github.com/antlr/antlr4/runs/7821060959?check_suite_focus=true#step:5:4):443
```

parrt:
Trying to rerun failed: https://github.com/antlr/antlr4/actions/runs/2852943941

parrt:
It looks like the time out did not happen upon rerunning it. I'm going to turn off the remote actions as it looks like we are good to go with hosted.

parrt:
I'm saying the opposite and agreeing with you. :) I have turned off the remote and left the hosted running, which only fail due to time out at the moment.

parrt:
Sounds good. Sorry for the confusion. self-hosted vs hosted / github-hosted.  Thanks for the assist.

parrt:
Done.

parrt:
@hs-apotell looks like we might have issues with master:

parrt:
Ok, sounds good.

PR #3797, opened 2022-07-23
Include swift & tool verification in CI workflow

hs-apotell:
Include swift & tool verification in CI workflow

* Adding CI for swift runtime
* Adding CI for tool verification (for java runtime target only)
* Also fixed a few minor issues identified in change #3785 

This change is a follow up to #3785.

Known Issue: There seems to be some sort of timeout issue with setting up Maven using Apache. It is reported here - https://github.com/stCarolas/setup-maven/issues/16

Attn: @KvanTTT @parrt


parrt:
thanks!

PR #3800, opened 2022-08-01
Add Go accessor for IntervalSet.intervals

JamesRTaylor:
Signed-off-by: James Taylor <jamestaylor@apache.org>

Unlike the Java version, in the Go-based version of interval_sets, there's no way to access the IntervalSets.intervals field. This field is useful when using Parser.getExpectedTokens(). The alternative is to use IntervalSets#StringVerbose and parse the string returned (which is much more expensive).


parrt:
Makes sense. thanks.

jimidle:
In Go, we should either make the field visible or use an accessor called Intervals() rather than GetIntervals(). However, I guess we can close #3801 now?

PR #3802, opened 2022-08-01
Fix #3508: Document the $parser attribute and its use in target-agnostic grammars.

### LINKED ISSUE #3508: Change "$self" to "$parser" in "target agnostic grammars" instructions
This is pretty critical.

The instructions for ["Target agnostic grammars"](https://github.com/antlr/antlr4/blob/896f90b22c485dfa5082c2e8fa8b2ec1326c726a/doc/python-target.md?plain=1#L109) are not correct. Following the [third option](https://github.com/antlr/antlr4/blob/896f90b22c485dfa5082c2e8fa8b2ec1326c726a/doc/python-target.md?plain=1#L123), after adding field "self" to a base class, and using `$self` in an action, the Antlr tool will output `"unknown attribute reference self in $self"`. In fact, there is no "$self" attribute. If one changes this to "self" (without the dollar sign), then the instructions would "work". But, as fate sometimes works, I tried something else.

If one changes `$self` to `$parser`, skip adding a field "self" to the base class altogether, "target agnostic" for a parser grammar will work for all but two targets! Note, this is only good for parser grammars. I don't think there is an equivalent "$lexer" for lexer actions, which really would be needed for actions in the lexer grammar. Or, really introduce "$self" for either lexer or parser actions.

The [documentation for attributes](https://github.com/antlr/antlr4/blob/master/doc/actions.md#parser-rule-attributes) really needs to include a discussion of "$parser". I don't know why it's missing because this attribute works.

Note, it's currently impossible to get a grammar with actions with a common syntax to work across all targets--at least for 4.9.3--but it is close using "$parser"--at least for a parser grammar. I would actually call the technique "not-quite target-agnostic grammars" because the common syntactic approach does not work for Cpp and PHP where there are include file issues and the non-conversion of "." to "->" for a pointer reference (see attached zip file for a proof of the technique working for CSharp, Dart, Go, Java, JavaScript, Python3). I have not tested "target agnostic grammars" for the old Antlr4cs port, which people shouldn't even be using at this point.
[target-agnostic2.zip](https://github.com/antlr/antlr4/files/7935242/target-agnostic2.zip)

RossPatterson:
Since I had to work all this out too, and found @kaby76's description in #3508 so helpful, I made an attempt at closing it out.

### DIFF
--- a/doc/actions.md
+++ b/doc/actions.md
@@ -98,6 +98,7 @@ returnStat : 'return' expr {System.out.println("first token "+$start.getText());
 |start|Token|The first token to be potentially matched by the rule that is on the main token channel; in other words, this attribute is never a hidden token. For rules that end up matching no tokens, this attribute points at the first token that could have been matched by this rule. When referring to the current rule, this attribute is available to any action within the rule.|
 |stop|Token|The last nonhidden channel token to be matched by the rule. When referring to the current rule, this attribute is available only to the after and finally actions.|
 |ctx|ParserRuleContext|The rule context object associated with a rule invocation. All of the other attributes are available through this attribute. For example, `$ctx.start` accesses the start field within the current rules context object. It’s the same as `$start`.|
+|parser|Parser|The parser itself.  This attribute can be used, for example, to invoke a method defined in the parser's `@members` section from a semantic predicate.|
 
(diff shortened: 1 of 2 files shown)


ericvergnaud:
@parrt blessed

PR #3809, opened 2022-08-09
init Reproducible Builds

hboutemy:
Signed-off-by: Hervé Boutemy <hboutemy@apache.org>

equivalent to https://github.com/antlr/antlr3/pull/210 that I did for ANTLR3

given ANTLR4 uses ANTLR3, to get a full reproducible build for ANTLR4, ANTLR3 version will require to be upgraded once https://github.com/antlr/antlr3/pull/209 has been merged and a release done

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -97,6 +97,7 @@
 	<properties>
 		<project.build.sourceEncoding>UTF-8</project.build.sourceEncoding>
 		<project.reporting.outputEncoding>UTF-8</project.reporting.outputEncoding>
+		<project.build.outputTimestamp>10</project.build.outputTimestamp>
 		<antlr.testinprocess>true</antlr.testinprocess>
 		<maven.compiler.source>11</maven.compiler.source>
 		<maven.compiler.target>11</maven.compiler.target>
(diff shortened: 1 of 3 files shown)


parrt:
This seems harmless enough.

pom.xml
```diff
@@ -163,6 +164,16 @@
 					<groupId>org.apache.maven.plugins</groupId>
 					<artifactId>maven-javadoc-plugin</artifactId>
 				</plugin>
+				<plugin>
```
parrt, line 167: Why do we need the extra dependencies here? Does the outputTimestamp thing need these?

hboutemy: @parrt
it's not really new dependencies: it's an update of plugin versions for plugins that are already used (default Maven lifecycle for jars), to use plugins that support Reproducible Builds (use the outputTimestamp)

hboutemy:
@parrt I added a second commit to workaround antlr3 non-reproducible output, as discussed in https://github.com/antlr/antlr3/pull/209
now, the build for ANTLR4 is fully reproducible on my side, as tested with:
```
mvn clean install -DskipTests && mvn clean package artifact:compare -DskipTests
```

PR #3817, opened 2022-08-12
Create emulation tracer

parrt:
Create new tracer class for ATN simulation; currently it just tracks DFA State construction and edge construction.   The goal is to create a machine readable traits from simulation so that we can compare across target.

See https://github.com/antlr/antlr4/discussions/3814

Rationale to create new classes in the runtime: highlights the target developers need to generate output upon some key events.

Happy to get thoughts from @kaby76 about suitability of this output.


parrt:
@ericvergnaud py2/3 output now same-same as Java's and C++'s :) woot!

parrt:
closing in favor of https://github.com/antlr/antlr4/pull/3957

PR #3819, opened 2022-08-13
Rm remote actions

parrt:
Looks like hosted github actions I'm working and so I'm going to try turning off the remote actions; we can always back this out if necessary and turn them back on. See https://github.com/antlr/antlr4/pull/3792


parrt:
Hmm... I see this running circleci but not the hosted, despite the hosted.yml being there.

PR #3820, opened 2022-08-14
Caching support

hs-apotell:
Add caching support for maven & dependencies. Also, include caching for
cpp builds using actions/ccache.

Builds are more reliable (avoids the archive.apache server which
intermittently reports timeouts) and also significantly improves the
overall builds times (down from 46 mins to 28 mins).

The slowest part of the build now is the Windows+cpp builds because
there is no reliable cache implementation yet. MacOS+cpp (65% cache hit) is
also relatively slow compared to Ubuntu+cpp (99% cache hit).

### DIFF
--- a/.github/workflows/hosted.yml
+++ b/.github/workflows/hosted.yml
@@ -172,6 +172,20 @@ jobs:
             target: swift
 
     steps:
+      # Check out the code before setting the environment since some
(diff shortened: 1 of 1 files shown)


parrt:
pulled into master as well.

Issue #3823, opened 2022-08-15
Disable failing CI tests in master


parrt:
@hs-apotell wrote:

> Ahh there are other changes that will have to come to master for the builds to succeed here. Master & dev have digressed too far. The reported issues in the build failures are all fixed in dev.

> Is a release due any time soon? If yes, then I would leave master as is and let the merge take care of it. If not, then I can disable the failing builds for master. Fixing those builds in master alone would be throw away work since it's going to get overwritten by dev anyways.

I'm not sure when we will do a release but hopefully soon. Regardless, if we could disable the stuff in master that would be useful for now. Thanks! It won't let me assign to you yet as I think it's waiting for your acceptance of membership but @hs-apotell ping :) thanks again for all your help

parrt:
Fixed by https://github.com/antlr/antlr4/pull/3824

PR #3824, opened 2022-08-15
Issue #3823: Temporarily disable a few tests on CI

hs-apotell:
Issue #3823: Temporarily disable a few tests on CI

The tests are currently failing. The underlying issues have been fixed
on dev and so the builds will be turned back with the next release.


parrt:
Thanks @hs-apotell !!!

PR #3825, opened 2022-08-17
Make swift runtime available through cocoapods

spinnerok:
This PR adds a CocoaPod for Swift runtime. I saw in the documentation that it is recommended to compile the runtime from sources together with generated files for grammar but CocoaPod provides more flexibility in case we want to distribute a library which depends on Antlr4.

Signed-off-by: spinnerok <serhiy.bobyr@zoho.eu>

### DIFF
--- a/.github/workflows/hosted.yml
+++ b/.github/workflows/hosted.yml
@@ -3,6 +3,8 @@ name: antlr4
 on:
   push:
     branches: [ master, dev, hostedci ]
+  tags:
+      - '*'
   pull_request:
     branches: [ master, dev ]
 
@@ -179,6 +181,15 @@ jobs:
     - name: Check out code
       uses: actions/checkout@v2
 
+    - name: Get tag
+      id: tag
+      run: |
+        if [[ $GITHUB_REF == 'refs/tags/'* ]]; then
+            echo "TAG=${GITHUB_REF/refs\/tags\//}" >> $GITHUB_ENV
+        else
+            echo "TAG=" >> $GITHUB_ENV
+        fi
+
     - name: Checkout antlr PHP runtime
       if: matrix.target == 'php'
       uses: actions/checkout@v2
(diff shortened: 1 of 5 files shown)


.github/workflows/hosted.yml
```diff
@@ -3,6 +3,8 @@ name: antlr4
 on:
   push:
     branches: [ master, dev, hostedci ]
+  tags:
```
spinnerok, line 6: The push of a tag will trigger a build which may be not a good solution because this is deviating from the main flow of releases for the other platforms. May be there is a better way to do it.

.github/workflows/hosted.yml
```diff
@@ -331,3 +342,14 @@ jobs:
       with:
         name: antlr_${{ matrix.os }}_${{ matrix.target }}
         path: antlr_${{ matrix.os }}_${{ matrix.target }}.tgz
+
+    - name: Publish CocoaPod
+      if: (env.TAG != null) && startsWith(matrix.os, 'macos') && (matrix.target == 'swift')
+      env:
+        COCOAPODS_TRUNK_TOKEN: ${{ secrets.COCOAPODS_TRUNK_TOKEN }}
```
spinnerok, line 349: `COCOAPODS_TRUNK_TOKEN` should be added to secrets on github and it should be obtained for the author email address. More information can be found here https://github.com/marketplace/actions/deploy-to-cocoapod-action

parrt: hiya! This seems kind of complicated and I have 10 languages I have to support. Is there anyway to enable the functionality you live without me having to register secrets and register with pod etc?  I dont' have `pod` command available at the moment from the commandline it appears, despite having xcode.

spinnerok: Unfortunately, to have it available on cocoapods someone has to generate a secret and publish it. I have published Antlr4 already as an unofficial pod to see how it will work and test it. I did it from my fork because I needed to add a secret to the project settings on github. It also requires adding an account for which the secret has been generated to the list of authors. 

So I can also publish a proper Antlr4 pod and maintain it but only from my fork where I have access to project setting. It also should be possible to publish it from Travis or another CI/CD tool but it still requires to set a secret somehow.

A resulting unofficial podspec is available here:  https://github.com/CocoaPods/Specs/blob/master/Specs/3/c/1/Antlr4-Unofficial/4.10.1.1/Antlr4-Unofficial.podspec.json

parrt: Well I think at this point I'd rather not and yet another artifact or anything I need to manage so it would be great if you could keep doing it from your fork.

PR #3829, opened 2022-08-20
Feature/newhash

jimidle:
fix: #3718 Rework of all Hash() and Equals() methods, and replace broken collections with generics

- Implement new collections using generics that implement the functionality
   required by the Java runtime in a more idiomatic Go way.
- Fix Hash() and Equals() for all objects in the runtime
- Fix getConflictingAlts so that it behaves the same way as Java, using a
   new generic collection
- Replaces the use of the array2DHashSet, which was causing unneeded memory
   allocations. Replaced with generic collection that allocates minimally
   (though, I think I can improve on that with a little analysis).

This PR fixes quite a few hidden bugs that were only exposed after replacing the various
implementations of collections with the new generic implementations.

Jim Idle - jimi@idle.ws

Signed-off-by: Jim.Idle <jimi@gatherstars.com>


jimidle:
I am not sure why there were merge conflicts from dev. I resolved them, but I guess my rebase and the original merge didn't quite work. Anyway, I will lose this branch once you merge it in to dev, and start with a new feature branch for a few improvements I have in mind.

Issue #3758, opened 2022-06-21
Grammar Name Conflict Golang with SPARQL issue


saucesaft:
Hi! So im trying to build a simple sparql parser for golang. Got the grammar for it from here: https://github.com/antlr/grammars-v4/tree/master/sparql

Then i ran this command to generate the code:
`
java -jar antlr-4.10.1-complete.jar -Dlanguage=Go -o parser Sparql.g4`

Now, after i included it into a Go program, without even using it, it errors out with this:
```# dkg-client-go/parser
parser/sparql_parser.go:9806:9: cannot use s (variable of type *RdfLiteralContext) as type antlr.RuleContext in return statement:
	*RdfLiteralContext does not implement antlr.RuleContext (wrong type for String method)
		have String() IStringContext
		want String([]string, antlr.RuleContext) string
parser/sparql_parser.go:9829:13: cannot use NewRdfLiteralContext(p, p.GetParserRuleContext(), p.GetState()) (value of type *RdfLiteralContext) as type IRdfLiteralContext in assignment:
	*RdfLiteralContext does not implement IRdfLiteralContext (wrong type for String method)
		have String() IStringContext
		want String([]string, antlr.RuleContext) string
```

From what i've seen in another issue, i think this might be related to some grammar name conflicts with the code generation, but i dont know how to fix it. If someone recognizes the problem, i'll be glad to make the changes myself and open a pull request. 

Thanks :)

kaby76:
Correct. This is the result of a "symbol conflict".

After looking over the generated files, and finding nothing obvious, I decided to check for the problem using automated means. I created a script (see below) that creates a new grammar with  all parser rule symbols with a trailing underscore. It renames a symbol back to the original name without the underscore, and test how it builds.

The problem is "string". A workaround is to rename it to something else, like "string_". This needs to be added to the escaping code in the Antlr tool for Go. @KvanTTT 

```
#!/bin/sh

names=`trparse Sparql.g4 | trxgrep ' //parserRuleSpec/RULE_REF/text()' | sort -u | dos2unix | grep -v -E -e '_$' | grep -v -E -e '^query$'`
newnames=""
for i in $names
do
	newnames="$newnames"$i,$i"_;"
done
newnames="${newnames::-1}"
trparse Sparql.g4 | trrename -r "$newnames" | trsponge -c
for i in $names
do
	fixed=$i'_'
	echo Testing $fixed '=>' $i
	rm -rf testit
	mkdir testit
[...]

PR #3831, opened 2022-08-22
fix: #3758 Allow for string being a keyword and fix go template to us…

jimidle:
fix: #3758 Consistently apply `.escapedName` instead of just `.name` for the go target. Add `string` to reserved words

  - The Go code generation template mostly ignored the use of `escapedName`, hence it would generate rules and so on
     that were keywords in go.
  - Added `string` as a keyword for go

### DIFF
--- a/runtime/Go/antlr/go.mod
+++ b/runtime/Go/antlr/go.mod
@@ -1,3 +1,5 @@
 module github.com/antlr/antlr4/runtime/Go/antlr
 
 go 1.18
+
+require golang.org/x/exp v0.0.0-20220722155223-a9213eeb770e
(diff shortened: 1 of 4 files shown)


KvanTTT:
Could you please add a test that covers the case?

jimidle:
The go.mod file is updated to include the go runtime extensions and the go.sum file is added to the repo, which I had not noticed was missing.

runtime/Go/antlr/go.mod
```diff
@@ -1,3 +1,5 @@
 module github.com/antlr/antlr4/runtime/Go/antlr
 
 go 1.18
+
+require golang.org/x/exp v0.0.0-20220722155223-a9213eeb770e
```
parrt, line 5: You might have fixed later...do we want to require exp stuff?

jimidle: Yes - I added this and go.sum as it is required, though go works it out. We do need to require it, but this package is part of the runtime that comes with go.

jimidle:
I will add a string rule to the existing keyword test. I should have done that with this PR.

jimidle:
@parrt I thin we can close this now?

Issue #3653, opened 2022-04-13
C#, Antlr4.Runtime.Standard 4.10.0, invalid signature


podprad:
Hello,

I've upgraded Antlr4.Runtime.Standard from 4.8.0 to 4.10.0 and now my web project (that run on IIS) complains about invalid strong name signature:
> Could not load file or assembly 'Antlr4.Runtime.Standard' or one of its dependencies. Strong name signature could not be verified.  The assembly may have been tampered with, or it was delay signed but not fully signed with the correct private key. (Exception from HRESULT: 0x80131045)

I verified the signature manually by sn.exe:

> where sn.exe
> C:\Program Files (x86)\Microsoft SDKs\Windows\v10.0A\bin\NETFX 4.8 Tools\sn.exe

> sn -v Antlr4.Runtime.Standard.dll 

Old version 4.8.0:

> netstandard1.3
> Assembly 'Antlr4.Runtime.Standard.dll' is valid

> net35
> Assembly 'Antlr4.Runtime.Standard.dll' is valid

New version 4.10.0:

> netstandard2.0
> Failed to verify assembly -- Strong name validation failed.

So it looks there is really something wrong with the signature.

I used the DLL included in the nuget package:
> https://www.nuget.org/packages/Antlr4.Runtime.Standard/

Best regards,
Michael

KvanTTT:
It's probably realted to #3360

cam-m:
Seems to also affect v4.10.1.

Any chance of a patch version bump with the correctly signed package?

ItCommitsTheCode:
This issue starts precisely at 4.9.3 and continues through to current releases.  The last, correctly-signed version is 4.9.2, based on my testing.

Barteling:
> This issue starts precisely at 4.9.3 and continues through to current releases. The last, correctly-signed version is 4.9.2, based on my testing.

I believe so too, 

Any change to solve this in a patch?

ericvergnaud:
There is no bug fix in C# runtime > 4.9.2 so you can stick to that version until the code signing issue is fixed in the next release

Barteling:
That is clear, thanks for all the awesome work you do 

Just wondering if you found a cause, because I reported it already for 4.9.3?  #3360

Issue #3360, opened 2021-11-20
antlr.runtime.standard 4.9.3 invalid strong name.


Barteling:
Using the latest nuget package 4.9.3 I get invalid strong name exceptions. 
Both the .NETStandard 2.0 and the NET45 version do not seem to be correctly signed. 
4.9.2 seems to be ok, verified with sn.exe

ericvergnaud:
@Barteling I don't think there is any fix in this version for C#, only released for versioning consistency, so you should stick to 4.9.2 until 4.9.4 is available
@parrt I guess this happened because you and I don't follow the same release process (I released 4.9.3, you released 4.9.2) ? When the next release gets prepared, let's make sure we align to avoid this issue.

parrt:
Hi guys. Yep, sorry about that.  I think i grab the artifact from appveyor as that's the limit of my abilities with C# haha.

parrt:
I guess I'll add link to 4.9.4

ericvergnaud:
> Hi guys. Yep, sorry about that. I think i grab the artifact from appveyor as that's the limit of my abilities with C# haha.

4.9.3 was released to Nuget by me so it has nothing to do with AppVeyor

KvanTTT:
It looks like we encountered a similar issue before: https://github.com/antlr/antlr4/issues/2511

parrt:
@ericvergnaud @Barteling @KvanTTT  Can you guys help me figure out how to get the proper artifact and with the right signatures etcetera? I'm going to build and release 4.10 soon.   it seems to be problematic on my mac so I've been downloading the artifact created at circleci.

parrt:
Will close. Updated doc from @ericvergnaud notes to sign properly.

PR #3836, opened 2022-08-23
feat: Bump poms to use 4.11 Snapshot

jimidle:
Move to new version numbers as Go target has had significant changes and needs a new minor version

### DIFF
--- a/antlr4-maven-plugin/pom.xml
+++ b/antlr4-maven-plugin/pom.xml
@@ -8,7 +8,7 @@
   <parent>
     <groupId>org.antlr</groupId>
     <artifactId>antlr4-master</artifactId>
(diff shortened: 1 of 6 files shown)


ericvergnaud:
@parrt as part of releasing 4.11, we also need to cater for the .net code signing issue, see #3653

parrt:
We just have to follow your process in doc right @ericvergnaud? (per https://github.com/antlr/antlr4/issues/3360)

ericvergnaud:
It seems the publishing needs to be done from a winter box otherwise the naming is incomplete

parrt:
Ok, i have two windows boxes (turned off) sitting here :)  Will be painful.

Issue #2016, opened 2017-09-18
Cpp, Go, JavaScript, Python2/3: Template rendering error.


AndreiPashkin:
Here is an example grammar:
```
grammar Test;

expression: op=NOT args+=expression
          | args+=expression (op=AND args+=expression)+
          | args+=expression (op=OR args+=expression)+
          | IDENTIFIER
          ;
AND : 'and' ;
OR : 'or' ;
NOT : 'not' ;
IDENTIFIER : [a-zA-Z_][a-zA-Z0-9_]* ;
WS : [ \t\r\n]+ -> skip ;
```
For Cpp, Go, JavaScript, Python2/3 target languages version 4.7 throws such errors while generating code:
```
error(32):  code generation template recRuleAltStartAction has missing, misnamed, or incomplete arg list; missing isListLabel
```

For the rest of the languages it works.

I believe that the error is in the templates since in the `recRuleAltStartAction` templates for the problematic languages argument `isListLabel` is not present while in the templates of the ok-languages it is present.

ericvergnaud:
Hi,

this grammar also fails in the Java reference implementation, although at a later stage:

./var/folders/7z/rdw_873x3_s5zr8pb81k9z7c0000gn/T/BaseJavaTest-main-95208223142060/TParser.java:173: error: incompatible types: ExpressionContext cannot be converted to List<ExpressionContext> 
						_localctx.args = _prevctx;
						                 ^
/var/folders/7z/rdw_873x3_s5zr8pb81k9z7c0000gn/T/BaseJavaTest-main-95208223142060/TParser.java:206: error: incompatible types: ExpressionContext cannot be converted to List<ExpressionContext>
						_localctx.args = _prevctx;

Not sure if this is bug in ANTLR or in the grammar, but I’d suggest getting this right with the reference implementation first.

@parrt can you have a look (I’ve added a test descriptor, see enclosed)

Eric

parrt:
With the Java target, all is ok:

```bash
varmint:/tmp $ alias a4.7.1
alias a4.7.1='java -cp "/Users/parrt/.m2/repository/org/antlr/antlr4/4.7.1-SNAPSHOT/antlr4-4.7.1-SNAPSHOT-complete.jar:$CLASSPATH" org.antlr.v4.Tool'
varmint:/tmp $ a4.7.1 Test.g4
varmint:/tmp $ 
```

Looks like Swift, Java, C# are ok and have this argument. Probably should add to others.

[...]

PR #3837, opened 2022-08-23
Test for "List label inside closure"

KvanTTT:
#2016


KvanTTT:
Moreover, my test does not cover correct execution. C# tests are green but it looks like list label works incorrectly. I'll think about the full coverage.

PR #3838, opened 2022-08-26
Replace smart-quote with single-quote in code examples

timmc:
This seems to be all the affected locations. I left smart-quotes alone in prose like `Here’s`.

Looks like my editor added a trailing newline on files that were missing one. I can back that out, but it's probably for the best. :-)


parrt:
thanks!

PR #3843, opened 2022-08-27
Augment error message during testing to include full cause of problem.

parrt:
Cause of compile/link errors were hidden during tests; bad in C++ as can't see anything.


parrt:
See any issues @KvanTTT ?

KvanTTT:
Sorry, didn't notice the message at the time. Interesting, I thought `toString` returns exhaustive message.

parrt:
no prob.

PR #3847, opened 2022-08-29
GitHub Workflows security hardening

sashashura:
This PR adds explicit [permissions section](https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#permissions) to workflows. This is a security best practice because by default workflows run with [extended set of permissions](https://docs.github.com/en/actions/security-guides/automatic-token-authentication#permissions-for-the-github_token) (except from `on: pull_request` [from external forks](https://securitylab.github.com/research/github-actions-preventing-pwn-requests/)). By specifying any permission explicitly all others are set to none. By using the principle of least privilege the damage a compromised workflow can do (because of an [injection](https://securitylab.github.com/research/github-actions-untrusted-input/) or compromised third party tool or action) is restricted.
It is recommended to have [most strict permissions on the top level](https://github.com/ossf/scorecard/blob/main/docs/checks.md#token-permissions) and grant write permissions on [job level](https://docs.github.com/en/actions/using-jobs/assigning-permissions-to-jobs) case by case.

### DIFF
--- a/.github/workflows/hosted.yml
+++ b/.github/workflows/hosted.yml
@@ -6,6 +6,9 @@ on:
   pull_request:
     branches: [ master, dev ]
 
+permissions:
+  contents: read
+
 jobs:
   cpp-builds:
     runs-on: ${{ matrix.os }}


sashashura:
Do I need to do it against dev?

parrt:
Hi! That change seems like a reasonable idea and simple enough. I guess we protecting against an action that somehow managed to overwrite the repository or related artifacts?

sashashura:
It protects from the repository compromise in case of a toolchain/dependency compromise. Of couse if your dependency is compromised you may build a compromised release. Although release builds happen not often, in this attack scenario, if a dependency is compromised for a short period of time, before it is noticed by maintainers, it may be enough to the workflow to start (it starts on every push). If permissions are unrestricted, the compromised tool/dependency may modify the repository content and even previously released artifacts.
To put it the other way, all-write permissions are like sudo. Why running builds with sudo unnecessarily? By restricting permissions the potential impact of compromised dependency is reduced.

PR #3848, opened 2022-08-30
fix: #2016 Fix Go template list reference, go runtime and got test te…

jimidle:
fix: #2016 Fix runtime, test templates and go codegen template

The go Runtime was confused about the EmprtyPredictionContext vs SIngletonPredictionContext.

For now, I have fixed up the go code for merge(), but I think longer term I will have to change the way the structures for
prediction context are organized, which is a piece of work that isn't worth it for v4.11. I will come back to that.

### DIFF
--- a/runtime-testsuite/resources/org/antlr/v4/test/runtime/templates/Go.test.stg
+++ b/runtime-testsuite/resources/org/antlr/v4/test/runtime/templates/Go.test.stg
@@ -18,7 +18,16 @@ AppendStr(a,b) ::= "<a> + <b>"
 
 Concat(a,b) ::= "<a><b>"
 
-AssertIsList(v) ::= "TODO!!"
+AssertIsList(v) ::= <<
+// A noddy range over the list will not compile if it is not getting a slice
+// however, Go will not compile the generated code if the slice vs single value is wrong.
+// Makes the Java based tests suite work though.
+j1__ := make([]interface{}, len(<v>))
+j2__ := <v>
+for j3__ := range j2__ {
+   j1__[j3__] = j2__[j3__]
+}
+>>
 
 AssignLocal(s, v) ::= "<s> = <v>;"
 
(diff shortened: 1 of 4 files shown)


runtime-testsuite/resources/org/antlr/v4/test/runtime/templates/Go.test.stg
```diff
@@ -18,7 +18,16 @@ AppendStr(a,b) ::= "<a> + <b>"
 
 Concat(a,b) ::= "<a><b>"
 
-AssertIsList(v) ::= "TODO!!"
+AssertIsList(v) ::= <<
+// A noddy range over the list will not compile if it is not getting a slice
+// however, Go will not compile the generated code if the slice vs single value is wrong.
+// Makes the Java based tests suite work though.
```
KvanTTT, line 24: Not sure comments should be presented in generated code thus they should not be presented in template.

jimidle: Not sure why that matters? It is only generated for the test. I have found the generated test code wanting - a few comments would have helped understand what was going on much more quickly.

jimidle:
So what are we doing with this one guys? I can take out the generated comments if you wish, not a big deal for me. If we can merge and close this, then Go is is good to go.

runtime-testsuite/resources/org/antlr/v4/test/runtime/templates/Go.test.stg
jimidle: @parrt @KvanTTT You guys want me to take out the generated comments before you merge it? No worries either way for me.

parrt: We can just leave it since that's easier.

KvanTTT: Ok, nevermind.

Issue #3840, opened 2022-08-27
Head towards 4.11 release


parrt:
Heh, we have to get 4.11 out so that people can use the amazing Go target what day is done by @jimidle.

Stuff to do:

- [ ] Make sure to generate a Windows dotnet package from Windows to get the signing right
- [ ] Update antlr.org quick start guide to use https://github.com/antlr/antlr4-tools
- [x] @jimidle any more Go Target stuff that needs to go in?
- [x] Make sure 4.11.0 tag is also they're not just 4.11 (Go requirement)
- [x] C++ issue regarding EOF symbol: https://github.com/antlr/antlr4/issues/2914
- [x] @KvanTTT's https://github.com/antlr/antlr4/pull/3701
- [x] https://github.com/antlr/antlr4/issues/3733 ST uses 3.5.2 antlr not most recent.
- [x] Fix old error https://github.com/antlr/antlr4/issues/2016, merging https://github.com/antlr/antlr4/pull/2022
- [x] https://github.com/antlr/antlr4/pull/3841
- [x] https://github.com/antlr/antlr4/pull/3794

What else needs to go in 4.11?

@jimidle @ericvergnaud @mike-lischke @marcospassos @pboyer, @janyou, @ewanmellor, @lingyv-li have comments?

KvanTTT:
I wish the following issues were resolved:

* https://github.com/antlr/antlr4/pull/3676 It decreases the size of generated files and improves performance
* New action-related warnings: https://github.com/antlr/antlr4/pull/3626 (it's better to reimplement them on ATN instead of grammar)
* Fixes related to "mutually left recursive rules" errors: https://github.com/antlr/antlr4/pull/3779
* Last but not least refactoring: https://github.com/antlr/antlr4/pull/3780

parrt:
Hmm..not sure I have the time to go through those for now. It will require a massive effort to remember all of the ATN stuff before I can evaluate those.  The last one is something I probably won't accept because it's just refactoring and incurs a risk.

lingyv-li:
🙌  Nothing to add for Dart.

mike-lischke:
The EOF macro problem is tricky to solve. Either you annoy users who use C libs with ANTLR4 or you annoy existing project owners if you rename that symbol. Given that there are only very few complains I tend to keep the current solution. For the threads patch @jcking can probably help.

parrt:
[...]

PR #3852, opened 2022-08-31
Feature/gomod4110

jimidle:
feat: #3840  Implement go runtime v4 path correctly so that release tags work.

Though this PR will seem like a huge change, in fact most of the changes are just copying the existing runtime module to a new module path `github.com/antlr/antlr4/runtime/Go/antlr/v4` so that Go runtime tags will actually work in this release.

The actual changes are:

  - Updated documentation
  - Change Go codegen template to use the new `v4` path
  - Change go test suite to use new v4 path
  - Change v4/go.mod to use the new runtime path

*NB* Though the runtime path changes to `v4` the public runtime API has not changed and it is very little work to change
to use the new module path. Most go programmers will know how to do this, but it is also documented.


parrt:
Looks good to me but let's synchronize so I don't screw this up.  

Oh, sorry emerged something related to the test rig from @KvanTTT and so there's a conflict you might want to resolve quickly.

parrt:
Ok, looking good on Go target...merging!

parrt:
@jimidle shouldn't we delete the non-v4 version of the Go code?

jimidle:
Next release. I want to play it safe for existing projects. When I am sure
that v4 is good, we can then delete the code and clean up the repo.

PR #3857, opened 2022-09-01
Declare @param in generated visitor methods

gthb:
Helps out JavaScript projects which use JSDoc-based type checking using tsc.

Without this:

* the type checker does not know the type of `ctx` and editor integrations
  highlight that with the message "Parameter 'ctx' implicitly has an 'any'
  type, but a better type may be inferred from usage.ts(7044)".

* Invoking an editor/IDE "Find usages" on the `ParserRuleContext` identifier
  will not find these visitor methods.

Signed-off-by: Gunnlaugur Þór Briem <gunnlaugur@gmail.com>


gthb:
Closing this because

1. it's not done right, breaking tests
2. to be useful it should really declare the more specific context type for each visitor method
3. I don't really need it, turns out it's good enough to manually add `@param` declarations to my visitor subclass, so I'm not investing effort in fixing items 1 and 2 😊

Issue #3212, opened 2021-07-01
Provide .NET Framework target in the csharp nuget package


appel1:
Currently the Antlr4.Runtime.Standard NuGet package only contains a netstandard 2.0 build. If you consume such a package targeting .NET Framework prior to 4.7.2 it adds additional dependencies that causes us some issues. Instead of working around those it would be easier if the C# runtime package included a .NET Framework 4.6.2 target. 

`<TargetFrameworks>net462;netstandard2.0</TargetFrameworks>`

We cannot upgrade to 4.7.2 or newer right now due to outside constraints for some of our projects.

KvanTTT:
What issues? According to the [table](https://docs.microsoft.com/en-us/dotnet/standard/net-standard) netstandard2.0 is compatible with .NET Framework >= 4.6.1.

appel1:
It also says
> 2 The versions listed here represent the rules that NuGet uses to determine whether a given .NET Standard library is applicable. While NuGet considers .NET Framework 4.6.1 as supporting .NET Standard 1.5 through 2.0, there are several issues with consuming .NET Standard libraries that were built for those versions from .NET Framework 4.6.1 projects. For .NET Framework projects that need to use such libraries, we recommend that you upgrade the project to target .NET Framework 4.7.2 or higher.

What happens is that it pulls in additional dependencies, like System.Runtime.InteropServices.RuntimeInformation, that needs to be deployed and so on.

appel1:
I'm fine with either. It compiles for .net 4.5 so I guess it makes sense to go for that.

PR #3861, opened 2022-09-02
C#: remove redundant functionality and Windows net to 462

KvanTTT:
* Remove redundant functionality (use native .NET api instead of hand-written classes and methods)
* Upgrade net45 to net462 since net45 doesn't support `Array.Empty<T>()` and it's too old


KvanTTT:
@ericvergnaud I hope it's much easier for you to review this pull request (it's mostly manual refactoring).

runtime/CSharp/src/Atn/ConflictInfo.cs
```diff
@@ -1,80 +0,0 @@
-/* Copyright (c) 2012-2017 The ANTLR Project. All rights reserved.
```
parrt, line 1: Is it possible someone is using this from the outside world?

runtime/CSharp/src/Atn/ParserATNSimulator.cs
```diff
@@ -2246,18 +2244,21 @@ protected DFAState AddDFAState(DFA dfa, DFAState D)
...
 
-				D.stateNumber = dfa.states.Count;
+				D.stateNumber = dfaStates.Count;
 				if (!D.configSet.IsReadOnly)
 				{
 					D.configSet.OptimizeConfigs(this);
 					D.configSet.IsReadOnly = true;
 				}
-				dfa.states.Put(D, D);
+				dfaStates.Add(D, D);
```
parrt, line 2261: are put/add really the same thing in C#?

runtime/CSharp/src/Dfa/AcceptStateInfo.cs
```diff
@@ -1,71 +0,0 @@
-/* Copyright (c) 2012-2017 The ANTLR Project. All rights reserved.
```
parrt, line 1: Makes me nervous to delete files.

runtime/CSharp/src/Misc/Arrays.cs
```diff
@@ -2,46 +2,12 @@
  * Use of this file is governed by the BSD 3-clause license that
  * can be found in the LICENSE.txt file in the project root.
  */
-namespace Antlr4.Runtime.Sharpen
```
parrt, line 5: oh i see. You are moving files around. I guess I'm just worried somebody might be using this but it's pretty deep inside so unlikely they are using right?

runtime/CSharp/src/Atn/ConflictInfo.cs
KvanTTT: Good question, but I don't think so. There is no documentation about this class, and there is no mention in other runtimes about such entity. The purpose of `ConflictInfo` is unclear.

parrt:
Nice work. I don't see anything scary in there other than we're moving some files around. Most of the code does look like just clean up. Happy if @ericvergnaud could take a quick scan.

runtime/CSharp/src/Misc/Arrays.cs
KvanTTT: Everything form `Sharpen` directory is utility classes and methods. Moreover, native API should be used instead of utility methods if possible, that's why most part of them can be removed. They should not be used out of ANTLR runtime.

runtime/CSharp/src/Atn/ParserATNSimulator.cs
KvanTTT: C# doesn't have `Put` method, it's utility method that is going to be removed: https://github.com/antlr/antlr4/pull/3861/commits/4108810d4e94cc40e6fda8d9d139d18f4748750a#diff-a40512909ef89c597eebbddd134e6895d6ad7eae2b2c9e5748b80c85ae2e2d47L21

C# has `[]` operator that replaces the existing value and `Add` method that throws an exception if the value is already presented in the dictionary. In this code, `Add` method is more semantically correct since checking on the existing value is written above:

```cs
if (dfaStates.TryGetValue(D, out var existing))
{
    return existing;
}
```

runtime/CSharp/src/Dfa/AcceptStateInfo.cs
KvanTTT: Also, no mention in other runtimes and no doc.

runtime/CSharp/src/Atn/ATNConfigSet.cs
```diff
@@ -385,16 +384,11 @@ public ConfigHashSet()
 
 		public ATNConfig GetOrAdd(ATNConfig config)
 		{
-			ATNConfig existing;
-			if (this.TryGetValue(config, out existing))
+			if (TryGetValue(config, out var existing))
 				return existing;
-			else
```
ericvergnaud, line 391: Please keep else/if it's more readable

runtime/CSharp/src/Atn/LexerActionExecutor.cs
```diff
@@ -98,9 +99,11 @@ public static Antlr4.Runtime.Atn.LexerActionExecutor Append(Antlr4.Runtime.Atn.L
             {
                 return new Antlr4.Runtime.Atn.LexerActionExecutor(new ILexerAction[] { lexerAction });
             }
-            ILexerAction[] lexerActions = Arrays.CopyOf(lexerActionExecutor.lexerActions, lexerActionExecutor.lexerActions.Length + 1);
-            lexerActions[lexerActions.Length - 1] = lexerAction;
-            return new Antlr4.Runtime.Atn.LexerActionExecutor(lexerActions);
+            var lexerActions = lexerActionExecutor.lexerActions;
+            ILexerAction[] newLexerActions  = new ILexerAction[lexerActions.Length + 1];
```
ericvergnaud, line 103: Please use Array.Resize for consistency

runtime/CSharp/src/Atn/ParserATNSimulator.cs
```diff
@@ -1399,7 +1397,7 @@ protected int getAltThatFinishedDecisionEntryRule(ATNConfigSet configSet)
 		 *  Assumption: the input stream has been restored to the starting point
 		 *  prediction, which is where predicates need to evaluate.
 		 */
-		protected Pair<ATNConfigSet, ATNConfigSet> SplitAccordingToSemanticValidity(
+		protected KeyValuePair<ATNConfigSet, ATNConfigSet> SplitAccordingToSemanticValidity(
```
ericvergnaud, line 1400: Not sure I'm supportive. Although KVP is technically equivalent, it is semantically inappropriate. Pair is simple enough that we can keep it.

KvanTTT: Ok, [`Tuple`](https://docs.microsoft.com/en-us/dotnet/api/system.tuple-2?view=net-6.0) is equivalent in this case. But it's `class` instead of `struct` it may be less effective.

ericvergnaud:
> @ericvergnaud I hope it's much easier for you to review this pull request (it's mostly manual refactoring).

Well not sure why you insist on making big PRs. I'm reviewing, but this should have been about dropping Sharpen only.

runtime/CSharp/src/Dfa/AcceptStateInfo.cs
ericvergnaud: It's in the repo, but not even part of the project
@parrt Remember this was inherited from a diverging implementation

runtime/CSharp/src/Recognizer.cs
```diff
@@ -264,7 +264,7 @@ public virtual string GetTokenErrorDisplay(IToken t)
         /// </exception>
         public virtual void AddErrorListener(IAntlrErrorListener<Symbol> listener)
         {
-            Args.NotNull("listener", listener);
+            if (listener == null) throw new NullReferenceException(nameof(listener));
```
ericvergnaud, line 267: No! this changes the behavior (Args.NotNull throws a different exception and is shorter anyway)

runtime/CSharp/src/Atn/LexerActionExecutor.cs
KvanTTT: I'm not sure it's correct code. Should `Append` method change internal state of instance if consider it returns another instance? It looks like unwanted side effect.

Also, there is no state changing in Java runtime, only copying: https://github.com/antlr/antlr4/blob/b5648f405b746d87ec559132ad1939fb6d402053/runtime/Java/src/org/antlr/v4/runtime/atn/LexerActionExecutor.java#L72 The code is just incorrect.

runtime/CSharp/src/Tree/Pattern/ParseTreePatternMatcher.cs
```diff
@@ -648,32 +647,34 @@ internal virtual IList<Chunk> Split(string pattern)
             // collect into chunks now
             if (ntags == 0)
             {
-                string text = Sharpen.Runtime.Substring(pattern, 0, n);
-                chunks.Add(new TextChunk(text));
+                chunks.Add(new TextChunk(pattern));
```
ericvergnaud, line 650: Wouldn't `new TextChunk(pattern.Substring(0, n))` be faster ?

runtime/CSharp/src/Recognizer.cs
KvanTTT: Thanks, it should be `ArgumentNullException`. `Args.NotNull` is used only one time, there is no need to keep entire class with method only for one line. Also, it's commonly used pattern in .NET, see https://referencesource.microsoft.com/#mscorlib/system/string.cs,86 for example.

runtime/CSharp/src/Tree/Pattern/ParseTreePatternMatcher.cs
```diff
@@ -648,32 +647,34 @@ internal virtual IList<Chunk> Split(string pattern)
...
                 // copy text up to first tag into chunks
-                string text = Sharpen.Runtime.Substring(pattern, 0, starts[0]);
+                string text = pattern.Substring(0, starts[0]);
                 chunks.Add(new TextChunk(text));
             }
             for (int i_1 = 0; i_1 < ntags; i_1++)
             {
                 // copy inside of <tag>
-                string tag = Sharpen.Runtime.Substring(pattern, starts[i_1] + start.Length, stops[i_1]);
+                int offset = starts[i_1] + start.Length;
```
ericvergnaud, line 661: so now we have more code ? What's the benefit ?

runtime/CSharp/src/Tree/Pattern/ParseTreePatternMatcher.cs
```diff
@@ -648,32 +647,34 @@ internal virtual IList<Chunk> Split(string pattern)
...
+                string tag = pattern.Substring(offset, stops[i_1] - offset);
                 string ruleOrToken = tag;
                 string label = null;
                 int colon = tag.IndexOf(':');
                 if (colon >= 0)
                 {
-                    label = Sharpen.Runtime.Substring(tag, 0, colon);
-                    ruleOrToken = Sharpen.Runtime.Substring(tag, colon + 1, tag.Length);
+                    label = tag.Substring(0, colon);
+                    offset = colon + 1;
```
ericvergnaud, line 669: same here

runtime/CSharp/src/Tree/Pattern/ParseTreePatternMatcher.cs
```diff
@@ -648,32 +647,34 @@ internal virtual IList<Chunk> Split(string pattern)
...
+                    ruleOrToken = tag.Substring(offset, tag.Length - offset);
                 }
                 chunks.Add(new TagChunk(label, ruleOrToken));
                 if (i_1 + 1 < ntags)
                 {
                     // copy from end of <tag> to start of next
-                    string text = Sharpen.Runtime.Substring(pattern, stops[i_1] + stop.Length, starts[i_1 + 1]);
+                    offset = stops[i_1] + stop.Length;
+                    string text = pattern.Substring(offset, starts[i_1 + 1] - offset);
```
ericvergnaud, line 677: same here

runtime/CSharp/src/Tree/Pattern/ParseTreePatternMatcher.cs
KvanTTT: `pattern.Substring(0, n)` is equivalent to `pattern`. I don't see the reason why it should be faster.

runtime/CSharp/src/Tree/Trees.cs
```diff
@@ -36,9 +37,7 @@ public static string ToStringTree(ITree t)
         /// </remarks>
         public static string ToStringTree(ITree t, Parser recog)
         {
-            string[] ruleNames = recog != null ? recog.RuleNames : null;
-            IList<string> ruleNamesList = ruleNames != null ? Arrays.AsList(ruleNames) : null;
-            return ToStringTree(t, ruleNamesList);
+            return ToStringTree(t, recog?.RuleNames);
```
ericvergnaud, line 40: Do we have a test proving this still works ?

runtime/CSharp/src/Tree/Pattern/ParseTreePatternMatcher.cs
KvanTTT: Getting rid of confusing. When .NET developer reads code, he expects some familiar behavior of methods with similar signatures (Java `substring` and C# `Substring` have the same naming but different semantics). Also, `Sharpen.Runtime.Substring` is used only in several places.

runtime/CSharp/src/Atn/LexerActionExecutor.cs
ericvergnaud: ok

runtime/CSharp/src/Atn/ParserATNSimulator.cs
ericvergnaud: Tuple seems appropriate

runtime/CSharp/src/Tree/Trees.cs
KvanTTT: We don't have, but it's trivial rewriting. `AsList` method just does nothing: https://github.com/antlr/antlr4/pull/3861/files#diff-a5b1c1aea0d0e2b922a00605b9181ebdf38ea00cef8a58464bfa777f0c51d8beL22

runtime/CSharp/src/Recognizer.cs
ericvergnaud: There is also no need to change working code if it does not improve it... it's a recipe for this kind of pitfall. Our test coverage is pretty poor, hence why both @parrt and myself are reluctant to big changes

runtime/CSharp/src/Tree/Pattern/ParseTreePatternMatcher.cs
ericvergnaud: Yes, I was thinking of a different scenario

runtime/CSharp/src/Tree/Pattern/ParseTreePatternMatcher.cs
ericvergnaud: I want to see the same algorithms in Java and C#, and that was the original thinking behind Sharpen I believe. That is way more important. I don't disagree with your observation, but then we should rename Substring to SubstringUsingOffset.

runtime/CSharp/src/Tree/Trees.cs
KvanTTT: It looks like `AsList` was added for imitation of Java code. But C# is a different language, and the same code can be written in a more simple way.

ericvergnaud:
> * Remove redundant functionality (use native .NET api instead of hand-written classes and methods)
> * Upgrade net45 to net462 since net45 doesn't support `Array.Empty<T>()` and it's too old
we should move to net472. Net462 is no longer supported

runtime/CSharp/src/Tree/Pattern/ParseTreePatternMatcher.cs
KvanTTT: > I don't disagree with your observation, but then we should rename Substring to SubstringUsingOffset.

I don't clearly understand. You mean reverting my changes and using renamed `Sharpen.Runtime.Substring` -> `Sharpen.Runtime.SubstringUsingOffset`? System `Substring` can no be renamed.

runtime/CSharp/src/Tree/Trees.cs
ericvergnaud: Java does not have a ? operator, that's my concern

ericvergnaud: I don't trust reasoning, that's why I'm always requiring tests

KvanTTT:
> if we move, it should be to net472

* I don't understand completely if net462 is no longer supported. According to the table, net462 and net472 have the same status: https://en.wikipedia.org/wiki/.NET_Framework_version_history#Overview
* Some users still need net462, see https://github.com/antlr/antlr4/issues/3212 @appel1 is that true?

runtime/CSharp/src/Tree/Pattern/ParseTreePatternMatcher.cs
ericvergnaud: I would rename Sharpen.Runtime.Substring and convert it to an extension method

runtime/CSharp/src/Recognizer.cs
KvanTTT: Anyway, Go and C++ runtimes were massively rewritten in the latest year without tests. And I doubt they strictly imitate Java code. It even might be not efficient.

appel1:
> * Some users still need net462, see [Provide .NET Framework target in the csharp nuget package #3212](https://github.com/antlr/antlr4/issues/3212) @appel1 is that true?

@KvanTTT Yes. At least for now. Hopefully it'll change next year once the extended support for Windows 7 ends. 

As I understand it there's no explicit end date for 4.6.2 because it follows its parent OS. https://docs.microsoft.com/en-us/lifecycle/products/microsoft-net-framework

ericvergnaud:
> > if we move, it should be to net472
> 
> * I don't understand completely if net462 is no longer supported. According to the table, net462 and net472 have the same status: https://en.wikipedia.org/wiki/.NET_Framework_version_history#Overview
> * Some users still need net462, see [Provide .NET Framework target in the csharp nuget package #3212](https://github.com/antlr/antlr4/issues/3212) @appel1 is that true?

See https://dotnet.microsoft.com/en-us/download/dotnet-framework/net46.
C# users can stick to older versions since we're not fixing any bug.

runtime/CSharp/src/Recognizer.cs
ericvergnaud: Go and C++ runtimes were rewritten to fix genuine problems. And bad practice from others is certainly no excuse for bad practice from oneself. Improvements in code coverage would be much more useful than improvements in code style.

appel1:
> > > if we move, it should be to net472
> > 
> > 
> > 
> > * I don't understand completely if net462 is no longer supported. According to the table, net462 and net472 have the same status: https://en.wikipedia.org/wiki/.NET_Framework_version_history#Overview
> > * Some users still need net462, see [Provide .NET Framework target in the csharp nuget package #3212](https://github.com/antlr/antlr4/issues/3212) @appel1 is that true?
> 
> See https://dotnet.microsoft.com/en-us/download/dotnet-framework/net46. C# users can stick to older versions since we're not fixing any bug.

That's 4.6 not 4.6.2. 

4.6 is no longer supported. 4.6.2 very much is.

KvanTTT:
> C# users can stick to older versions since we're not fixing any bug.

If we follow this logic, we had to deprecate Python 2 a long time ago since it was discontinued in 2020, but we still support it: https://www.python.org/doc/sunset-python-2/

ericvergnaud:
> > > > if we move, it should be to net472
> > > 
> > > 
> > > 
> > > * I don't understand completely if net462 is no longer supported. According to the table, net462 and net472 have the same status: https://en.wikipedia.org/wiki/.NET_Framework_version_history#Overview
> > > * Some users still need net462, see [Provide .NET Framework target in the csharp nuget package #3212](https://github.com/antlr/antlr4/issues/3212) @appel1 is that true?
> > 
> > 
> > See https://dotnet.microsoft.com/en-us/download/dotnet-framework/net46. C# users can stick to older versions since we're not fixing any bug.
> 
> That's 4.6 not 4.6.2.
> 
> 4.6 is no longer supported. 4.6.2 very much is.

Fair enough

ericvergnaud:
> > C# users can stick to older versions since we're not fixing any bug.
> 
> If we follow this logic, we had to deprecate Python 2 a long time ago since it was discontinued in 2020, but we still support it: https://www.python.org/doc/sunset-python-2/
> 
> Also, we changed the ATN format in ANTLR 4.10, new generated lexers and parser are not back compatible.

That's not the best hasty decision we ever made, that breaking change did not create any practical value, and we had to provide specific support which would have been avoided had we done this change more calmly.

My principles ave very simple:
 - don't change code that works
 - support legacy versions unless something is in the way

runtime/CSharp/src/Atn/ParserATNSimulator.cs
parrt: true. KVP might be misleading.

parrt:
Sounds like this is too big of a change for no outward improvement.  I agree that refactoring PRs should be viewed skeptically.  Sometimes it's worth it. The atn int vs string has been a thorn in my side forever and was a big difference / confusion point between target code.

Ok, i'm holding off on merging this one. Maybe we split it up?

KvanTTT:
I can push `!fixup` commits to make further review easier (review only changes instead of entire PR). Before the final merging, I'll squash `!fixup` commit with ordinary ones.

parrt:
I think @ericvergnaud's comment about risk-reward equation is valid though. Refactoring has a high risk and little reward in this case. Yes, it's much better to have "clean" code but...

Corniel:
What is the status of this?

PR #3870, opened 2022-09-07
[WIP]Typescript target

### LINKED ISSUE #3868: 'native' import statements in @parser:header are dropped
Consider the following grammar:

grammar G;

@parser:header {
// import X from './X';
import X from './X';
class Y extends X {}
}

a: 'a' EOF;

The generated code will look as follows:

.../...
// import X from './X';
class Y extends X {}
class GParser extends Parser { .../... }

notice that the native statement "import X from './X';" has been dropped.
it doesn't happen with preceding comment or the succeeding declaration.
I suspect it is because the tool sees the 'import...' line as an antlr import statement...

ericvergnaud:
All ts tests pass locally.
Submitting a PR to ensure nothing else is broken, and try ts CI.
Also fixes #3868
tsc generates tons of warnings which I need to deal with


ericvergnaud:
mmm... not sure why my commits aren't signed

ericvergnaud:
@parrt hey, here's a beta of the typescript target
the CI works on Linux but needs to be fixed for Mac/Windows
Also fixes #3868

KvanTTT:
> mmm... not sure why my commits aren't signed

Every commit should have a sign, for example:

```
Signed-off-by: Ivan Kochurkin <kvanttt@gmail.com>
```

Also, is it possible to squash most part of commits? Commits like

```
renamed for clarity
fix issues
progressing
...
```

are not useful for reviewers and git history since they are not informative.

ericvergnaud:
> > mmm... not sure why my commits aren't signed
> 
> Every commit should have a sign, for example:
> 
> ```
> Signed-off-by: Ivan Kochurkin <kvanttt@gmail.com>
> ```
> 
> Also, is it possible to squash most part of commits? Commits like
> 
> ```
> renamed for clarity
> fix issues
> progressing
> ...
> ```
> 
> are not useful for reviewers and git history since they are not informative.

As written, I know what DCO expects, but I had automated this on my end, and it seems to no longer work..  
This is WIP, so no need to review/squash yet.

tool/src/org/antlr/v4/codegen/target/TypeScriptTarget.java
```diff
@@ -0,0 +1,204 @@
...
+	}
+
+	@Override
+	public boolean supportsOverloadedMethods() {
+		return true;
+	}
+
+	@Override
+	public boolean isATNSerializedAsInts() {
+		return true;
```
KvanTTT, line 201: No need to override the default value with the same value. It returns `false` only for Java.

tool/src/org/antlr/v4/codegen/target/TypeScriptTarget.java
```diff
@@ -0,0 +1,204 @@
...
+				}
+			}
+
+			i += toAdvance;
+		}
+
+		if ( addQuotes ) sb.append('"');
+
+		return sb.toString();
+	}
```
KvanTTT, line 177: This method has a lot of in common with [`getTargetStringLiteralFromANTLRStringLiteral`](https://github.com/antlr/antlr4/blob/7e8c4f4d9b5659673f00289c1ee60c7a74007042/tool/src/org/antlr/v4/codegen/Target.java#L237) from the base class. Please merge logic.

tool/src/org/antlr/v4/tool/Grammar.java
```diff
@@ -430,6 +430,10 @@ public void defineAction(GrammarAST atAST) {
         }
     }
 
+	public void removeParserHeader() {
+		namedActions.remove("header");
+	}
```
KvanTTT, line 435: I'm against adding such method to `Grammar`. Grammar should not have such responsibility and it looks like workaround.

tool/src/org/antlr/v4/codegen/CodeGenPipeline.java
```diff
@@ -50,6 +50,8 @@ public void process() {
 				writeRecognizer(parser, gen, false);
 			}
 
+			g.removeParserHeader();
```
KvanTTT, line 53: What's the meaning of such removing? It doesn't look clear.

KvanTTT (changes requested):

KvanTTT:
> I had automated this on my end, and it seems to no longer work..

Looks like so because not every commit has a sign.

tool/src/org/antlr/v4/codegen/CodeGenPipeline.java
ericvergnaud: See #3868

tool/src/org/antlr/v4/tool/Grammar.java
ericvergnaud: Grammar should definitely have responsibility for managing its members. That's basic OOP.

runtime-testsuite/test/org/antlr/v4/test/runtime/typescript/TscRunner.java
```diff
@@ -0,0 +1,114 @@
...
+	private final static String NORMALIZED_JAVASCRIPT_RUNTIME_PATH = getRuntimePath("JavaScript").replace('\\', '/');
+
+	@Override
+	public String getLanguage() {
+		return "TypeScript";
+	}
+
+	@Override
+	protected void initRuntime() throws Exception {
+		npmLinkRuntime();
```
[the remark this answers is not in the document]
ericvergnaud: I guess we have different views on code quality. It is much better to put the code in a function because it self-documents the code, and when one needs to add code in init they can add a call to another method. <init> is self-explanatory of _when_ it is called, but not on _what_ it does.

runtime-testsuite/test/org/antlr/v4/test/runtime/typescript/TscRunner.java
```diff
@@ -0,0 +1,114 @@
...
+					RuntimeTestUtils.getTextFromResource("org/antlr/v4/test/runtime/helpers/package_ts.json"));
+
+			writeFile(getTempDirPath(), "tsconfig.json",
+					RuntimeTestUtils.getTextFromResource("org/antlr/v4/test/runtime/helpers/tsconfig.json"));
+
+			npmInstall();
+
+			npmLinkAntlr4();
+
+			tscCompile();
```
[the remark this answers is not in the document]
ericvergnaud: Sorry, bad suggestion, see me comment re npmLinkRuntime

runtime-testsuite/test/org/antlr/v4/test/runtime/typescript/TscRunner.java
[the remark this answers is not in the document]
KvanTTT: From my point of view, it's obvious what the following code does, it's self-explanatory:

```java
Processor.run(new String[] {"npm", "--silent", "link"}, NORMALIZED_JAVASCRIPT_RUNTIME_PATH);
```

There is no new info from method name `npmLinkRuntime`.

> when one needs to add code in init they can add a call to another method.

At present time, there is no other code inside the method, in future it can be another situation, but it dosn't matter for now.

runtime-testsuite/test/org/antlr/v4/test/runtime/typescript/TscRunner.java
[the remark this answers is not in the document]
KvanTTT: At least `Processor.run` instead of `ProcessBuilder` should be used for brevity and unification with other runtimes (see other targets). The code with `ProcessBuilder` generates duplication.

tool/src/org/antlr/v4/tool/Grammar.java
KvanTTT: It should but related methods. It's unclear the need of method of removing `header` named action.

ericvergnaud:
@parrt hooray! all tests successful across all targets. Typescript CI is very slow, but we can look into it later (it doesn't affect the runtime). Otherwise, this is almost ready to go in beta, just missing code from #3878, and squashing the commits (some of them are unsigned due a config issue at my end).

parrt:
Hooray! Lemme look at #3878 tomorrow and then take a look at this.

KvanTTT:
> Typescript CI is very slow, but we can look into it later (it doesn't affect the runtime).

Now I see a lot of 

```
java.io.IOException: Failed to delete file: C:\Users\RUNNER~1\AppData\Local\Temp\TscRunner-ForkJoinPool-1-worker-3-1662740528124
```

on CI: https://github.com/antlr/antlr4/runs/8274285204?check_suite_focus=true#step:22:3510 It should be fixed before merging, otherwise typescript tests are useless.

ericvergnaud:
Same happens I suspect with JavaScript. Not sure why it fails but it does not affect the runtime nor the tests

Envoyé de mon iPhone

KvanTTT:
I don't see such errors in `dev` branch for JavaScript: https://github.com/antlr/antlr4/runs/8288355910?check_suite_focus=true

runtime-testsuite/test/org/antlr/v4/test/runtime/typescript/TscRunner.java
```diff
@@ -0,0 +1,146 @@
...
+
+import static org.antlr.v4.test.runtime.FileUtils.writeFile;
+import static org.antlr.v4.test.runtime.RuntimeTestUtils.isWindows;
+
+public class TscRunner extends RuntimeRunner {
+
+	/* TypeScript runtime is the same as JavaScript runtime */
+	private final static String NORMALIZED_JAVASCRIPT_RUNTIME_PATH = getRuntimePath("JavaScript").replace('\\', '/');
+	private final static String NPM_EXEC = "npm" + (isWindows() ? ".cmd" : "");
+	private final static String TSC_EXEC = "tsc" + (isWindows() ? ".cmd" : "");
```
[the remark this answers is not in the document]
ericvergnaud: No, that's only true for node

runtime/Swift/Package.swift
```diff
@@ -9,6 +10,10 @@ let package = Package(
             name: "Antlr4",
             type: .dynamic,
             targets: ["Antlr4"]),
+        .library(
+            name: "Antlr4",
+            type: .static,
+            targets: ["Antlr4"])
```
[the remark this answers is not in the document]
ericvergnaud: that's a merge side effect

ericvergnaud:
> I don't see such errors in `dev` branch for JavaScript: https://github.com/antlr/antlr4/runs/8288355910?check_suite_focus=true It affects CI servers since test data is not being removed and may take a lot of resource.

Only the top folder encounters this error, the contents are cleared

ericvergnaud:
@parrt thanks for merging #3878 (which was missing some tweaks, my bad), however I noticed you merged it into master rather dev ? So I had to merge back from master which brings in this PR a few unrelated files, I guess they'll fade away once you merge master onto dev. 
Runtime wise, as far as I'm concerned, this is ready to go. 
However not sure about the API (the d.ts files) since we don't know what people need. 
So my preferred path would be to publish a beta tool and js/ts runtime only, and call for testers. I expect these will come primarily from antlr4ts, so hopefully they will have mature requirements. I'll also be doing my own testing, but that may take a while since it requires migrating my code from js to ts. I expect the beta to last 2-3 months. Let me know what you think.

runtime/Swift/Package.swift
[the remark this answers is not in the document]
KvanTTT: Try using interactive rebase next time to get rid of issues like these.

runtime-testsuite/test/org/antlr/v4/test/runtime/typescript/TscRunner.java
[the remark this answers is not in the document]
KvanTTT: I don't think so. We don't have `.cmd` suffix for `node` in current `dev` branch: https://github.com/antlr/antlr4/blob/dev/runtime-testsuite/test/org/antlr/v4/test/runtime/javascript/NodeRunner.java#L35 It works fine on Winodws.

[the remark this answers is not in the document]
ericvergnaud: you are wrong, node and npm are different executables

runtime-testsuite/resources/org/antlr/v4/test/runtime/helpers/Test.ts.stg
```diff
@@ -0,0 +1,53 @@
+import { default as antlr4 } from 'antlr4';
+import <lexerName> from './<lexerName>.js';
+<if(parserName)>
+import <parserName> from './<parserName>.js';
+<if(useListener)>
+import <grammarName>Listener from './<grammarName>Listener.js';
+<endif>
+<if(useVisitor)>
+import <grammarName>Visitor from './<grammarName>Visitor.js';
+<endif>
```
KvanTTT, line 10: Why generated files are written in JavaScript instead of TypeScript? It looks weird.

runtime-testsuite/test/org/antlr/v4/test/runtime/typescript/TscRunner.java
```diff
@@ -0,0 +1,98 @@
...
+
+
+	@Override
+	protected void initRuntime() throws Exception {
+		installTsc();
+		npmLinkRuntime();
+	}
+
+	private void installTsc() throws Exception {
+		Processor.run(new String[] {NPM_EXEC, "--silent", "install", "-g", "typescript"}, null);
```
KvanTTT, line 38: Please remove `--silent` option here and everywhere. Because of it, runner swallows errors and other useful messages. There is no need to hide compiler messages.

tool/src/org/antlr/v4/codegen/target/TypeScriptTarget.java
KvanTTT: The issue actually is not resolved, I don't understand why it was marked as resolved.

runtime-testsuite/resources/org/antlr/v4/test/runtime/helpers/Test.ts.stg
ericvergnaud: Agreed it looks weird, but that's the way it works

runtime-testsuite/test/org/antlr/v4/test/runtime/typescript/TscRunner.java
ericvergnaud: There is because messages go to the expected output. Errors are not swallowed

KvanTTT: No. When I run `npm install --silent -g typescript` on my Ubuntu machine, it prints nothing but surprisingly do nothing. But when I run `npm install -g typescript` it prints the following:

```
npm ERR! code EACCES
npm ERR! syscall mkdir
npm ERR! path /usr/local/lib/node_modules/typescript
npm ERR! errno -13
npm ERR! Error: EACCES: permission denied, mkdir '/usr/local/lib/node_modules/typescript'
npm ERR!  [Error: EACCES: permission denied, mkdir '/usr/local/lib/node_modules/typescript'] {
npm ERR!   errno: -13,
npm ERR!   code: 'EACCES',
npm ERR!   syscall: 'mkdir',
npm ERR!   path: '/usr/local/lib/node_modules/typescript'
npm ERR! }
npm ERR! 
npm ERR! The operation was rejected by your operating system.
npm ERR! It is likely you do not have the permissions to access this file as the current user
npm ERR! 
npm ERR! If you believe this might be a permissions issue, please double-check the
npm ERR! permissions of the file and its containing directories, or try running
npm ERR! the command again as root/Administrator.

npm ERR! A complete log of this run can be found in:
npm ERR!     /home/ivankochurkin/.npm/_logs/2022-09-13T20_53_54_181Z-debug-0.log

```

tool/src/org/antlr/v4/codegen/target/TypeScriptTarget.java
ericvergnaud: Because it's not an issue. Java and Javascript support slightly different literals.
See https://docs.oracle.com/javase/specs/jls/se7/html/jls-3.html#jls-3.10.5 and https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String.
IMHO it's a bug to have just one translation method for all targets because inevitably users will complain, and we need to safely fix the translation on a per target basis.

runtime-testsuite/test/org/antlr/v4/test/runtime/typescript/TscRunner.java
ericvergnaud: yeah it requires sudo, which is implicit in the CI

ericvergnaud: but if you remove --silent and is successful it prints all sorts of things, which interfere with the testing

tool/src/org/antlr/v4/codegen/target/TypeScriptTarget.java
KvanTTT: > IMHO it's a bug to have just one translation method for all targets because inevitably users will complain, and we need to safely fix the translation on a per target basis.

Firstly, it's not bug, but maybe incorrect implementation. Secondly, IMHO it's not optimal to create duplicated functionality because of slightly different things. They can be encapsulated to separated methods according to OOP. The less code we have, the less potential errors we have. Moreover, this method not only escapes chars but also joins them. Joining logic should be the same. For that matter, `charCount`, `codePointAt` and other char methods are Java-specific, they should not be used for TypeScript or other target literals, but actually it doesn't matter.

ericvergnaud:
It’s an incorrect implementation therefore it’s a design bug. I’m not really interested in arguing about code style. Your suggestions are welcome but they should remain suggestions.

Envoyé de mon iPhone

runtime-testsuite/test/org/antlr/v4/test/runtime/typescript/TscRunner.java
KvanTTT: But it should run fine also locally, without CI.

parrt:
> beta

Well, sounds reasonable to me but I need to evaluate the merits of either approach discussed in https://github.com/antlr/antlr4/discussions/3867

What's involved in creating a suitable beta? A zip?  A branch?

tool/src/org/antlr/v4/codegen/target/TypeScriptTarget.java
parrt: Hi guys. @KvanTTT and I discussed having one method a while ago; is there a way to share common functionality but still have targets do special things when necessary?

ericvergnaud:
It involves publishing a beta in maven central and npmEnvoyé de mon iPhoneLe 16 sept. 2022 à 02:08, Terence Parr ***@***.***> a écrit :﻿

beta

Well, sounds reasonable to me but I need to evaluate the merits of either approach discussed in #3867
What's involved in creating a suitable beta? A zip?  A branch?

tool/src/org/antlr/v4/codegen/target/TypeScriptTarget.java
ericvergnaud: Wouldn't that create unnecessary complexity ?

ericvergnaud: Actually, since the default inherited method sets escapeSpecial to false, the 2 methods end up behaving the same (they shouldn't since the list of escapes should be different)

parrt: hmm...yeah, I guess I'm trying to avoid duplicating code, particularly since most targets will be similar. If they are sufficiently different then it warrants having separate sections of functionality. I think those sections used to be separate overridden functions but are now combined into a single function.  @KvanTTT can you remind me about the history of the combined function here? thanks!

KvanTTT: Sorry for the delay, I was figuring out of API of Target and how `getTargetStringLiteralFromString` and `getTargetStringLiteralFromANTLRStringLiteral` methods work. These methods are responsible for the conversion from grammar code (string literals or predicated) to target string literals. It looks like the single difference is the second method should conserve escaping characters since they are used in token names (that why `\n` from ANTLR string becomes `\\n` in target string).

I've clarified API of Target, fixed potential conversion bugs and removed duplication code in https://github.com/antlr/antlr4/pull/3891, please take a look.

jonaskello:
I was going to propose an issue to make the javascript target work better in typescript (exporting context classes so we can import them as typescript types). However having an official typescript target is of course a lot better :-). Just to know which way I should proceed I wanted to check if this PR will have a chance of getting merged in the foreseeable future or if it is more of an experiment?

PR #3871, opened 2022-09-07
fix: Restore missing changes to v4 of go runtime

jimidle:
I had not thought about this, but I guess some of the fix PRs were out of order and did not have one fix that was in the legacy version of the runtime, copied in to the v4 branch. This PR fixes that.

It is not worth doing yet another minor release as if anyone needs this exact fix, they can use the dev branch until the next release.

Signed-off-by: Jim.Idle <jimi@gatherstars.com>

### DIFF
--- a/runtime/Go/antlr/antlrdoc.go
+++ b/runtime/Go/antlr/antlrdoc.go
@@ -48,7 +48,7 @@ And the generate.sh file will look similar to this:
 	#!/bin/sh
 
 	alias antlr4='java -Xmx500M -cp "./antlr4-4.11.1-complete.jar:$CLASSPATH" org.antlr.v4.Tool'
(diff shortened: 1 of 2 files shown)


jimidle:
@parrt If you get a chance would please merge this, otherwise I will lose track of it ;)

parrt:
Will this break a bunch of Go-target parsers? In other word should I try to incorporate this into the galactic overlords codebase yet?

jimidle:
No mate. The fix was already in the 1.x version. If anyone upgraded without
moving to v4, then they already had the fix. But as the fix was for
something that didn’t even compile before this, we can be assured that
nobody was using it.

parrt:
Cool. thanks.

PR #3878, opened 2022-09-08
the scope (parser or lexer) in @parser::header was dropped, so keep t…

### LINKED ISSUE #3877: Parser header is incorrectly inserted in Listener and Visitor files
When adding a `@header` or `@parser::header` to a grammar, the enclosed text block is inserted not only in the generated parser code, but also in the generated listener and visitor code.
This is incorrect because listener and visitor do not contain any production code and as such would never have a need for additional code that is never invoked.

ericvergnaud:
the scope (parser or lexer) in @parser::header was dropped, so keep track of it and only include @header in Listener and Visitor code

Fixes #3877

### DIFF
--- a/tool/src/org/antlr/v4/codegen/model/OutputFile.java
+++ b/tool/src/org/antlr/v4/codegen/model/OutputFile.java
@@ -13,6 +13,7 @@ import org.antlr.v4.tool.ast.ActionAST;
 
 import java.util.HashMap;
 import java.util.Map;
+import java.util.function.Predicate;
 
 public abstract class OutputFile extends OutputModelObject {
 	public final String fileName;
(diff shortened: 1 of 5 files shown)


parrt:
Based upon your analysis in the bug report, I wonder if we should ever generate header stuff into the listener and visitor files. They can't have any user defined types in there. I'm trying to think of a need for ever injecting header stuff into listener and visitor generated files.

tool/src/org/antlr/v4/codegen/model/OutputFile.java
```diff
@@ -32,10 +33,15 @@ public OutputFile(OutputModelFactory factory, String fileName) {
...
 	public Map<String, Action> buildNamedActions(Grammar g) {
+		return buildNamedActions(g, null);
+	}
+
+	public Map<String, Action> buildNamedActions(Grammar g, Predicate<ActionAST> filter) {
 		Map<String, Action> namedActions = new HashMap<String, Action>();
 		for (String name : g.namedActions.keySet()) {
 			ActionAST ast = g.namedActions.get(name);
-			namedActions.put(name, new Action(factory, ast));
+			if(filter==null || filter.test(ast))
```
parrt, line 43: I think the project style uses curly braces even if there's only one statement. It reduces one possibility of coding bugs.

tool/src/org/antlr/v4/codegen/model/VisitorFile.java
```diff
@@ -58,7 +58,8 @@ public VisitorFile(OutputModelFactory factory, String fileName) {
 			}
 		}
 		ActionAST ast = g.namedActions.get("header");
-		if ( ast!=null ) header = new Action(factory, ast);
```
parrt, line 61: haha looks I or someone else already violated the rules about if statement... I guess it's OK if it's on one line like that.

PR #3879, opened 2022-09-09
Golang: Fixed Index Out of Range Bug With GetChild

adambcomer:
The ANTLR Golang runtime has a index out of range bug with GetChild. To ensure this is a bug, I checked the [python3](https://github.com/antlr/antlr4/blob/master/runtime/Python3/src/antlr4/ParserRuleContext.py#L116) and [javascript](https://github.com/antlr/antlr4/blob/master/runtime/JavaScript/src/antlr4/context/ParserRuleContext.js#L122) runtimes. The logic for the Golang runtime differs here causing a index out of range error when a user gets the child at the length of the array.


jimidle:
@adambcomer This PR is in the wrong code. It should be in the /v4 version of the go runtime now. 

However, the range check is incorrect, so  I will fix it as per the Java code, which also checks for >=0.

If you or @parrt will please close this PR, I will include it in my PR for another fix in a related area

Issue #3875, opened 2022-09-08
Using a "skip" token in parser rule causes major performance problems in the Go target parser


kaby76:
I am updating grammars-v4 for 4.11.1. Unfortunately, the [asm8080](https://github.com/antlr/grammars-v4/tree/master/asm/asm8080) and [asmZ80](https://github.com/antlr/grammars-v4/tree/master/asm/asmZ80) grammars now do not terminate (actually killed after 5 minutes by the watchdog program trwdog). For 4.10.1, they terminated. Neither of these grammars contain target-specific code, and are really really simple.

kaby76:
Just in case I'm doing something wrong, I tested 4.11.1 on the [SimpleBoolean](https://github.com/manango/perf-repro/blob/master/evaluate/simpleboolean/SimpleBoolean.g4) [issue that was reported](https://github.com/antlr/antlr4/issues/3718). Indeed, 4.11.1 does fix that problem (fast!), but asm8080 now takes 376s for [cpm22.asm](https://github.com/antlr/grammars-v4/blob/master/asm/asm8080/examples/CPM22.ASM).

kaby76:
I did a side-by-side comparison of the debug output from ParserATNSimulator between CSharp and Go targets for asm8080/cpm22.asm. The outputs are essentially identical which means they are computing the same results. But, clearly, the Go implementation is much much slower.

jimidle:
I will take a look. There is still an issue with way too many allocations in parse_context. I know what it is (failure to 'emulate' Java objects), but I felt it was too much to try and cram in as a fix at the last minute- programmer checks in code and then gets on a plane type situation ;)

There was one fix that made it to what is the non-v4 version that did not get in to the v4 version (thought they were supposed to be identical). I have submitted a PR for that and is awaiting Ter's approval. But I doubt it is that because that fixes something that just did not work before. 

Tomorrow (public holiday here), I will take a look at those grammars and work out what is going on - I suspect it is the allocation problem, but could be a hidden bug that I did not find yet. 

I am surprised that they do not terminate (or did they just need more time to run?).

Thanks for the heads up. This code is taking some time to wrangle in to shape.

jimidle:
BTW - I get this when generating, which is a correct warning:

```
warning(154): asm8080.g4:38:0: rule prog contains an optional block with at least one alternative that can match an empty string
```

[...]

PR #3880, opened 2022-09-10
fix: #3758 Extrememly poor performance on poorly written grammars

jimidle:
fix: #3875 Extremely poor performance on poorly written grammars

The problem was once again a poorly written hash function, which caused ATN config sets to
hash poorly. The poor grammar caused many thousands of DFA states to hash to the same 
bucket. In your average grammar, this situation would not arise.

For now, I have changed that hash function to use the memory address of the slice that
contains the configs, though i think that this is sub-optimal.  I will revisit this as part of a
general attack at further performance gains.

### DIFF
--- a/runtime/Go/antlr/v4/atn_config_set.go
+++ b/runtime/Go/antlr/v4/atn_config_set.go
@@ -4,7 +4,10 @@
 
 package antlr
 
-import "fmt"
+import (
+	"fmt"
+	"strconv"
+)
 
 type ATNConfigSet interface {
 	Hash() int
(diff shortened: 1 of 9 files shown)


jimidle:
@parrt Ready to merge when you get a chance

parrt:
> hash function to use the memory address of the slice

But won't that cause no sharing of any kind?  I believe two DFA states with the exact same content should get the same hash otherwise it will look like no DFA state has ever been seen before and will prevent DFA usage I think.

jimidle:
Yes, but what had been done was that they all got exactly the same hash - I have to come back to this one  I think. Java uses the hash of the enclosing set object. Go does not have that and it was using the hash of each config in the set. I don't think that using the address is quite right, but I have not thought of a better way to reflect Java hash just yet. I do not like using the address because even though Go does not relocate when garbage collecting, it might not always be the case.

We can leave this unmerged for now until I get a little time to re-look. It is the only place where it seems a bit tricky to get hashing working like in Java.

jimidle:
@parrt This last change fixes Hash() and optimizes allocations of comparators. DFA reuse works as it should.

parrt:
Ah. missed the update. I'll try it out.

parrt:
Works! Great. thanks.

PR #3889, opened 2022-09-15
Fixing reserve word NULL for cpp targets

1sand0s:
PR Summary

1. The [ANTLR grammar for Cypher](https://opencypher.org/resources/) contains a Token named `NULL` which is a reserve word in `Cpp`. This PR adds `NULL` to the list of reserve words in `CppTarget.java`.  
2. I passed the `String` arguments of `getTargetStringLiteralFromString` and `getTargetStringLiteralFromANTLRStringLiteral` in `Target.java`,  `s` and `literal` respectively, to `Target's` `escapeIfNeeded` function since potential reserve words must be handled before being accumulated into the StringBuilders
3. I passed the  Key<String> of `g.tokenNameToTypeMap` in `Recognizer` to `Target's` `escapeIfNeeded` function  before being inserted into `tokens` in `Recognizer`

Best

### DIFF
--- a/runtime/Cpp/README.md
+++ b/runtime/Cpp/README.md
@@ -35,12 +35,12 @@ The C++ target has been the work of the following people:
 
 ### Build + Usage Notes
 
-The minimum C++ version to compile the ANTLR C++ runtime with is C++11. The supplied projects can built the runtime either as static or dynamic library, as both 32bit and 64bit arch. The macOS project contains a target for iOS and can also be built using cmake (instead of XCode).
+The minimum C++ version to compile the ANTLR C++ runtime with is C++17. The supplied projects can built the runtime either as static or dynamic library, as both 32bit and 64bit arch. The macOS project contains a target for iOS and can also be built using cmake (instead of XCode).
 
 Include the antlr4-runtime.h umbrella header in your target application to get everything needed to use the library.
 
 If you are compiling with cmake, the minimum version required is cmake 2.8.
-By default, the libraries produced by the CMake build target C++11. If you want to target a different C++ standard, you can explicitly pass the standard - e.g. `-DCMAKE_CXX_STANDARD=17`.
+By default, the libraries produced by the CMake build target C++17. If you want to target a different C++ standard, you can explicitly pass the standard - e.g. `-DCMAKE_CXX_STANDARD=17`.
 
 #### Compiling on Windows with Visual Studio using he Visual Studio projects
 Simply open the VS project from the runtime folder (VS 2019+) and build it.
--- a/tool/src/org/antlr/v4/codegen/Target.java
+++ b/tool/src/org/antlr/v4/codegen/Target.java
@@ -243,6 +243,7 @@ public abstract class Target {
 	{
 		StringBuilder sb = new StringBuilder();
 
+		literal = escapeIfNeeded(literal);
 		if ( addQuotes ) sb.append('"');
 
 		for (int i = 1; i < literal.length() -1; ) {
--- a/tool/src/org/antlr/v4/codegen/Target.java
+++ b/tool/src/org/antlr/v4/codegen/Target.java
@@ -243,7 +243,6 @@ public abstract class Target {
 	{
 		StringBuilder sb = new StringBuilder();
 
-		literal = escapeIfNeeded(literal);
 		if ( addQuotes ) sb.append('"');
 
 		for (int i = 1; i < literal.length() -1; ) {
--- a/tool/src/org/antlr/v4/codegen/Target.java
+++ b/tool/src/org/antlr/v4/codegen/Target.java
@@ -126,7 +126,7 @@ public abstract class Target {
 	 *  to a token type in the generated code.
 	 */
 	public String getTokenTypeAsTargetLabel(Grammar g, int ttype) {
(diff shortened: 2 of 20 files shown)


tool/src/org/antlr/v4/codegen/Target.java
```diff
@@ -242,6 +243,7 @@ public String getTargetStringLiteralFromANTLRStringLiteral(
 	{
 		StringBuilder sb = new StringBuilder();
 
+		literal = escapeIfNeeded(literal);
```
KvanTTT, line 246: Could you explain why do you need escaping here? It doesn't look correct since the method always returns target string but not identifier.

KvanTTT (changes requested):
Please add or extend the test [ReservedWordsEscaping.txt](https://github.com/antlr/antlr4/blob/master/runtime-testsuite/resources/org/antlr/v4/test/runtime/descriptors/LexerExec/ReservedWordsEscaping.txt) that covers suggested changes.

tool/src/org/antlr/v4/codegen/Target.java
1sand0s: @KvanTTT You are right, there is no escaping needed here. I have removed it and added a test case for checking the escaping of reserve word `NULL` for cpp targets.

tool/src/org/antlr/v4/codegen/Target.java
```diff
@@ -167,6 +167,7 @@ public String getTargetStringLiteralFromString(String s, boolean quoted) {
 			return null;
 		}
 
+		s = escapeIfNeeded(s);
```
KvanTTT, line 170: Escaping is also not needed here since output is string literal.

KvanTTT:
Could you please get rid of unrelated commits?

1sand0s:
> Could you please get rid of unrelated commits (using interactive rebase)? Some changes are unrelated to the topic.

@KvanTTT Done !

1sand0s:
> Could you please get rid of unrelated commits (using interactive rebase)? Some changes are unrelated to the topic.

@KvanTTT Just a reminder that I have removed unrelated commits.

Thanks !

KvanTTT:
I see unrelated commits anyway. @parrt the changes look good to me, but I suggest squashing all commits since most of them are useless.

runtime/Cpp/README.md
```diff
@@ -35,12 +35,12 @@ The C++ target has been the work of the following people:
 
 ### Build + Usage Notes
 
-The minimum C++ version to compile the ANTLR C++ runtime with is C++11. The supplied projects can built the runtime either as static or dynamic library, as both 32bit and 64bit arch. The macOS project contains a target for iOS and can also be built using cmake (instead of XCode).
+The minimum C++ version to compile the ANTLR C++ runtime with is C++17. The supplied projects can built the runtime either as static or dynamic library, as both 32bit and 64bit arch. The macOS project contains a target for iOS and can also be built using cmake (instead of XCode).
```
parrt, line 38: Is this minimum truly 17 now? Did we make a choice to do this or is this a new requirement?

tool/src/org/antlr/v4/codegen/Target.java
```diff
@@ -126,7 +126,7 @@ protected void genFile(Grammar g, ST outputFileST, String fileName)
 	 *  to a token type in the generated code.
 	 */
 	public String getTokenTypeAsTargetLabel(Grammar g, int ttype) {
-		String name = g.getTokenName(ttype);
+	    String name = this.escapeIfNeeded(g.getTokenName(ttype));
```
parrt, line 129: I'm surprised this used to work at all. Did we not need the escape previously?

runtime/Cpp/README.md
1sand0s: @parrt The constructor of ANTLRInputStream declared in ANTLRInputStream.h, uses std::string_view which is available only post [C++17](https://en.cppreference.com/w/cpp/string/basic_string_view) so the minimum required CMAKE_CXX_STANDARD to build the runtime library for C++ targets would be 17.

tool/src/org/antlr/v4/codegen/Target.java
parrt: Was this not being escaped before? Hmm... Seems like without the escape none of the reserved words would have been rewritten.

1sand0s: @parrt  Because all the reserve words so far for CppTargets are lower case and thus are never expected to appear as token definition as per antlr specification (rulenames in lower case and tokens in uppercase).

parrt: oh ok, got it!

parrt:
Looks like we have a merge conflict. Can you take a look / rebase?

1sand0s:
> Looks like we have a merge conflict. Can you take a look / rebase?

I compared the `antlr4/dev` branch with `1sand0s/antlr4/Fixing_Reserve_Word_NULL_for_Cpp_Targets` branch and did not find merge conflicts on my end.

parrt:
AH. it was set at rebase not merge. A merge will work.  thanks!

PR #3891, opened 2022-09-18
Optimize conversion from ANTLR literals to target string literals

KvanTTT:
* Merge `getTargetStringLiteralFromString` and `getTargetStringLiteralFromANTLRStringLiteral` into `getTargetStringLiteralFromAntlrGrammar` since they do almost the same. `getTargetStringLiteralFromAntlrGrammar` has `keepEscaping` paramter. If it's `true`, then escaped chars are being escaped in target code (it's used for token names).
* Clarify API of Target, remove duplicated code, extend `EscapedCharactersInTokenLiterals` test


parrt:
can you back out whitespace onlychanges so it's smaller to review?

KvanTTT:
Done.

tool/src/org/antlr/v4/codegen/target/GoTarget.java
```diff
@@ -59,42 +56,6 @@ protected Set<String> getReservedWords() {
...
 
-	@Override
-	protected void genFile(Grammar g, ST outputFileST, String fileName) {
-		super.genFile(g, outputFileST, fileName);
-		if (DO_GOFMT && !fileName.startsWith(".") /* criterion taken from gofmt */ && fileName.endsWith(".go")) {
-			gofmt(new File(getCodeGenerator().tool.getOutputDirectory(g.fileName), fileName));
-		}
-	}
-
-	private void gofmt(File fileName) {
```
parrt, line 70: @jimidle do we want to run go fmt or should this be removed?

tool/src/org/antlr/v4/codegen/UnicodeEscapes.java
```diff
@@ -11,12 +11,6 @@
  * languages' syntax.
  */
 public class UnicodeEscapes {
-	public static String escapeCodePoint(int codePoint, String language) {
```
parrt, line 14: so it's not used anywhere but TestUnicodeEscapes?

tool/src/org/antlr/v4/codegen/target/GoTarget.java
KvanTTT: It looks strange to me that ANTLR core depends on `gofmt` tool and I don't clearly understand why this formatting is required at all.

tool/src/org/antlr/v4/codegen/UnicodeEscapes.java
KvanTTT: Yes.

parrt:
So this is to support the goal of less code duplication across Target objects to convert antlr literals to target literals?  On it's own, this looks ok (although big) but not worth the risk of refactor unless we have a goal in mind.  @ericvergnaud this seems related to the other discussion about target literals for ts.

PR #3892, opened 2022-09-20
feat: Add contributed tree.go walker to the v4 code base

jimidle:
Adds user contributed basic tree walker to the v4 code base


parrt:
For cleanliness of the license. Maybe we need the original contributor to sign off? Can you remind me where we got it?

PR #3906, opened 2022-09-24
Package.swift: provide an option to link against ANTLR 4 statically

edigaryev:
Problem: currently there's no way to link against ANTLR 4 statically from Swift, which is a deal-breaker if you want to ship a single binary. See https://github.com/cirruslabs/tart/pull/230 for example.

Solution: normally, [the `type: .dynamic` should be omitted](https://developer.apple.com/documentation/packagedescription/product/library(name:type:targets:)) from library product specification, but changing this now might break things for Swift users. Solve this by introducing a separate Swift PM product `Antlr4Static` that explicitly specifies `type: .static`.


edigaryev:
@parrt mind taking a look at this?

We need this to be able to use ANTLR 4 in [Tart](https://github.com/cirruslabs/tart), see https://github.com/cirruslabs/tart/pull/230.

Thanks!

parrt:
ok, though we normally do PRs on dev. can you rebase on that?

KvanTTT:
> ok, though we normally do PRs on dev. can you rebase on that?

I'd say to create a new PR on `dev` and close this one.

edigaryev:
> ok, though we normally do PRs on dev. can you rebase on that?

Done 👌

edigaryev:
@parrt sorry to nudge you, but mind taking a look at this once more?

parrt:
Thanks for the nudge. I get way too much email from way too many open source packages to maintain ha ha

PR #3913, opened 2022-09-26
JavaScript: Export named exports in addition to default.

### LINKED ISSUE #3912: JavaScript: Named exports
Currently the JavaScript runtime for Antlr4 only provides a default export:
```js
const antlr4 = {
    atn, dfa, tree, error, Token, CommonToken, CharStreams, InputStream, FileStream, CommonTokenStream, Lexer, Parser,
    PredictionContextCache, ParserRuleContext, Interval, IntervalSet, LL1Analyzer, Utils
};

export default antlr4;
```

This can be problematic as a consumer as outlined nicely by the esbuild documentation: 
https://esbuild.github.io/content-types/#default-interop
```
If you are a library author: When writing new code, you should strongly consider avoiding the default export entirely. It has unfortunately been tainted with compatibility problems and using it will likely cause problems for your users at some point.
```

This is indeed the case for us currently with trying to test against our usage of antlr and how babel is handling the default export versus the imports etc.

The simplest solution would be to add named exports (without removing the default, so a non-breaking change for consumers):
```js
const antlr4 = {
    atn, dfa, tree, error, Token, CommonToken, CharStreams, InputStream, FileStream, CommonTokenStream, Lexer, Parser,
    PredictionContextCache, ParserRuleContext, Interval, IntervalSet, LL1Analyzer, Utils
};

export default antlr4;

export {
    atn, dfa, tree, error, Token, CommonToken, CharStreams, InputStream, FileStream, CommonTokenStream, Lexer, Parser,
    PredictionContextCache, ParserRuleContext, Interval, IntervalSet, LL1Analyzer, Utils
};
```

This would:
- Allow consumers successfully using the default import to continue with no change.
- Allow named imports to be handled more easily.
- Should also help with dead code elimination / tree-shaking

Codex-:
This change adds named exports to the JavaScript runtime, the default export is left to maintain compatibility.

### Motivation
There is a better outline of the issues encountered when using solely default exports as described by the esbuild team: 

https://esbuild.github.io/content-types/#default-interop
```
If you are a library author: When writing new code, you should strongly consider avoiding the default export entirely. It has unfortunately been tainted with compatibility problems and using it will likely cause problems for your users at some point.
```

See #3912 for more info

Resolves #3912

### DIFF
--- a/runtime/JavaScript/src/antlr4/index.js
+++ b/runtime/JavaScript/src/antlr4/index.js
@@ -30,3 +30,8 @@ const antlr4 = {
 };
 
 export default antlr4;
+
+export {
+    atn, dfa, tree, error, Token, CommonToken, CharStreams, InputStream, FileStream, CommonTokenStream, Lexer, Parser,
+    PredictionContextCache, ParserRuleContext, Interval, IntervalSet, LL1Analyzer, Utils
+}


parrt:
I will have to rely on @ericvergnaud to take a peek at this.

parrt:
@ericvergnaud ?

ericvergnaud:
> @ericvergnaud ?

Already superseded. You can close.

PR #3915, opened 2022-10-02
[Cpp]Fix crashes in TokenStreamRewriter

tcy-bitstothings:
Signed-off-by: Tomasz Cybulski <98831780+tcy-bitstothings@users.noreply.github.com>

A simple PR fixing crashes in cpp TokenStreamRewriter.
Rewrite operation objects were accessed after deletion.


parrt:
Hmm... why is `iop` not available after `delete`?  My C++ is rusty.

Issue #3874, opened 2022-09-08
Support in-memory transpilation


KOLANICH:
It'd be nice to have a capability to transpile grammars into source code in-memory, without touching storage, via API.

I have even created a PR about that (#2774), but it has been rejected because of "no big changes" (I had to do a bit of refactoring in order to support both into-file and in-memory transpilation cleanly enough).

So currently I have to maintain a separate branch (https://github.com/UniGrammar/antlr4/tree/tool_refactoring) and I'm not happy about that, because sometimes merge conflicts happen and pinpointing why and where exactly they happen and how to fix that and reapplying the changes manually without any good semantic AST-based version control system is hard and time-consuming. Also I'm not happy with shipping an own version of ANTLR that has no chance to land into distros.

Could someone in the project examine my changes and integrate them into the code base (or give me an advice on what I should do) to get them merged?

jimidle:
Perhaps there are other ways to achieve your end goals - what are your reasons for needing this?

KOLANICH:
I use ANTLR4 as a library from my own tools. In multiple ways (the other way I use ANTLR is using its internal representation of grammars, but it has nothing to do with this issue), one of which is to compile a grammar in-memory (the grammar for ANTLR is itself transpiled from my own DSL for grammars, allowing them to be portable across different parser generators with own advantages and drawbacks, for example ANTLR has a benefit that it is convenient to debug grammars in it). Then multiple things can happen.

1. the generated java sources are passed into in-memory Java compiler and compiled into Java bytecode and the resulting classes are fed into a test rig (basically I type `UniGrammar viz antlr4 grammar.yug string-to-parse` and get the test rig GUI with AST)
2. the generated python sources are evaled and run through my wrapper
3. the generated java sources are compiled and run through my wrapper (I use the awesome JPype lib (through my abstraction layer JAbs) to interface java code from python)
4. the generated files are saved into the places my tool expects them to find (yet another abstraction I call a `ParserBundle`)

I don't want to touch SSD/HDD when it is completely unnecessary. Of course I can allow ANTLR to save files to some dir, and then read them from it, and you can say that it is a way to achieve the end goals of compiling a file, but if we add "without doung unnecessary I/O when it is unneeded", then it turns out that goal is unachievable without much very dirty workarounds like using `ramfs`.

IMHO this issue should better be fixed on ANTLR side by providing the needed API.

ericvergnaud:
Hi,

[...]

PR #3924, opened 2022-10-11
Extracted getting the interp file name into a separate method `Tool::getInterpFileName`.

KOLANICH:


parrt:
As with my other comment... We really need to have a good reason to change something for fear we might break somebody's code that relies on the tool.

KvanTTT:
What's the motivation of extracting the method with only single using?

KOLANICH:
>What's the motivation of extracting the method with only single using?

To make maintaining my fork with the commits implementing in-memory transpilation #3874 easier. Thanks for mentioning this! I think I need to add a comment saying that this function shouldn't be inlined.

>We really need to have a good reason to change something for fear we might break somebody's code that relies on the tool.

I don't know if it is considered a good reason, but I just try to merge some of the changes (usually equivalent ones) needed for #3874 in a careful step by step way.

PR #3925, opened 2022-10-11
Refactor CodeGenerator a bit: move target initialization code into a separate method.

KOLANICH:


parrt:
Hi. Thanks for the effort, but any change has to have a very good reason at this point in its life cycle. Is there a problem this is solving?

KOLANICH:
I don't know if it is considered a good reason, but I just try to merge some of the changes (usually equivalent ones) needed for #3874 in a careful step by step way.

Issue #3926, opened 2022-10-11
[Go target] Generated parser contains interfaces without requisite methods for tree navigation


kaby76:
This is a placeholder for a problem someone noted in StackOverflow ([so74022012](https://stackoverflow.com/q/74022012/4779853). It might be somewhere in the antlr/antlr4 issues, but I cannot find it.

The Go target doesn't generate code with accessor functions using the RHS symbol name as is done with other targets. Consequently, code that makes navigations using the accessor functions have to be type casted in order to go from one node to another in the parse tree. 

#### Grammar
[abb](https://github.com/antlr/grammars-v4/blob/master/abb/abbParser.g4)

```
module  : moduleData EOF   ;
moduleData : MODULE moduleName NEWLINE  dataList  NEWLINE* ENDMODULE ;
```

#### Generated Java
```
public static class ModuleContext extends ParserRuleContext {
		public ModuleDataContext moduleData() { ... }
}
public static class ModuleDataContext extends ParserRuleContext {
		public TerminalNode MODULE() { return getToken(abbParser.MODULE, 0); }
		public ModuleNameContext moduleName() { ... }
}
```
##### Navigation in Java
```
module.moduleData().moduleName() ...
```

#### Generated Go
```
func (s *ModuleContext) ModuleData() IModuleDataContext { ... }
type IModuleDataContext interface {
	antlr.ParserRuleContext

	// GetParser returns the parser.
	GetParser() antlr.Parser

	// IsModuleDataContext differentiates from other interfaces.
	IsModuleDataContext()
}
type ModuleDataContext struct {
	*antlr.BaseParserRuleContext
	parser antlr.Parser
}
```
##### Navigation in Go
```
module.(*parser.ModuleContext).moduleData().(*parser.ModuleDataContext).moduleName()
```
In the generated Go, the accessors return interfaces (unlike that in Java or C#), and the interfaces do not have methods declared for the accessors contained in that interface.

j3r3miah:
Hey @kaby76, thanks for the quick response to my StackOverflow post! What are your thoughts on this attempt to fix this issue? https://github.com/antlr/antlr4/pull/3927

kaby76:
@j3r3miah Yes, that is exactly what needs to be generated for the interface. @jimidle can help out if needed.

jimidle:
Let me take a look. Generally I have been using the Listener, which works well. I have not really explored the tree walker. It seems like maybe it was left behind a little.

[...]

PR #3927, opened 2022-10-11
[Go target] Fix for #3926: Add accessors for tree navigation to interfaces in generated parser

j3r3miah:
# What does this PR fix?

* #3926 

# Discussion

This is an attempt to remove the need to always cast these interfaces to concrete types. I considered two options:

1. Update accessors to return concrete types instead of interfaces
2. Add the missing method signatures to the existing interfaces

This PR implements option (2), as (1) would be a breaking change for existing applications. I believe (2) is backwards compatible.

The strategy here is to add a flag, `signature`, to each of the `ContextGetterDecl` subclasses so that each of their corresponding template definitions can conditionally generate a method signature (for the interface) versus the complete method declaration.

The `getSignatureDecl()` functions on each of the `ContextGetterDecl` subclasses are pretty tedious... open to suggestions for a more elegant option...

### Grammar
[abb](https://github.com/antlr/grammars-v4/blob/master/abb/abbParser.g4)

```
module  : moduleData EOF   ;
moduleData : MODULE moduleName NEWLINE  dataList  NEWLINE* ENDMODULE ;
```

### Generated Go
```
func (p *abbParserParser) Module() (localctx IModuleContext) { ... }
func (s *ModuleContext) ModuleData() IModuleDataContext { ... }

type IModuleDataContext interface {
	antlr.ParserRuleContext

	// GetParser returns the parser.
	GetParser() antlr.Parser

	// IsModuleDataContext differentiates from other interfaces.
	IsModuleDataContext()
}

type ModuleDataContext struct {
	*antlr.BaseParserRuleContext
	parser antlr.Parser
}
```

### Navigation in Go
```
module.(*parser.ModuleContext).ModuleData().(*parser.ModuleDataContext).ModuleName()
```

### Generated Go
```
func (p *abbParserParser) Module() (localctx IModuleContext) { ... }
func (s *ModuleContext) ModuleData() IModuleDataContext { ... }

type IModuleDataContext interface {
    antlr.ParserRuleContext

    // GetParser returns the parser.
    GetParser() antlr.Parser

    // Getter signatures
    MODULE() antlr.TerminalNode
    ModuleName() IModuleNameContext
    AllNEWLINE() []antlr.TerminalNode
    NEWLINE(i int) antlr.TerminalNode
    DataList() IDataListContext
    ENDMODULE() antlr.TerminalNode

    // IsModuleDataContext differentiates from other interfaces.
    IsModuleDataContext()
}

type ModuleDataContext struct {
	*antlr.BaseParserRuleContext
	parser antlr.Parser
}
```

### Navigation in Go
```
module.ModuleData().ModuleName()
```

# Why is this PR important?
This change increases the usability of the Go parsers. Without this change, code littered with type assertions is difficult to read/maintain.

### DIFF
--- a/tool/resources/org/antlr/v4/tool/templates/codegen/Go/Go.stg
+++ b/tool/resources/org/antlr/v4/tool/templates/codegen/Go/Go.stg
@@ -919,25 +919,25 @@ RuleContextListDecl(rdecl) ::= "<rdecl.escapedName> []I<rdecl.ctxName>"
 AttributeDecl(d) ::= "<d.escapedName> <d.type><if(d.initValue)>// TODO = <d.initValue><endif>"
 
 ContextTokenGetterDecl(t) ::= <<
-<t.escapedName; format="cap">() antlr.TerminalNode {
+<t.escapedName; format="cap">() antlr.TerminalNode<if(!t.signature)> {
 	return s.GetToken(<parser.name><t.escapedName>, 0)
-}
+}<endif>
 >>
 
 ContextTokenListGetterDecl(t) ::= <<
-All<t.escapedName; format="cap">() []antlr.TerminalNode {
+All<t.escapedName; format="cap">() []antlr.TerminalNode<if(!t.signature)> {
 	return s.GetTokens(<parser.name><t.escapedName>)
-}
+}<endif>
 >>
 
 ContextTokenListIndexedGetterDecl(t) ::= <<
-<t.escapedName; format="cap">(i int) antlr.TerminalNode {
+<t.escapedName; format="cap">(i int) antlr.TerminalNode<if(!t.signature)> {
 	return s.GetToken(<parser.name><t.escapedName>, i)
-}
+}<endif>
 >>
 
 ContextRuleGetterDecl(r) ::= <<
-<r.escapedName; format="cap">() I<r.ctxName> {
+<r.escapedName; format="cap">() I<r.ctxName><if(!r.signature)> {
 	var t antlr.RuleContext;
 	for _, ctx := range s.GetChildren() {
 		if _, ok := ctx.(I<r.ctxName>); ok {
@@ -951,11 +951,11 @@ ContextRuleGetterDecl(r) ::= <<
 	}
 
 	return t.(I<r.ctxName>)
-}
+}<endif>
 >>
 
 ContextRuleListGetterDecl(r) ::= <<
-All<r.escapedName; format="cap">() []I<r.ctxName> {
+All<r.escapedName; format="cap">() []I<r.ctxName><if(!r.signature)> {
 	children := s.GetChildren()
 	len := 0
 	for _, ctx := range children {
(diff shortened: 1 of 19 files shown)


j3r3miah:
Hi @jimidle, I'd appreciate your review here

jimidle:
I will look at this today. I notice one check does not pass though. We will need that to pass. On the face of it, this looks like the correct thing to generate

j3r3miah:
Thanks, @jimidle 

I saw that failure as well - it looks like infra flake to me:
> ERROR: The remote file either doesn't exist, is unauthorized, or is forbidden for url 'https://github.com/llvm/llvm-project/releases/download/llvmorg-15.0.2/LLVM-15.0.2-win64.exe'. Exception calling "GetResponse" with "0" argument(s): "The remote server returned an error: (503) Server Unavailable." 

I'm familiar with Github Actions: is there a way to retry that job (I don't see a retry button [anywhere](https://github.com/antlr/antlr4/actions/runs/3229253748/jobs/5286335998)) or do I have to push another commit to trigger a full rebuild?

tool/resources/org/antlr/v4/tool/templates/codegen/Go/Go.stg
```diff
@@ -1014,7 +1014,7 @@ ListLabelName(label) ::= "<label>"
 CaptureNextToken(d) ::= "<d.varName> = p.GetTokenStream().LT(1)"
 CaptureNextTokenType(d) ::= "<d.varName> = p.GetTokenStream().LA(1)"
 
-StructDecl(struct, ctorAttrs, attrs, getters, dispatchMethods, interfaces, extensionMembers) ::= <<
+StructDecl(struct, ctorAttrs, attrs, getters, dispatchMethods, interfaces, extensionMembers, signatures) ::= <<
```
j3r3miah, line 1017: Hmm I should probably also add this new `signatures` parameter to the `StructDecl` template in all the other targets as well, eh?

j3r3miah: (Added them)

j3r3miah:
Hey @jimidle, I'd appreciate that review when you have a chance. Thank you!

kaby76:
Looks wonderful. The generated code is exactly what we need, and mirrors what the other targets do. @parrt

parrt:
I can't tell if it's backward compatible...i don't wanna break code in google. ;)

I'll wait for @jimidle to sign off on this but looks better.

j3r3miah:
The only breaking scenario I can imagine is if user code is relying on _implementing_ the parser-generated interfaces for some reason, which seems unlikely - but I'll defer to the experts! Thankya.

jimidle:
@parrt We can merge this PR - it will not break anything as far as I can see.

parrt:
Thanks @jimidle !

parrt:
@perezd this should help the visitor, if you have time to check.

tool/src/org/antlr/v4/codegen/model/decl/ContextGetterDecl.java
```diff
@@ -11,14 +11,23 @@
...
+		this.signature = signature;
 	}
 
 	/** Not used for output; just used to distinguish between decl types
 	 *  to avoid dups.
 	 */
 	public String getArgType() { return ""; }; // assume no args
 
+	private boolean signature = false;
+	public boolean getSignature() { return signature; }
```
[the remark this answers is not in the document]
j3r3miah: Good call: https://github.com/antlr/antlr4/pull/3947

PR #3951, opened 2022-11-05
Issue 3845

kaby76:
Fix for the crashes in Cpp for various grammars. You can't dereference a null point.


kaby76:
Wrong "merge to" branch.

PR #3952, opened 2022-11-05
Fix for issue 3845 -- null pointer deref in Cpp target

kaby76:
Very simply, the code that compares two PredictionContext must test before dereferencing. https://github.com/antlr/antlr4/blob/aa1f1f12a846846fcb78544bdb2ad135471376db/runtime/Cpp/runtime/src/atn/ArrayPredictionContext.cpp#L25 . If both pointers are null then the compare is "true".

### DIFF
--- a/runtime/Cpp/runtime/src/atn/ArrayPredictionContext.cpp
+++ b/runtime/Cpp/runtime/src/atn/ArrayPredictionContext.cpp
@@ -22,12 +22,10 @@ namespace {
   }
 
   bool predictionContextEqual(const Ref<const PredictionContext> &lhs, const Ref<const PredictionContext> &rhs) {
-    if (lhs == nullptr && rhs == nullptr)
-        return true;
-    if (lhs == nullptr && !(rhs == nullptr))
-        return false;
-    if (!(lhs == nullptr) && rhs == nullptr)
-        return false;
+    if (lhs == nullptr)
+      return rhs == nullptr;
+    else if (rhs == nullptr)
+      return false;
     return *lhs == *rhs;
   }
 
--- a/runtime/Cpp/runtime/src/atn/ArrayPredictionContext.cpp
+++ b/runtime/Cpp/runtime/src/atn/ArrayPredictionContext.cpp
@@ -22,6 +22,12 @@ namespace {
   }
 
   bool predictionContextEqual(const Ref<const PredictionContext> &lhs, const Ref<const PredictionContext> &rhs) {
+    if (lhs == nullptr && rhs == nullptr)
+        return true;
+    if (lhs == nullptr && !(rhs == nullptr))
+        return false;
+    if (!(lhs == nullptr) && rhs == nullptr)
+        return false;
     return *lhs == *rhs;
   }
 


runtime/Cpp/runtime/src/atn/ArrayPredictionContext.cpp
```diff
@@ -22,6 +22,12 @@ namespace {
   }
 
   bool predictionContextEqual(const Ref<const PredictionContext> &lhs, const Ref<const PredictionContext> &rhs) {
+    if (lhs == nullptr && rhs == nullptr)
+        return true;
+    if (lhs == nullptr && !(rhs == nullptr))
+        return false;
+    if (!(lhs == nullptr) && rhs == nullptr)
+        return false;
```
KvanTTT, line 30: Please simplify:

```suggestion
    if (lhs == nullptr)
        return rhs == nullptr;
    else if (rhs == nullptr)
        return false;
```

kaby76: Sure.

But, I bet there are other places like this. For example, these two statements are very similar to this bug, with parameters of type `Ref<> &`.

* https://github.com/antlr/antlr4/blob/aa1f1f12a846846fcb78544bdb2ad135471376db/runtime/Cpp/runtime/src/atn/ATNConfig.h#L39
* https://github.com/antlr/antlr4/blob/aa1f1f12a846846fcb78544bdb2ad135471376db/runtime/Cpp/runtime/src/atn/PredictionContextCache.cpp#L55

People make changes that "fix" a tiny little problem, but don't read the damn code. @parrt what about these occurrences?

parrt: Yep, that's why I'm opposed to refactoring for refactoring sake. It can introduce all sorts of subtle bugs.

We have to be careful here for example because the real solution is that context objects should never be null... There is a sentinel that indicates empty stack. That is why I didn't simply add the null checks... There's a larger bug and we shouldn't paint over it.

kaby76: > Yep, that's why I'm opposed to refactoring for refactoring sake. It can introduce all sorts of subtle bugs.
> 
> We have to be careful here for example because the real solution is that context objects should never be null... There is a sentinel that indicates empty stack. That is why I didn't simply add the null checks... There's a larger bug and we shouldn't paint over it.

Then, these should be at least assert()'s so that it crashes with a message.

Is there a method that anyone wrote that checks the integrity of the data structure so I can percolate the check around and snuff this bug out? Someone is creating an array with a null pointer at the end of the array.

parrt: That's what I was hoping to do with the Data dump from the ATN simulator... We could use it to compare how the different targets work. 

BTW, here is the start of me trying to fix C++ by refracturing the complex code: https://github.com/antlr/antlr4/pull/3863

I gave up trying to debug when a print statement exploded all over my screen with compile errors.

Here is my attempt to start a tracer for ATN: https://github.com/antlr/antlr4/pull/3817

kaby76: Actually, I think your assumption that there cannot be null parents in an array of PredicateContext is wrong--or some of the ports are doing it wrong.

I wrote a Check() function that checks the integrity of the ArrayPredictionContext and it led to this code. This code creates a "parents" array containing a null pointer.

* Cpp:
https://github.com/antlr/antlr4/blob/16cfc6b122b184d1e412c7e175d4e12370b0b99d/runtime/Cpp/runtime/src/atn/PredictionContext.cpp#L320

Compare this with CSharp, Java, PHP, and Python3.

* CSharp:
https://github.com/antlr/antlr4/blob/16cfc6b122b184d1e412c7e175d4e12370b0b99d/runtime/CSharp/src/Atn/PredictionContext.cs#L391

* Java:
https://github.com/antlr/antlr4/blob/16cfc6b122b184d1e412c7e175d4e12370b0b99d/runtime/Java/src/org/antlr/v4/runtime/atn/PredictionContext.java#L324

* PHP:
https://github.com/antlr/antlr-php-runtime/blob/64fd1e2e8ca6fd80b57f53a0729e74b984bd767f/src/PredictionContexts/PredictionContext.php#L371

* Python3:
https://github.com/antlr/antlr4/blob/16cfc6b122b184d1e412c7e175d4e12370b0b99d/runtime/Python3/src/antlr4/PredictionContext.py#L420

For Dart, null pointers are disallowed (I think??), so the code looks very different.
* Dart:
https://github.com/antlr/antlr4/blob/16cfc6b122b184d1e412c7e175d4e12370b0b99d/runtime/Dart/lib/src/prediction_context.dart#L247

For Go, I think it too was written without using null pointers:
* Go:
https://github.com/antlr/antlr4/blob/16cfc6b122b184d1e412c7e175d4e12370b0b99d/runtime/Go/antlr/prediction_context.go#L536

kaby76:
Changing back from "draft" to "review" because ArrayPredictionContext can contain null pointers.

runtime/Cpp/runtime/src/atn/ArrayPredictionContext.cpp
parrt: Well, dang. You're right. Hmm...yeah, touching the ATN simulation is really risky as I don't remember the details well. :(

kaby76: Yes, I can relate to forgetting details. It definitely is concerning that a couple of the targets don't have nulls in the parents array. I think to be "safe", I should test this against the regression tests of grammars-v4 for Cpp.

parrt: I would love to do a comparison of the java target against other targets. I have good trust in java behavior.

runtime/Cpp/runtime/src/atn/ArrayPredictionContext.cpp
```diff
@@ -22,6 +22,10 @@ namespace {
   }
 
   bool predictionContextEqual(const Ref<const PredictionContext> &lhs, const Ref<const PredictionContext> &rhs) {
+    if (lhs == nullptr)
```
parrt, line 25: I wonder how badly this will hit performance

kaby76: > I wonder how badly this will hit performance

I think this question is difficult to answer without changing the Cpp runtime to not create "parent" arrays with null pointers. For tests that call the comparison function, the parse crashes; for tests that don't call the function, the parse works. Presumably, in Java when it calls [equals()](https://github.com/antlr/antlr4/blob/aa1f1f12a846846fcb78544bdb2ad135471376db/runtime/Java/src/org/antlr/v4/runtime/atn/ArrayPredictionContext.java#L78)] on comparing two "parents" arrays, it will test whether the element in the array is null (see [Arrays.equals()](https://github.com/openjdk/jdk/blob/0ee25de704eb8f179888812bd9b1e7bebc384fd6/src/java.base/share/classes/java/util/Arrays.java#L2979) => [Object.equals()](https://github.com/openjdk/jdk/blob/0ee25de704eb8f179888812bd9b1e7bebc384fd6/src/java.base/share/classes/java/util/Objects.java#L64)). This change makes Cpp operate the same way as the Java code. The choice is either to get the Cpp working, or rewrite all the runtimes to not use null pointers in "parents".

ericvergnaud: I suspect it doesn't make any sense to have nulls in parents arrays, so I would opt for x-target cleanup, which as a side effect is likely to slightly improve performance

parrt: I could swear that we had a sentinel node that represented top of the stack rather than null but I have to look at the code more carefully because @kaby76 found something that makes it look like it really is using null.  I am terrified of changing anything in the ATN simulation given just how much code in the world relies on it.

kaby76: The old code compare called [Arrays::equals()](https://github.com/antlr/antlr4/blob/53b83e329b67e7095a37bc155e172aea79f41558/runtime/Cpp/runtime/src/atn/ArrayPredictionContext.cpp#L55). That in turn would [test null pointers](https://github.com/antlr/antlr4/blob/master/runtime/Cpp/runtime/src/support/Arrays.h#L35). The nullptr in the parents array has existed for 6 years at least.

parrt:
Fixed by https://github.com/antlr/antlr4/pull/3958

PR #3958, opened 2022-11-14
Fix 3845 nullptr cpp

parrt:
Fixed while doing https://github.com/antlr/antlr4/pull/3817  Also see  @kaby76's fix that I implemented: https://github.com/antlr/grammars-v4/issues/2909

### DIFF
--- a/runtime/Cpp/runtime/src/atn/ArrayPredictionContext.cpp
+++ b/runtime/Cpp/runtime/src/atn/ArrayPredictionContext.cpp
@@ -22,6 +22,11 @@ namespace {
   }
 
   bool predictionContextEqual(const Ref<const PredictionContext> &lhs, const Ref<const PredictionContext> &rhs) {
+    // parent PredictionContext pointers can be null during full context mode and
+    // the ctxs are in an ArrayPredictionContext.  If both are null, return true
+    // if just one is null, return false. If both are non-null, do comparison.
+    if (lhs == nullptr && rhs == nullptr) return true;
+    if (lhs != nullptr || rhs != nullptr) return false;
     return *lhs == *rhs;
   }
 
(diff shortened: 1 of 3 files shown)


runtime/Cpp/runtime/src/atn/ArrayPredictionContext.cpp
```diff
@@ -22,6 +22,11 @@ namespace {
   }
 
   bool predictionContextEqual(const Ref<const PredictionContext> &lhs, const Ref<const PredictionContext> &rhs) {
+    // parent PredictionContext pointers can be null during full context mode and
+    // the ctxs are in an ArrayPredictionContext.  If both are null, return true
+    // if just one is null, return false. If both are non-null, do comparison.
+    if (lhs == nullptr && rhs == nullptr) return true;
+    if (lhs != nullptr || rhs != nullptr) return false;
     return *lhs == *rhs;
```
KvanTTT, line 30: It's incorrect. Last line is never executed since it `return false` if any or both arguments are not null. Also, it's not optimal. I suggest fixing in the following way:

```suggestion
    if (lhs == nullptr)
        return rhs == nullptr;
    else
        return rhs != nullptr;
    return *lhs == *rhs;
```

parrt: dang! you're right. I tried to get too clever. grrr...will fix. thanks.

parrt: is your line 32 dead code?

parrt: maybe this?

```
if ( lhs == nullptr ) return rhs == nullptr;
if ( rhs == nullptr ) return false; // lhs!=null and rhs==null
return *lhs == *rhs;                // both nonnull
```

KvanTTT: Not dead, it's comparison by dereference (if I remember and understand C++ correctly).

parrt: meaning your line 32 is never reached.

KvanTTT: Sorry, my first suggestion is also incorrect. The second version:

```c++
    if (lhs == nullptr)
        return rhs == nullptr;
    if (rhs == nullptr)
        return false;
    return *lhs == *rhs;
```

parrt: yep trying similar now. seems to work. this one still fails though, but this time with a SEGV https://github.com/antlr/antlr4/issues/3959 rather than infinite loop.

KvanTTT: Yes, it's correct. Didn't get update in time.

parrt: pushed the fix. thanks for finding!  I updated trace diff for failing test: https://github.com/antlr/antlr4/issues/3959

PR #3972, opened 2022-11-20
Fix issue 3965

ericvergnaud:


ericvergnaud:
@parrt fixes the issue
I'd expect it will only marginally affect performance since it didn't affect the algorithm itself

PR #3957, opened 2022-11-11
Generate identical atn simulation output across targets to compare parsing functionality

parrt:
Per https://github.com/antlr/antlr4/discussions/3814 trying to debug ATN sim.  For ref, previous thoughts: https://github.com/antlr/antlr4/pull/3817

PR #3974, opened 2022-11-20
Add ATN tracing support for PHP target

marcospassos:
Add ATN tracing support for the PHP target, as requested [here](https://github.com/antlr/antlr4/pull/3957#event-7850225906).

Companion PR:
https://github.com/antlr/antlr-php-runtime/pull/32

@parrt, could you please check if I'm missing something?


parrt:
Looks great.

PR #3982, opened 2022-11-25
[C#]Simplified code, faster performance

toolgood:
Simplified code, faster performance
One of my projects has the fastest running speed of 220ms, and the fastest running speed after modification is 198ms.


KvanTTT:
Hi. I suggest closing it in favor of #3983 since it fixes issues over the entire repository.

PR #3985, opened 2022-11-26
!Exit methods optional

incident-recipient:


incident-recipient:
The failures I'm seeing in the checks seem to be due to environment issues in the build system rather than problems with the code we're submitting.

python2 seems not to exist as a package (which is not too surprising given it has been deprecated by the python community)

The swift environment seems to have a problem with GPG.

Ubuntu latest complains of an unsupported version.

I'm not seeing anything actionable on my end here, but feel free to ask if it there is some place I can contribute.

incident-recipient:
Cross linking to issue https://github.com/antlr/antlr4/issues/3986

Issue #3988, opened 2022-11-26
github actions now fail for python2 and ubuntu clang and ubuntu swift


parrt:
Hi @hs-apotell any chance you could take a quick look to see about getting python2 / clang / swift up and running again?

https://github.com/antlr/antlr4/actions/runs/3551377074/jobs/5965500626

We're working on getting swift to reuse the same runtime lib but it works locally (just really slowly): https://github.com/antlr/antlr4/pull/3984

Here are the errors related to install for actions:

```
Run egor-tensin/setup-clang@v1
Run New-Variable os -Value 'Linux' -Option Constant
Warning: apt-key is deprecated. Manage keyring files in trusted.gpg.d instead (see apt-key([8](https://github.com/antlr/antlr4/actions/runs/3551377074/jobs/5965499687#step:4:9))).
OK
Exception: /home/runner/work/_temp/95ed61fb-bcf9-4641-bdf3-1b57700101c5.ps1:90
Line |
  90 |          default { throw "Unsupported distribution: $distro" }
     |                    ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
     | Unsupported distribution: Ubuntu-22.04
```

```
Run actions/setup-python@v4
Version 2.x was not found in the local cache
Error: Version 2.x with arch x64 not found
The list of all available versions can be found here: https://raw.githubusercontent.com/actions/python-versions/main/versions-manifest.json
```

```
gpg: no valid OpenPGP data found.
gpg: the signature could not be verified.
Please remember that the signature file (.sig or .asc)
should be the first file given on the command line.
Error: Unexpected error, unable to continue. Please report at https://github.com/swift-actions/setup-swift/issues
The process '/usr/bin/gpg' failed with exit code 2
```

parrt:
Fixed by https://github.com/antlr/antlr4/pull/3989

PR #3989, opened 2022-11-27
Issue #3988: Builds failing on Ubuntu

hs-apotell:
Issue #3988: Builds failing on Ubuntu

The workflow is setup to work on Ubuntu-20.04 and the default 'latest' is migrating to Ubuntu-22.04 which isn't supported yet (all the dependent actions needs to support the platform before the workflow can use it).

Switching all platforms to use explicit versions rather than the '-latest' tag so the transition is more voluntary than something that just happens behind the scene.

Also, upgraded all setup-xx actions to latest versions to remove some of the deprecation warnings on dashboard.

Setup concurrency as well (kill any existing build running on the same branch branch and using the same workflow).

### DIFF
--- a/.github/workflows/hosted.yml
+++ b/.github/workflows/hosted.yml
@@ -1,5 +1,9 @@
 name: antlr4
 
+concurrency:
+  group: ${{ github.repository }}-${{ github.workflow }}-${{ github.ref }}
+  cancel-in-progress: true
+
 on:
   push:
     branches: [ master, dev, hostedci ]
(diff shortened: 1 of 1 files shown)


parrt:
thanks @hs-apotell !!!

parrt:
love this part:

> Setup concurrency as well (kill any existing build running on the same branch branch and using the same workflow).

parrt:
BTW, swift actually runs python ;)  That's what we discovered then realized swift didn't build.

>  Running org.antlr.v4.test.runtime.python3.Python3RuntimeTests

PR #3999, opened 2022-12-03
Go: InputStream GetTextFromTokens can't get the content of the tokens correctly

lemon-1997:
I'm using antrl4 these days, and I found this problem, when I try to get the original content from two tokens, I can't get it correctly,
```go
start := ctx.TokenBegin()
stop := ctx.TokenEnd()
content := ctx.TokenBegin().GetInputStream().GetTextFromTokens(start, stop)
```
so I can only do it this way, I think it's a bug
```go
start := ctx.TokenBegin().GetStart()
stop := ctx.TokenEnd().GetStop()
content := ctx.TokenBegin().GetInputStream().GetText(start, stop)
```

### DIFF
--- a/runtime/Go/antlr/v4/input_stream.go
+++ b/runtime/Go/antlr/v4/input_stream.go
@@ -94,7 +94,7 @@ func (is *InputStream) GetText(start int, stop int) string {
 
 func (is *InputStream) GetTextFromTokens(start, stop Token) string {
 	if start != nil && stop != nil {
-		return is.GetTextFromInterval(NewInterval(start.GetTokenIndex(), stop.GetTokenIndex()))
+		return is.GetTextFromInterval(NewInterval(start.GetStart(), stop.GetStop()))
 	}
 
 	return ""


runtime/Go/antlr/input_stream.go
```diff
@@ -94,7 +94,7 @@ func (is *InputStream) GetText(start int, stop int) string {
 
 func (is *InputStream) GetTextFromTokens(start, stop Token) string {
 	if start != nil && stop != nil {
-		return is.GetTextFromInterval(NewInterval(start.GetTokenIndex(), stop.GetTokenIndex()))
+		return is.GetTextFromInterval(NewInterval(start.GetStart(), stop.GetStop()))
```
kaby76, line 97: Shouldn't this go in runtime/Go/antlr/***v4/***input_stream.go, not runtime/Go/antlr/input_stream.go?

lemon-1997: Sorry, it's ok now, I push the new code

jimidle:
GetTextFromTokens() is correct in v4 - I just tested it - the change above is not required. 

The OP does not say what is wrong with the text returned, but I suspect that the issue is with the parameters being give to GetTextFromTokens(), which may imply that ctx.TokenEnd() is not returning the correct token.

Please close this PR as it is the wrong fix. I will check TokenStart() and TokenEnd() and fix any issue there...

parrt:
Yes I am supreme dictator for life ;) closing

PR #4027, opened 2022-12-18
Typescript target

ericvergnaud:
Candidate PR for the typescript target.

PR #4030, opened 2022-12-20
Extract FileUtils updates by @ericvergnaud

parrt:
Extract FileUtils updates by @ericvergnaud in https://github.com/antlr/antlr4/pull/4027 then he can rebase after we merge this into dev.  All tests pass locally (didn't check python2 actually but python3 works).

Signed-off-by: Terence Parr <parrt@antlr.org>

### DIFF
--- a/runtime-testsuite/test/org/antlr/v4/test/runtime/FileUtils.java
+++ b/runtime-testsuite/test/org/antlr/v4/test/runtime/FileUtils.java
@@ -8,9 +8,15 @@ package org.antlr.v4.test.runtime;
 
 import org.antlr.v4.runtime.misc.Utils;
 
-import java.io.*;
+import java.io.File;
+import java.io.IOException;
+import java.io.PrintWriter;
 import java.nio.charset.StandardCharsets;
-import java.nio.file.*;
+import java.nio.file.Files;
+import java.nio.file.LinkOption;
+import java.nio.file.Path;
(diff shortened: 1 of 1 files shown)


parrt:
@ericvergnaud ready for your review once tests pass. Then we can merge before ts.

runtime-testsuite/test/org/antlr/v4/test/runtime/FileUtils.java
```diff
@@ -55,7 +61,7 @@ public static void mkdir(String dir) {
 	}
 
 	public static void deleteDirectory(File f) throws IOException {
-		if (f.isDirectory()) {
+		if (f.isDirectory() && !isLink(f.toPath())) {
```
KvanTTT, line 64: Could you clarify why link checking during deleting is important? What does it affect on?

ericvergnaud: During tests setup, the Antlr package is webpacked.
For each test, that package is installed as a dependency of the project in node_modules, using npm link.
Without this change, the antlr package gets deleted as part of destroying the test folder, which makes all subsequent tests fail.

parrt:
During tests setup, the Antlr package is webpacked.
For each test, that package is installed as a dependency of the project in node_modules, using npm link.
Without this change, the antlr package gets deleted as part of destroying the test folder, which makes all subsequent tests fail.

PR #4034, opened 2022-12-20
Eric file objs update

parrt:
Bug #3878 was missing some scenarios regarding scope of header actions.


parrt:
Passes locally @ericvergnaud so merging.

PR #4039, opened 2022-12-22
Tree agregateresult python3

pawelmagnu:
fixed python3 tree aggregate result function to append results from children rather than returning last child result


ericvergnaud:
Hi,
thanks for the proposal but unfortunately it would not only break backwards compatibility but also consistency with other targets (which all return just `nextResult `).
I believe it's easy enough to override this behavior in your own ParseTreeVisitor, so closing.

Issue #2209, opened 2018-01-24
Imported Lexer Grammar is lower priority


kasbah:
Example.g4
```antlr
grammar Example;
import Example2;

main: NUMBER;

UNKNOWN: . ;
```
Example2.g4
```antlr
lexer grammar Example2;

NUMBER: [0-9];
```

Running: 
```bash
echo "10 abc" | grun Example main -diagnostics -tree -tokens
```

```
[@0,0:0='1',<UNKNOWN>,1:0]
[@1,1:1='0',<UNKNOWN>,1:1]
[@2,2:2=' ',<UNKNOWN>,1:2]
[@3,3:3='a',<UNKNOWN>,1:3]
[@4,4:4='b',<UNKNOWN>,1:4]
[@5,5:5='c',<UNKNOWN>,1:5]
[@6,6:6='\n',<UNKNOWN>,1:6]
[@7,7:6='<EOF>',<EOF>,2:0]
line 1:0 mismatched input '1' expecting NUMBER
(main 1 0   a b c \n)
```

Moving the UNKNOWN rule into Example2:

```diff
--- a/Example.g4
+++ b/Example.g4
@@ -3,4 +3,3 @@ import Example2;
 
 main: NUMBER;
 
-UNKNOWN: . ;

--- a/Example2.g4
+++ b/Example2.g4
@@ -1,3 +1,5 @@
 lexer grammar Example2;
 
 NUMBER: [0-9];
+
+UNKNOWN: . ;
```

Results in the expected output:

```
[@0,0:0='1',<NUMBER>,1:0]
[@1,1:1='0',<NUMBER>,1:1]
[@2,2:2=' ',<UNKNOWN>,1:2]
[@3,3:3='a',<UNKNOWN>,1:3]
[@4,4:4='b',<UNKNOWN>,1:4]
[@5,5:5='c',<UNKNOWN>,1:5]
[@6,6:6='\n',<UNKNOWN>,1:6]
[@7,7:6='<EOF>',<EOF>,2:0]
(main 1)
```

kasbah:
Files: [import-test-case.zip](https://github.com/antlr/antlr4/files/1658242/import-test-case.zip)

ericvergnaud:
[...]

PR #4044, opened 2022-12-25
Fix precedence of imported rules #2209

ericvergnaud:


ericvergnaud:
Fixes #2209

ericvergnaud:
FYI the failing tests all relate to token rule precedence, easy to fix

PR #4051, opened 2022-12-31
Typescript target fix missing apis

ericvergnaud:


ericvergnaud:
@parrt this is ready to merge

PR #4053, opened 2023-01-02
[TypeScript] Fixed typings, added typescript testing

erayhanoglu:


ericvergnaud:
Hi,
thanks for this, but not sure what problem you are trying to solve ?
(I've been using the beta for 2 months without any problem)

erayhanoglu:
Hi Eric,
Thank you for fast response. I will try to explain the problems below,

1- First problem is, in JS files all things are exported using "default export" but in TypeScript definition files, things are exported as "named export". The two system are not same. 

2- Exports in index.js and index.d.ts are different. Many exported things are not exported in d.ts file, and some of then are exported wrong.

For example when you try to import ParseTreeVisitor, typescript will give you no error. But index.js there is not such export named ParseTreeVisitor. It is exported in "tree" namesace.

The code below will give you no error in typescript. But when you run the app, you will see that ParseTreeVisitor is undefined.
```ts
import { ParseTreeVisitor} from 'antlr4';
console.log(ParseTreeVisitor); // prints undefined
```

In index.js
```
export default {
    ..., tree, error, ....
}
```
This line adds a "tree" property into module.exports object which contains exported members.

In index.d.ts
```ts
export * from './tree';
```
This line merges all exported members in ./tree into module.exports object.

ericvergnaud:
Thanks for the explanation. Here are my comments:
 - a 'difference' is not a problem
 - your PR fail the CI completely, and the number of changes makes it impossible to determine which one went wrong
As suggested, let's deal with genuine problems 1 by 1, starting with ParseTreeVisitor (the 1 thing that our CI doesn't test and I'm not currently using) ?

PR #4055, opened 2023-01-03
[TypeScript] Fixed typings, added typescript testing

erayhanoglu:
This patch fixes invalid definition files.

### DIFF
--- a/runtime/JavaScript/package.json
+++ b/runtime/JavaScript/package.json
@@ -39,7 +39,7 @@
   },
   "scripts": {
     "build": "webpack",
-    "test": "jasmine"
+    "test": "jasmine && tsc"
   },
   "engines": {
     "node": ">=16"
--- a/runtime/JavaScript/src/antlr4/BufferedTokenStream.d.ts
+++ b/runtime/JavaScript/src/antlr4/BufferedTokenStream.d.ts
@@ -1,8 +1,10 @@
-import { TokenStream } from './TokenStream';
-import { Lexer } from "./Lexer";
+import TokenStream from './TokenStream';
+import Lexer from "./Lexer";
 
-export declare class BufferedTokenStream extends TokenStream {
+declare class BufferedTokenStream extends TokenStream {
 
     tokenSource: Lexer;
 
 }
+
+export default BufferedTokenStream;
--- a/runtime/JavaScript/src/antlr4/CharStream.d.ts
+++ b/runtime/JavaScript/src/antlr4/CharStream.d.ts
@@ -1,4 +1,4 @@
-export declare class CharStream {
+declare class CharStream {
 
     index: number; // defined as property
     size: number;// defined as property
@@ -15,3 +15,5 @@ export declare class CharStream {
     getText(start: number, stop: number): string;
     toString(): string;
 }
+
+export default CharStream;
--- a/runtime/JavaScript/src/antlr4/CharStreams.d.ts
+++ b/runtime/JavaScript/src/antlr4/CharStreams.d.ts
@@ -1,9 +1,11 @@
-import {CharStream} from "./CharStream";
+import CharStream from "./CharStream";
 
-export declare class CharStreams {
+declare class CharStreams {
     static fromString(data: string, decodeToUnicodeCodePoints?: boolean): CharStream;
     static fromBuffer(buffer: Buffer, encoding?: string): CharStream;
     static fromBlob(blob: Blob, encoding: string, onLoad: (stream: CharStream) => void, onError: (error: Error) => void): void;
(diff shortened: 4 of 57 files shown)


erayhanoglu:
Hi Eric. I sat the problem in failing tests and applied a fix.

About our last conversation.
1-
For a simple project which only imports base classes exported by index.js, it may work. For complex projects which needs to import other classes and methods it doesn't work. Because export definitions are wrong.
All js files in the project using default export, and d.ts files uses named exports. If you try to import a file in src/ folder you will see the problem. Here is an example
```ts
import {CharStream} from 'antlr4/src/antlr4/CharStream.js';
console.log(CharStream);
```
Thescript will give you no error, because according to d.ts file export = {CharStream} but in js export = CharStream . Whe you run here is what you will see: 
```sh
import { CharStream } from 'antlr4/src/antlr4/CharStream.js';
         ^^^^^^^^^^
SyntaxError: The requested module 'antlr4/src/antlr4/CharStream.js' does not provide an export named 'CharStream'
    at ModuleJob._instantiate (node:internal/modules/esm/module_job:124:21)
    at async ModuleJob.run (node:internal/modules/esm/module_job:190:5)
```

And also export definitions in index.d.ts are wrong also. For example, Index.js exports "atn, dfa, context, misc, tree, error" namespaces. There is no such namespaces in index.d.ts. Index.js exports LL1Analyzer there is no such definition in d.ts

In the example below you will get typescript error. 
```ts
import antlr4, {LL1Analyzer} from 'antlr4';
```
```sh
TS2305: Module '"antlr4"' has no exported member 'LL1Analyzer'.
```
But if you run the code below,
```ts
import antlr4 from 'antlr4';
console.log((antlr4 as any).LL1Analyzer);
```
```sh
[class LL1Analyzer] { HIT_PRED: 0 }
```
it is there!

2- I see that "atn, dfa, context, misc, tree, error" namespaces need to be exported in to root. This is my mistake
3- About the failing test, 
```sh
Error: L.ts(7,2): error TS2305: Module '"antlr4"' has no exported member 'DecisionState'.
```
index.d.ts exports all members in './state'. So DecisionState exported as a type. But index.js does not exports that. So this should be added in to index.js
```js
..
import { default as error } from './error/index.js';
import { default as state } from './state/index.js';
..
export default {
    atn, dfa, context, misc, tree, error, state, ...
}
```

ericvergnaud:
Well you’re supposed to write Import { CharStream } from ‘antlr4’Why are you not doing that ?Envoyé de mon iPhoneLe 3 janv. 2023 à 08:36, Eray Hanoglu ***@***.***> a écrit :﻿
Hi Eric. I sat the problem in failing tests and applied a fix.
About our last conversation.
1-
For a simple project which only imports base classes exported by index.js, it may work. For complex projects which needs to import other classes and methods it doesn't work. Because export definitions are wrong.
All js files in the project using default export, and d.ts files uses named exports. If you try to import a file in src/ folder you will see the problem. Here is an example
import {CharStream} from 'antlr4/src/antlr4/CharStream.js';
console.log(CharStream);
Thescript will give you no error, because according to d.ts file export = {CharStream} but in js export = CharStream . Whe you run here is what you will see:
import { CharStream } from 'antlr4/src/antlr4/CharStream.js';
         ^^^^^^^^^^
SyntaxError: The requested module 'antlr4/src/antlr4/CharStream.js' does not provide an export named 'CharStream'
    at ModuleJob._instantiate (node:internal/modules/esm/module_job:124:21)
    at async ModuleJob.run (node:internal/modules/esm/module_job:190:5)
And also export definitions in index.d.ts are wrong also. For example, Index.js exports "atn, dfa, context, misc, tree, error" namespaces. There is no such namespaces in index.d.ts. Index.js exports LL1Analyzer there is no such definition in d.ts
In the example below you will get typescript error.
import antlr4, {LL1Analyzer} from 'antlr4';
TS2305: Module '"antlr4"' has no exported member 'LL1Analyzer'.
But if you run the code below,
import antlr4 from 'antlr4';
console.log((antlr4 as any).LL1Analyzer);
[class LL1Analyzer] { HIT_PRED: 0 }
it is there!
2- I see that "atn, dfa, context, misc, tree, error" namespaces need to be exported in to root. This is my mistake
3- About the failing test,
Error: L.ts(7,2): error TS2305: Module '"antlr4"' has no exported member 'DecisionState'.
index.d.ts exports all members in './state'. So DecisionState exported as a type. But index.js does not exports that. So this should be added in to index.js
..
import { default as error } from './error/index.js';
import { default as state } from './state/index.js';
..
export default {
    atn, dfa, context, misc, tree, error, state, ...
}

erayhanoglu:
I can not execute tests in github actions in my own. Now i am closing this pull request and starting from the scratch. Lets do this fix from like you say, one by one.

PR #4076, opened 2023-01-16
Javascript fix missing export

ericvergnaud:


ericvergnaud:
@parrt this one is ready too

Issue #3708, opened 2022-05-09
Cpp: std::call_once requires linking to pthreads (on Linux)


Krzmbrzl:
When using the C++ target, the generated lexer will use a call to `std::call_once` in its `initialize` method. However, without further action, this will throw `std::runtime_error` with the message `Unknown error -1`.

Thanks to https://stackoverflow.com/questions/51584960/stdcall-once-throws-stdsystem-error-unknown-error-1, I figured out that it is necessary to link against pthreads, since `std::call_once` requires a function from it that is not found otherwise (triggering the error).

The following cmake snippet is enough to accomplish this:
```cmake
find_package(Threads REQUIRED)
target_link_libraries(myTarget PRIVATE Threads::Threads)
```

Now, I think that this dependency should be resolved on ANTLR's side such that for any downstream user it is enough to just link against the ANTLR runtime and thanks to cmake's magic the dependency will propagate. Thus, my suggestion would be to link ANTLR's C++ runtime against pthreads using above snippet.

At this point, I am not sure whether on other platforms other libraries might be required in order for this to work. Does someone have experience with C++ ANTLR on macOS and Windows?

EDIT: Based on the [docs](https://cmake.org/cmake/help/latest/module/FindThreads.html) it seems that above cmake snippet is actually cross-platform and not pthread-specific, so maybe we don't have to pay any special attention here?

zhaochenyuangit:
You are a life-saver

Krzmbrzl:
Since this appears to affect multiple users and nobody seems to want to answer my questions here, I went ahead and created a PR as I see fit: #3794

PR #4086, opened 2023-01-22
cpp runtime: Remove pthread dependency.

ArthurSonzogni:
ANTLR doesn't use threads, and it used not to depend on pthread library either.

It changed recently in 2022: https://github.com/antlr/antlr4/issues/3708 The patch linked against pthread, because the GNU libstdc++ used to depend on it in their implementation of `std::call_once`.

By the way, the libstdc++ stopped it in 2020:
https://gcc.gnu.org/git/gitweb.cgi?p=gcc.git;h=93e79ed391b9c636f087e6eb7e70f14963cd10ad So this is not more needed.

I would like to stop depending on pthread. I am using ANTLR in C++ WebAssembly for the website:
https://arthursonzogni.com/Diagon/

It doesn't compile with emscripten anymore, because by default pthread is not enabled. It could be enabled, but it would force me to deploy cross-origin-isolation:
https://web.dev/cross-origin-isolation-guide/

Solutions:
1. Stop linking against pthread, because the libstdc++ stopped depending on it for std::call_once.
2. Implement std::call_once ourself using std::atomic_flag
3. Implement std::call_once ourself using a boolean flag, assuming we don't need to support threads.

I chose to do (2) in this patch.

### DIFF
--- a/runtime/Cpp/runtime/CMakeLists.txt
+++ b/runtime/Cpp/runtime/CMakeLists.txt
@@ -28,13 +28,6 @@ file(GLOB libantlrcpp_SRC
 add_library(antlr4_shared SHARED ${libantlrcpp_SRC})
 add_library(antlr4_static STATIC ${libantlrcpp_SRC})
 
-# Make sure to link against threads (pthreads) library in order to be able to
-# make use of std::call_once in the code without producing runtime errors
-# (see also https://github.com/antlr/antlr4/issues/3708 and/or https://stackoverflow.com/q/51584960).
-find_package(Threads REQUIRED)
-target_link_libraries(antlr4_shared Threads::Threads)
-target_link_libraries(antlr4_static Threads::Threads)
-
 if (ANTLR_BUILD_CPP_TESTS)
   include(FetchContent)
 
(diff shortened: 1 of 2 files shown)


ArthurSonzogni:
@Krzmbrzl WDYT?

Krzmbrzl:
Sounds good to me. Option 1) would be an issue imo, because it restricts usability of ANTLR to somewhat recent systems only. However, implementing this ourselves appears as a solid workaround for our use case :+1:

ArthurSonzogni:
Okay. Then, I think this is ready. Thanks!
The latest CQ run failed, but that was unrelated to my patch. I rebased to run them again.

parrt:
Thanks!

parrt:
Hmm..seems we have a failure:

https://github.com/antlr/antlr4/actions/runs/4031407036/jobs/6930678126

```
CMake suite maintained and supported by Kitware (kitware.com/cmake).
/usr/local/bin/ninja
1.11.1
/usr/local/opt/ccache/libexec/gcc-9
ccache: error: Could not find compiler "gcc-9" in PATH
```

parrt:
I'll retry it.

ArthurSonzogni:
Thanks for the code review!

parrt:
damn. this was on master not dev.

parrt:
Can you base this on latest dev and resubmit? thanks. @ArthurSonzogni

PR #4094, opened 2023-02-02
Python2+3 runtimes: Define getTokenNames.

### LINKED ISSUE #1043: getTokenNames method is missing in the Recognizer class
The getTokenNames method is missing in the Recognizer class.

getTokenNames method was marked as `Deprecated` [here](http://www.antlr.org/api/Java/org/antlr/v4/runtime/Recognizer.html). It says that getVocabulary() should be used instead. At the same time this method is also missing.

Also there is a method that uses missing getTokenNames():

```
def getTokenTypeMap(self):
        tokenNames = self.getTokenNames()
        if tokenNames is None:
            from antlr4.error.Errors import UnsupportedOperationException
            raise UnsupportedOperationException("The current recognizer does not provide a list of token names.")
        result = self.tokenTypeMapCache.get(tokenNames, None)
```

Zannick:
Fixes #1043: a list of token names can be easily formulated by combining the lists of literalNames and symbolicNames, since they use the same indexing system.

Also similarly fixes the absence of getRuleNames.

Also fixes attempting to use the result of zip() as a map which broke both getTokenTypeMap and getRuleIndexMap. In Python 2, zip() returns a list of tuples. In Python3, zip returns an object that can only be iterated over once (and thus would not be useful to cache).

### DIFF
--- a/runtime/Python2/src/antlr4/Recognizer.py
+++ b/runtime/Python2/src/antlr4/Recognizer.py
@@ -45,6 +45,12 @@ class Recognizer(object):
     def removeErrorListeners(self):
         self._listeners = []
 
+    def getTokenNames(self):
+        return tuple(self.literalNames + self.symbolicNames[len(self.literalNames):])
+
+    def getRuleNames(self):
+        return tuple(self.ruleNames)
+
     def getTokenTypeMap(self):
         tokenNames = self.getTokenNames()
         if tokenNames is None:
@@ -52,7 +58,7 @@ class Recognizer(object):
             raise UnsupportedOperationException("The current recognizer does not provide a list of token names.")
         result = self.tokenTypeMapCache.get(tokenNames, None)
         if result is None:
-            result = zip( tokenNames, range(0, len(tokenNames)))
+            result = dict(zip(tokenNames, range(0, len(tokenNames))))
             result["EOF"] = Token.EOF
             self.tokenTypeMapCache[tokenNames] = result
         return result
@@ -68,7 +74,7 @@ class Recognizer(object):
             raise UnsupportedOperationException("The current recognizer does not provide a list of rule names.")
         result = self.ruleIndexMapCache.get(ruleNames, None)
         if result is None:
-            result = zip( ruleNames, range(0, len(ruleNames)))
+            result = dict(zip(ruleNames, range(0, len(ruleNames))))
             self.ruleIndexMapCache[ruleNames] = result
         return result
 
--- a/runtime/Python3/src/antlr4/Recognizer.py
+++ b/runtime/Python3/src/antlr4/Recognizer.py
@@ -49,6 +49,12 @@ class Recognizer(object):
     def removeErrorListeners(self):
         self._listeners = []
 
+    def getTokenNames(self):
+        return tuple(self.literalNames + self.symbolicNames[len(self.literalNames):])
+
+    def getRuleNames(self):
+        return tuple(self.ruleNames)
+
     def getTokenTypeMap(self):
         tokenNames = self.getTokenNames()
         if tokenNames is None:
@@ -56,7 +62,7 @@ class Recognizer(object):
             raise UnsupportedOperationException("The current recognizer does not provide a list of token names.")
         result = self.tokenTypeMapCache.get(tokenNames, None)
         if result is None:
-            result = zip( tokenNames, range(0, len(tokenNames)))
+            result = dict(zip(tokenNames, range(0, len(tokenNames))))
             result["EOF"] = Token.EOF
             self.tokenTypeMapCache[tokenNames] = result
         return result
@@ -72,7 +78,7 @@ class Recognizer(object):
             raise UnsupportedOperationException("The current recognizer does not provide a list of rule names.")
         result = self.ruleIndexMapCache.get(ruleNames, None)
         if result is None:
-            result = zip( ruleNames, range(0, len(ruleNames)))
+            result = dict(zip(ruleNames, range(0, len(ruleNames))))
             self.ruleIndexMapCache[ruleNames] = result
         return result
 


runtime/Python2/src/antlr4/Recognizer.py
```diff
@@ -68,7 +74,7 @@ def getRuleIndexMap(self):
             raise UnsupportedOperationException("The current recognizer does not provide a list of rule names.")
         result = self.ruleIndexMapCache.get(ruleNames, None)
         if result is None:
-            result = zip( ruleNames, range(0, len(ruleNames)))
+            result = dict(zip(ruleNames, range(0, len(ruleNames))))
```
ericvergnaud, line 77: According to your submission, this is not required in Python 2 ?

runtime/Python2/src/antlr4/Recognizer.py
```diff
@@ -45,14 +45,20 @@ def removeErrorListener(self, listener):
     def removeErrorListeners(self):
         self._listeners = []
 
+    def getTokenNames(self):
+        return tuple(self.literalNames + self.symbolicNames[len(self.literalNames):])
```
ericvergnaud, line 49: why return a tuple rather than a list ?

runtime/Python2/src/antlr4/Recognizer.py
```diff
@@ -45,14 +45,20 @@ def removeErrorListener(self, listener):
...
+
+    def getRuleNames(self):
+        return tuple(self.ruleNames)
```
ericvergnaud, line 52: same here

runtime/Python3/src/antlr4/Recognizer.py
```diff
@@ -49,14 +49,20 @@ def removeErrorListener(self, listener):
     def removeErrorListeners(self):
         self._listeners = []
 
+    def getTokenNames(self):
+        return tuple(self.literalNames + self.symbolicNames[len(self.literalNames):])
```
ericvergnaud, line 53: same here

runtime/Python3/src/antlr4/Recognizer.py
```diff
@@ -49,14 +49,20 @@ def removeErrorListener(self, listener):
...
+
+    def getRuleNames(self):
+        return tuple(self.ruleNames)
```
ericvergnaud, line 56: same here

runtime/Python2/src/antlr4/Recognizer.py
ericvergnaud: and why not just use self.ruleNames ?

runtime/Python2/src/antlr4/Recognizer.py
Zannick: The function name does, however, imply that it is returning a map. Given that we're caching it, it's probably better to cache the map rather than the list of pairs?

runtime/Python2/src/antlr4/Recognizer.py
Zannick: This is being used as a dict key in the calling function and so needs to be a hashable type, of which list is not one; the alternative is to cast it to a tuple in getTokenTypeMap/getRuleIndexMap. Would you prefer that?

runtime/Python2/src/antlr4/Recognizer.py
Zannick: Defining the getRuleNames function allows better backwards compatibility for anyone overriding it, rather than just replacing the one call site. Although my change to return a dict from getRuleIndexMap may already be a breaking change... let me know what you'd prefer.

runtime/Python3/src/antlr4/Recognizer.py
Zannick: Let's discuss in the other thread and I'll update here as necessary.

runtime/Python3/src/antlr4/Recognizer.py
Zannick: Let's discuss in the other thread and I'll update here as necessary.

runtime/Python2/src/antlr4/Recognizer.py
ericvergnaud: a collection of token names used as a dict key ? not sure I understand...

ericvergnaud:
this PR proposes too many unrelated changes, please split:
 - dict vs zip
 - token names
 - rule names

runtime/Python2/src/antlr4/Recognizer.py
ericvergnaud: since the function doesn't currently exist, not sure what you mean by "backwards compatibility " ?

runtime/Python2/src/antlr4/Recognizer.py
ericvergnaud: also I have some doubt about your logic. The name of a token can be its literal name or its symbolic name, concatenating both lists makes no sense to me.

ericvergnaud: have you tested this ?

Zannick: 1.   Yes, the implementations of getTokenTypeMap/getRuleIndexMap use the collection as a dict key, see lines 69 and 72: https://github.com/antlr/antlr4/blob/76fa05c21b12b96a6c12a0a82e611ed9d87d5af4/runtime/Python2/src/antlr4/Recognizer.py#L64-L73
2.   Literals and symbols are two different types of tokens, none have both a literal and a symbolic name. Taking from the grammar I've been working on, the lexer's ruleNames are like `[ "T__0", "T__1", ... "AND", "OR", "NOT" ]`, and these correspond to the properties on the generated Lexer class, where `T__0 = 1`, `AND = 20`, etc. The lexer's symbolic names are only "AND", etc, and not "T__0" etc, while the literal names are `"'('"` etc. So this concatenation would generate a list that can be used like `tokenNames[20]` produces `"AND"` and `tokenNames[1]` produces `"'('"`, with `"T__0"` etc not in the list at all.
    Now, there is a difference between the generated Lexer and Parser classes that I hadn't noticed before and I will have to fix; the Parser class's `symbolicNames` contains the string `"<INVALID>"` for index 0 and all the literal token types, so my first symbol "AND" is at index 20. But the Lexer class has one `"<INVALID>"` at index 0 and then "AND" is at index 1.
3.    Only manually, with the Parser subclass, and I ran the existing parse tests for the Python targets. I couldn't find a convenient way to test any other functionality specific to the Python runtimes.

runtime/Python2/src/antlr4/Recognizer.py
Zannick: Users of the library can still define it, supposing that getRuleIndexMap is functioning correctly otherwise.

ericvergnaud:
I think the right approach here would be to implement the vocabulary concept that is provided with Java and C# runtimes.

PR #4096, opened 2023-02-02
fix dev branch TypeScript target some typo

windwiny:
fix dev branch TypeScript target some typo

### DIFF
--- a/doc/typescript-target.md
+++ b/doc/typescript-target.md
@@ -89,7 +89,7 @@ class CustomVisitor extends MyGrammarVisitor {
   }
 }
 
-tree.accept(new Visitor());
+tree.accept(new CustomVisitor());
 ````
 
 ## How do I create and run a custom listener?
--- a/runtime-testsuite/resources/org/antlr/v4/test/runtime/helpers/Test.ts.stg
+++ b/runtime-testsuite/resources/org/antlr/v4/test/runtime/helpers/Test.ts.stg
@@ -1,5 +1,7 @@
+import * as fs from 'fs';
 import {
-	FileStream,
+	//FileStream,
+	CharStreams,
 	CommonTokenStream,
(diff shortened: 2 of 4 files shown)


ericvergnaud:
Thanks but your proposed changes break the build.

ericvergnaud:
Hi, thanks for this.
We will NOT drop the FileStream so rollback the corresponding changes.
Also a PR should fix 1 issue, so please split this PR between docs and templates.

windwiny:
> Hi, thanks for this. We will NOT drop the FileStream so rollback the corresponding changes. Also a PR should fix 1 issue, so please split this PR between docs and templates.

Hi, i see on the master branch ,  `webpack.config.cjs` `target` is `web`, 
on dev branch add TS target commits change it to `node16`, and add module FileStream, it use `fs` , this works on Node environment only, NOT work on web browser script.  

If this package drop fs/FileStream, or restore target to `web`, node environment user itself can import and use `fs`. 
Or we divide this `antlr4` into two packages, corresponding to the Node/Web environment ? 
otherwise every web env user must rebuild dist itself.


Maybe you can use git cherry-pick merge these small changes?  and then i close this PR.

ericvergnaud:
Hi, thanks for your contribution. I have found a better solution, you can close this PR.

windwiny:
In my case `const fs = isNode ? await eval("import('fs')") : null;`  not is better solution,

i use `vite` build html and ts files to js,  `await is only valid in async functions and the top level bodies of modules` ,

`<script>` add `type='module'` attr,  it need new version browser, and NOT work on offline html file `file:///xxx.html`

ericvergnaud:
Thus purpose here is NOT for people being able to call this code in a browser but to have the same package for browser and node.
Node 16 supports top level await.

Issue #4095, opened 2023-02-02
Python LexerNoViableAltException should not be created with None mgs


Sofyan-SU:
**Grammar bug** : 
Identifier: [A-Za-z][A-Za-z0-9_]* ;  
expr: Identifier ;

**Trigger** :
- Input to expr is given an illegal character e.g : "id@"
- one has overriden the recover function of the lexer. e.g. :
`class ErrorLexer(bugLexer):
    """
    New Lexer class that overrides the bugLexer and crashes in case of lexer error
    """

    def recover(self, r):
        raise r  # Raise the error for the upper level`

**Analysis**
LexerNoViableAltException is created with message = None
Exception is caught by ErrorStrategy that reportError with a None message
`            print("unknown recognition error type: " + type(e).__name__)
            recognizer.notifyErrorListeners(e.message, e.offendingToken, e)`
Then the ConsoleErrorListener will display the syntaxError as follow:
`print("line " + str(line) + ":" + str(column) + " " + msg, file=sys.stderr)`
This raise an error as msg is None.

**Proposal** is to create the LexerNoViableAltException with a message
e.g. :
`    def __init__(self, lexer:Lexer, input:InputStream, startIndex:int, deadEndConfigs:ATNConfigSet):
        super().__init__(message=None, recognizer=lexer, input=input, ctx=None)
        self.startIndex = startIndex
        self.deadEndConfigs = deadEndConfigs
        self.message = self.__str__()`

ericvergnaud:
I think your proposal would be too big a change. We definitely want to prevent the crash, so I'd simply set self.message to "" if None. If that's ok with you, please submit a PR.

Sofyan-SU:
Issue is corrected

PR #4097, opened 2023-02-03
Python LexerNoViableAltException is created with messages (#4095)

Sofyan-SU:
- Added a message to LexerNoViableAltException (python2 & 3) : this avoids exception to be raised if the exception message is printed
- Corrected the ErrorStrategy (Python2) : 
   -> notifyErrorListeners first argument should be the message, and second the Offending Token 
   -> Exception does not have getMessage method


ericvergnaud:
Thanks, I am fine with the PR, but you haven't signed off your commit, please follow instructions in https://github.com/antlr/antlr4/blob/master/CONTRIBUTING.md

Sofyan-SU:
Commit is now signed. There is one check that fails, however, I do not see any relation with my commit

ericvergnaud:
@parrt blessed

PR #4101, opened 2023-02-08
Fully qualify std::move invocations to fix -Wunqualified-std-cast-call

rupprecht:
Clang implemented a warning last year to detect bare `move` and `forward` calls, as they have the potential to be error prone: https://github.com/llvm/llvm-project/commit/70b1f6de539867353940d3dcb8b25786d5082d63. This warning should be available in clang-15.


rupprecht:
Updated the commit with DCO

parrt:
@hzeller @mike-lischke this seems like a reasonable change as it's just cleaning up right?

hzeller:
LGTM

PR #4109, opened 2023-02-11
[Cpp] CMake: use variables for specifying install paths consistently

jmairboeck:
This allows overriding them if needed when the system uses different paths than usual (e.g. Haiku, which uses "develop/headers" for includes instead of "include").

Signed-off-by: Joachim Mairböck <j.mairboeck@gmail.com>

### DIFF
--- a/runtime/Cpp/CMakeLists.txt
+++ b/runtime/Cpp/CMakeLists.txt
@@ -138,10 +138,11 @@ if(WITH_DEMO)
  add_subdirectory(demo)
 endif(WITH_DEMO)
 
+include(GNUInstallDirs)
+
 # Generate CMake Package Files only if install is active
 if (ANTLR4_INSTALL)
 
-  include(GNUInstallDirs)
   include(CMakePackageConfigHelpers)
 
   if(NOT ANTLR4_CMAKE_DIR)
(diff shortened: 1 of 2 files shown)


runtime/Cpp/CMakeLists.txt
```diff
@@ -222,14 +223,14 @@ endif(ANTLR4_INSTALL)
 
 if(EXISTS LICENSE.txt)
 install(FILES LICENSE.txt
-        DESTINATION "share/doc/libantlr4")
+        DESTINATION ${CMAKE_INSTALL_DOCDIR})
```
Krzmbrzl, line 226: Shouldn't this use an ANTLR-specific sub-directory? :thinking:

jmairboeck: `CMAKE_INSTALL_DOCDIR` contains a project specific sub-directory by default, see https://cmake.org/cmake/help/latest/module/GNUInstallDirs.html#result-variables. It defaults to `DATAROOTDIR/doc/PROJECT_NAME`.

Krzmbrzl: Oh, interesting. Didn't know that.

I wonder whether changing the casing of the project name has other side effects though :thinking:

jmairboeck:
I rebased this on top of the changes from #4661.

hzeller:
@parrt this is sitting here for a while, should it be merged ?

parrt:
I closed my eyes and hit merge :)

PR #4119, opened 2023-02-16
Fix for Issue 4110

kaby76:
These are some changes for Issue 4110, which is for the TypeScript target.
* CommonToken is needed. https://github.com/antlr/grammars-v4/blob/f1b4fc39279ce6a9e3878057fba935642f22c364/python/python3/JavaScript/Python3LexerBase.js#L59
* reset() is needed. https://github.com/kaby76/Domemtech.Trash/blob/2431705bdd3a2f3d64f1de5917ec07cd8f195c48/src/trgen/templates/TypeScript/Test.ts#L189
* skip() is needed. https://github.com/antlr/grammars-v4/blob/f1b4fc39279ce6a9e3878057fba935642f22c364/python/python3/JavaScript/Python3LexerBase.js#L109
* I'm including the export of other methods from Lexer as I know they are needed elsewhere.

### DIFF
--- a/runtime/JavaScript/src/antlr4/CommonToken.d.ts
+++ b/runtime/JavaScript/src/antlr4/CommonToken.d.ts
@@ -1,0 +1,8 @@
+import { Token } from "./Token";
+
+export declare class CommonToken extends Token {
+    constructor(source: number, type: number, channel: number, start: number, stop: number);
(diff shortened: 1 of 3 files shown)


ericvergnaud:
@parrt ready to merge

PR #4121, opened 2023-02-17
cmake: Fix output dir for Visual Studio generator

Leo1690:
Fix output dir for Visual Studio generator and for when ANTLR_BUILD_CPP_TESTS is not set

The antlr4 binaries are copied to `${ANTLR4_ROOT}/runtime/Cpp/dist` only when ANTLR_BUILD_CPP_TESTS is set. However, when linking antlr4_shared to a new target, the target expects to find the `. lib` binary in `${ANTLR4_ROOT}/runtime/Cpp/dist/$(Configuration)` for Visual Studio generator and it expects `${ANTLR4_ROOT}/runtime/Cpp/dist` for other generator like Ninja. In any case, the binaries are always generated in  `${ANTLR4_ROOT}/runtime/Cpp/runtime/$(Configuration)` or `${ANTLR4_ROOT}/runtime/Cpp/runtime` depending on the generator.

We can observe in this [cmake file](https://github.com/antlr/antlr4/blob/master/runtime/Cpp/runtime/CMakeLists.txt#L122) that the binaries are copied to `dist` only when ANTLR_BUILD_CPP_TESTS is set.


parrt:
Was on master...i reset to dev but now it's pulling in weird stuff. maybe recreate?

Leo1690:
> Was on master...i reset to dev but now it's pulling in weird stuff. maybe recreate?

Sorry, I did not realize that. Now may patch is in dev.

PR #4129, opened 2023-02-21
README: Update required Java version

Krzmbrzl:
The badge in the README suggests that ANTLR requires Java 7 or higher, whereas since ANTLR v4.10 Java 11 or higher is required.

NOTE: This is a commit against the `master` branch, but given that the latest versions already have this requirement, I figured it makes more sense to base this against `master` rather than `dev`.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,6 +1,6 @@
 # ANTLR v4
 
-[![Java 7+](https://img.shields.io/badge/java-7+-4c7e9f.svg)](http://java.oracle.com)
+[![Java 11+](https://img.shields.io/badge/java-11+-4c7e9f.svg)](http://java.oracle.com)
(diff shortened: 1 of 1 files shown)


parrt:
Hmm...the runtime only requires 8 i think but fair enough.

parrt:
i copied to dev.

Krzmbrzl:
> Hmm...the runtime only requires 8 i think but fair enough.

I can say with certainty that a JRE 8 is jot sufficient to generate C++ parsers from grammars - is that what you were referring to?

ericvergnaud:
@Krzmbrzl the tool requires v11, the runtime (runs the generated parsers) requires v8

parrt:
I meant a Java runtime. The tool itself requires 11.

PR #4137, opened 2023-02-23
java runtime: add module-info.java and multi-release build to support…

### LINKED ISSUE #2946: ANTLR as a named Java module
Hi there,
it is planned to release ANTLR4 as a named module for Java? Currently the ANTLR runtime does not have a "module-info.class". This causes problems when using it with "jlink". As we have Java 15 now and modules can/should be used since Java 9 it would be great to convert ANTLR into a named module too.

Sure, I could inject a "module-info.java" myself with "jdeps --generate-module-info" and "javac --patch-module", but this does not feel right. And it creates new problems if the jar is provided by Maven...

alexberdnik:
This PR fixes #2946 by adding module-info.java to antlr-runtime module of the project. This is done as "multi-release jar" which allows to specify multiple versions of classes for different java versions. In this case the module is compiled for java 8 with java 9 classes stored in META-INF/versions/9.

Module-info.java exports module named `org.antlr.antlr.runtime`. This differs from the original automatic-module-name (`org.antlr.antlr4.runtime`) because JPMS doesn't like trailing digits in names.

Module exports all packages except `org.antlr.v4.runtime.tree.pattern` and `org.antlr.v4.runtime.tree.xpath`. Please let me know if these are used outside the module and should be included as well.

### DIFF
--- a/runtime/Java/pom.xml
+++ b/runtime/Java/pom.xml
@@ -119,8 +119,7 @@
                             <goal>compile</goal>
                         </goals>
                         <configuration>
-                            <source>1.8</source>
-                            <target>1.8</target>
+                            <release>8</release>
                         </configuration>
                     </execution>
 
--- a/runtime/Java/pom.xml
+++ b/runtime/Java/pom.xml
@@ -101,6 +101,9 @@
 						<manifest>
 							<addDefaultImplementationEntries>true</addDefaultImplementationEntries>
 						</manifest>
+                        <manifestEntries>
+                            <Multi-Release>true</Multi-Release>
+                        </manifestEntries>
 						<manifestFile>${project.build.outputDirectory}/META-INF/MANIFEST.MF</manifestFile>
 					</archive>
 				</configuration>
--- a/runtime/Java/pom.xml
+++ b/runtime/Java/pom.xml
@@ -83,7 +83,6 @@
 						<phase>process-classes</phase>
 						<configuration>
 							<instructions>
-								<Automatic-Module-Name>org.antlr.antlr4.runtime</Automatic-Module-Name>
 								<Bundle-SymbolicName>org.antlr.antlr4-runtime</Bundle-SymbolicName>
 								<Import-Package>org.antlr.v4.gui;resolution:=optional, *</Import-Package>
 							</instructions>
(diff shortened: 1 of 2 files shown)


parrt:
i think we're stuck at jdk 8 for now...

alexberdnik:
That's ok, the runtime will continue to work under java 8. module-info.java will be visible only on java 9+ - see https://openjdk.org/jeps/238

alexberdnik:
> i think we're stuck at jdk 8 for now...

I just noticed that antlr4-maven-plugin requires java 11+, otherwise I'm getting wrong class version error...

I tested locally with these code (sample json parser):
```java
        final JSONLexer lexer = new JSONLexer(CharStreams.fromString(json));
        final JSONParser parser = new JSONParser(new CommonTokenStream(lexer));
        final JSONParser.JsonContext ctx = parser.json();
        System.out.println(ctx.toStringTree());
```
... and the snapshot version of the runtime (one with module-info.java)
The code works perfectly in modular environment (with module path) and also on java 8 jvm. For java 8 build, `release` option of the compiler is important, otherwise we could face phantom issues like [this](https://stackoverflow.com/questions/61267495/exception-in-thread-main-java-lang-nosuchmethoderror-java-nio-bytebuffer-flip) one.


So to sum-up:
- runtime works both on java 8 and java 9 +
- the project itself should be built by java 9+
- (found during testing) latest antlr4-maven-plugin requires java 11 (tested 4.12.0 and 4.11.1, both are not working on java 8).

bowbahdoe:
I think I might have duplicated this one #4383

But in that PR, small difference is that I don't think Java 9 is required to build, so the JDK wouldn't need to be changed anywhere.

alexberdnik:
I'm not sure moditect works without java 9 - you need to compile module-info.java somehow and java 8 compiler just can't do it.
Anyway, you won't be able to compile the project with java < 11, as I mentioned before.
Moditect-based approach looks a bit hacky for me - if the module-info.java can be written directly in the project, why don't just do that? I do understand that this may be the only way if you use a third-party library  and don't have access to the source code.

xazap:
> I'm not sure moditect works without java 9 - you need to compile module-info.java somehow and java 8 compiler just can't do it.

Moditect supports both Java 8 and Java 9+. It does either:
- compile a module-info.java. This requires a Java 9+ compiler. This is what  061196b
 does since it defines the module-info content in `<moduleInfo>` tags in the moditect plugin configuration.
- Compile a module-info.java with Java 8. It will use a byte code generating library to generate class version 53.0  (Java 9+) byte code using Java 8. This would require changing `<moduleInfo>` to `<moduleInfoSource>` in the moditect plugin configuration.

>  Moditect-based approach looks a bit hacky for me - if the module-info.java can be written directly in the project, why don't just do that? 

Agreed, since the build for the antlr4 runtime is currently already is on Java 11 with Java 8 as a target. I would vote to merge this instead of the PR with the moditect plugin. Much simpler.

Issue #4102, opened 2023-02-09
Cpp Target CMake Overhaul


jnhyatt:
In the C++ ANTLR target, the build system is using somewhat outdated CMake throughout. It tends to avoid modern CMake patterns in favor of older, lower-level concepts. On the other hand, it doesn't seem to do anything that modern CMake isn't capable of doing more cleanly. In addition, build files for other environments (Xcode, VS) are also present in the Cpp runtime source. This isn't ideal because CMake is a build generator designed to output build files specific to the platform and target the user chooses. Having environment-specific build files in the source tree is data duplication and must be kept up to date with each build, and in some cases (#4088), they aren't updated with releases.

A short summary of current CMake issues:
- Uses low-level environment-specific compiler flags/directories
- Macros fetch content or otherwise act as package manager
- Doesn't install targets according to [`FindPackage` rules](https://cmake.org/cmake/help/latest/command/find_package.html#config-mode-search-procedure)
- Relies on module mode (`FindANTLR.cmake`) rather than config mode
- Requires relatively new CMake version
- Overrides default configuration settings

I've found several issues in the project that could potentially be resolved with a CMake overhaul. In fact, [in @mike-lischke's response](https://github.com/antlr/antlr4/issues/3134#issuecomment-815666314) to #3134, he mentions that the CMake files could use an overhaul. I propose (and volunteer) to redesign the Cpp target build files with the following goals:

- Port to modern CMake
- Properly configure install rules
- Simplify client-side workflow
- Maximize backwards compatibility

My current branch would change the client-side workflow outlined [here](https://github.com/antlr/antlr4/blob/master/runtime/Cpp/cmake/README.md) to the following:
```cmake
# ANTLR must be built and installed
find_package(ANTLR REQUIRED)

# antlr executable must be set in CMake or an environment variable
antlr_target(MyLexerAndParser src/MyGrammar.g4)

add_executable (MyParser src/main.cpp)

# Depend on ANTLR runtime (static or dynamic) and depend on generated parser/lexer files
target_link_libraries (MyParser
    PUBLIC
        ANTLR_MyLexerAndParser # target created by antlr_target, builds generated files
        antlr4::runtime
)
```

[...]

PR #4138, opened 2023-02-24
Modern CMake

jnhyatt:
Addresses #4102.

This PR is not currently ready to merge. However, I wanted it here for reference and for anyone who'd like to contribute to the work in progress.

Some of the major edits so far:

### Build-specific files have been removed.

CMake is a build generator, so it should be used to generate platform-specific files. This simplifies the runtime and the demo to a single workflow: CMake -> \<your build system here\>. It also consolidates sources and clarifies the build process.

### FindANTLR.cmake has been removed.

FindXXX.cmake is an artifact from an older era of CMake. In modern CMake, such scripts are used as a way of finding packages that do not use CMake as their primary build system and will not or cannot be updated to use it as such. In projects that build primarily with CMake (Antlr included), projects are instead recommended to export targets and then install those targets to the user's system (or other install location determined by the user -- the golden rule is to put the build workflow into the client's hands). I won't go into further detail here about the benefits of the modern approach, but [here](https://pabloariasal.github.io/2018/02/19/its-time-to-do-cmake-right/) is an excellent writeup of some of the differences between legacy and modern CMake, and [here](https://gist.github.com/mbinna/c61dbb39bca0e4fb7d1f73b0d66a4fd1) is a summary of the points presented there.

### README.md and demo have been modified.

The README.md in `runtime/Cpp/cmake` outlines a somewhat complex process to get Antlr working in a new project. With these changes, that workflow is simplified and reflects in the new README. The demo will need to be modified as well, but I've held off so far because it contains XCode-specific files, and I don't currently have access to an OSX development environment. I will put some work toward unifying the build/source files for the demo, but if someone would like to test the changes on OSX, I'd gladly accept the help. Another major change to the README is promoting the use of [CPM](https://github.com/cpm-cmake/CPM.cmake). I'd welcome feedback concerning this addition. CPM has worked very well for me in the past, and it's unintrusive: Antlr's build files do not need to be modified in any way to support it. In addition, its use is commonly recommended as a standard tool for modern C++. Barring the use of CPM, the typical CMake workflow (build -> install -> target_link_libraries) is presented.

It's worth listing here some of the high-level steps that still need to be implemented before this should reasonably be considered for merging:

- Unify and rewrite demo
- Reintroduce install rules
- Reintroduce some configuration (static/dynamic libraries)
- Update tests


runtime/Cpp/runtime/antlr4cpp-vs2022.vcxproj
```diff
@@ -1,652 +0,0 @@
-﻿<?xml version="1.0" encoding="utf-8"?>
```
jmairboeck, line 1: Please don't remove the native project files. There are users of them (including myself).

I am not against modernizing the CMake build but this has nothing to with CMake.

KvanTTT: Sure. Also, Windows tests infrastructure use this file that's why build is failing.

jnhyatt: I can add these back in. I wonder about the intent here, though -- is CMake the primary build system? I would argue CMake should be the default, and that CI should be updated to use it. There are reasons not to keep native build files around, but I understand this would break projects which currently depend on them. Could we potentially add a deprecation of some sort?

jmairboeck: To be fair, CMake can generate project files for Visual Studio, but the CMake generated project files are not "portable" like the existing native ones because they use absolute paths. It would require quite some effort to switch to a CMake based build.

PR #4140, opened 2023-02-24
Limit use of Posix threads to Unix

hs-apotell:
Posix threads are available only on Unix (and unix like) platforms. Wrap the dependency accordingly so builds don't fail on other platforms.


parrt:
hyperbusy but soon.  we have similar PR btw.

jmairboeck:
`Threads` is a built-in meta-target of CMake and should resolve to nothing on platforms which don't require a separate threading library anyway. So this is probably not needed.

PR #4142, opened 2023-02-25
Fix PathElement for Token type

HiroShinke:
symbolicName must be appropriate for Token


ericvergnaud:
Hi,

thanks for this.

Would you mind:
 - adding a unit test
 - provide the same for Python 2 target
 - commit using DCO

Eric

PR #4145, opened 2023-02-27
[Typescript] Add missing methods typing declaration

hieunguyen2211:
Add missing typing declaration methods and properties that I need to use in my project.

### DIFF
--- a/runtime/JavaScript/src/antlr4/Lexer.d.ts
+++ b/runtime/JavaScript/src/antlr4/Lexer.d.ts
@@ -6,6 +6,9 @@ import {Token} from "./Token";
 export declare class Lexer extends Recognizer<number> {
 
     static DEFAULT_MODE: number;
+    static literalNames: (string | null)[];
+    static symbolicNames: (string | null)[];
+    static ruleNames: string[];
 
     _input: CharStream;
     _interp: LexerATNSimulator;
@@ -28,4 +31,5 @@ export declare class Lexer extends Recognizer<number> {
     emitToken(token: Token): void;
     emit(): Token;
     emitEOF(): Token;
+    getAllTokens(): Token[];
 }
--- a/runtime/JavaScript/src/antlr4/Lexer.d.ts
+++ b/runtime/JavaScript/src/antlr4/Lexer.d.ts
@@ -6,9 +6,6 @@ import {Token} from "./Token";
 export declare class Lexer extends Recognizer<number> {
 
     static DEFAULT_MODE: number;
-    static literalNames: (string | null)[];
-    static symbolicNames: (string | null)[];
-    static ruleNames: string[];
 
     _input: CharStream;
     _interp: LexerATNSimulator;
--- a/runtime/JavaScript/src/antlr4/context/ParserRuleContext.d.ts
+++ b/runtime/JavaScript/src/antlr4/context/ParserRuleContext.d.ts
@@ -11,7 +11,6 @@ export declare class ParserRuleContext extends RuleContext {
     parentCtx: ParserRuleContext | undefined;
     exception?: RecognitionException;
     parser?: Parser;
-    ruleIndex: number;
 
     constructor(parent?: ParserRuleContext, invokingStateNumber?: number);
     copyFrom(ctx: ParserRuleContext): void;
--- a/runtime/JavaScript/src/antlr4/context/ParserRuleContext.d.ts
+++ b/runtime/JavaScript/src/antlr4/context/ParserRuleContext.d.ts
@@ -11,6 +11,7 @@ export declare class ParserRuleContext extends RuleContext {
     parentCtx: ParserRuleContext | undefined;
     exception?: RecognitionException;
     parser?: Parser;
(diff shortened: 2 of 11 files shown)


runtime/JavaScript/src/antlr4/context/ParserRuleContext.d.ts
```diff
@@ -11,6 +11,7 @@ export declare class ParserRuleContext extends RuleContext {
     parentCtx: ParserRuleContext | undefined;
     exception?: RecognitionException;
     parser?: Parser;
+    ruleIndex: number;
```
ericvergnaud, line 14: this breaks the build, and there is no ruleIndex defined in ParserRuleContext.js

runtime/JavaScript/src/antlr4/error/DefaultErrorStrategy.d.ts
```diff
@@ -16,4 +16,11 @@ export declare class DefaultErrorStrategy implements ErrorStrategy {
 
     sync(recognizer: Parser): void;
 
+    constructor();
```
ericvergnaud, line 19: an empty constructor is not needed

runtime/JavaScript/src/antlr4/error/RecognitionException.d.ts
```diff
@@ -13,6 +14,6 @@ export interface ExceptionParams {
 
 export declare class RecognitionException extends Error {
     ctx: RuleContext;
-
+    offendingToken: Token;
```
ericvergnaud, line 17: should be Token | null

runtime/JavaScript/src/antlr4/state/ATNState.d.ts
```diff
@@ -1,6 +1,8 @@
 import {ATN} from "../atn";
+import {Transition} from "../transition";
 
 export declare class ATNState {
     atn: ATN;
     stateNumber: number;
+    transitions: Transition[];
```
ericvergnaud, line 7: can you explain why you need this ?

runtime/JavaScript/src/antlr4/context/ParserRuleContext.d.ts
hieunguyen2211: In DefaultErrorStrategy.js, in reportFailedPredicate method, at line 295, I see this code:
`
const ruleName = recognizer.ruleNames[recognizer._ctx.ruleIndex]
`
As far as my understanding, recognizer in this case is a Parser instance, _ctx is a ParserRuleContext. Therefore, ruleIndex should be available in ParserRuleContext.

ericvergnaud: That works in js because in the above context, the _ctx is always a subclass of ParserRuleContext for which ruleIndex is defined. But there are subclasses for which it is not defined. See the ParserRuleContext code, there is no ruleIndex. Exposing one would be misleading, and it breaks the build.

ericvergnaud: If you need to write typescript code that mimics DefaultErrorStrategy, you can create an IndexedRuleContext interface with a ruleContext field, and cast the _ctx.

runtime/JavaScript/src/antlr4/Lexer.d.ts
```diff
@@ -6,6 +6,9 @@ import {Token} from "./Token";
 export declare class Lexer extends Recognizer<number> {
 
     static DEFAULT_MODE: number;
+    static literalNames: (string | null)[];
```
ericvergnaud, line 9: I'm not sure this makes any sense... these are defined with the generated XXXLexer, they don't exist in the base class.

runtime/JavaScript/src/antlr4/error/DefaultErrorStrategy.d.ts
hieunguyen2211: Removed in my last commit.

runtime/JavaScript/src/antlr4/state/ATNState.d.ts
hieunguyen2211: Removed in my last commit.

runtime/JavaScript/src/antlr4/Lexer.d.ts
hieunguyen2211: Removed in my last commit.

nghiaviettran:
[ericvergnaud](https://github.com/ericvergnaud) Are there any open points on this PR? We need this changes for our project release. Many thanks!

ericvergnaud:
> [ericvergnaud](https://github.com/ericvergnaud) Are there any open points on this PR? We need this changes for our project release. Many thanks!
We're missing a DCO.

hieunguyen2211:
>

hieunguyen2211:
> 

I updated, please check my PR again.

ericvergnaud:
@parrt blessed

PR #4147, opened 2023-02-28
[TypeScript] Fix types in template (missing null-union)

Sharknoon:
Hello,

the arrays `literalNames` and `symbolicNames` do not only contain strings, but also null. Therefore we should reflect that in the TypeScript type to prevent build errors in strict mode.

image
image

Thank you for your great work on the TypeScript target, loving it :)

Kind regards
Sharknoon

### DIFF
--- a/tool/resources/org/antlr/v4/tool/templates/codegen/TypeScript/TypeScript.stg
+++ b/tool/resources/org/antlr/v4/tool/templates/codegen/TypeScript/TypeScript.stg
@@ -873,8 +873,8 @@ export default class <lexer.name> extends <superClass; null="Lexer"> {
 	<rest(lexer.modes):{m | public static readonly <m> = <i>;}; separator="\n">
 
 	public static readonly channelNames: string[] = [ "DEFAULT_TOKEN_CHANNEL", "HIDDEN"<if (lexer.channelNames)>, <lexer.channelNames:{c| "<c>"}; separator=", ", wrap, anchor><endif> ];
-	public static readonly literalNames: string[] = [ <lexer.literalNames:{t | <t>}; null="null", separator=", ", wrap, anchor> ];
-	public static readonly symbolicNames: string[] = [ <lexer.symbolicNames:{t | <t>}; null="null", separator=", ", wrap, anchor> ];
+	public static readonly literalNames: (string | null)[] = [ <lexer.literalNames:{t | <t>}; null="null", separator=", ", wrap, anchor> ];
+	public static readonly symbolicNames: (string | null)[] = [ <lexer.symbolicNames:{t | <t>}; null="null", separator=", ", wrap, anchor> ];
 	public static readonly modeNames: string[] = [ <lexer.modes:{m| "<m>",}; separator=" ", wrap, anchor> ];
 
 	public static readonly ruleNames: string[] = [


ericvergnaud:
@parrt blessed

matthew-dean:
A lot of good TypeScript and JavaScript runtime improvements for 4.12.1. Might there be a release soon?

Sharknoon:
I hope so, waiting for a release too :)

Codex-:
Also encountering this issue causing type checking to fail, hoping for 4.12.1 🙏

parrt:
@jimidle @ericvergnaud are at a good spot for release?

jimidle:
I have lots of stuff I am working on, on and off for performance and
memory, but it’s experimental really and there is plenty in go for a new
release.

I will work on documentation today and see if there is anything I want to
send to dev. I’ll send  GitHub actions for the new go release repo too.
Then I am ready when you are.

jimidle:
Working on it right now and will have my go stuff ready by end of day
Taipei time.

You release mechanism will need to copy the main branch of:
github.com/antlr/antlr4/runtime/Go/antlr/v4

to the main branch of

github.com/antlr-go/antlr

So, the top level of the repo.

Then the v4.x.y. tag should be issued against it. THe main development
source can be now be tagged exactly as you need it to be.

The go.mod for the runtime takes care of this being v4 and now the go get
command will resolve tags and so on correctly.

You may bring in the go runtime from the main branch of either repo of
course, depending how screwed up is Google's monorepo and the fact that
they don't use modules (they must jump through a lot of hoops to get that
to work, as modules are more or less required these days).

Jim

ericvergnaud:
@parrt all green here, I just submitted 2 minor PRs for you to merge. Nothing else pending on my side.

parrt:
> You release mechanism will need to copy the main branch of:

ok, is this something we can document in the release .md file?

jimidle:
Ter,

The release of v4.13.0 is now live in the new release only repo and works
like a bought one.

The main .md file does already draw attention to the new repo, but of
course feel free to add to that if you don't think it is prominent enough.

My own projects are now using the new repo and pass all their tests. More
importantly, the tags work as they should and our users will now see real
tag numbers in their go.mod files:


module github.com/jimidle/mvparse

go 1.19

require (
   github.com/antlr4-go/antlr/v4 v4.13.0
   github.com/dominikbraun/graph v0.21.0
   github.com/pyroscope-io/client v0.7.0
)


require (
   github.com/pyroscope-io/godeltaprof v0.1.0 // indirect
   golang.org/x/exp v0.0.0-20230515195305-f3d0a9c9a5cc // indirect
)

So, when you make the release this weekend, there is nothing extra for you
to do over what you already do.

The go release notes tell you what to do to upgrade the go code within
Google to use the new repo - it is pretty trivial. Please look out for the
one change where you might need more than the import change. It is subtle
but trivial in that you may need to change the parser and lexer variable
definitions as they don't return an interface any more.

Most people won't even notice, but if they declared the type instead of
just using auto typing, then you will need to change the declaration to be
a * pointer. There are new interfaces into the runtime, but the runtime
interfaces are exactly the same. I doubt that you will even see this, but
regen the grammars and compile. If you get an error, then it is just this:

If your code looks like this:

var lexer = parser.NewMySqlLexer(nil)
var p = parser.NewMySqlParser(nil)

Or this:

lexer := parser.NewMySqlLexer(nil)
p := parser.NewMySqlParser(nil)

Then no changes need to be made. However, if you predeclared the parser and
lexer variables with their type, such as like this:

var lexer parser.MySqlLexer
var p parser.MySqlParser
// ...
lexer = parser.NewMySqlLexer(nil)
p = parser.NewMySqlParser(nil)
You will need to change your variable declarations to pointers (note the
introduction of the * below.

var lexer *parser.MySqlLexer
var p *parser.MySqlParser
// ...
lexer = parser.NewMySqlLexer(nil)
p = parser.NewMySqlParser(nil)


And that's it.

I will look out for your release announcement. This should be a nice
release for the go folk!

Jim

PS: I have not seen a lick of work for nearly 6 months now - the market is
broken. If Google wants a contractor or you hear of anyone else needing
anything, please bear me in mind!

matthew-dean:
@jimidle 

> The release of v4.13.0 is now live in the new release only repo

Is there a way to pull this into a project? Can that be documented?

jimidle:
It is documented.

https://github.com/antlr/antlr4/blob/master/doc/go-target.md

jimidle:
Also, of course:

https://pkg.go.dev/github.com/antlr4-go/antlr/v4

Issue #3687, opened 2022-04-29
Optimize `Contains` call for all runimes (O(N) -> O(log(N))


KvanTTT:
It's actual for rules with multiple single-token alternatives. It's quite common case, for instance:

```g4
id
    : ID
    | Abstract
    | Array
    | As
    | BinaryCast
    | BoolType
    | BooleanConstant
    | Break
    | Callable
    | Case
    | Catch
    ...
```

We have such rules for [PHP](https://github.com/antlr/grammars-v4/blob/master/php/PhpParser.g4#L745), [PlSql](https://github.com/antlr/grammars-v4/blob/master/sql/plsql/PlSqlParser.g4#L4621), [C#](https://github.com/antlr/grammars-v4/blob/master/csharp/CSharpLexer.g4#L393) and other grammars.

It's optimized for [Java](https://github.com/antlr/antlr4/blob/dev/runtime/Java/src/org/antlr/v4/runtime/misc/IntervalSet.java#L385-L405), C++, PHP but not optimized for some other runtimes: [C#](https://github.com/antlr/antlr4/blob/4b649103f31e0c0bc3bd0bb964efd0dea6b9f5a0/runtime/CSharp/src/Misc/IntervalSet.cs#L476-L496), [Go](https://github.com/antlr/antlr4/blob/88c2159aab9e5bf281b4680b62060b3b2c0dd23c/runtime/Go/antlr/interval_set.go#L129-L139), [JavaScript](https://github.com/antlr/antlr4/blob/d26cb63ea13649b37d37389fb0404d505a9d378c/runtime/JavaScript/src/antlr4/misc/IntervalSet.js#L92-L103), [Python](https://github.com/antlr/antlr4/blob/e1e5969e3154c456fcd85edf64114238681181f4/runtime/Python3/src/antlr4/IntervalSet.py#L89-L93), [Swift](https://github.com/antlr/antlr4/blob/4e2bb5a0ccda044e50b15c3a176226dcea49296e/runtime/Swift/Sources/Antlr4/misc/IntervalSet.swift#L402-L414).

Also, for single-value intervals complexity can be improve up to O(1) if use `HashMap` but it's efficient only for quite big collections (>= 20 elements).

vityaman:
The question is: does it really speed up the code?

I have a hypothesis that O(n) may work better with small sizes of IntervalSet::intervals than O(logn) due to constants, branch prediction and prefetching.

[...]

PR #4148, opened 2023-03-01
[optimization] runtime/java/IntervalSet::remove(int el)

vityaman:
Optimized IntervalSet::remove(int el) method to use binary search instead of linear scan to find needed interval where element is located.

But it may work slowly because of binary search issue with branch mispredictions that are noticeable on small array sizes:
https://stackoverflow.com/questions/57482831/linear-search-vs-binary-search-efficiency

Are there some statistics of IntervalSet::intervals array size? It is possible that linear scan on IntervalSet::intervals have a better performance. I think that in case there are not so many elements in this array, methods like 'contains', 'add' and 'remove' can perform better with linear scan implementation.


KvanTTT:
Relevant issue: https://github.com/antlr/antlr4/issues/3687

PR #4154, opened 2023-03-05
fix: Some code that should have been destined for /v4 was instead put…

jimidle:
fix: Code that should have gone to the /v4 files was instead placed in to the root (pre v4) files

This PR merely copies the changes in to the correct place - code has already been reviewed. 

 - Incorporate an iterative tree walker
 - Expose Reset() and String() in lexer/token and add test

@parrt - please merge so I can then proceed with removing the pre-v4 files from the repo


parrt:
thanks!

PR #4155, opened 2023-03-05
Remove redundant source code from go runtime

jimidle:
Removes the non /v4 source code from the runtime and marks the original go.mod file as redundant


jimidle:
@parrt This is the cleanup PR that will leave us with only one v4 source code for the go runtime.

parrt:
Thanks, Jim!

PR #4162, opened 2023-03-09
Avoid usage of expensive FlexibleHashMap

dreis2211:
Hi 👋 

I was profiling `checkstyle` the other day and noticed that the usage of `PredictionMode.getConflictingAltSubsets` is responsible for almost ~13% of all CPU frames inside my async-profiler flamegraphs. 
image

When zooming in this can be largely traced back to the usage of `FlexibleHashMap`
image

From the comments I couldn't find the exact reason for this Map, but it seems to be used in conjunction with `...EqualityComparator` to customize hash-codes.

Imho, the same can be achieved by using a normal `HashMap` and wrapping the key as proposed in this PR. By doing so and building checkstyle locally, I get the following improvements.

**Old**
```
       14,41 real        24,28 user         1,27 sys
```
**New**
```
       12,68 real        20,99 user         1,05 sys
```
And the flamegraphs look also cleaner
image

Let me know if you think this is a viable solution.

Cheers,
Christoph

### DIFF
--- a/runtime/Java/src/org/antlr/v4/runtime/misc/FlexibleHashMap.java
+++ b/runtime/Java/src/org/antlr/v4/runtime/misc/FlexibleHashMap.java
@@ -15,7 +15,9 @@ import java.util.Set;
 
 /** A limited map (many unsupported operations) that lets me use
  *  varying hashCode/equals.
+ * @deprecated in favor of using standard maps with custom keys.
  */
+@Deprecated
 public class FlexibleHashMap<K,V> implements Map<K, V> {
 	public static final int INITAL_CAPACITY = 16; // must be power of 2
 	public static final int INITAL_BUCKET_CAPACITY = 8;
--- a/runtime/Java/src/org/antlr/v4/runtime/atn/PredictionMode.java
+++ b/runtime/Java/src/org/antlr/v4/runtime/atn/PredictionMode.java
@@ -6,8 +6,6 @@
 
 package org.antlr.v4.runtime.atn;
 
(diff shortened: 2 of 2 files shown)


runtime/Java/src/org/antlr/v4/runtime/misc/FlexibleHashMap.java
```diff
@@ -15,7 +15,9 @@
 
 /** A limited map (many unsupported operations) that lets me use
  *  varying hashCode/equals.
+ * @deprecated in favor of using standard maps with custom keys.
  */
+@Deprecated
 public class FlexibleHashMap<K,V> implements Map<K, V> {
```
dreis2211, line 21: Since this is public API I didn't want to remove it right away. But it's basically unused inside the project now. Let me know if I should completely remove it already.

parrt:
hi! Sorry but we can't possibly change that critical algorithm at this point. I don't have time to analyze your change, but this is part of the secret sauce where things have to hash to the same set in a very special way.  Differences from this approach in the other targets have consistently been misunderstandings and bugs. Thank you for the contribution, but I will have to close.

ericvergnaud:
@parrt this PR is not changing the algorithm, it's the same hash coding. Rather it's using a different strategy for providing custom hashCode/equals. I think it's worth considering.

parrt:
Ah. Reopening. I didn't have time to look in depth... So I will leave for a different day

dreis2211:
Any chance this could be re-evaluated?

parrt:
Unfortunately, I just don't have the time to evaluate this. sorry!

PR #4176, opened 2023-03-13
Feature/fixembedding

jimidle:
feat: Convert major structs to use embedded parent structs rather than allocating pointers

Many of the structs representing things like transitions, contexts, atn etc emulated Java inheritance by using pointers to instances of parent structs within their definitions. This doubles, triples and even quadruples the number of allocations needed to be made to build the recognizer structs and the prediction contexts.

This PR replaces the pointers with true embedding in the structs that are the hotspots within the runtime.

Seems like a big PR but in fact all the changes are doing the same thing.


jimidle:
@parrt - Builds are still failing in the cpp artifact upload stage. The go builds are all good, so I think that this PR is good to go. It is a major improvement on memory allocations, with quite a bit more to come.

PR #4180, opened 2023-03-15
Fix #3967 Merge of Prediction Context arrays was broken

### LINKED ISSUE #3967: Go mergeArrays output differs from java
[test2.zip](https://github.com/antlr/antlr4/files/10047566/test2.zip)

jimidle:
fix: Fixes merge arrays to perform an actual comparison rather than comparing two pointers!
Fixes: #3967
Closes #3967

 - A small error was made whereby the comparison with a newly created merge `M` of two prediction contexts
  `a` and `b`, was comparing the pointers of `a` and `M` and not using the internal `Equals()` method.
  The ATN output trace is now equal.

 - Also adds a new `testrig` internal project (so it won't be used by `go get`) that allows quick setup
   and testing of ATN tracing for a test grammar and test input.

 - Excludes go performance profiles from git

 - Corrects a small error in one of the ATN tracing scripts.


jimidle:
@parrt Usual CPP tests failing - who can we get to fix this?

jimidle:
@parrt I am backed up on this PR, if you have a chance to merge it? Cheers

parrt:
No idea how to fix those C++ build errors. some github actions crap. :(  @ericvergnaud have any ideas?

ericvergnaud:
@hs-apotell is looking into it

jimidle:
I was wondering what we are actually doing in that action. Is there an
actual need to upload the C++ output artifacts? As the Tommy Cooper joke
goes:

Doctor, doctor, I’ve broken my arm in three places.
Don’t go to those three places again!

Maybe we just stop uploading and let everyone build the target runtimes
theirselves? I realize that this comment should be on the thread about this
by @hs-apotell.

jimidle:
Hmm. I hear what you’re saying, but if users are using the C++ runtime,
they will surely have the ability to build it? Do we know if anyone is
actually using these artifacts?

jimidle:
Ok that’s cool. If it isn’t stopping the build anymore, then I guess it
doesn’t matter, but I am skeptical that storing the build outputs is
useful. If you don’t have a Mac, then I can’t see the value of being able
to download it. But seeing that it builds and tests correctly is necessary
of course.

I don’t have a windows machine anymore so I don’t have any use for
downloading the C++ library built for windows. Maybe I am missing something
here?

There’s no artifact for Go, as the source code is what the user downloads.

But as I said, if it doesn’t stop the build, then no harm no foul.

PR #4181, opened 2023-03-15
Improve runtime performance of the Go runtime test suite

jimidle:
feat: Speed up the go runtime tests

Prior to this change, after writing the go.mod file for each test, we had to run a `go mod tidy` command. This is an extra process sparked for each test, which can be quite the overhead.

This change runs a `go mod tidy` only if we have not yet cached the results (`go.mod` and `go.sum`) for any prior test and therefore saves sparking a process for each test.

@parrt When you get a chance, could you change the Github test runtime environment to use go 1.20.1 ? I would like to keep our development and testing versions of the go compiler up to date, even if I decide not to require go 1.20 at this time. Cheers.

### DIFF
--- a/runtime-testsuite/resources/junit-platform.properties
+++ b/runtime-testsuite/resources/junit-platform.properties
@@ -1,3 +1,4 @@
 junit.jupiter.execution.parallel.enabled = true
 junit.jupiter.execution.parallel.mode.default = concurrent
 junit.jupiter.execution.parallel.mode.classes.default = concurrent
+junit.jupiter.execution.parallel.config.fixed.parallelism=4
--- a/runtime-testsuite/resources/junit-platform.properties
+++ b/runtime-testsuite/resources/junit-platform.properties
@@ -1,4 +1,3 @@
 junit.jupiter.execution.parallel.enabled = true
 junit.jupiter.execution.parallel.mode.default = concurrent
 junit.jupiter.execution.parallel.mode.classes.default = concurrent
-junit.jupiter.execution.parallel.config.fixed.parallelism=4
--- a/runtime-testsuite/test/org/antlr/v4/test/runtime/RuntimeRunner.java
+++ b/runtime-testsuite/test/org/antlr/v4/test/runtime/RuntimeRunner.java
@@ -26,25 +26,45 @@ public abstract class RuntimeRunner implements AutoCloseable {
 
 	public abstract String getLanguage();
 
-	protected String getExtension() { return getLanguage().toLowerCase(); }
+	protected String getExtension() {
+		return getLanguage().toLowerCase();
+	}
 
-	protected String getTitleName() { return getLanguage(); }
+	protected String getTitleName() {
+		return getLanguage();
+	}
 
-	protected String getTestFileName() { return "Test"; }
+	protected String getTestFileName() {
+		return "Test";
+	}
 
-	protected String getLexerSuffix() { return "Lexer"; }
+	protected String getLexerSuffix() {
+		return "Lexer";
+	}
 
-	protected String getParserSuffix() { return "Parser"; }
+	protected String getParserSuffix() {
+		return "Parser";
+	}
 
-	protected String getBaseListenerSuffix() { return "BaseListener"; }
+	protected String getBaseListenerSuffix() {
+		return "BaseListener";
+	}
 
-	protected String getListenerSuffix() { return "Listener"; }
+	protected String getListenerSuffix() {
+		return "Listener";
+	}
 
-	protected String getBaseVisitorSuffix() { return "BaseVisitor"; }
+	protected String getBaseVisitorSuffix() {
+		return "BaseVisitor";
+	}
 
-	protected String getVisitorSuffix() { return "Visitor"; }
+	protected String getVisitorSuffix() {
+		return "Visitor";
+	}
 
-	protected String grammarNameToFileName(String grammarName) { return grammarName; }
+	protected String grammarNameToFileName(String grammarName) {
+		return grammarName;
+	}
 
 	private static String runtimeToolPath;
 	private static String compilerPath;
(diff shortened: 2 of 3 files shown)


runtime-testsuite/test/org/antlr/v4/test/runtime/go/GoRunner.java
```diff
@@ -111,36 +114,61 @@ protected String grammarParseRuleToRecognizerName(String startRuleName) {
 		return rn;
 	}
 
+	@Override
+	protected  List<String> getTargetToolOptions() {
+		ArrayList<String> options = new ArrayList<String>();
+		options.add("-o");
+		options.add(tempTestDir.resolve("parser").toString());
+		return options;
```
KvanTTT, line 122: It can be calculated once and used for every test since `tempTestDir` is calculated once.

runtime-testsuite/resources/junit-platform.properties
```diff
@@ -1,3 +1,4 @@
 junit.jupiter.execution.parallel.enabled = true
 junit.jupiter.execution.parallel.mode.default = concurrent
-junit.jupiter.execution.parallel.mode.classes.default = concurrent
\ No newline at end of file
+junit.jupiter.execution.parallel.mode.classes.default = concurrent
+junit.jupiter.execution.parallel.config.fixed.parallelism=4
```
KvanTTT, line 4: Setting up a manual value may decrease runtime tests performance.

Also, I don't think `4` is suitable value, modern processors have many thread. I suggest using `8` or at least `6`.

KvanTTT (changes requested):

jimidle:
Ok. I’ll take care of your requests tomorrow:)

runtime-testsuite/test/org/antlr/v4/test/runtime/go/GoRunner.java
```diff
@@ -111,36 +114,61 @@ protected String grammarParseRuleToRecognizerName(String startRuleName) {
...
+//			try {
+//				Path originalFile = Paths.get(tempDirPath, generatedFile.name);
+//				Files.move(originalFile, Paths.get(tempDirPath, "parser", generatedFile.name));
+//			} catch (IOException e) {
+//				return new CompiledState(generatedState, e);
+//			}
+//		}
+
+		// We have already created a suitable go.mod file, though it may need to have go mod tidy run on it one time
+		//
```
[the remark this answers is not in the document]
jimidle: Yup - just an oversight - forgot to delete it

runtime-testsuite/test/org/antlr/v4/test/runtime/RuntimeRunner.java
```diff
@@ -162,6 +169,14 @@ public State run(RunOptions runOptions) {
 		if (runOptions.superClass != null && runOptions.superClass.length() > 0) {
 			options.add("-DsuperClass=" + runOptions.superClass);
 		}
+
+		// See if the target wants to add tool options
+		//
+		List<String>targetOpts = getTargetToolOptions();
```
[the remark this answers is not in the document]
jimidle: Changed the way I do this for caching as per your other comment. See next commit.

runtime-testsuite/resources/junit-platform.properties
jimidle: This was an oversight. This was not supposed to be pushed, and anyway, it seems to have no effect. I need to find out how to stop the go tests from trying to run every tests at the same time as this takes longer than limiting the threads to the number of cores (or thereabouts). 

@parrt Do you remember how to do this?

runtime-testsuite/test/org/antlr/v4/test/runtime/go/GoRunner.java
jimidle: I have cached it now.

runtime-testsuite/test/org/antlr/v4/test/runtime/RuntimeRunner.java
```diff
@@ -154,6 +154,13 @@ public static String getRuntimePath(String language) {
 		return runtimePath.toString() + FileSeparator + language;
 	}
 
+	// Allows any target to add additional options for the antlr tool such as the location of the output files
+	// which is useful for the Go target for instance to avoid having to move them before running the test
+	//
+	protected List<String> getTargetToolOptions() {
```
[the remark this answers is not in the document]
jimidle: I cannot agree as while the Go target only needs the output directory at the moment. That could change. It's a pretty trivial change to get the options, so I think let's not sweat it too much ;)

runtime-testsuite/test/org/antlr/v4/test/runtime/RuntimeRunner.java
```diff
@@ -26,25 +26,45 @@ public abstract class RuntimeRunner implements AutoCloseable {
 
 	public abstract String getLanguage();
 
-	protected String getExtension() { return getLanguage().toLowerCase(); }
+	protected String getExtension() {
+		return getLanguage().toLowerCase();
+	}
```
KvanTTT, line 31: Please don't change formatting of the existing code without big need. Especially if the request's topic doesn't relate to this change. Here and below.

KvanTTT (changes requested):

runtime-testsuite/test/org/antlr/v4/test/runtime/RuntimeRunner.java
jimidle: That's just IntelliJ doing it - I am not in to bikeshedding about code formats ;)

jimidle: IntelliJ just did it without me realizing. Formatting is reversed.

KvanTTT: I suspect it can be disabled in the options.

jimidle:
Yes. It can. But I happen to agree with it. If only we knew people that
worked at Jetbrains! ;)


Are we good with the PR now?

jimidle:
I’ll cycle back to this. I need to be able to change it just for go and
allow other target authors to decide what works for them.

parrt:
I'll wait for your signal Jim.

jimidle:
This is good to go Ter. I took a day to improve the Go performance, and
I’ve spent 4 days bikeshedding this PR ;). Let’s press the merge button oh
dictator supreme. ;)

parrt:
I hear and obey!

runtime-testsuite/resources/junit-platform.properties
parrt: @jimidle I can't remember what I had for lunch so.... haha

jimidle:
Fair enough ;). I’ll work it out. I think Ivan was on the right track.

Issue #3934, opened 2022-10-18
Go: exponentially bad/absymal performance as of 


kirides:
- target information: Go `go get -u github.com/antlr/antlr4/runtime/Go/antlr/v4@d5dd253` (same with plain `/v4`)
 - smallest possible grammar and code that reproduces the behavior

 ## Expected

Parsing 19 files should take around 0.2s

## Actual

Parsing 19 files takes more than 11s, and exponentially more the more files there are

Happens as of https://github.com/antlr/antlr4/commit/f19dc0ea3ede67483b6aec6128d5dd44053778f9 (~5.5s)
Got worse with 56df72ca6742 (~10s)

See example in attached zip: 
[issue.zip](https://github.com/antlr/antlr4/files/9809815/issue.zip)

To test old and new behaviour, just toggle the commented lines in `go.mod` and run
`go mod tidy && go test -timeout 30s -run ^TestParsing$ issue -count=1`

kaby76:
The grammar has one big ambiguity, but it's probably not the problem. `statement` can derive `funcCall` and `expression`, but `expression` can derive `value`, which can also derive `funcCall`. It reads the entire statement just to decide which one to pick of `funcCall` vs `expression`.

kirides:
removing `funcCall` from Statement and `nameNode` from expression didn't really do anything

kaby76:
I added code to separate the parse from the tree walker, since this has nothing to do with tree walking, and added code to time the parse call. I then ran the code with all 19 files in data/ vs only one file in data/, data/EngineAdr_G1.d, and compared the time performing the one parse versus the parse in an aggregate of 19 different parses. In the 19-file parse run, EngineAdr_G1.d ran in 3.500 s. When run alone, it ran in 0.564 s. That's a pretty significant difference. If anything, previous data from a parse should improve the performance of the parse. This says for Go, it's actually hurting. I haven't checked internally the actual closure computations side-by-side between Go and CSharp or Java, but that should be checked.

kaby76:
[...]

PR #4187, opened 2023-03-16
Fix the prediction cache and the comparison algorithms from the legacy code

jimidle:
Fixes: #3934
Closes: #3934 

While this is not the end of performance improvements and probably not the end of bug fixes for some of the structs that we are collecting, this PR massively (orders of magnitude) improves the performance characteristics of the Go runtime and also fixes the ATN prediction context cache, that just wasn't working.

This PR disables one test that is also disabled for many other targets. I will fix that in a follow-up PR.

This is ready to merge, with my other PRs (except my testing improvements - I must fix some minor issues for Ivan.


jimidle:
@parrt I could also do with this PR merging. I will have quite a few coming fast and furious new. Cheers.

PR #4192, opened 2023-03-18
Don't Panic! 50% plus performance improvement

jimidle:
feat: Another 50%+ performance improvement

Prior to this change, the runtime and generated code was using panic() and recover()
to perform error checking and reporting. This is extremely expensive and just not the
way to do it. 

This change now uses goto, and explicit checking for error state after selected calls
into the runtime. This has greatly improved parser performance. Using the test code
provided by a recent performance issue report, the parse is now twice as fast as the
issue raised was hoping for. 

This change may require code changes by users if they have added Go code in to their
parsers and have declared variables at random points in actions. They merely have to
declare the variables at the start of the rule. This is a minor price to pay for the massive
performance improvement.

@parrt If you can merge this when you see it, I have many more performance improvements
to send your way.

@kaby76 If you have time to run any of your performance tests, I would appreciate your feedback sir.

### DIFF
--- a/runtime-testsuite/resources/org/antlr/v4/test/runtime/descriptors/CompositeParsers/ImportedRuleWithAction.txt
+++ b/runtime-testsuite/resources/org/antlr/v4/test/runtime/descriptors/CompositeParsers/ImportedRuleWithAction.txt
@@ -18,3 +18,6 @@ s
 [input]
 b
 
+[skip]
+Go
+
--- a/runtime-testsuite/resources/org/antlr/v4/test/runtime/templates/Go.test.stg
+++ b/runtime-testsuite/resources/org/antlr/v4/test/runtime/templates/Go.test.stg
@@ -19,13 +19,9 @@ AppendStr(a,b) ::= "<a> + <b>"
 Concat(a,b) ::= "<a><b>"
 
 AssertIsList(v) ::= <<
-// A noddy range over the list will not compile if it is not getting a slice
-// however, Go will not compile the generated code if the slice vs single value is wrong.
-// Makes the Java based tests suite work though.
-j1__ := make([]interface{}, len(<v>))
-j2__ := <v>
-for j3__ := range j2__ {
-   j1__[j3__] = j2__[j3__]
+// Go will not compile this generated code if the slice vs single value is wrong.
+for i := range localctx.(*ExpressionContext).GetArgs() {
+    _ = localctx.(*ExpressionContext).GetArgs()[i]
 }
 >>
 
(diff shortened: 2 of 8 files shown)


parrt:
Seems OK to me. It looks like this would only improve performance when there's recognition Errors however right?

jimidle:
No mate. The generated code was using a deferred function, well, two
actually. Try catch is free in c++ for instance, but the deferred funcs in
go cost on their own, and performance of recover() has, rightly got more
expensive in deference to defer().  I basically reset to 17 years ago when
I used goto with C :)

This is >50% for all grammars. I was going to test against the Java target
today, as I now think go is at least on a par. Idly, perhaps on a Parr. But
I think I’ll let Ken have a mosey and look at his results.

kaby76:
I'll be testing this today. K

kaby76:
I guess the tool has changed:
```
$ make
bash build.sh
go: added github.com/antlr/antlr4/runtime/Go/antlr/v4 v4.0.0-20230318180256-adbe946416f5
go: added golang.org/x/exp v0.0.0-20220722155223-a9213eeb770e
# example.com/myparser/parser
parser\java_parser.go:1495:67: not enough arguments in call to p.GetInterpreter().AdaptivePredict
        have (antlr.TokenStream, number, antlr.ParserRuleContext)
        want (*antlr.BaseParser, antlr.TokenStream, int, antlr.ParserRuleContext)
```
This is going to take a little time to work out. The scripts not only need to use a specific repo/branch, but it will also need to generate a tool specific branch of Antlr because the template was updated. Currently, the perf scripts work against what is published, which limits things a bit.

parrt:
yay! back to the 80s with goto. love it.

jimidle:
Yes - you need to rebuild the tool as the codegen has changed quite a bit.
It will probably change more before the next release

kaby76:
I've been adjust my scripts but just looking over the grammars-v4/java/java grammar, In the meanwhile, I am just gathering some data for a sanity check. Unfortunately, I do not see and performance increase with the "dev" branch Go runtime. In fact, I now see some grammars/input no longer terminating (killed after 5 minutes vs. before completed in 55s). [grammars-v4/sql/mysql/Positive-Technologies](https://github.com/antlr/grammars-v4/tree/a244fda17cfdbd7ae1443ce2db87092f69f59e69/sql/mysql/Positive-Technologies), [examples/bitrix_queries_cut.sql](https://github.com/antlr/grammars-v4/blob/a244fda17cfdbd7ae1443ce2db87092f69f59e69/sql/mysql/Positive-Technologies/examples/bitrix_queries_cut.sql). I am using the published 4.12 version vs. the "dev" version. Please check.

jimidle:
It's possible that the grammars expose a hidden bug and don't terminate.
But if that is the case, the other grammars should be showing a a big
improvement. That's the case in my local system. BUt I will of course work
out why the bitrix queries parser is hanging.

kaby76:
I tried grammars-v4/java/java first, and saw a slight perf slowdown for the dev-branch, too. That's when I tried the MySQL grammar. I'll try some other grammars.

jimidle:
OK - I may still have work to do then with these. Every grammar that I have
tried has been very much including the recently reported one that supplied
the grammar and test inputs. That parser was taking 13 seconds, and now
takes 70-90ms on my system.

It could always be the case that there are other problems that I did not
hit with the grammars I used of course, but I can't see how/why the
performance would slow down at the moment.

PR #4193, opened 2023-03-18
Improve JS build to better support esmodules/commonjs/node

jharris4:
Hi! We've been using the antlr4 JavaScript runtime for a few years now, and recently got around to modernizing our cypher code editor which is built on top of an antlr4 grammar.

The antlr4 package has been working great for us, but recent churn on the build/packaging has been causing us some headaches. That was the motivation for this PR! :-)

This PR modernizes & cleans up the build for the JavaScript runtime so that it properly exports all of the following:
- commonsjs module via package.json `main` field
- es module via package.json `module` field
- typescript types via the package.json `types` field

Additionally, the package.json `exports` field is used to handle use of the library via nodejs so that exports and types of FileStream etc are never seen in the browser context.

I've published a fork of the antlr JavaScript runtime to npm as `antlr4-js-exports`, and tested it in our editor monorepo to verify that it works as expected from nodejs, and for the web in the both commonjs / esmodule format.


jharris4:
This PR also removes 6+ year old polyfills that aren't need anymore, and cleans up code that was modifying the String.prototype which is not an encouraged practice for JavaScript libraries...

jharris4:
Before I worry about splitting PRs I’d like to get the CI fixed. I got JavaScript fixed but the typescript one is baffling me a little bit. It seems it’s just doing a build with webpack and not actually using any typescript input?!

jharris4:
Now that I got the CI tests passing for this PR, I'm opening a smaller PR against dev with the git commit signature:

https://github.com/antlr/antlr4/pull/4199

jharris4:
And here's the 2nd smaller PR against dev with the git commit signature that removes code that polutes the String.prototype:

#4200

ericvergnaud:
Closing in favor of #4199

PR #4196, opened 2023-03-20
Improve PredictionContext performance, fix bugs

jimidle:
The PredictionContext in Go was an interface prior to this change. However, there was no need for an interface as the differences in the types were minor and can be catered for much faster with a simple struct and a few case statements. Interfaces are also not ideal for Generics. We can end up needing 3 pointer dereferences just to look at an int.

This PR also changes Intervals to stop using pointers. They are two simple to need that.


jimidle:
@parrt - I would appreciate it if you could merge this - I am now waiting for it to complete a big bug fix for config caching. Cheers.

PR #4199, opened 2023-03-20
make javascript package support both esm & commonjs

jharris4:


jharris4:
I fixed the conflicts with dev branch on this PR, and tests are all passing now, except for some failures on cpp-builds for gcc. I don't think that's related to this PR though...?

runtime/JavaScript/src/antlr4/index.web.js
```diff
@@ -8,7 +8,6 @@ import { default as context } from './context/index.js';
 import { default as misc } from './misc/index.js';
 import { default as tree } from './tree/index.js';
 import { default as error } from './error/index.js';
-import { default as CharStreams } from './CharStreams.js';
```
ericvergnaud, line 11: Why on earth would you want to remove this ?

runtime/JavaScript/package.json
```diff
@@ -20,27 +34,29 @@
...
-    "webpack": "^5.75.0",
-    "webpack-cli": "^4.10.0"
+    "del-cli": "^5.0.0",
+    "rollup": "^3.20.0",
+    "typescript": "^4.9.5"
   },
   "scripts": {
-    "build": "webpack",
+    "build-commonjs": "babel --config-file ./babel.commonjs.config.js --extensions \".js\" --out-dir lib ./src/antlr4",
+    "build-es": "babel --extensions \".js\" --out-dir es ./src/antlr4",
```
ericvergnaud, line 53: 'es' should be 'ism' for clarity

runtime/JavaScript/package.json
```diff
@@ -20,27 +34,29 @@
...
-    "typescript": "^4.8.3",
-    "webpack": "^5.75.0",
-    "webpack-cli": "^4.10.0"
+    "del-cli": "^5.0.0",
+    "rollup": "^3.20.0",
+    "typescript": "^4.9.5"
   },
   "scripts": {
-    "build": "webpack",
+    "build-commonjs": "babel --config-file ./babel.commonjs.config.js --extensions \".js\" --out-dir lib ./src/antlr4",
```
ericvergnaud, line 52: commonjs should be cjs and build into cjs, not into lib

runtime/JavaScript/package.json
```diff
@@ -20,27 +34,29 @@
...
-    "@types/node": "^18.7.23",
-    "babel-loader": "^8.2.5",
+    "@babel/cli": "^7.21.0",
+    "@babel/core": "^7.21.3",
+    "@babel/eslint-parser": "^7.21.3",
+    "@babel/preset-env": "^7.20.2",
+    "@rollup/plugin-babel": "^6.0.3",
+    "@types/node": "^18.15.3",
     "c8": "^7.12.0",
-    "compression-webpack-plugin": "^10.0.0",
```
ericvergnaud, line 29: so what happens to compression ?

runtime/JavaScript/package.json
```diff
@@ -1,12 +1,26 @@
 {
   "name": "antlr4",
   "version": "4.12.0",
-  "type": "module",
```
ericvergnaud, line 4: are you proposing to no longer build a module ?

ericvergnaud:
Hi,
thanks for this.
Not sure what problem you are solving...
commonjs and umd are dead, the new standard is esm
the generated parsers are also targeting esm
why are cjs and umd needed when apps can simply consume the esm module ?

runtime/JavaScript/package.json
jharris4: ok, I can change it, but you do recall it was just `dist` before anything else. And I took `es` and `lib` from common examples from other packages, but sure I can change to to `esm` and `cjs` if you prefer.

runtime-testsuite/resources/org/antlr/v4/test/runtime/helpers/package_ts.json
```diff
@@ -1,9 +1,6 @@
 {
   "type": "module",
-   "devDependencies": {
+  "devDependencies": {
     "@types/node": "^18.0.5"
-   },
-  "dependencies": {
```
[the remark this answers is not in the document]
jharris4: nope, all the tests passed without it. You don't need to actually have a package installed to link it. If that were the case nobody would be able to link packages that weren't published first... :-)

runtime/JavaScript/package.json
jharris4: No, the `type` module is now only in the `es` directory. Previously this library wasn't actually exporting a module (webpack doesn't output esm modules) but said it was. That was part the motivation of this PR, this kind of stuff breaks builds for other libraries that are consuming antlr4.

Now this package will export both commonjs & esm and support both.

runtime/JavaScript/package.json
jharris4: compression doesn't really make sense here. Webpack is used for shipping web applications primarily, and this project is a library... In any case, this only impacts the umd bundle, which nobody really uses.

runtime/JavaScript/src/antlr4/index.web.js
jharris4: It has a dependency on `FileStream` which is node only, so the web version needs to have this removed.

runtime-testsuite/resources/org/antlr/v4/test/runtime/helpers/package_ts.json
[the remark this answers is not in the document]
ericvergnaud: that's not my experience at all 
it doesn't need to be installed but it has to be declared

runtime/JavaScript/package.json
ericvergnaud: it was dist when there was a single build. We don't want users to think that cjs is the recommended way of consuming antlr

runtime/JavaScript/package.json
ericvergnaud: why build umd if 'nobody really uses it' ? And what makes you think that antlr is not used in web apps ? I do.

runtime-testsuite/resources/org/antlr/v4/test/runtime/helpers/package_ts.json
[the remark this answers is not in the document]
jharris4: The tests are all passing without it, and I use it all the time without first installing, so I guess we have different experiences here :-)

runtime/JavaScript/package.json
jharris4: ok, I'll change it all to `lib` -> `cjs` and `es` -> `esm`

ericvergnaud:
> @ericvergnaud thanks for the review comments, happy to address or discuss further if that'll help move this PR forward. :-)

Well you'd have to explain what problem you are solving, with a concrete example. 'Conforming to standards' is not a compelling reason for changing the build when those standards are legacy and tripling the maintenance required to keep it working in the future is definitely a show stopper...

jharris4:
> 

The problem I'm solving is that we maintain an editor library for the front-end that uses antlr4 for the grammar.

Recently antlr4 changed (in minor version change no less!) from exporting an esm module, to now in the latest version exporting commonjs but saying it's an esm module.

We also have lots of unit tests on our project that use node require for antlr, whereas the frontend apps use esm imports. It's been whack a mole lately trying to fix one or the other.

Suffice it to say, I got tired of working around this, and figured I was better to contributed back a proper fix and help this project get its JavaScript library using best practices for bundling.

TLDR I don't want to care about bundling, and my goal here is for nobody else to have to care about want bundle antlr is exporting either. With this PR, both commonjs & esm are supported.

If you look carefully, you'll see that in fact the JS tests are now using esm, and the TS tests are still using commonjs because of the tsconfig telling it to here:

https://github.com/antlr/antlr4/blob/8dcc6526cfb154d688497f31cf1e0904801c6df2/runtime-testsuite/resources/org/antlr/v4/test/runtime/helpers/tsconfig.json#L4

jharris4:
Ok, that's a bit of hyperbole saying this PR "triples the maintenance". I put a lot of my own hard (unpaid) work into this, and I spent quite a few hours cleaning things up to try to make it better & easier to maintain. If I failed in that task, it'd be good if you could share specifics.

runtime/JavaScript/package.json
jharris4: we can remove the umd if you'd like. I basically only left it there because it was there before and I was trying to minimize the changes in this PR.

I created this whole PR since we are using antlr4 on the web, so trust me I know how it's used, and this PR is designed specifically to make antlr4 less painful to consume in a web application by correctly setting up the `type`, `main`, `module` & `exports` fields.

runtime/JavaScript/src/antlr4/index.web.js
ericvergnaud: this PR won't make it if it removes functionality

runtime/JavaScript/package.json
ericvergnaud: thanks

ericvergnaud:
not sure where you take it that the current build exports a commonjs module when it actually uses esm imports, doesn't tolerate require and doesn't work at all if you remove "type : module"...

ericvergnaud:
re versioning, the readme tells you that antlr can't follow semver practice

ericvergnaud:
so you're telling me that something is not working as expected from an esm module, and your 'fix' is to make antlr available as a cjs one ?

ericvergnaud:
the reason I say "triple maintenance" is because from the number of attempts you've been through I can anticipate how much pain it's going to be each time anything changes in the tools involved in the build chain. With 1 build chain it's already a pain, you're proposing to have 3, 1 for each module type...

ericvergnaud:
how about just fixing whatever you think is wrong in esm, and updating your tests to use esm too ?

ericvergnaud:
I have plenty of esm stuff using the esm build (that you say is commonjs) and it works like a charm.
So maybe you need to provide examples of what exactly is not working as expected ?

runtime/JavaScript/package.json
ericvergnaud: there was no umd before, so not sure I follow...

ericvergnaud:
I also get the feeling that the problems you encounter relate to your specific situation where the front-end is esm but the back-end is cjs ?

runtime/JavaScript/package.json
jharris4: the bundle you used to create with webpack was also in umd format

jharris4:
As for the motivation of this PR, I offered to have a chat to discuss the motivation & reasoning behind all these changes. A code walkthrough could really help with that, but here we are in text...

Some backstory:

1) We maintain Neo4j's Cypher editor, which uses an antlr4 grammar behind the scenes, and is consumed in both esm & commonjs format (we don't get to decide what our users use esm or cjs)
2) A bunch of users have been complaining to us that their builds break in weird unexpected places like when trying to use the editor in test framework that run in node.
3) I did some investigation and pinpointed the issue to this library, and decided I'd do the right thing and make a PR instead of just forking it.
4) We also had issues where we couldn't use antlr4 in web context without build errors and weird workarounds to silence those errors.

As per your comments about lots of commits getting this working, YOU SURE ARE RIGHT! :-)

The actual changes for this PR were mostly done by me in about 20 minutes, and the rest of the time & commits was understanding tests & fixing them since I couldn't see how to run them locally.

This PR also makes an improvement in that the JavaScript platform test now uses the code that is published to npm instead of just the source, which is a better practice.

I'd prefer to see that this PR LESSENS the maintenance burden because it's bringing the JavaScript library's build system more in line with best practices as well as fixing some issues.

If you'd like to see the kind of thing this PR is trying to solve, here's an example of the kind of workaround we'd like to stop using to be able to use this repository:

https://github.com/neo4j/cypher-editor/blob/main/packages/antlr4-browser/src/index.js

TLDR on above example, without this we can't use antlr4 for the web. So please don't assume I'm inventing problems here that don't exist.

Believe me I'd very much like to be able to not think about it again, but if it'd be helpful, I'd be happy to help maintain the JS build on this package.

jharris4:
Also, FYI unkpg can be really useful for examining package build output.

For example here is the current main entrypoint for antrl4:

https://unpkg.com/browse/antlr4@4.12.0/dist/antlr4.node.js

You can see from the above (which is the build output from the JavaScript package) that the code is not pure esm. it's some weird webpack generated thing that imports the require function from a node module...

ericvergnaud:
The current build follows official webpack practice for esm modules, taken from here:
https://webpack.js.org/guides/ecma-script-modules/
So not sure where you take it that it's not esm...
Yes webpack does replace imports with __require, but that thing doesn't come from node, it's 100% local (which is the whole purpose of webpack).
And a consumer can still write: import { x, y } from 'antlr4'; and is required to use import rather than require.
So it is exposed as an esm package.

AFAIK the build works fine for both node and web, thanks to webpack's multi-module feature.
The dependency on FileStream is dropped as part of webpacking, and there is no need to drop CharStreams to achieve that (obviously CharStreams.fromFile will throw). This has been tested successfully with both node and browsers in esm apps.

I understand that you have a specific need to generate a commonjs library, not just an esm one.
You still haven't provided concrete examples of what doesn't work when using the webpack build. https://github.com/neo4j/cypher-editor/blob/main/packages/antlr4-browser/src/index.js gives me no clue re what breaks for your users. All I see is it doesn't export FileStream and CharStreams.

I'd welcome a cjs build but it can't be at the cost of breaking the simplicity of the current build and dropping features or compression. 

Since cjs is dead anyway, why not stick to your currently working solution (antlr4-browser) until you no longer need to ship cjs ?

runtime/JavaScript/src/antlr4/index.web.js
ericvergnaud: You're duplicating code. CharStreams.node.js should reuse the code from CharStreams.web.js, adding the 2 file related entry points.

runtime/JavaScript/package.json
ericvergnaud: internally, but exposed as esm

jharris4: I'm referring to [this umd formatted build output](https://unpkg.com/browse/antlr4@4.12.0/dist/antlr4.web.js) that was pointed to by the [package.json](https://unpkg.com/browse/antlr4@4.12.0/package.json) `browser` field.

runtime/JavaScript/src/antlr4/index.web.js
jharris4: ok, I can make that change. I duplicated code because it followed the pattern of what was done for the `index.node.js` and `index.web.js` split.

If we want to keep the code base consistent, doesn't this also mean that the same should be done to the two index files to remove duplication? There is quite a hefty amount of duplication there, and in the type definitions too.

If you can accept this duplication temporarily, I think it might be better to open a separate PR if/after this PR is merged to remove duplication in all places. It'd help limit the scope of this PR, do you agree? :-)

ericvergnaud:
Please avoid sentences such as _'If this package wants to be consumable by both node and web, there is no other documented way to make this work that I'm aware of._' or 'some weird webpack generated thing that imports the require function from a node module...', followed by "I understand webpack quite well". 

This package **_is_** consumable by both node and web, provided that they comply with esm, which is required by the generated parsers anyway. The reason I can say this is that **I'm doing it**. It's not OSS stuff so unfortunately I can't share any links, but since you're the only user facing issues with the current packaging (and many have expressed their satisfaction), I know as a fact that what's at stake is not what you say it is...

Your issue is different and specific: you want this package to be available as a cjs package such that it can be bundled as part of a library that you provide for both node using cjs and web using esm. Very few users have that requirement I believe.

I don't yet have an answer to the question of whether we want antlr4 to be consumable by cjs libraries, or more precisely how much effort I'm willing to invest in it.
I'm not sure I want to learn rollup or babel (again) just to satisfy 1 specific scenario that will naturally disappear in a few months.

jharris4:
The sentences that you're asking me to avoid were an attempt to respond to your request to explain and justify why this PR is useful. I apologize if that came across the wrong way, but I have good intentions.

In some sense you are correct that this package is consumable for web currently, but with this massive caveat that it's in umd format only.

Providing only umd is problematic for consumers of antlr4 for the web because it means their code cannot use es6 imports with antlr4, and this makes it harder to consume or test against. A umd bundle also means that bundler tools such as webpack, vite, parcel can no longer do any import optimization on antlr4 which can lead to substantially large bundle file sizes and build time.

The antlr-browser package we added is a horrible hack that only works with version `4.10.1` of antlr4 which did export pure esm modules for node, and is very brittle (it's broken in `4.12.0`).

The changes I have made in this PR resolved all of these issues, and made it so that I could use antr4 with esm for the web.

If you don't mind my asking, did any of the web usages you've seen use version `4.12.0` with es6 imports like `import antlr4 from "antlr4"`? Because for me that gives a compile error without the changes in this PR.

ericvergnaud:
Here is an example of what works in the projects I mentioned earlier:

`import {CharStream, CommonTokenStream, ParserRuleContext, ParseTree, ParseTreeWalker, Token} from "antlr4";`

This is used in an esm typescript module, which is itself used in a typescript esm webapp, running in a worker for background parsing and checking of code being edited in ace editor. The web app is a CRA.

The project also uses `import {FileStream} from "antlr4"; `in tests implemented in mocha, running in node.

This is using 4.12.0 from npm. Out of the box. No tweak.

Appreciate the above is all typescript. So I just tried the following:
 - npx create-react-app test-antlr4-web-js
 - cd test-antlr4-web-js
 - yarn add 'antlr4'
 - in App.js:
   import antlr4 from "antlr4";
   console.log(antlr4)
 - yarn start

Image 2

I suggest you try it too. You'll find that contrarily to what you keep saying, it works out of the box for both the web and node. There's no caveat. Which is why I can affirm that the issue is specific to your setup.

If version 4.11 works, then why switch to 4.12 ? As already mentioned, these upgrades could be breaking, see our README. Maybe this one was because some changes were required to get the TypeScript target to work. There is no bug fix or improvement in 4.12 (for javascript).

jharris4:
I did some more experimentation, and I misspoke saying `4.12.0` is broken for web. I had assumed some build tools would error if they only had the `browser` field in `package.json` because this is not at all common practice, but a quick check shows it does compile fine with both `webpack` and `vite`. I didn't try more picky build tools like `jest` or `vitest` that sometimes tripped over past versions of `antlr4`.

So basically this PR is to change the mode of consumption for web of this package from umd to esm.

This matters because it has a large impact on bundle size and build times.

To illustrate this, I put together [a little test repo](https://github.com/jharris4/antlr4-cjs-esm-tests) showing the webpack & vite build output for both `antlr4` and the `antlr-js-exports` fork I published to npm.

It shows that using true esm imports for the same example saves about 200KB from the build output which is pretty substantial. React is smaller than that for example!

Screenshot 2023-03-22 at 3 58 21 PM

ericvergnaud:
I'm afraid that test is irrelevant. Nobody uses just a Lexer. A typical usage would pretty much consume all of the antlr runtime: Parser, InputStream, CommonTokenStream and ParseTreeWalker at a minimum, so the gain would rarely be significant.

jharris4:
Those imports are bringing in the entire default import, so it's not just the Lexer, it's all the functionality a consuming web library such as our cypher editor needs. So the savings are considerable!

But in any case, it seems like we're reaching the end of the road here. I feel it's a shame that my hard work and commitment to making things better wasn't better received, but I guess part of that is on me for failing to make the case for why this PR is more than just a spurious improvement.

In any case, I guess it's not a total waste since I now see an upgrade path for us from `4.10.1` to `4.12.0` that unblocks us, we just have to accept that the esm for antlr4 isn't really esm, and accept the extra couple of hundred KB bundle size increase as unavoidable.

Thanks for all your great work on this library, I really do appreciate that it builds properly for node & web now in the latest versions thanks to your improvements.

If you want to discuss this PR further I'd be happy to, but I also understand if you'd just rather stick with stuff you're comfortable with.

ericvergnaud:
Indeed, thanks for your efforts. Please close.

jharris4:
ok closing, it'll be around as a reference for anyone else who wants to look into this stuff :-)

I'd consider making another PR to just fix the esm exports (without adding cjs), if there's any interest in having a fix for that merged.

If this library is open to that possibility let me know, otherwise I won't spend the effort.

NilsBaumgartner1994:
stuck at the same problem currently:


```
const antlr4 = require('antlr4');
//import antlr4 from 'antlr4';

async function main() {
  new antlr4.InputStream("Hallo");
```


```
internal/modules/cjs/loader.js:1080
      throw new ERR_REQUIRE_ESM(filename, parentPath, packageJsonPath);
      ^

Error [ERR_REQUIRE_ESM]: Must use import to load ES Module: /Users/nbaumgartner/Documents/GitHub/data-clumps/src/api/node_modules/antlr4/dist/antlr4.web.js
require() of ES modules is not supported.
require() of /Users/nbaumgartner/Documents/GitHub/data-clumps/src/api/node_modules/antlr4/dist/antlr4.web.js from /Users/nbaumgartner/Documents/GitHub/data-clumps/src/api/build/ignoreCoverage/development.js is an ES module file as it is a .js file whose nearest parent package.json contains "type": "module" which defines all .js files in that package scope as ES modules.
Instead rename antlr4.web.js to end in .cjs, change the requiring code to use import(), or remove "type": "module" from /Users/nbaumgartner/Documents/GitHub/data-clumps/src/api/node_modules/antlr4/package.json.

    at Object.Module._extensions..js (internal/modules/cjs/loader.js:1080:13)
    at Module.load (internal/modules/cjs/loader.js:928:32)
    at Function.Module._load (internal/modules/cjs/loader.js:769:14)
    at Module.require (internal/modules/cjs/loader.js:952:19)
    at require (internal/modules/cjs/helpers.js:88:18)
    at Object.<anonymous> (/Users/nbaumgartner/Documents/GitHub/data-clumps/src/api/build/ignoreCoverage/development.js:14:16)
    at Module._compile (internal/modules/cjs/loader.js:1063:30)
    at Object.Module._extensions..js (internal/modules/cjs/loader.js:1092:10)
    at Module.load (internal/modules/cjs/loader.js:928:32)
    at Function.Module._load (internal/modules/cjs/loader.js:769:14) {
  code: 'ERR_REQUIRE_ESM'
}
```

NilsBaumgartner1994:
> Out of the box

Unfortunately this only works for react apps but not for plain node projects. 

Following this simple tutorial to create a npm package: https://itnext.io/step-by-step-building-and-publishing-an-npm-typescript-package-44fe7164964c

And having this simple "index.ts" file:

```
const antlr4 = require('antlr4');
const input = "Hello, world!";
const chars = new antlr4.InputStream(input);

export const Greeter = (name: string) => `Hello ${name}`;
``` 

results in the same error.

NilsBaumgartner1994:
> Why don't you use import { InputStream } from 'antlr4' ?

Because it still show the same error. Would be realy nice if antlr4 would support this as mentioned in this PR.

```
internal/modules/cjs/loader.js:1080
      throw new ERR_REQUIRE_ESM(filename, parentPath, packageJsonPath);
      ^

Error [ERR_REQUIRE_ESM]: Must use import to load ES Module: /Users/nbaumgartner/Documents/GitHub/Playground/node_modules/antlr4/dist/antlr4.node.js
require() of ES modules is not supported.
require() of /Users/nbaumgartner/Documents/GitHub/Playground/node_modules/antlr4/dist/antlr4.node.js from /Users/nbaumgartner/Documents/GitHub/Playground/lib/index.js is an ES module file as it is a .js file whose nearest parent package.json contains "type": "module" which defines all .js files in that package scope as ES modules.
Instead rename antlr4.node.js to end in .cjs, change the requiring code to use import(), or remove "type": "module" from /Users/nbaumgartner/Documents/GitHub/Playground/node_modules/antlr4/package.json.

    at Object.Module._extensions..js (internal/modules/cjs/loader.js:1080:13)
    at Module.load (internal/modules/cjs/loader.js:928:32)
    at Function.Module._load (internal/modules/cjs/loader.js:769:14)
    at Module.require (internal/modules/cjs/loader.js:952:19)
    at require (internal/modules/cjs/helpers.js:88:18)
    at Object.<anonymous> (/Users/nbaumgartner/Documents/GitHub/Playground/lib/index.js:4:16)
    at Module._compile (internal/modules/cjs/loader.js:1063:30)
    at Object.Module._extensions..js (internal/modules/cjs/loader.js:1092:10)
    at Module.load (internal/modules/cjs/loader.js:928:32)
    at Function.Module._load (internal/modules/cjs/loader.js:769:14) {
  code: 'ERR_REQUIRE_ESM'
}
```


I will check your other comment

> Your package must be configured for ESM. Check this: https://gist.github.com/sindresorhus/a39789f98801d908bbc7ff3ecc99d99c

jharris4:
Yeah, this example is trying to use commonjs node module resolution to import antlr4 which afaik just won't work since it currently has `"type": "module"` in the package.json.

This PR fixes the above error (and makes the esm build sizes smaller too), but I seem to have done a poor job explaining why the fix was important enough to be merged.

I wish we were in a pure esm world where we could get rid of commonjs, but too many of our consumers are still using commonjs for us to make that choice for them. It's a tough sell telling them they have to go modernize all their JavaScript to esm from cjs just to use our code, but without this PR we don't seem to have any other options.

It's so strange to hate commonjs and want to get rid of it, and be here trying to argue for it's inclusion in a PR, but here we are! haha

NilsBaumgartner1994:
> Yeah, this example is trying to use commonjs node module resolution to import antlr4 which afaik just won't work since it currently has `"type": "module"` in the package.json.
> 
> This PR fixes the above error (and makes the esm build sizes smaller too), but I seem to have done a poor job explaining why the fix was important enough to be merged.
> 
> I wish we were in a pure esm world where we could get rid of commonjs, but too many of our consumers are still using commonjs for us to make that choice for them. It's a tough sell telling them they have to go modernize all their JavaScript to esm from cjs just to use our code, but without this PR we don't seem to have any other options.
> 
> It's so strange to hate commonjs and want to get rid of it, and be here trying to argue for it's inclusion in a PR, but here we are! haha

I think transitioning to modern projects is good. But I think what would help here would be for people a clean small example package. For me I want to develop a npm package which has typescript and can be imported and used in other projects like react-native. Therefore the building process was a bit hard and to have a possibility to test my package. This makes transitions easier if you have a MWP (minimal working project).

ericvergnaud:
Feel free to propose a sample MWP that can be downloaded from the docs.

NilsBaumgartner1994:
Trying to make it work

NilsBaumgartner1994:
Can use antlr4 in directus because of esm package: https://github.com/directus/directus/issues/17997

NilsBaumgartner1994:
@jharris4 Your package works like a charm <3 Your points are exaclty correct.

> As for commonjs support. Believe me, I would LOVE to move into a world where it's not needed anymore. The trouble is, we are maintaining packages for OTHERS to use. For example with our cypher editor, we have users who are still using commonjs so we can't drop support yet unfortunately.

That is so 100% true. 

@ericvergnaud I hope you guys see that this is a real need :-) I love antlr4 since it is very powerful.

NilsBaumgartner1994:
> antlr-js-exports

For all out there facing a similar problem this package could help from @jharris4

jharris4:
@NilsBaumgartner1994 thanks for confirming the fix in the forked package, but please note that package was purely for testing, and not recommended for real world usage.

People should stick with the official `antlr4` package with the caveat that there's no support for commonjs module resolution.

NilsBaumgartner1994:
Hi there, Sure but as you said, it shows that the fix works.Beste Grüße,Nils BaumgartnerAm 30.03.2023 um 14:19 schrieb Jon Harris ***@***.***>:﻿
@NilsBaumgartner1994 thanks for confirming the fix in the forked package, but please note that package was purely for testing, and not recommended for real world usage.
People should stick with the official antlr4 package with the caveat that there's no support for commonjs module resolution.

antoniom:
We have a similar problem:

Our frontend application (written in TS) is using @reduxjs/toolkit and other dependencies which are not available as ESM. 
Upgrading antlr to 4.12 causes tests that use the package to fail. To fix them I followed [Jest's guide on testing ESM modules](https://jestjs.io/docs/ecmascript-modules) which caused tests that were using non-ESM modules to fail from their side.

NilsBaumgartner1994:
Which antlr4 version did you used prior?Beste Grüße,Nils BaumgartnerAm 31.03.2023 um 14:09 schrieb Antonis Balasas ***@***.***>:﻿
We have a similar problem:
Our frontend application (written in TS) is using @reduxjs/toolkit and other dependencies which are not available as ESM.
Upgrading antlr to 4.12 causes tests that use the package to fail. To fix them I followed Jest's guide on testing ESM modules which caused tests that were using non-ESM modules to fail from their side.

antoniom:
@NilsBaumgartner1994 using 4.8.0 using js/babel, no TS

ericvergnaud:
@antoniom @NilsBaumgartner1994 please check #4217

ericvergnaud:
@jharris4 please check #4217

JohnDoeAntler:
@jharris4 You're my lifesaver, thanks for your NPM package! I was spending half a day trying to get the ANTLR4 runtime running without the ESM error, then I saw the comment from @NilsBaumgartner1994, just in a half minute of installing the package then I got rid all issues! :D

PR #4200, opened 2023-03-20
do not modify String.prototype in js package

jharris4:


runtime/JavaScript/src/antlr4/CharStream.js
```diff
@@ -4,8 +4,6 @@
  */
 
 import Token from './Token.js';
-import './polyfills/codepointat.js';
```
ericvergnaud, line 7: how does this relate to the PR ?

runtime/JavaScript/src/antlr4/index.node.js
```diff
@@ -58,9 +58,3 @@ export {
     RecognitionException, NoViableAltException, FailedPredicateException, ErrorListener, DiagnosticErrorListener, BailErrorStrategy,
     arrayToString
 }
-
-/* eslint no-unused-vars: [ "off"] */
-// need to import unused to force loading
-import StringHashCode from './utils/stringHashCode.js';
```
ericvergnaud, line 64: Did you read the comment ?

runtime/JavaScript/src/antlr4/index.web.js
```diff
@@ -57,9 +57,3 @@ export {
     RecognitionException, NoViableAltException, FailedPredicateException, ErrorListener, DiagnosticErrorListener, BailErrorStrategy,
     arrayToString
 }
-
-/* eslint no-unused-vars: [ "off"] */
-// need to import unused to force loading
-import StringHashCode from './utils/stringHashCode.js';
```
ericvergnaud, line 63: Same here

runtime/JavaScript/src/antlr4/CharStream.js
jharris4: Those polyfills are modifying the String.prototype if you look at the code.

For example here:

https://github.com/antlr/antlr4/blob/master/runtime/JavaScript/src/antlr4/polyfills/codepointat.js#L58

I did check that these polyfills are no longer needed, we've come a long way since they were added around 6 years ago! Now those functions that were polyfilled are available everywhere.

https://caniuse.com/mdn-javascript_builtins_string_codepointat
https://caniuse.com/mdn-javascript_builtins_string_fromcodepoint

ericvergnaud:
Yes THIRD-PARTY-NOTICES should be removed

PR #4206, opened 2023-03-23
Added setErrorHandler

pl7y:


pl7y:
Opened in error, closing.

ericvergnaud:
Hey
thanks, but as mentioned we don't use setters in JS. One can just set _errHandler directly.

Issue #4211, opened 2023-03-28
Esbuild fail to bundle from 4.10 to 4.12


richard-julien:
- [x] I have reproduced my issue using the latest version of ANTLR
- [x] I have done a search of the existing issues to make sure I'm not sending in a duplicate

Please include the following information:
 - target information: **Javascript**
 - smallest possible grammar and code that reproduces the behavior: **N/A related only to code import**
 - description of the expected behavior and actual behavior: No error bundling code with esbuild

Hi Antlr team,

I will try to expose my problem in a clearer way but sorry if information looks like partial. Difficult to explain correctly my problem with my current knowledge of ANTLR.

I try to upgrade ANTLR in my project (nodejs) from 4.10 to 4.12.
My project use esbuild to bundle the source code.
After upgrading to 4.12, the bundled code refuse to start with the error.

```
node:internal/modules/cjs/loader:1367
    throw new ERR_INVALID_ARG_VALUE('filename', filename, createRequireError);
    ^

TypeError [ERR_INVALID_ARG_VALUE]: The argument 'filename' must be a file URL object, file URL string, or absolute path string. Received undefined
```

Looking into the bundled code, its looks like the esbuild transform the file and i have this line.

`var ha = (0, import_module.createRequire)(import_meta.url)("fs");`

And as the error said, the import_meta is empty initialized so url is undefined.

Looking into esbuild, the documentation said that import.meta.url is not supported when bundling:

`You also may not want to bundle your dependencies with esbuild. There are many node-specific features that esbuild doesn't support while bundling such as __dirname, import.meta.url, fs.readFileSync, and *.node native binary modules. You can exclude all of your dependencies from the bundle by setting [packages](https://esbuild.github.io/api/#packages) to external:`

So my question is, do you change something in ANTLR that now use the meta import feature that break the compatibility with esbuild bundling? 

Thanks in advance.

ericvergnaud:
Yes the packaging changed in 4.12 for TypeScript to work.Envoyé de mon iPhoneLe 28 mars 2023 à 11:18, Julien Richard ***@***.***> a écrit :﻿

 I have reproduced my issue using the latest version of ANTLR
 I have done a search of the existing issues to make sure I'm not sending in a duplicate

Please include the following information:

target information: Javascript
smallest possible grammar and code that reproduces the behavior: N/A related only to code import
description of the expected behavior and actual behavior: No error bundling code with esbuild

Hi Antlr team,
[...]

PR #4217, opened 2023-04-01
[Javascript] Added support for commonjs in nodejs package

Tarjei400:
No package should enforce upon creator a need to migrate to ESM. As per this pull request, I added webpack configuration that produces cjs and mjs files Correspoding to CommonJS and ESM module. After this pull request if will be up to a project which one to use.

Cheers!

### DIFF
--- a/runtime/JavaScript/spec/imports/NodeCommonJSImportSpec.cjs
+++ b/runtime/JavaScript/spec/imports/NodeCommonJSImportSpec.cjs
@@ -1,29 +1,6 @@
 "use strict";
-var __createBinding = (this && this.__createBinding) || (Object.create ? (function(o, m, k, k2) {
-    if (k2 === undefined) k2 = k;
-    var desc = Object.getOwnPropertyDescriptor(m, k);
-    if (!desc || ("get" in desc ? !m.__esModule : desc.writable || desc.configurable)) {
-      desc = { enumerable: true, get: function() { return m[k]; } };
-    }
-    Object.defineProperty(o, k2, desc);
-}) : (function(o, m, k, k2) {
-    if (k2 === undefined) k2 = k;
-    o[k2] = m[k];
-}));
-var __setModuleDefault = (this && this.__setModuleDefault) || (Object.create ? (function(o, v) {
-    Object.defineProperty(o, "default", { enumerable: true, value: v });
-}) : function(o, v) {
-    o["default"] = v;
-});
-var __importStar = (this && this.__importStar) || function (mod) {
-    if (mod && mod.__esModule) return mod;
-    var result = {};
-    if (mod != null) for (var k in mod) if (k !== "default" && Object.prototype.hasOwnProperty.call(mod, k)) __createBinding(result, mod, k);
-    __setModuleDefault(result, mod);
-    return result;
-};
-Object.defineProperty(exports, "__esModule", { value: true });
-const antlr4 = __importStar(require("antlr4"));
+
+const antlr4 = require("antlr4");
 describe('Antlr4 Node CommonJs', () => {
     it('should use the CommonJS module on Node.js', () => {
         expect(antlr4).toBeDefined();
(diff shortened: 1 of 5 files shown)


ericvergnaud:
Hey, nobody's enforcing anything, and we welcome contributors. Let's see how this works for other devs looking for CommonJS support.

NilsBaumgartner1994:
@Tarjei400 thank you :-) will have a look. @ericvergnaud thanks for notification.

Tarjei400:
@enessoylu I just mean that at the moment library requires ESM in a project and this is "enforcing" ESM in a project, I didnt mean anyone per se :P Anyways thanks for taking a look, I think it will be a usefull addition for this package!

jharris4:
I think adding `exports` to the `package.json` is a step in the right direction.

Following down this path should help with the TypeScript target as well. For example as of TypeScript 5.0 you can do the following with `exports` in your `package.json` to export separate types for node and web:

```
{
  "exports": {
    ".": {
      "node": {
        "types": "src/index.node.d.ts",
        "default": "dist/bundle.node.js"
      },
      "browser": {
        "types": "src/index.web.d.ts",
        "default": "dist/bundle.web.js"
      }
    }
  }
}
```

And with TypeScript 5 the package entry point & types can be toggled in the `tsconfig.json` by changing `compilerOptions.moduleResolution` between `nodenext` and `bundler`.

[TypeScript docs]( https://devblogs.microsoft.com/typescript/announcing-typescript-5-0/#moduleresolution-bundler)

[NodeJS docs](https://nodejs.org/api/packages.html#conditional-exports)

Tarjei400:
@jharris4 Sorry it took some time. I checked that export solution and added tests for nodejs esm/commonjs

runtime/JavaScript/spec/imports/NodeCommonJSImportSpec.cjs
```diff
@@ -0,0 +1,31 @@
+"use strict";
+var __createBinding = (this && this.__createBinding) || (Object.create ? (function(o, m, k, k2) {
```
ericvergnaud, line 2: I could use an explanation of the code in this file. Why can't we just write "require('antlr4')" ?

ericvergnaud:
If I understand correctly, this PR would generate esm and cjs for web and node, without touching any runtime code ? Sounds promising! Awaiting for my inline comments to be addressed.

runtime/JavaScript/spec/imports/NodeCommonJSImportSpec.cjs
Tarjei400: Right, I just transpiled it with ts, but you are right require will do the job. Changed :+1:

runtime/JavaScript/webpack.config.js
```diff
@@ -5,68 +5,63 @@ import {fileURLToPath} from "url";
 const __filename = fileURLToPath(import.meta.url);
 const __dirname = path.dirname(__filename);
 
-const nodeConfig = {
+
+const buildConfig = ( platform, extentions ) => ({
```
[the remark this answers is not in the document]
Tarjei400: Type, fixed

Tarjei400:
Correct @ericvergnaud :D Addressed your comments!

ericvergnaud:
Good stuff! Thanks so much, I know for sure this will make many developers happy.
@parrt blessed

Tarjei400:
Myself included :D Also thanks for this great tool!

ericvergnaud:
Fixes #4211

jedwards1211:
@Tarjei400 big thanks for this 😊

NilsBaumgartner1994:
NiceBeste Grüße,Nils BaumgartnerAm 10.11.2023 um 09:43 schrieb Andy Edwards ***@***.***>:﻿
@Tarjei400 big thanks for this 😊

PR #4222, opened 2023-04-10
Remove redundant __name__ check in multiple files

Tameem97:
Removed the redundant `__name__` checks in Python 3 files. In Python 3, `__name__` is implicitly set and will never be None, making this condition always true. Did not modify any of Python 2 files since I am not sure about Python 2, but I think the same logic may also apply.


ericvergnaud:
@parrt blessed

PR #4223, opened 2023-04-11
Feature/memory

jimidle:
feat: Memory usage improvements, performance and statistics

 - Implements a statistics build configuration which gathers useful statistics for maintainers. No impact if not selected for build. Documentation to follow in separate PR
 - Improves implementation of certain collections and provides dedicated non-generic collections in some cases.
 - Adds configuration by function for the runtime, which is used for trace builds etc. No change to normal invocation API.
 - Unfortunately this seems like a lot of changes but they are mainly func signatures for new structs etc.
 - Go has slightly change the formatting requirements - best to keep up to date but some files are merely formatting changes. Git can ignore whitespace changes on diffs etc though
 - Comment improvements for both me and any future maintainers

Badly formed grammars will see some small improvements in memory usage and performance, though I doubt that this is a lot. I have been using the mysql contributed grammar as an example of the worst runtime case. But I also used release build flags rather than the default go ones:

```sh
 go build -ldflags="-s -w" -gcflags="-N -l -d inlfuncswithclosures,inlstaticinit" . 
```

@kaby76 - maybe you can change your build rig to use the go flags above. It can make a difference in some cases, especially on good grammars.

From here, I will concentrate on rewarding well formed grammars. I am disinclined to spend more time on the worst cases for the moment, until the statistics package that this PR implements is fully implemented.

@parrt - if you would please merge this when you have a chance?


parrt:
yay!

PR #4231, opened 2023-04-16
Add support to Golang codegen for booleans

mdehoog:
Currently Golang codegen passes through `boolean` types without type conversion. This generates invalid Golang, as `boolean` is not Golang keyword.

This PR copies Swift's implementation of a type map: https://github.com/antlr/antlr4/blob/8dcc6526cfb154d688497f31cf1e0904801c6df2/tool/resources/org/antlr/v4/tool/templates/codegen/Swift/Swift.stg#L40-L48

### DIFF
--- a/tool/resources/org/antlr/v4/tool/templates/codegen/Go/Go.stg
+++ b/tool/resources/org/antlr/v4/tool/templates/codegen/Go/Go.stg
@@ -1,3 +1,8 @@
+GoTypeMap ::= [
+	"boolean":"bool",
+	default : key
+]
+
 fileHeader(grammarFileName, ANTLRVersion) ::= <<
 // Code generated from <grammarFileName> by ANTLR <ANTLRVersion>. DO NOT EDIT.
 >>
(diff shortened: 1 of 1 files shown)


jimidle:
Where is that input coming from? Something in your grammar itself? Can you give me an example? I am not saying this isn't a good idea, but I have forgotten where the type map comes in to it to be honest?

jimidle:
> Here's some examples: https://github.com/ethereum/solidity/blob/c6aab84a0c25a9348e22326f12c2ebbf407989dc/docs/grammar/SolidityParser.g4#L161-L164 (see the [PR](https://github.com/ethereum/solidity/pull/14128) linked to this one for context). Also related: [ethereum/solidity#9679](https://github.com/ethereum/solidity/issues/9679), which is relevant to the C++ codegen.
> 
> Reason for force-pushing is to keep the commit history clean (there was a small typo in the original commit i pushed that the CI flagged) (and yes, didn't realize the `-s` was a requirement for this repo).

Ah, I see. I thought it was about embedded code in the parser.  On the face of it, this does not seem unreasonable. Let me check your changes...

jimidle:
@parrt I approve of this PR.

jimidle:
@parrt This can be merged

PR #4233, opened 2023-04-17
Feat: Clean up go input streams to be a little more efficient

jimidle:
feat: Clean up input streams

  - The input streams were inefficient, making more copies of the incoming data than they needed to.
  -  Add new Go idiomatic IO stream,  though this cannot be open-ended such as a socket, it is easier for go programmers
     in certain situations.


jimidle:
@parrt Would you merge please?

Issue #3938, opened 2022-10-25
Performance Issues Running Cpp Runtime on Many Threads


wangtao9:
When I use the Cpp parser in multiple threads (around 30+), the performance drops a lot compared to using a single thread.

I'm guessing it's the lock/mutex, so I simply removed all locks and re-tested, and it did run a lot faster.

But removing locks makes the parser no longer thread-safe, I would like to ask is there any way to improve the performance of multi-threaded parser?

andreasbuhr:
@wangtao9 is this solved as https://github.com/antlr/antlr4/pull/4237 is merged?

Issue #2584, opened 2019-06-30
Parsing with C++ target 5x slower than Java


olowo726:
I recently migrated a project from Java to C++.  I built the C++ runtime with Visual Studio Community 2019 which I also use for everything else in the C++ project. All my tests passes which shows that the parsing itself works the same for both targets. 

The actual parsing for C++ is about 5x slower than Java even though the C++ code is built with release configuration optimized for speed and Java is executed in Eclipse debug mode!

The files parsed in the tests are about 500k. I can provide the grammar and data to parse if needed.

Anyone with a clue on why and what can be done?

ericvergnaud:
Hi,

the place for support is the google discussion group.

Cheers,

olowo726:
Hi,

Perhaps I have misunderstood the purpose of this forum. I thought it was for bugs and improvement suggestions.

Well, for sure this isn't a total crash and neither does the C++ implementation fail to parse the data. But I still feel that when a C++ implementation is more than 5x slower than the Java one there is probably some issue lurking which deserves investigation. I doubt that this can be solved by using the tool differently, i.e. a code generation for or compiler flag.

Kind Regards,
Olof

ericvergnaud:
Hi,

sorry for the tone, but you’re making a lot of assumptions.
Every target language comes with its strengths and weaknesses, and Antlr cannot therefore guarantee that the performance of a given grammar will be consistent across targets.
Using the tool differently may very well solve your problem.

Eric

KvanTTT:
I think GitHub tracker is the correct place for such and other performance issues. Especially, considering the fact that C++ runtime is 5x slower than Java one. There are many issues [exist](https://github.com/antlr/antlr4/issues?utf8=%E2%9C%93&q=is%3Aissue+is%3Aopen+performance). C++ is a fast language compare to Python and JavaScript.

ericvergnaud:
Sorry Ivan, but without a narrowed use case i.e. very small grammar that reproduces the issue and a profiler log, not sure anyone will look into this.
This slowdown could result from a poor grammar design, a compiler setting...

Nic30:
I do have similar issue with this [grammar](https://github.com/Nic30/hdlConvertor/blob/master/grammars/sv2017Parser.g4)
The grammar is complicated and contains all sort of problems (including many ambiguities, context sensitivity etc.). I am currently trying to find out where the issue is.

[...]

Issue #2454, opened 2018-12-26
antlr4 doesn't scale on a computer with 8 or more cores


waterguo:
version: 4.5
language: java

The problem is caused by a synchronization block in LexerATNSimulator.addDFAState(...). The block is preventing the antlr lexer from scaling on a multiple core computer.
```
synchronized (dfa.states) {
...
}
```

My syntax file can be found at https://github.com/waterguo/antsdb/blob/master/fish-server/src/main/antlr4/com/antsdb/saltedfish/lexer/Mysql.g4

Input file is a database dump file generated by mysqldump. It has lots of blob and clob fields. Thus some tokens generated by lexer can be as large as 16mb. 

My current workaround is to create a custom interpreter like below
setInterpreter(new LexerATNSimulator(this, _ATN, deepClone(_decisionToDFA), new PredictionContextCache()));

The workaround makes the parsing as close to 10 times better on a 16 core computer.

Though my workaround works but I think this should be an something addressed by antlr. Or maybe something to think about for the next version of antlr. 

I am not able to provide the code to reproduce the problem for now due to sheer size of the input data - 100gb and complexity of my project. I will try to come up with a simplified code example that can demonstrate the problem later.

waterguo:
@sharwell Thanks. I will try it.

baran37:
@sharwell Any chance of this being addressed in the mainline?  I am hitting this also and have had some trouble with the optimized fork.

sharwell:
@baran37 it's possible, but not likely. The changes could be challenging to implement here and verify for correctness. Please feel free to file an issue on [tunnelvisionlabs/antlr4](https://github.com/tunnelvisionlabs/antlr4) if you'd like help with the optimized fork.

baran37:
Ok, thanks for the quick response!

PR #4237, opened 2023-04-18
Optimize the concurrent performance of Cpp target by more than 10 times

wangtao9:
**Usage:**
~~add `-lock-free-cpp-target` option when generating parser, e.g.~~
~~java -jar ${ANTLR_JAR} -Dlanguage=Cpp -lock-free-cpp-target Cypher.g4~~
add compile option ` -DANTLR4_USE_THREAD_LOCAL_CACHE=1` when compiling the Cpp lexer & parser.

**Related issues:**
https://github.com/antlr/antlr4/issues/2454
https://github.com/antlr/antlr4/issues/2584
https://github.com/antlr/antlr4/issues/3938
[Why the C++ target is 6X slower than the Java target](https://stackoverflow.com/questions/52528982/why-the-c-target-is-6x-slower-than-the-java-target)

**Optimization result:**
image

**Test configuration:**
Intel(R) Xeon(R) CPU E5-2682 v4 @ 2.50GHz ; Cores: 16 ; Logical processors: 32
256GB memory
grammar file: https://s3.amazonaws.com/artifacts.opencypher.org/M21/Cypher.g4
test query:
```cypher
    MATCH (n:Person {id:2000})-[:knows]->(friend)-[:located_in]->(city)
    RETURN friend, count(city) ORDER BY friend.id LIMIT 100
```


KvanTTT:
This optimization looks great.

jimidle:
I have done a similar build config for go. It does make a difference. I didn’t get 10x from go, but there could be all sorts of reasons for that. 

Is your input example all you tried? Your mileage may vary on different input. I’ll try the same thing with go as well. 

It’s 22:45 where I live, so I’ll try tomorrow

wangtao9:
> I have done a similar build config for go. It does make a difference. I didn’t get 10x from go, but there could be all sorts of reasons for that.
> 
> Is your input example all you tried? Your mileage may vary on different input. I’ll try the same thing with go as well.
> 
> It’s 22:45 where I live, so I’ll try tomorrow

I've also tried simpler inputs such as "RETURN 1" and more complex examples of over 800 characters, both with significant improvements.

But more importantly, this optimization can achieve a huge performance improvement from a mechanical point of view. C++'s runtime handles locks differently from JVM, and it will fall into kernel calls more frequently, which is one of the main reasons why the concurrency performance of c++ target is much slower than that of java.

jimidle:
Understood. Go used atomically mostly, it is actually pretty good at this
stuff.

wangtao9:
@parrt Can this PR be merged?

tool/src/org/antlr/v4/tool/Grammar.java
```diff
@@ -82,6 +82,7 @@ public class Grammar implements AttributeResolver {
 		parserOptions.add("language");
 		parserOptions.add("accessLevel");
 		parserOptions.add("exportMacro");
+		parserOptions.add("lockFreeCppTarget");
```
KvanTTT, line 85: What the meaning of such option? If it improves performance, why just don't turn on the optimization unconditionally?

KvanTTT (changes requested):

parrt:
Perhaps @hzeller has an opinion here, but frankly I'm terrified of Multi threaded version of the parsing strategy... It is incredibly tricky to get right and there are many people that rely on the C++ runtime.

parrt:
OK I just looked at the code. Adding threads you are simply removing a lock I correct? I guess the question is how does it work without the lock in a multithreaded environment?

hzeller:
I had a brief look - it changes one global varibale with multiple thread_local ones. I don't know exactly the call sequence of the cpp runtime, so it is hard to know what happens then: it seems like now initialize is called unconditionally every time instead of once which would be strange, but I think I first have to look exactly what that means by looking at the generated code. Maybe this evening.

Personally, I would anywayavoid doing something like
 ```c++
static Foo foo = nullptr;
call_once(initialize_foo);
```

But more something like
```
  static Foo foo = new Foo();
```
Then the memory model takes care of initializing that exactly once and possibly be more efficient than any call_once implementation, which was more a thing needed before the c++11 clarification of the memory model.
I know @jcking was looking at multi-threaded performance, maybe he has come across this part of the code.

**Comment by hzeller** [comment]
@jcking changed the once implementation to be either a local one or an absl one in [this change](https://github.com/antlr/antlr4/commit/ae666e6865a9ffb454a2f99f084c0a2727e8f4fc#diff-302ffe9cafc70e47b2454064237ec1cfea1336126412caba9268dbccd8142bde), but it was not changing the call_once() need per-se. I suspect he did that because things are faster with `absl`.

With suggestions in my [previous comment](https://github.com/antlr/antlr4/pull/4237#issuecomment-1520837195) we can eliminate the pre-c++11 need to `call_once()` entirely and just do the static initialization.

**Comment by wangtao9** [comment]
> OK I just looked at the code. Adding threads you are simply removing a lock I correct? I guess the question is how does it work without the lock in a multithreaded environment?

@parrt  Not exactly. It does not simply remove a lock, but changes the static data shared by multi-threads into one copy for each thread. Since the data is owned by each thread, locks are no longer relied upon to keep the data safe (although locks still exist).

**Comment by wangtao9** [comment]
@hzeller 
"eliminate the pre-c++11 need to call_once()" is a good thing, but it's not directly related to what this optimization does.

The idea of this optimization is to turn the static data shared by multi-threads into thread-owned, so as to avoid competition for locks.

`<lexer.name>::initialize()`/`<parser.name>::initialize()` is actually only called once in the constructor (called once per thread), so this optimization does not require `call_once()` or the equivalent semantics of C++11.

**Comment by hzeller** [comment]
Is the data structure modified in each thread ?

If not, we don't need locks, and can make the data structure `const` to make sure nobody does that in the future.

But  if so, then it being static sounds like a bad idea. Thread local will fix that particular situation to not require locks then but it also means that there is something else going on and changing it to thread local wil change the semantics as now every thread sees a different content.

File: tool/src/org/antlr/v4/tool/Grammar.java (hunk 1 above)
Comment (wangtao9) [reply tool/src/org/antlr/v4/tool/Grammar.java:85]: This optimization is not costless, requiring more memory space for storing thread-owned static data. For scenarios with low throughput, the implementation before optimization is also OK.

**Comment by wangtao9** [comment]
@hzeller 
The static data (mainly DFA, ATN, etc.) is constructed during the parse process, and will not be modified after the construction is completed. Most importantly, the above data structures constructed by multi-thread and single-thread are completely consistent.

I also verified this with experiments, as shown in the figure below, the left is the log of building DFA states with a single thread, and the right side is 4 threads doing the same thing. It can be seen that except for the different thread ids, the constructed DFA states are exactly the same.
image

**Comment by wangtao9** [comment]
@hzeller 
The static data (mainly DFA, ATN, etc.) is constructed during the parse process, and will not be modified after the construction is completed. Most importantly, the above data structures constructed by multi-thread and single-thread are completely consistent.

I also verified this with experiments, as shown in the figure below, the left is the log of building DFA states with a single thread, and the right side is 4 threads doing the same thing. It can be seen that except for the different thread ids, the constructed DFA states are exactly the same.
image

**Comment by ericvergnaud** [comment]
So basically you're sacrificing reuse to avoid locks? I suspect this might be counter beneficial in terms if performance with complex grammars because each thread needs to rebuild the complete DFA instead of it being built just once? Is it possible that the locks that currently protect concurrent updates of the DFA are protecting too much code ?

File: tool/src/org/antlr/v4/codegen/model/Recognizer.java [hunk 2]
```diff
@@ -77,6 +78,7 @@ public Recognizer(OutputModelFactory factory) {
 		else {
 			superClass = null;
 		}
+		lockFreeCppTarget = "ON".equals(g.getOptionString("lockFreeCppTarget"));
```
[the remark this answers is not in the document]
wangtao9: Indeed, I have fixed it and updated the usage.

tool/src/org/antlr/v4/tool/Grammar.java
KvanTTT: I don't think target name (in this case `Cpp`) should be included to the option name. What if the similar behaivor will be implemented in another target?

wangtao9:
@ericvergnaud I think what you say makes sense, it could happen.
However, in my usage scenario (ultra-high concurrent requests), I care more about the scalability of multi-core performance, and this optimization is necessary. And in several scenarios I tested, it is effective.
To be on the safe side, I added an option `-lock-free-cpp-target` and turned off this optimization by default.

tool/src/org/antlr/v4/tool/Grammar.java
wangtao9: Exactly! Another name is appropriate, how about `-thread-local-recognizer`? Have other recommendations?

However, it seems that only the Cpp target need this. Targets such as Go and Java will not be dragged down by excessive lock competition.

hzeller:
So if the state is constructed once and then never modified, then the object can be `const` right ? If so, there is no need for any locks (and thus thread local).
So the fact that this apparently makes a difference sounds like the state is indeed modified in the various threads, so that there need to be locks ?

KvanTTT:
I thought about it again and now I'm not sure it's a good idea to include the option to generator. Why this runtime option can't be activated in runtime? It's more flexible solution since it doesn't require regeneration. At least, in C++ it's possible to use preprocessor directive.

tool/resources/org/antlr/v4/tool/templates/codegen/Cpp/Cpp.stg
```diff
@@ -144,7 +144,11 @@ struct <lexer.name; format = "cap">StaticData final {
 };
 
 ::antlr4::internal::OnceFlag <lexer.grammarName; format = "lower">LexerOnceFlag;
+<if (lexer.lockFreeCppTarget)>
+static thread_local <lexer.name; format = "cap">StaticData *<lexer.grammarName; format = "lower">LexerStaticData = nullptr;
+<else>
 <lexer.name; format = "cap">StaticData *<lexer.grammarName; format = "lower">LexerStaticData = nullptr;
+<endif>
```
KvanTTT, line 151: Please extract common part:

```suggestion
<if (lexer.lockFreeCppTarget)>
static thread_local
<endif>
<lexer.name; format = "cap">StaticData *<lexer.grammarName; format = "lower">LexerStaticData = nullptr;
```

ericvergnaud:
I’m supportive of this evolution where the behavior is actionable via a macro when compiling the runtime rather than when generating the parser Envoyé de mon iPhoneLe 25 avr. 2023 à 20:39, Ivan Kochurkin ***@***.***> a écrit :﻿
I thought about it again and now I'm not sure it's a good idea to include the option to generator. Why this runtime option can't be activated in runtime? It's more flexible solution since it doesn't require regeneration. At least, in C++ it's possible to use preprocessor directive.

tool/resources/org/antlr/v4/tool/templates/codegen/Cpp/Cpp.stg
```diff
@@ -364,7 +372,11 @@ struct <parser.name; format = "cap">StaticData final {
 };
 
 ::antlr4::internal::OnceFlag <parser.grammarName; format = "lower">ParserOnceFlag;
+<if (parser.lockFreeCppTarget)>
+static thread_local <parser.name; format = "cap">StaticData *<parser.grammarName; format = "lower">ParserStaticData = nullptr;
+<else>
 <parser.name; format = "cap">StaticData *<parser.grammarName; format = "lower">ParserStaticData = nullptr;
+<endif>
```
KvanTTT, line 379:
```suggestion
<if (parser.lockFreeCppTarget)>
static thread_local
<endif>
<parser.name; format = "cap">StaticData *<parser.grammarName; format = "lower">ParserStaticData = nullptr;
```

KvanTTT (changes requested):
I thought about it once again and now I'm don't think introducing the new option is a good idea. Why this runtime option can't be activated in runtime? It's more flexible solution since it doesn't require regeneration. At least, in C++ it's possible to use preprocessor directives:

```cpp
#if USE_THREAD_LOCAL_CACHE
static thread_local 
#endif
<lexer.name; format = "cap">StaticData *<lexer.grammarName; format = "lower">LexerStaticData = nullptr;
```

jimidle:
In Go, I use build configurations that allow the runtime to be built in
single threaded mode. Because the runtime is built at the same time as the
generated parser etc then this works on a per project basis. C++ needs a
different mechanism I guess as the library is pre-built. So -D build
configuration and building different versions of the library seems
reasonable.

But in Go, my purpose is to elide locks altogether when the user knows that
each instance will be independent - the default build is with mutexes.
Things that are just statically initialized are done via a do.Once() and
are not locked after that. I would change the static to just declare in
line, but it needs to deserialize the lexer etc. and is small beer time
wise. The locks are used for prediction cache etc. They are needed if
multiple go routines are calling the same lexer/parser. I did not feel any
need to bother changing code gen as the do.Once() mechanism is
extremely fast anyway, and in single threaded mode it is basically just an
atomic read once it has happened once.

I have not found it overly useful to reuse the parser in multiple go
routines (equivalent to threads) as there tends to be a lot of other work
going on around it, which would mean I have to start mutexing all of that.
It is likely the same for other targets - is anyone really using multiple
threads on the same recognizer? If the grammar is well formed, then the
overhead of one instance of my toolchains per go routine is essentially
irrelevant. Poor grammars are basically just that and I am not going to do
much more work improving things for poor grammars when the answer is to fix
the grammar.

Go is a bit different though as channels make it easy to spark up N workers
and distribute the compile work in parallel. The parsing is generally
trivial compared to everything else that needs doing, which needs multiple
tree walks. So one go routine per translation unit and in single threaded
build configuration works well.

Locking contention with many go routines using one parser is measurable.
However, I did not achieve the same speed increase that this PR suggests
for the C++ runtime. Perhaps the locking in the C++ runtime needs some
examination?

wangtao9:
@hzeller In C++, these data cannot be simply declared const, because we have no way to determine their values during the declaration phase, they are assigned during the parse process.

But I think there is a way (with major changes) to completely remove the use of locks. That would be a big upgrade, and similar optimizations could be achieved.

wangtao9:
@KvanTTT I also support this modification, changing the generate-time option to compile-time.
I will make it happen.

wangtao9:
@jimidle What is the multithreading speedup you measured in Go?

I suppose that the scalability should be much better than Cpp target (the speedup ratio of 32 threads compared to single thread is only 1.34)

jimidle:
I have not done any formal testing of this as I prefer to think of it as
experimental for this release. For one parser I did not get a huge
improvement, but I know that when I use something like 32 routines (usually
there is not much point using more than the number of cores though) there
is an overall performance benefit because there is no lock contention. I
will do some 'reality' testing at some point down the line. The other thing
with so many threads is of course memory cache which takes exploring as to
whether the effects are large. They sometimes can be very large. Hence I
don't require the users to pick one or the other - they can change per
project. That's easier in Go than most languages.

tool/resources/org/antlr/v4/tool/templates/codegen/Cpp/Cpp.stg
wangtao9: Done. Please review the new code.

tool/src/org/antlr/v4/tool/Grammar.java
wangtao9: I haven't compared it directly, but I have compared it with my colleagues who use Java target. The scalability of Java target is far better than that of C++. To be precise, the scalability of the C++ target is poor--the speedup ratio of 32 threads is only 1.34.

In addition, someone has done a comparison of C++ vs. Java: *Why the C++ target is 6X slower than the Java target* https://stackoverflow.com/questions/52528982/why-the-c-target-is-6x-slower-than-the-java-target

wangtao9: The reason I mentioned before is, C++'s runtime handles locks differently from JVM, and it will fall into kernel calls more frequently

wangtao9:
@parrt @KvanTTT @jimidle @hzeller @ericvergnaud 
Hi all, I created a test repo for this PR, following the steps in README, you can easily verify the optimization effect.

The test repo: https://github.com/wangtao9/antlr4-perfopt-test

KvanTTT (changes requested):
Also, please add information about the new directive to the doc ([cpp-target.md](https://github.com/antlr/antlr4/blob/master/doc/cpp-target.md)).

hzeller:
My main concern is if it is ok to not share the state between threads, in other words if there are no correctness constraints violated.

If (and only if) the data structure mostly just needs initialization over a period of 'warm-up time'  but in fact is not dependent at all on the input and does not need to share between inputs but essentially ends up being exactly the same content in each thread, then it is a valid temporary hack to declare it `thread_local`.

Of course, it should then be considered to solve this another way that is more explicit; for instance actually have a per-thread datastructure in which then such object is kept, then also all locks can be removed. This might be a larger change as @wangtao9 points out. It should still be done, and we should file an issue that contains all the necessary details.

TL;DR: 
  * do the threads need communication for correctness ? (then this change is problematic)
  * Or is the data structure independent and just now is shared for memory reasons, then this change is ok (but needs some follow-up refactoring to better reflect the per-thread ownerships).

parrt:
I remember that @sharwell and I had to think really hard about this for the java target. Our goal was to avoid locking when possible. I think one of the key ideas was that we allowed some race conditions to fill edges of the DFA multiple times from different threads because they would all arrive at the same answer. Ultimately one of them would win, but it didn't matter which one.

For the C++ target, the same concepts would apply but I'm not qualified to say whether the data structures are set up and whether the algorithm is exactly the same. It strikes me as a severe risk. An option could be used to turn it on but then we have another option to deal with, which doubles the state space of our testing.

sharwell:
One thing that would be a potential problem on C++ but not on Java/C# is the latter allows us to compute a value multiple times and then quietly discard all but one of them. The C++ target would be responsible for determining which one among multiple values actually ended up in the cache, and releasing the memory from all the remaining ones.

parrt:
Thanks, @sharwell ! Yeah, this sounds risky w/o someone doing a deep dive.  This could be a major contribution but we'd have to be sure, to avoid screwing C++ targeted parser folks.

hzeller:
Is it possible to entirely pre-compute the needed data structure once at compile time and then compile into the system as a big static array of sorts ?

parrt:
Unfortunately, the nature of the data structure is that we dynamically construct the DFA according to the needs of the input for a particular run or runs.

hzeller:
Ok, but it sounds like it is essentially a caching infrastructure to make things more efficient over time. That means, making it thread-local and have each thread come to their own conclusion should not violate any correctness constraints.

If cache-sharing results in such lock-contention then better have each thread use a bit more CPU to fill the datastructure by itself sounds beneficial.

@parrt @sharwell Does this sound correct ? Are there situations in which each thread might come to different conclusions that are conflicting ?

ericvergnaud:
There are 2 caveats with this proposal:- redundant computation of predictions - memory consumption But I also don’t think correctness is affected.Envoyé de mon iPhoneLe 30 avr. 2023 à 18:10, Henner Zeller ***@***.***> a écrit :﻿
Ok, but it sounds like it is essentially a caching infrastructure to make things more efficient over time. That means, making it thread-local and have each thread come to their own conclusion should not violate any correctness constraints.
If cache-sharing results in such lock-contention then better have each thread use a bit more CPU to fill the datastructure by itself sounds beneficial.
@parrt @sharwell Does this sound correct ? Are there situations in which each thread might come to different conclusions that are conflicting ?

hzeller:
So if there is no correctness issue, then this LGTM.

The choice comes with more memory consumption and redundant computation, but that is a tradeoff the user can make. And for the single-threaded case, there is no disadvantage.

parrt:
Certainly each thread could come up with her own completely separate copy of the entire DFA. Each DFA should be exactly the same. If they are sharing a DFA, it should still be OK assuming C++ doesn't have caching issues or whatever that Sam mentioned.  Assuming the algorithm is correct, the only way to extend a DFA with a new edge will be unique edge/label pair and every thread should come up with that same answer. So it's really just a matter of race conditions. I'm not qualified to say whether removing the specific locks is a problem however.

Let me ping @mike-lischke.  Any thoughts on this lock removal?

hzeller:
You can't remove locks in C++, the outcome is undefined.

But the current solution is not modifying locks, it just makes sure every thread has its own copy and generates its own datastructure. A bit more CPU intensive overall, but now there are no locks to compete for.

mike-lischke:
Not sharing the DFA means each thread has to do the warmup phase on its own. What exactly does this approach improve? Keep in mind that the locks are only used while building up the DFA. It doesn't affect normal runtime behavior once this is done (well, actually, if new input comes up that wasn't parsed before there can still be modifications, but this has a very low overhead, compared to the initial warmup). Additionally, making the DFA thread local can have a serious memory impact.

To me the attempts to improve that part of the code seem to be efforts spent in the wrong spots. It's not the (very few) locks there that are only used while building up the DFA (whose impact could be lowered by running a parser in a single thread and prime it with typical input, until the DFA no longer grows significantly and **after** that start the other threads). Instead work should be spent in the `adaptivePredict` part, which has a much higher overall impact. First convert the recursion to iterations to avoid deep call stacks (which can also become a serious problem in threads, where you cannot set a bigger stack size) and then remove the need of `std::shared_ptr`. Manage the pointers in a different way and work only with raw pointers in this hot path. Maybe an own shared pointer class would be useful here, which does not have to be thread safe and hence doesn't use locks?

mike-lischke:
> One thing that would be a potential problem on C++ but not on Java/C# is the latter allows us to compute a value multiple times and then quietly discard all but one of them. The C++ target would be responsible for determining which one among multiple values actually ended up in the cache, and releasing the memory from all the remaining ones.

@sharwell, do you refer to the fact that sometimes references are replaced (e.g. during optimization/merge runs in `PredictionContext`)? I think this is handled properly by the shared pointers (and is one of the main reasons why shared pointers are used at all).

wangtao9:
@mike-lischke 
> Keep in mind that the locks are only used while building up the DFA. It doesn't affect normal runtime behavior once this is done

Most of what you say is correct, but I think this key point is not quite right. The locks are not "only used while building up the DFA", after the built is done, the read lock is used for EVERY read, resulting in poor concurrent performance of the C++ target (the speedup ratio of 32 threads is less than 2).
That's why I submitted this PR.

wangtao9:
@parrt 
I measured one warmup plus many parses. Although not specific measured, warmup should be time-consuming. But warmup is a one-time job (whether it is all threads call once(before this optimization), or each thread call once(after this optimization)), as @mike-lischke mentioned, the subsequent incremental warmup "has a very low overhead, compared to the initial warmup".

I agree with @hzeller, this is a tradeoff, and the cost is more memory consumption and redundant computation.

wangtao9:
> So if there is no correctness issue, then this LGTM.
> 
> The choice comes with more memory consumption and redundant computation, but that is a tradeoff the user can make. And for the single-threaded case, there is no disadvantage.

@hzeller Totally agree, exactly what I meant.

wangtao9:
> Also, please add information about the new directive to the doc ([cpp-target.md](https://github.com/antlr/antlr4/blob/master/doc/cpp-target.md)).

@KvanTTT Done.

mike-lischke:
@wangtao9 OK, thanks for taking care to create such a patch, after analysing the code thoroughly. I also believe that the C++ could benefit very much from removing locks (including shared_ptr). So I'm fine with your PR.

doc/cpp-target.md
```diff
@@ -102,6 +102,11 @@ In order to create a static lib in Visual Studio define the `ANTLR4CPP_STATIC` m
 
 For gcc and clang it is possible to use the `-fvisibility=hidden` setting to hide all symbols except those that are made default-visible (which has been defined for all public classes in the runtime).
 
+### Compile Aspects
+When compiling generated files, you can configure a compile option according to your needs (also optional):
+
+* A **thread local DFA macro**: Add `-DANTLR4_USE_THREAD_LOCAL_CACHE=1` to the compilation options will enable using thread local DAF cache (disabled by default), after that, each thread uses its own DFA. This will increase memory usage to store thread local DFAs and redundant computation to build thread local DFAs (not too much). The benefit is that it can improve the concurrent performance running with multiple threads. In other words, when you find your concurent throughput is not enough, you should consider turning on this option.
```
mike-lischke, line 108: typo: "DAF cache"

mike-lischke (approved):
Just fix that little typo.

doc/cpp-target.md
wangtao9: fixed

mike-lischke (approved):

wangtao9:
@mike-lischke Thanks for your review. Like you I also think that removing locks is worth doing, maybe it can be put on the agenda in the near future? Until this big action is done,  C++ runtime users can get similar benefits from this optimization :D

doc/cpp-target.md
```diff
@@ -102,6 +102,11 @@ In order to create a static lib in Visual Studio define the `ANTLR4CPP_STATIC` m
 
 For gcc and clang it is possible to use the `-fvisibility=hidden` setting to hide all symbols except those that are made default-visible (which has been defined for all public classes in the runtime).
 
+### Compile Aspects
+When compiling generated files, you can configure a compile option according to your needs (also optional):
+
+* A **thread local DFA macro**: Add `-DANTLR4_USE_THREAD_LOCAL_CACHE=1` to the compilation options will enable using thread local DFA cache (disabled by default), after that, each thread uses its own DFA. This will increase memory usage to store thread local DFAs and redundant computation to build thread local DFAs (not too much). The benefit is that it can improve the concurrent performance running with multiple threads. In other words, when you find your concurent throughput is not high enough, you should consider turning on this option.
```
KvanTTT, line 108: Please follows Markdown best practices:

* Use line breaks after headers ([Heading Best Practices](https://www.markdownguide.org/basic-syntax/#heading-best-practices))
* Split long lines on several small ones (not longer than 120 chars). It makes git diffs much more convenient. ([Line Break Best Practices](https://www.markdownguide.org/basic-syntax/#line-break-best-practices)). At least split by sentences.

```suggestion
### Compile Aspects

When compiling generated files, you can configure a compile option according to your needs (also optional):

* A **thread local DFA macro**: Add `-DANTLR4_USE_THREAD_LOCAL_CACHE=1` to the compilation options
will enable using thread local DFA cache (disabled by default), after that, each thread uses its own DFA.
This will increase memory usage to store thread local DFAs and redundant computation to build thread local DFAs (not too much).
The benefit is that it can improve the concurrent performance running with multiple threads.
In other words, when you find your concurent throughput is not high enough, you should consider turning on this option.
```

KvanTTT (commented):
Now it's OK for me but please also fix minor issues with `cpp-target.md`.

KvanTTT (approved):
Thanks a lot!

wangtao9:
Glad to contribute to antlr4 project! :D

parrt:
Thanks, everyone, especially @wangtao9 !

taodongl:
@wangtao9 @parrt 
```
static thread_local JSONLexerStaticData *jsonlexerLexerStaticData = nullptr;
```
Does it lead to "memory leak", who responsible to release the memory when thread destroyed?
Similar post about `thread_local` : https://stackoverflow.com/questions/46429861/c-how-to-use-thread-local-to-declare-a-pointer-variable

wangtao9:
@taodongl 
Great you found it! There is indeed a risk of memory leaks when threads exit prematurely. I actually fixed this a few weeks ago, see this commit https://github.com/wangtao9/antlr4/commit/ce6649a0c4aea040920d5c55331d630d8bec98ed. I'll submit a new PR later @parrt.

I solved the problem with `thread_local static std::unique_ptr`, the `unique_ptr` is responsible for managing memory.

Now no memory leaks detected, you can verify it with this link: https://github.com/wangtao9/antlr4-perfopt-test/tree/sanitizer_check

PR #4238, opened 2023-04-18
feat: Adds actual usable @ actions to the go templates

jimidle:
feat: Add usable @ actions to the go code generator

As I implement my own parsers in the Go runtime, it became obvious that the existing @members and @header were implemented incorrectly and in fact were not usable.

I cannot change the use of those existing actions as I have no way to know if people have already used them for something. So this PR adds the following new actions:

```antlr
@structmembers {} // allows insertion of items in to the parser or lexer instance struct
@includes {} // allows specifying additional includes in the generated source code
```

There may be other actions that will prove to be useful down the line.


jimidle:
@parrt  - another one for yah!

PR #4241, opened 2023-04-19
 [JavaScript] publish getNumberOfSyntaxErrors method.

killerall:
Implemented getNumberOfSyntaxErrors for javascript runtime.

### DIFF
--- a/runtime/JavaScript/src/antlr4/Parser.d.ts
+++ b/runtime/JavaScript/src/antlr4/Parser.d.ts
@@ -40,4 +40,5 @@ export declare class Parser extends Recognizer<Token> {
     setTokenStream(input: TokenStream): void;
     notifyErrorListeners(msg: string, offendingToken: Token, err: RecognitionException | undefined): void;
     getCurrentToken(): Token;
+    getNumberOfSyntaxErrors(): number;
 }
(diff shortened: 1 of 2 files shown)


killerall:
> _syntaxErrors

Hello, for 3 reasons:
1-) Compatibility between java and javascript version.

2-) When using Lib with typescript, parser._syntaxErrors, I got this error:

  "property '_syntaxErrors' does not exist on type"
Although at runtime, the value is accessible and correct.

3-) There is actually a long history of using dangling underscores to indicate “private” members of objects in JavaScript (though JavaScript doesn’t have truly private members, this convention served as a warning).

ericvergnaud:
How about renaming _syntaxErrors to syntaxErrorsCount and exposing it in TS ?

killerall:
I can definitely do this PR with this change from _syntaxErrors to syntaxErrorsCount, but wouldn't that break the code of someone already using _syntaxErrors natively(JS)?

ericvergnaud:
You're right. Maybe add it as a "get" fn ?

killerall:
I'm sorry, but I'm not a javascript expert. Your suggestion is to do something like:

```
get syntaxErrorsCount() {
        return this._syntaxErrors;
    }
```

But I didn't find similar construction in Parser.js

ericvergnaud:
Indeed you're not seeing much of this because when I started the project 15 years ago it wasn't supported. But that's the way to go nowadays for simple getters.

killerall:
New PR with requested changes https://github.com/antlr/antlr4/pull/4261

PR #4248, opened 2023-04-27
OpenBSD also uses clang.

3405691582:
There is probably a nicer way to do this, but this at least gets projects that depend on this building.


3405691582:
I'm going to submit this pr again atop the dev branch; I can't figure out how to get github to recognize that I rebased this.

Issue #4078, opened 2023-01-16
Insertion of new CommonToken(int type) in parse tree causes crash in toStringTree()


kaby76:
This is a problem with the [grammars-v4/php](https://github.com/antlr/grammars-v4/tree/56dcd3d9251506d7514a68e86a4c192735c48f59/php) grammar, but in reality, it is a problem with the Antlr4 API since the API allows you to create a [CommonToken(int type) with a null text field](https://github.com/antlr/antlr4/blob/76fa05c21b12b96a6c12a0a82e611ed9d87d5af4/runtime/Java/src/org/antlr/v4/runtime/CommonToken.java#L87) and put it in a parse tree by returning the token via an overridden [nextToken()](https://github.com/antlr/antlr4/blob/76fa05c21b12b96a6c12a0a82e611ed9d87d5af4/runtime/Java/src/org/antlr/v4/runtime/Lexer.java#L112), which is what was done with this grammar.

When [toStringTree()](https://github.com/antlr/antlr4/blob/76fa05c21b12b96a6c12a0a82e611ed9d87d5af4/runtime/Java/src/org/antlr/v4/runtime/tree/Trees.java#L48) is called, the text for the token is attempted to be computed since it's a null value. But, instead, it crashes.
```
Exception in thread "main" java.lang.NullPointerException
        at org.antlr.v4.runtime.misc.Utils.escapeWhitespace(Utils.java:63)
        at org.antlr.v4.runtime.tree.Trees.toStringTree(Trees.java:49)
        at org.antlr.v4.runtime.tree.Trees.toStringTree(Trees.java:58)
        at org.antlr.v4.runtime.tree.Trees.toStringTree(Trees.java:58)
        at org.antlr.v4.runtime.tree.Trees.toStringTree(Trees.java:58)
        at org.antlr.v4.runtime.tree.Trees.toStringTree(Trees.java:58)
        at org.antlr.v4.runtime.tree.Trees.toStringTree(Trees.java:58)
        at org.antlr.v4.runtime.tree.Trees.toStringTree(Trees.java:42)
        at org.antlr.v4.runtime.RuleContext.toStringTree(RuleContext.java:189)
        at Test.DoParse(Test.java:206)
        at Test.ParseFilename(Test.java:135)
        at Test.main(Test.java:107)
```
Surprising, if we as to perform "token.toString()", we get a string with "<no text>" as the text.

[...]

PR #4251, opened 2023-04-30
#4078 [runtime/java] fixed npe due to CommonToken.getText() returning null when inputStream is null

vityaman:
Solution of Issue 4078: https://github.com/antlr/antlr4/issues/4078

The problem was that CommonToken::getText() returns null if and only if it has InputStream field null, what happens when we call CommonToken(int type) as there source is equal to 
```
protected static final Pair<TokenSource, CharStream> EMPTY_SOURCE =
    new Pair<TokenSource, CharStream>(null, null);
```

```
// CommonToken.java

@Override
public String getText() {
	if ( text!=null ) {
		return text;
	}
	CharStream input = getInputStream();
	if ( input==null ) return null; // <-- oops
	int n = input.size();
	if ( start<n && stop<n) {
		return input.getText(Interval.of(start,stop));
	}
	else {
		return "<EOF>";
	}
}
// ...
public String toString(Recognizer<?, ?> r) {
	String channelStr = "";
	if ( channel>0 ) {
		channelStr=",channel="+channel;
	}
	String txt = getText();
	if ( txt!=null ) { // <-- this is what saved us in toString
		txt = txt.replace("\n","\\n");
		txt = txt.replace("\r","\\r");
		txt = txt.replace("\t","\\t");
	}
	else {
		txt = "<no text>";
	}
	String typeString = String.valueOf(type);
	if ( r!=null ) {
		typeString = r.getVocabulary().getDisplayName(type);
	}
	return "[@"+getTokenIndex()+","+start+":"+stop+"='"+txt+"'<"+typeString+">"+channelStr+","+line+":"+getCharPositionInLine()+"]";
}
```

### DIFF
--- a/runtime/CSharp/src/Atn/LexerATNSimulator.cs
+++ b/runtime/CSharp/src/Atn/LexerATNSimulator.cs
@@ -779,7 +779,7 @@ namespace Antlr4.Runtime.Atn
 		}
 
 
-		public void Consume(ICharStream input)
+		public virtual void Consume(ICharStream input)
 		{
 			int curChar = input.LA(1);
 			if (curChar == '\n')
--- a/runtime/Java/src/org/antlr/v4/runtime/misc/Utils.java
+++ b/runtime/Java/src/org/antlr/v4/runtime/misc/Utils.java
@@ -61,10 +61,10 @@ public class Utils {
 	public static String escapeWhitespace(String s, boolean escapeSpaces) {
 		StringBuilder buf = new StringBuilder();
 		for (char c : s.toCharArray()) {
-			if ( c==' ' && escapeSpaces ) buf.append('\u00B7');
-			else if ( c=='\t' ) buf.append("\\t");
-			else if ( c=='\n' ) buf.append("\\n");
-			else if ( c=='\r' ) buf.append("\\r");
+			if (c == ' ' && escapeSpaces) buf.append('\u00B7');
+			else if (c == '\t') buf.append("\\t");
+			else if (c == '\n') buf.append("\\n");
+			else if (c == '\r') buf.append("\\r");
 			else buf.append(c);
 		}
 		return buf.toString();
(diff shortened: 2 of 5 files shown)


runtime/CSharp/src/Atn/LexerATNSimulator.cs
```diff
@@ -779,7 +779,7 @@ public int Column
 		}
 
 
-		public void Consume(ICharStream input)
+		public virtual void Consume(ICharStream input)
```
KvanTTT, line 782: It relates neither to PR nor to the Java target.

runtime/Java/src/org/antlr/v4/runtime/misc/Utils.java
```diff
@@ -61,10 +61,10 @@ public  static <T> void removeAllElements(Collection<T> data, T value) {
...
 		StringBuilder buf = new StringBuilder();
 		for (char c : s.toCharArray()) {
-			if ( c==' ' && escapeSpaces ) buf.append('\u00B7');
-			else if ( c=='\t' ) buf.append("\\t");
-			else if ( c=='\n' ) buf.append("\\n");
-			else if ( c=='\r' ) buf.append("\\r");
+			if (c == ' ' && escapeSpaces) buf.append('\u00B7');
+			else if (c == '\t') buf.append("\\t");
+			else if (c == '\n') buf.append("\\n");
+			else if (c == '\r') buf.append("\\r");
```
KvanTTT, line 67: Please don't mix whitespace and semantics changes (ideally, don't change whitespaces without need).

runtime/Java/src/org/antlr/v4/runtime/tree/TerminalNodeImpl.java
```diff
@@ -60,8 +60,10 @@ public String toStringTree(Parser parser) {
 
 	@Override
 	public String toString() {
-			if ( symbol.getType() == Token.EOF ) return "<EOF>";
-			return symbol.getText();
+		if (symbol.getType() == Token.EOF) {
+			return "<EOF>";
+		}
+		return symbol.getText();
```
KvanTTT, line 66: Also, please revert this change.

KvanTTT (changes requested):

vityaman:
Will close this and resolve issues in PR https://github.com/antlr/antlr4/pull/4252 as here I have problems with commit history and it is easier to solve them by creating a fresh one.

PR #4252, opened 2023-04-30
#4078 [runtime/java] made CommonToken::getText return nonnull

vityaman:
Solution of Issue https://github.com/antlr/antlr4/issues/4078.
Fix of PR https://github.com/antlr/antlr4/pull/4251.

The problem was that `CommonToken::getText()` returns `null` if and only if it has `InputStream` field is `null`, what happens when we call `CommonToken(int type)` as there source is equal to

```java
protected static final Pair<TokenSource, CharStream> EMPTY_SOURCE =
    new Pair<TokenSource, CharStream>(null, null);
```

```java
// CommonToken.java

@Override
public String getText() {
	if ( text!=null ) {
		return text;
	}
	CharStream input = getInputStream();
	if ( input==null ) return null; // <-- oops
	int n = input.size();
	if ( start<n && stop<n) {
		return input.getText(Interval.of(start,stop));
	}
	else {
		return "<EOF>";
	}
}
// ...
public String toString(Recognizer<?, ?> r) {
	String channelStr = "";
	if ( channel>0 ) {
		channelStr=",channel="+channel;
	}
	String txt = getText();
	if ( txt!=null ) { // <-- this is what saved us in toString
		txt = txt.replace("\n","\\n");
		txt = txt.replace("\r","\\r");
		txt = txt.replace("\t","\\t");
	}
	else {
		txt = "<no text>";
	}
	String typeString = String.valueOf(type);
	if ( r!=null ) {
		typeString = r.getVocabulary().getDisplayName(type);
	}
	return "[@"+getTokenIndex()+","+start+":"+stop+"='"+txt+"'<"+typeString+">"+channelStr+","+line+":"+getCharPositionInLine()+"]";
}
```

### DIFF
--- a/runtime/Java/src/org/antlr/v4/runtime/CommonToken.java
+++ b/runtime/Java/src/org/antlr/v4/runtime/CommonToken.java
@@ -7,6 +7,7 @@ package org.antlr.v4.runtime;
 
 import org.antlr.v4.runtime.misc.Interval;
 import org.antlr.v4.runtime.misc.Pair;
+import org.antlr.v4.runtime.misc.Utils;
 
 import java.io.Serializable;
 
@@ -164,7 +165,7 @@ public class CommonToken implements WritableToken, Serializable {
 		}
 
 		CharStream input = getInputStream();
-		if ( input==null ) return null;
+		if ( input==null ) return "<no text>";
 		int n = input.size();
 		if ( start<n && stop<n) {
 			return input.getText(Interval.of(start,stop));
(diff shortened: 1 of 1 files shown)


ericvergnaud:
Hi,
Thanks for this but this would not only be a breaking change for Java but it would also introduce inconsistency across targets.
I'm not sure why having getText return null is seen as a problem when the underlying cause is the explicit creation of tokens without a text.
I'd suggest deriving CommonToken to ensure that getText returns whatever makes sense in the context of its usage.

vityaman:
@ericvergnaud, hm. You are right. 

Therefore I think that we need more info about this issue https://github.com/antlr/antlr4/issues/4078. 

This PR can be closed so.

ericvergnaud:
Closing per author's instructions

PR #4255, opened 2023-05-05
Export generated data structures

jimidle:
feat: Make static data available outside the generated parser code

  - Allows Go code to access things like RuleNames outside the generated code
    instead of non-idiomatic and non-existent Getxxx() funcs


jimidle:
@parrt Ready to merge when you get a chance

PR #4256, opened 2023-05-05
Change closureCheckingStopState from recursive to iterative

jimidle:
feat: closureCheckingStopState is now an iterative function

  - Recursive function retained but will be removed next release
  - Improves stack usage on poor grammars
  - Provides general speed up for all grammars

First of a few potential changes to use iteration instead of recursion


jimidle:
@parrt Another ready for merge please :)

PR #4257, opened 2023-05-05
Go/antl4/interval_set: OutOfRange panic fixes

artemseleznev:
Go/antl4/interval_set: OutOfRange panic fixes


jimidle:
I’ll look at this of course but I think I might have already fixed it. I
have a bunch of changes wired up that are quite big so I have to test the
bejesus out of them. I’ll bring this in if I haven’t already fixed it.

Thanks for your effort!

runtime/Go/antlr/v4/token.go
```diff
@@ -137,7 +137,7 @@ func NewCommonToken(source *TokenSourceCharStreamPair, tokenType, channel, start
 		},
 	}
 
-	if t.source.tokenSource != nil {
+	if t.source != nil && t.source.tokenSource != nil {
```
artemseleznev, line 140: we have call of this function when source is directly `nil`. So I've added this condition

Issue #3418, opened 2021-12-19
C++ runtime demo fails to build


felixn:
In the current state of the master branch (dd623ec), the C++ runtime demo fails to build:
```
$ cd <snip>/runtime/Cpp
$ mkdir build && mkdir run && cd build
$ cmake .. -DANTLR_JAR_LOCATION=<snip>/antlr-4.9.3-complete.jar -DWITH_DEMO=True
$ make
...
[ 97%] Building CXX object demo/CMakeFiles/antlr4-demo.dir/Linux/main.cpp.o
In file included from /home/felix/src/antlr4/runtime/Cpp/demo/Linux/main.cpp:16:
/home/felix/src/antlr4/runtime/Cpp/demo/generated/TLexer.h:57:43: error: ‘virtual const std::vector<std::__cxx11::basic_string<char> >& antlrcpptest::TLexer::getTokenNames() const’ marked ‘override’, but does not override
   57 |   virtual const std::vector<std::string>& getTokenNames() const override; // deprecated, use vocabulary instead
      |                                           ^~~~~~~~~~~~~
In file included from /home/felix/src/antlr4/runtime/Cpp/demo/Linux/main.cpp:17:
/home/felix/src/antlr4/runtime/Cpp/demo/generated/TParser.h:44:43: error: ‘virtual const std::vector<std::__cxx11::basic_string<char> >& antlrcpptest::TParser::getTokenNames() const’ marked ‘override’, but does not override
   44 |   virtual const std::vector<std::string>& getTokenNames() const override { return _tokenNames; }; // deprecated: use vocabulary instead.
      |                                           ^~~~~~~~~~~~~
make[2]: *** [demo/CMakeFiles/antlr4-demo.dir/build.make:97: demo/CMakeFiles/antlr4-demo.dir/Linux/main.cpp.o] Error 1
make[1]: *** [CMakeFiles/Makefile2:414: demo/CMakeFiles/antlr4-demo.dir/all] Error 2
make: *** [Makefile:166: all] Error 2
```

There's also a new warning:
```
[ 12%] Building CXX object runtime/CMakeFiles/antlr4_shared.dir/src/Vocabulary.cpp.o
/home/felix/src/antlr4/runtime/Cpp/runtime/src/Vocabulary.cpp: In static member function ‘static void antlr4::dfa::Vocabulary::__static_initialization_and_destruction_0(int, int)’:
[...]

PR #4267, opened 2023-05-13
cpp runtime: Remove pthread dependency.

ArthurSonzogni:
Hello, this is an old patch merged in February:
https://github.com/antlr/antlr4/pull/4086

But, it got reverted because it merged into `master` instead of `dev`. @parrt asked me to do another PR, but his message get lost.

I coming back today, because I saw two related issues this patch would fix:
- https://github.com/ArthurSonzogni/Diagon/discussions/66
- https://github.com/antlr/antlr4/issues/3418#issuecomment-1451880253

### DIFF
--- a/runtime/Cpp/runtime/CMakeLists.txt
+++ b/runtime/Cpp/runtime/CMakeLists.txt
@@ -39,20 +39,6 @@ if (ANTLR_BUILD_STATIC)
   add_library(antlr4_static STATIC ${libantlrcpp_SRC})
 endif()
 
-if (CMAKE_HOST_UNIX)
-  # Make sure to link against threads (pthreads) library in order to be able to
-  # make use of std::call_once in the code without producing runtime errors
-  # (see also https://github.com/antlr/antlr4/issues/3708 and/or https://stackoverflow.com/q/51584960).
-  find_package(Threads REQUIRED)
-
-  if (TARGET antlr4_shared)
-    target_link_libraries(antlr4_shared Threads::Threads)
-  endif()
-  if (TARGET antlr4_static)
-    target_link_libraries(antlr4_static Threads::Threads)
-  endif()
-endif()
-
 IF(TRACE_ATN)
     ADD_DEFINITIONS(-DTRACE_ATN_SIM=1)
 ENDIF(TRACE_ATN)
--- a/runtime/Cpp/runtime/src/internal/Synchronization.h
+++ b/runtime/Cpp/runtime/src/internal/Synchronization.h
@@ -27,12 +27,13 @@
 
 #include "antlr4-common.h"
 
+#include <atomic>
+#include <functional>
 #include <mutex>
 #include <shared_mutex>
 #include <utility>
 
 #if ANTLR4CPP_USING_ABSEIL
-#include "absl/base/call_once.h"
 #include "absl/base/thread_annotations.h"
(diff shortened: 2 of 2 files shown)


hzeller:
This does not look like it waits until the call is finished. The idea is that call_once() can be called in multiple threads, and once it returns in all of these threads, it is guaranteed that the function has been called exactly once and returned. So this simplified implementation does not behave the same.

Instead of modifying call_once, we should not need this at all by initializing a static pointer; see my comment at

https://github.com/antlr/antlr4/pull/4237#issuecomment-1520837195

ArthurSonzogni:
> This does not look like it waits until the call is finished. The idea is that call_once() can be called in multiple threads, and once it returns in all of these threads, it is guaranteed that the function has been called exactly once and returned. So this simplified implementation does not behave the same.
> 
> Instead of modifying call_once, we should not need this at all by initializing a static pointer; see my comment at
> 
> [#4237 (comment)](https://github.com/antlr/antlr4/pull/4237#issuecomment-1520837195)

Hmmm, I assumed there was no multithreaded implementation of the CPP runtime. I guess it changed recently.

Thanks for the code/comments pointers! I will take a closer look when I am more available.

ArthurSonzogni:
@hzeller I addressed your comment.
I agree static initialization is probably better for ANTLR. However I noticed folks were arguing about it in the PR you linked, so I prefered for now to take the short path and make call_once thread safe instead. I would be happy if call_once is removed in the future of course.

Could you please review the implementation?

hzeller:
Since you're now introducing std::mutex, you are definitely need the pthread library as well, so this won't change things.

I would stay away of re-implementing `std::call_once` to not run into any threads subtleties and performance issues just to save an optional link to a library that probably is anyway linked under the hood. 
I'd keep it at that or entirely remove the need for call_once() with the static initialization.

The argument in the other pull request was mostly that my suggestion would not address what the focus was in the PR, but it is definitely a good step to do.

If you want to do something about not linking pthread unconditionally (I don't quite understand why yet though), then maybe add a test in the cmake file if using pthread is needed for call_once.

ArthurSonzogni:
Okay. I see. Difficult to say when pthread is needed and when it isn't.

I guess I have no choice other addressing your comment about using static initialization. I will try later.

Some projects need not to use pthread. In particular when targeting the web. I am doing it with:
https://arthursonzogni.com/diagon

The web has some additional restrictions when shared memory is used. It is because  you can use spectre attack. It is much easier when no threads are around.

hzeller:
(URL does not work, maybe Upper-case Diagnon ? https://arthursonzogni.com/Diagon/
Also thanks, I just discovered your [ftxui](https://github.com/ArthurSonzogni/FTXUI) library, @ArthurSonzogni , very interesting. I also like the terminal for [many things](https://timg.sh/))

ArthurSonzogni:
There are `antlr4::internal::Mutex` already using pthread. So the idea of removing pthread is hopeless here.

Moreover, my patch doesn't make sense. I tried to understand better why diagon regressed recently. I ran a bisection and found it was `fixed` by: https://github.com/antlr/antlr4/commit/f55d2513181d60aeef46aab61fc8ebf50281a733

I previously did it, but "inverted" the bad/good and concluded this was the culprit instead of the fix.

PR #4279, opened 2023-05-23
fix: #4128 - Prevent generating escape sequences when Windows directories start with  a 

### LINKED ISSUE #4128: ANTLR 4.12.0 may produce Java sources with illegal unicode escapes under Windows if a package name begins with 'u'
This is a regression, as 4.11.1 works correctly.

For files with a package starting in `u`, for instance `it.unibo.xxx`, when used under Windows, ANTLR 4.12.0 produces a header comment in the form:

```
// Generated from it\unibo\alchemist\model\internal\biochemistry\dsl\Biochemistrydsl.g4 by ANTLR 4.12.0
```

This is problematic, however, as the Java compiler reports it as `error: illegal unicode escape`

Here is a gradle scan showing the issue:
https://scans.gradle.com/s/uxdeejknwdmve/console-log?task=:alchemist-incarnation-biochemistry:compileJava

Quick reproduction can be achieved with the following:
```
git clone https://github.com/AlchemistSimulator/Alchemist.git
cd Alchemist
git checkout renovate/antlr4
./gradlew :alchemist-incarnation-biochemistry:compileJava
```

The fix should be reasonably easy; I'd suggest converting backslashes to slashes in the header (also, this would generate the same comment code for all the platforms).

jimidle:
When code generating for Java on Windows, a grammar file may be contained in a subdirectory that starts with a 'u'. Because the generated file is comment with the source file that generated it, and windows path escaping is not used,
then the comment  looks like this in Java:

```java
// Generated from .\users\jim.g by ANTLR 4.13.0
```
 The \u is seen by the Java lexer as an invalid escape sequence because Java processes the whole translation unit through pre-processing and lexing.
 
So, here, just for the template generating the comment, we use Unix style delimiters instead of the backslashes.

NB: I have no Windows box, so I will have to let the GitHub triggers ensure that Windows builds.

Fixes #4128 
Closes #4128

### DIFF
--- a/tool/src/org/antlr/v4/codegen/model/OutputFile.java
+++ b/tool/src/org/antlr/v4/codegen/model/OutputFile.java
@@ -26,7 +26,7 @@ public abstract class OutputFile extends OutputModelObject {
         super(factory);
         this.fileName = fileName;
         Grammar g = factory.getGrammar();
-		grammarFileName = g.fileName;
+		grammarFileName = g.fileName.replace("\\", "/"); // Prevent a path with windows delim and u breaking Java pre-parser on comments
 		ANTLRVersion = Tool.VERSION;
         TokenLabelType = g.getOptionString("TokenLabelType");
         InputSymbolType = TokenLabelType;


tool/src/org/antlr/v4/codegen/model/OutputFile.java
```diff
@@ -26,7 +26,7 @@ public OutputFile(OutputModelFactory factory, String fileName) {
         super(factory);
         this.fileName = fileName;
         Grammar g = factory.getGrammar();
-		grammarFileName = g.fileName;
+		grammarFileName = g.fileName.replace("\\", "/"); // Prevent a path with windows delim and u breaking Java pre-parser on comments
```
KvanTTT, line 29: As I understand, this bug is only relevant to Java target. Is it possible to fix the path only for Java target on Windows platform?

KvanTTT (changes requested):
Please also add a unit test that covers suggested changes.

jimidle:
I don’t see the point of fixing it only for Java. Windows supports the
syntax and it would be a total pain, but…

…it changes a generated comment!

So no test as I would have to embed the test somewhere, somehow anyway and
the test running code is fragile.

So I won’t make these changes for the sake of one generated comment that
affects only a few people using Java.

@parrt - can we merge this?

parrt:
`.\users` is hilarious. I can see how that windows path messes with unicode...  I'm ok with the switch.  I wonder if this will trigger a bunch of rebuilds or cause "incompatibilities".  Any thoughts on weird cases like that?

jimidle:
I’m always a bit wary about generated comments because of that. But the
comment does not put a time stamp on it.

It will show a diff because of the comment, but if you don’t regenerate,
then the generated code won’t change. I thought about having a poll to see
if we should just take the comment out, but I can see that it might be
useful, and I don’t have time to get involved with the ensuing 8 days of
bike shedding !

Java users will be using Maven usually, which tends to rebuild the world
anyway.

It’s six of one and a half dozen of the other I think.

KvanTTT:
> …it changes a generated comment!

Yes, but it still breaks compilation, and some users have encoutered this bug.

> So no test as I would have to embed the test somewhere, somehow anyway and
the test running code is fragile.

The more tests we have the less bugs we'll encounter in the future. Also, current ANTLR test coverege is quite poor.

Please just add the file `uStartingCharInGrammarDoesntBreakCompilation.txt` to [ParserExec](https://github.com/antlr/antlr4/tree/master/runtime-testsuite/resources/org/antlr/v4/test/runtime/descriptors/ParserExec) with the following content:

```
[type]
Parser

[grammar]
grammar u;
u : 'u' { <writeln("$text")> };

[start]
u

[input]
u

[output]
"""u
"""
```

This test covers all platforms and targets and currently it's failing.

KvanTTT:
Done: https://github.com/antlr/antlr4/pull/4295

Issue #4229, opened 2023-04-14
Runtime TypeScript tests performance should be imporved 


KvanTTT:
Currently, it runs very slow, ~15 times slower than JS.

It looks like it copies antlr runtime files for every test. I've tried to fixed it but failed, I'm not very experience in TS.

ericvergnaud:
I tried too and failed too.There are 2 culprits: - compiling TS to JS, this step is unavoidable and largely explains the slowness compared to JS - npm install, this step I tried to avoid repeating by linking the node_modules directory but for some reason this fails randomly but frequently enough to discourage me at the time (I was focusing on getting the TS runtime out)Envoyé de mon iPhoneLe 14 avr. 2023 à 20:52, Ivan Kochurkin ***@***.***> a écrit :﻿
Currently, it runs very slow, ~15 times slower than JS.
It looks like it copies antlr runtime files for every test. I've tried to fixed it but failed, I'm not very experience in TS.

KvanTTT:
> I tried to avoid repeating by linking the node_modules directory but for some reason this fails randomly but frequently enough to discourage me at the time

Do you have a branch with this improvements? I could take a look.

PR #4281, opened 2023-05-23
fix: Speed up  test suites

jimidle:
- Upgrades surefire
  - Limits amount of parallelism to stop swamping the test machines.

Because the test suite does not strictly follow the jupiter config, which was configured to use as many cores as possible, the machine would end up swamped and either IO bound, or just defeating any CPU cache advantages.

The change to limit parallelism to 3 means that everything goes faster, including the Java tests, which do not need to create external processes.

This took some working out.

Please note that the formatting in the pom was out of whack, hence there looks to be more changes to the pom than there are. Compare with whitespace ignored because the formatting needs to be done. 

I will go through the maven stuff next and see if we can't improve it. I don't think anyone has looked at the poms in quite some time.

### DIFF
--- a/runtime-testsuite/pom.xml
+++ b/runtime-testsuite/pom.xml
@@ -4,114 +4,124 @@
   ~ can be found in the LICENSE.txt file in the project root.
   -->
 
-<project xmlns="http://maven.apache.org/POM/4.0.0" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/maven-v4_0_0.xsd">
+<project xmlns="http://maven.apache.org/POM/4.0.0" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+         xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/maven-v4_0_0.xsd">
 
-	<modelVersion>4.0.0</modelVersion>
-	<parent>
-		<groupId>org.antlr</groupId>
-		<artifactId>antlr4-master</artifactId>
-		<version>4.13.1-SNAPSHOT</version>
-	</parent>
-	<artifactId>antlr4-runtime-testsuite</artifactId>
-	<name>ANTLR 4 Runtime Tests (4th generation)</name>
-	<description>A collection of tests for ANTLR 4 Runtime libraries.</description>
+    <modelVersion>4.0.0</modelVersion>
+    <parent>
+        <groupId>org.antlr</groupId>
+        <artifactId>antlr4-master</artifactId>
+        <version>4.13.1-SNAPSHOT</version>
+    </parent>
+    <artifactId>antlr4-runtime-testsuite</artifactId>
+    <name>ANTLR 4 Runtime Tests (4th generation)</name>
+    <description>A collection of tests for ANTLR 4 Runtime libraries.</description>
 
-	<prerequisites>
-		<maven>3.8</maven>
-	</prerequisites>
+    <prerequisites>
+        <maven>3.8</maven>
+    </prerequisites>
 
-	<inceptionYear>2009</inceptionYear>
+    <inceptionYear>2009</inceptionYear>
 
-	<properties>
-		<jUnitVersion>5.9.0</jUnitVersion>
-	</properties>
+    <properties>
+        <jUnitVersion>5.9.0</jUnitVersion>
+    </properties>
 
-	<dependencies>
-		<dependency>
-			<groupId>org.antlr</groupId>
-			<artifactId>ST4</artifactId>
-			<version>4.3.4</version>
-			<scope>test</scope>
-		</dependency>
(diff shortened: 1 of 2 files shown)


KvanTTT (changes requested):
Sorry, I completely disagree with this pull request.

I don't see significant performance difference for Java: it's even a bit worse: **~16.7 s** with 3 threads versus **~14.1** with default number of threads.

But I see significant degradation for Dart: **~92 s** vs **~54 s**.

Graphs of CPU load approve my words about the degradation:

| Dart (3) | Dart (default) |
|-|-|
| Dart with 3 threads | Dart with default number of threads |

Probably you measured performance in incorrect way.

Did you measure using IntelliJ IDEA? It looks like it aggregates spent time for all threads but not calculate the time of execution.

You can try using measuring using console, it prints relevant time:

```
mvn -Dtest='dart.**' test
```

I don't think that personal measurement should be considered a lot because I don't have problems with default number of threads on my local machine. In this way, restricting of number of threads looks weird.

Also, the problem with parallelism is specific to the Go runtime, that's why it should be fixed in Go not but not globally.

jimidle:
It is because I overestimated the number of threads that the Github runners
can handle, hence it is more swamped.

The performance is not just Go, it is pretty much anything that isn't Java
as most everything else has to spark a process. BUt it definitely needs
more work. First I wanted to get some numbers from the Github workers and
cycle back around.

The runners etc are really convoluted and over complicated. By the time we
know we are running Go tests, I think that it is too late. Basically the
build machine gets swamped by anything other than Java tests. If there is
nothing that can be done easily, then I am just going to leave it as it is.
Go takes about 2 minutes. Cpp takes an eternity.

But there's no need to be so emphatic about your disagreement mate ;)

KvanTTT:
> The runners etc are really convoluted and over complicated.

Do you have ideas how they can be simplified? I've tried but it looks like I've done everything I could.

> Cpp takes an eternity.

Yes, it's know fact. But it's unlikely can be fixed because C++ compiler is slow itself.

> But there's no need to be so emphatic about your disagreement mate ;)

Sorry for the exaggeration.

jimidle:
> > The performance is not just Go, it is pretty much anything that isn't Java
> > as most everything else has to spark a process.
> 
> I don't observe that, my Windows 10 scheduler efficiently handles such load.

When you run the go tests (it also happens on my system with other things) does it not swamp your system? It might be a behavioral difference on systems. My system has 12 cores, and when I run tests other than Java, it sparks up about 3x that number of processes at once. The system is fine handling it, but anything else I am trying to use at the same time becomes unusable. Basically any of the test suites that require external processes. 

> 
> > The runners etc are really convoluted and over complicated.
> 
> Do you have ideas how they can be simplified? I've tried but it looks like I've done everything I could.

I think that we would have to start again. I revamped the go ones to make them as fast as the external process can run, and to only need one external process, but that's a far as it can go. But, I don't think that it is really broken enough to redo it to be honest. The main thing is that it is a Java oriented system and we have had to shoe-horn in the other systems - so it's never going to be great for the other runtimes - but maybe it is fine and I am just annoyed by the machine swamp on OSX, when developing locally.

We could change it to each target implementing the tests in their idiomatic way. But I really don't see any advantage to that. And the ways to speed things up end up meaning they would not be individual unit tests.

> 
> > Go takes about 2 minutes.

Same for me on Mac. But any other process is swamped. Two minutes is an eternity when you are waiting, such as waiting for the referee to blow the final whistle.

> 
> It's quite fast, I don't see a reason to improve Go runner while TypeScript runner works much slower and it can be improved (see #4229).

It is as fast as I can make it, but it is faster on my system all around with process limiting. It doesn't really matter how long it takes within reason, but the swamping of the machines is annoying. This PR failed because pegging it to 3 (it's actually more like 9 when you set that setting to 3), is more than the github runners can handle.
> 
> > Cpp takes an eternity.
> 
> Yes, it's know fact. But it's unlikely can be fixed because C++ compiler is slow itself.

Yes. 
> 
> > But there's no need to be so emphatic about your disagreement mate ;)
> 
> Sorry for the exaggeration.

Don't worry, I am not offended. You made me chuckle.

I think I will back out of that bit of this PR and leave it as it is for now. Actually I think I will just close it as the pom changes are best done in  my Maven update PR, which is yet to be submitted.

KvanTTT:
> When you run the go tests (it also happens on my system with other things) does it not swamp your system?

Maybe quite a bit but it doesn't prevent me from working and I can wait. Also, I run full set of tests quite rarely.

Anyway, fixed number of thread numbers is not efficient (`junit.jupiter.execution.parallel.config.strategy=fixed`). I have 16 cores on my Windows 10 and many cores on my Mac M2 and most part of them are idling if only 3 threads are used.

It's better to use  `dynamic` thread strategy with `0.9` load factor (see details in the official [JUnit doc](https://junit.org/junit5/docs/snapshot/user-guide/#writing-tests-parallel-execution-config-properties). Maybe other settings also can help. Anyway, I don't this strategy is better compare to the max load. If you system is being swamped, you can change settings locally.

PR #4283, opened 2023-05-24
Update dotnet to the latest version in runtime tests (6 -> 7)

KvanTTT:
Surprisingly, the test `LargeLexer` is freezing on dotnet 7 (local machine). But it works fine if I change version of `.csproj` to 7. It looks like a bug in dotnet (there is something wrong with back compatibility) but it can be fixed by version upgrading.


KvanTTT:
@parrt it's ready for merging.

PR #4292, opened 2023-05-29
fix go-target.md, remove duplicate 'get'

css521:
fix go-target.md。

Remove duplicate 'get'

`go get -u get github.com/antlr4-go/antlr -> go get -u github.com/antlr4-go/antlr`


jimidle:
Good catch. This should be merged @parrt

PR #4294, opened 2023-05-31
Replace experimental dependency on golang.org/x/exp

TristonianJones:
The `golang.org/x/exp` module does not make any guarantees about API or functionality stability; however, the functions being used [`slices.Equal`](https://pkg.go.dev/golang.org/x/exp/slices#Equal) and [`slices.EqualFunc`](https://pkg.go.dev/golang.org/x/exp/slices#EqualFunc) are relatively straight-forward. Equivalent functionality has been reproduced in the contexts where these functions were used.

This change means that ANTLR will no longer have any go.mod dependencies which might otherwise affect minimum version selection (MVS) on an unstable API surface.

### DIFF
--- a/.github/workflows/hosted.yml
+++ b/.github/workflows/hosted.yml
@@ -257,7 +257,7 @@ jobs:
       if: matrix.target == 'csharp'
       uses: actions/setup-dotnet@v3.0.3
       with:
-        dotnet-version: '6.0.x'
+        dotnet-version: '7.0.x'
 
     - name: Setup Dart 2.12.1
       if: matrix.target == 'dart'
--- a/runtime-testsuite/resources/org/antlr/v4/test/runtime/helpers/Antlr4.Test.csproj.stg
+++ b/runtime-testsuite/resources/org/antlr/v4/test/runtime/helpers/Antlr4.Test.csproj.stg
@@ -1,7 +1,7 @@
 ﻿\<Project Sdk="Microsoft.NET.Sdk">
 
   \<PropertyGroup>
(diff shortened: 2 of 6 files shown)


jimidle:
I have already done this and used explicit functions. But that didn’t make
it in to .13 because I forgot. It is in a separate performance branch that
I have not put in as a PR yet.

TristonianJones:
@jimidle Any objections to submitting this narrowly scoped PR? Or, rather do you have an ETA for when you expect to produce a PR? 

Thanks,

-Tristan

jimidle:
Well, let me see what you have done vs what I have done. I will look today.
When I merge my PR it will likely then have a merge conflict etc. I will
get back to you. Thanks for your effort.

It’s worth raising an issue before doing work, so we all know it’s stuff we
need.

cbandy:
This would help me out as well. How can I help to get this small change merged?

jimidle:
Sorry for the delay in bringing in these minor changes, but I have been a little busy.

jimidle:
I did not bring that in to 4.13.1 as I have been too busy to get to it. If you wish to resubmit this, after rebase then I can bring this in. I would like to keep slices.Sort though. It is still in the exp package but it is obviously baked now. I do not know when that will make it out of exp but I am sure it will. I know the interface is not guaranteed, but I highly doubt that it will change.

cbandy:
Go 1.21 added slices.Sort to the stdlib; https://go.dev/doc/go1.21#slices

cbandy:
https://github.com/antlr/antlr4/pull/4688 uses slices from the stdlib.

Issue #4285, opened 2023-05-26
Antlr 4.13.0 typescript target introduced invalid module configuration


trmjoa:
Hi,

This [PR](https://github.com/antlr/antlr4/pull/4217/files#diff-7c99c0b7a2507f0ed32c83f5634f6d281157f04c1391820f3ccb734bbb7652fb) introduces a bug in the npm package definition.

In the `exports` section of the package.json file the file referenced in the `node`/`browser` sections does not exist.

```
  "exports": {
    ".": {
      "node": {
        "types": "src/index.node.d.ts",
        "import": "./dist/antlr4.node.mjs",
        "require": "./dist/antlr4.node.cjs",
        "default": "./dist/antlr4.node.mjs"
      },
      "browser": {
        "types": "src/index.web.d.ts",
        "import": "./dist/antlr4.web.mjs",
        "require": "./dist/antlr4.web.cjs",
        "default": "./dist/antlr4.web.mjs"
      }
    }
```

`src/index.node.d.ts` and `src/index.web.d.ts` does not exist in the shipped package. All files in the `src` directory in the released module is located in `src/antlr4`, and even here there is no combined definition files.

Unless you create a combined file the intention is maybe to point this path at `src/antlr4/index.d.ts`?

ericvergnaud:
mmm.... not sure what happened
@Tarjei400 Any chance you can look into this and submit a PR ?

Tarjei400:
@ericvergnaud I iwll take a look :+1:

Tarjei400:
@trmjoa @ericvergnaud  Seems it came in review and my missunderstanding of types file. I think you are right and we should stick to reference src/antlr4/index.d.ts in type, and remove node/browser override. Could you take a look at pr?

ericvergnaud:
@trmjoa Can you confirm if this change does not introduce other related issues ?

Tarjei400:
For reference, proper solution will be comming with this PR #4297

trmjoa:
Hi. Sorry for the late reply. I have been off-grid for a while.
I have tried locally changing the types pointer as in #4297, which seems to be an improvement. vscode/typescript compiler manages to resolve the file (as we would expect), but vscode is unable to resolved the named imports. 

Example, when you click a named export in your generated parser, vscode will not manage to look the types for that export, something it usually does. Anyhow, #4297 fixes that its at least is a valid path.

PR #4288, opened 2023-05-28
Removed non existing type declaration file references fixes #4285

Tarjei400:
As stated in mentioned ticket #4285 type declaration files don't exist. As there only exists index.d.ts, and src folder is shipped in then bundle it seems we can reference this file as root file

PR #4297, opened 2023-06-01
Fixed file paths used in `package.json#exports.types`

Andarist:
fixes https://github.com/antlr/antlr4/issues/4285

### DIFF
--- a/runtime/JavaScript/package.json
+++ b/runtime/JavaScript/package.json
@@ -51,13 +51,13 @@
   "exports": {
     ".": {
       "node": {
-        "types": "src/index.node.d.ts",
+        "types": "./src/antlr4/index.node.d.ts",
         "import": "./dist/antlr4.node.mjs",
         "require": "./dist/antlr4.node.cjs",
         "default": "./dist/antlr4.node.mjs"
       },
       "browser": {
-        "types": "src/index.web.d.ts",
+        "types": "./src/antlr4/index.web.d.ts",
         "import": "./dist/antlr4.web.mjs",
         "require": "./dist/antlr4.web.cjs",
         "default": "./dist/antlr4.web.mjs"
--- a/runtime/JavaScript/package.json
+++ b/runtime/JavaScript/package.json
@@ -51,13 +51,13 @@
   "exports": {
     ".": {
       "node": {
-        "types": "./src/antlr4/index.node.d.ts",
+        "types": "./src/antlr4/index.d.ts",
         "import": "./dist/antlr4.node.mjs",
         "require": "./dist/antlr4.node.cjs",
         "default": "./dist/antlr4.node.mjs"
       },
       "browser": {
-        "types": "./src/antlr4/index.web.d.ts",
+        "types": "./src/antlr4/index.d.ts",
         "import": "./dist/antlr4.web.mjs",
         "require": "./dist/antlr4.web.cjs",
         "default": "./dist/antlr4.web.mjs"


ericvergnaud:
Hi, thanks for this but the proposed file paths don't exist either...
This is a duplicate of #4288

Andarist:
Oh, you are totally right 🙈 I was on a PR spree through a list of affected repos and didn't recheck this one properly. I did recheck that you have `.node.` and `.web.` files in `src` but I didn't notice that those are just `.js` files (and not declaration files).

runtime/JavaScript/package.json
```diff
@@ -51,13 +51,13 @@
   "exports": {
     ".": {
       "node": {
-        "types": "src/index.node.d.ts",
+        "types": "./src/antlr4/index.node.d.ts",
```
Andarist, line 54:
```suggestion
        "types": "./src/antlr4/index.d.ts",
```

runtime/JavaScript/package.json
```diff
@@ -51,13 +51,13 @@
...
         "import": "./dist/antlr4.node.mjs",
         "require": "./dist/antlr4.node.cjs",
         "default": "./dist/antlr4.node.mjs"
       },
       "browser": {
-        "types": "src/index.web.d.ts",
+        "types": "./src/antlr4/index.web.d.ts",
```
Andarist, line 60:
```suggestion
        "types": "./src/antlr4/index.d.ts",
```

Andarist:
I fixed the paths now, sorry for that! Note that this still has some subtle issues but they are not easily solvable and if anything... I'd prefer to solve them in separate PRs.

runtime/JavaScript/package.json
```diff
@@ -51,13 +51,13 @@
...
       "node": {
-        "types": "src/index.node.d.ts",
+        "types": "./src/antlr4/index.d.ts",
         "import": "./dist/antlr4.node.mjs",
         "require": "./dist/antlr4.node.cjs",
         "default": "./dist/antlr4.node.mjs"
       },
       "browser": {
-        "types": "src/index.web.d.ts",
+        "types": "./src/antlr4/index.d.ts",
```
Andarist, line 60: Note that this is a small lie but one that isn't new. It was already here through the top-level `types`:
https://github.com/antlr/antlr4/blob/faef88016247cf4d29d968df806481206ecde36d/runtime/JavaScript/package.json#L8

The lie is that at runtime the browser entrypoint **doesn't** expose `FileStream` export but the TS thinks that it does. Note that I advise against diverging interfaces between environments because it gets really messy when a package gets used transitively etc. It's much easier to just have the same interface/exports list regardless of the environment etc.

runtime/JavaScript/package.json
```diff
@@ -51,13 +51,13 @@
   "exports": {
     ".": {
       "node": {
-        "types": "src/index.node.d.ts",
+        "types": "./src/antlr4/index.d.ts",
         "import": "./dist/antlr4.node.mjs",
         "require": "./dist/antlr4.node.cjs",
```
Andarist, line 56: Using `types` like this isn't exactly correct - but many packages do it and usually you can "get away with it". There are subtle issues with it though! When checking the package with https://arethetypeswrong.github.io (a site created by the TS team member who is responsible for module resolution) then you might see "Masquerading as CJS" problem being flagged on the package (not now though, but rather with this fix in place - now you have different problems reported there). 

I believe though that "Masquerading as CJS" problem should be fixed separately - or not at all. My preference is to fix it but it might complicate your setup so I'm not sure if your team would actually consider fixing it as something that you want to do.

ericvergnaud: If you ask me, JS itself masquerades common sense, and TS only partially fixes the problem
:-)
That said we definitely want small dedicated PRs, this is critical to understanding side effects.
Re this one, why do we even need a "types" entry in the "exports" section ? Is the top-level "types" not sufficient for tsc and ts tools (I haven't explored that tbh) ?

Andarist: There are 2 main ideas behind `package.json#exports`:
- list what you allow to be consumed from the package
- allow "selective" resolution based on conditions

So in node and other runtimes, once you add `package.json#exports` you essentially "hide" everything that is not listed. This basically works like that in all runtimes capable of resolving `package.json#exports` (it's a somewhat new addition, first added in node 12). So even the old `package.json#main` field has no effect in a package with `exports` (`exports` just take priority over it)

So it only makes sense that you have to also list types here (if you don't rely on the "sibling lookup" that I've mentioned before in the other comment). However, TS decided **not to enable** `package.json#exports` by default. You need to opt into it. So as long as you are not using the new module resolution options in TS, then your package seems to be working fine - the runtime looks into `package.json#exports`, resolves files according to the manifest there and that's fine. At the same time, `package.json#exports` are ignored by default in TS so it looks into `package.json#types` (using the "classic" node module resolution algorithm).

But once you opt-into the `package.json#exports` resolutions then `package.json#types` become **ignored** in packages with `package.json#exports`. I think that this is the right thing to do (as per the first idea behind `package.json#exports` that i listed. The bizarre part though is that TS decided to not include `package.json#exports` support by default. This cases a lot of confusion because most people stick to `moduleResolution: node` (they might not even know about the newer options), they might add `package.json#exports` **withoout** the `types` condition there and they might even recheck that with their local configuration and can only assume that things work OK since it works OK in their project. I can only presume that TS team decided to do this because otherwise many packages would just break/wouldnt be able to resolve types after consumers would upgrade their TS version (people already shipped `package.json#exports` in packages before TS implemented support for that). I'm not sure if that decision was worth it... but hey, here we are 🤷‍♂️

ericvergnaud: Finally an explanation I can understand :-)

ericvergnaud:
@parrt blessed

jonaskello:
I'm currently blocked by the invalid exports that are fixed by this PR. Thanks for fixing it! :-) Is there any estimate when this will be released?

parrt:
We released fairly recently so it'll be a few months I'd say.

jonaskello:
Since the invalid typing spec is a blocking bug for typescript users, I would suggest it be released in the nearest patch version whenever that is.

Andarist:
[patch-package](https://github.com/ds300/patch-package) is a wonderful tool that might unblock you in the meantime

PR #4298, opened 2023-06-01
Simplify Java runtime tests

KvanTTT:


KvanTTT:
There is one failed check on macos-11 and cpp. But it doesn't relate to my changes.

@parrt it can be reviewed.

parrt:
I didn't scrutinize in great detail, but looks like an improvement. thanks!

PR #4301, opened 2023-06-03
Improve testing infrastructure

KvanTTT:


KvanTTT:
@parrt please review.

parrt:
Hi. What is the goal here? I'm not sure I favor such a massive change w/o a really good reason. thanks!

KvanTTT:
It provides several small goals instead of single big one:

* Add support of separated lexer and parser in descriptors as I noted above:

```
[grammar]
parser grammar Parser;

options { tokenVocab=Lexer; } 
...

[grammar]
lexer grammar Lexer;
...
```

It allows to port [TestParserExec](https://github.com/antlr/antlr4/blob/master/tool-testsuite/test/org/antlr/v4/test/tool/TestParserExec.java#L95) hardcoded tests to descriptors and check all runtimes instead of only Java.

* Siplify runtime test descriptors. No need to use `type` section because grammar type can be detected automatically (`Lexer`, `Parser`, `CompositeLexer`, `CompositeParser`):

```diff
- [type]
- Parser
-
```

* Simplify tool tests. No need to specify grammar file, parser and lexer names because they are detected automatically:

```java
ExecutedState executedState = execParser(grammar, "a", "x", true);
```

instead of

```java
ExecutedState executedState = execParser("T.g4", grammar, "TParser", "TLexer", "a", "x", true);
```

* Add new checks on presence of generated files
* More than `-1000` lines of code. It's worth it

Also, it changes only testing infastructure but not core or runtimes. There are a lot of massive changes to test data because of removing redundant info.

It can be considered as follow-up of [Runtime tests refactoring](https://github.com/antlr/antlr4/pull/3736) and [Runtime tests refactoring 2](https://github.com/antlr/antlr4/pull/3750).

PR #4308, opened 2023-06-09
Update FileStream.js

huang-1234:
add wrap follow if


ericvergnaud:
Hi,
thanks for this, but we don't accept code-style driven changes. 
They vary too much from one developer to another.

Issue #4291, opened 2023-05-29
Some rule names cause runtime errors in Python target


xmcp:
[See this SO post](https://stackoverflow.com/questions/76353461/antlr4-python-target-not-working-with-some-special-rule-names?noredirect=1#comment134640110_76353461)

**Target:** `-Dlanguage=Python3`

**Grammar to reproduce:**

```
grammar Grammar;

program : state* EOF ;
state: 'break;' | 'continue;' | 'return;' ;
```

**Actual bahavior:**

The command `antlr4 -Dlanguage=Python3 Grammar.g4` silently successes, but the generated parser does not work. It will raise the below runtime exception:

```
Traceback (most recent call last):
  ...
  File ".../GrammarParser.py", line 96, in program
    self.state()
TypeError: 'int' object is not callable
```

**Expected behavior:**

Either the generated parser works, or the antlr4 command gives a warning about it.

**Bug cause:**

The rule name `state` conflicts with an internal field used by the generated target, leading to a naming conflict:

```python
class GrammarParser ( Parser ):
    def program(self):
        ...
            ...
                ...
                self.state = 4
                self.state()
                self.state = 9
                ...
```

Other rule names like `match`, `atn` or `consume` also lead to various runtime exceptions that looks confusing especially to beginners:

```
# match:
TypeError: GrammarParser.match() takes 1 positional argument but 2 were given

# atn:
AttributeError: 'function' object has no attribute 'states'

# consume:
RecursionError: maximum recursion depth exceeded while calling a Python object
```

Since that the Python target [already renames keywords](https://github.com/antlr/antlr4/blob/873a01ca2939e623f396d356f9e903b9afd4ca32/tool/src/org/antlr/v4/codegen/target/Python3Target.java#L15), it should also rename internal names like `state` and `match`.

kaby76:
Seeing that this may never be fixed, I have the beginnings of a [Trash](https://github.com/kaby76/Domemtech.Trash) script that finds parser symbol conflicts, at least for Python3. https://github.com/kaby76/g4-checks/blob/main/find-known-conflicts.sh. It does require version 0.20.24 of Trash tools.

kaby76:
> @kaby76 this is not high priority, but it could probably be fixed by adding 'state' and reserved fields to the list of keywords. Feel free to give it a try.

[...]

PR #4312, opened 2023-06-13
name 'state' conflict with Python3 runtime.

cyqw:
This PR fixes https://github.com/antlr/antlr4/issues/4291

This change add keyword tests.
add function names in parser as reserved keywords for targets.
use rule.escapedName instead  rule.name in php.stg


ericvergnaud:
Hi,
thanks for this.
Unfortunately it's very likely that some developers are relying on `ctx.state` (a typical usage scenario is code completion).
The proposed refactoring would break their code. Plus it doesn't solve the problem, rather it moves it (now I can't use _state as a rule name).
I'm afraid the only reasonable solution here is to disallow usage of `state` as a rule name...

KvanTTT (changes requested):
This fix is incorrect. Please add `state` to Python3 and Python2 [reservedWords](https://github.com/antlr/antlr4/blob/master/tool/src/org/antlr/v4/codegen/target/Python3Target.java#L15) list without touching runtime files.

cyqw:
I think add the name of function or attribute of Parser, lexer to reservedWords is not a good idea.
The functions or attributes of parser, lexer should not expose to generated parser, lexer directly by inherit except some functions can be override. 
The parser should be split, one is the interface of generated parser, other one is the implement of basic parser, it is a member of generated parser.

cyqw:
> This fix is incorrect. Please add `state` to Python3 and Python2 [reservedWords](https://github.com/antlr/antlr4/blob/master/tool/src/org/antlr/v4/codegen/target/Python3Target.java#L15) list without touching runtime files.
> 
> Also, it looks like `state` should be added to JavaScript, TypeScript and Dart reservedWords because CI tests are red: https://github.com/antlr/antlr4/actions/runs/5261330620/jobs/9509322855?pr=4312

@KvanTTT please review the new changes,  thanks

KvanTTT:
Sorry, I don't review MR in this repository since I don't have maintainer rights and I can't merge requests.

ericvergnaud:
@parrt blessed

ericvergnaud:
@parrt kind reminder

PR #4313, opened 2023-06-15
Remove stray type annotations from Python2 runtime

agatti:
XPath support still retained Python3 type annotations. This yields a partial installation when using setuptools on both plain Python2

```
byte-compiling build/bdist.linux-x86_64/egg/antlr4/xpath/XPath.py to XPath.pyc
  File "build/bdist.linux-x86_64/egg/antlr4/xpath/XPath.py", line 69
    def __init__(self, parser:Parser, path:str):
                             ^
SyntaxError: invalid syntax
```

or Jython

```
Extracting antlr4_python2_runtime-4.13.0-py2.7.egg to build/jython/Lib/site-packages
SyntaxError: ("mismatched input ':' expecting RPAREN", ('build/jython/Lib/site-packages/antlr4_python2_runtime-4.13.0-py2.7.egg/antlr4/xpath/XPath.py', 69, 29, '    def __init__(self, parser:Parser, path:str):\n'))
```

### DIFF
--- a/runtime/Python2/src/antlr4/xpath/XPath.py
+++ b/runtime/Python2/src/antlr4/xpath/XPath.py
@@ -66,12 +66,12 @@ class XPath(object):
     WILDCARD = "*" # word not operator/separator
     NOT = "!" # word for invert operator
 
-    def __init__(self, parser:Parser, path:str):
+    def __init__(self, parser, path):
         self.parser = parser
         self.path = path
         self.elements = self.split(path)
 
-    def split(self, path:str):
+    def split(self, path):
         input = InputStream(path)
         lexer = XPathLexer(input)
         def recover(self, e):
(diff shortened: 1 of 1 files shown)


ericvergnaud:
@parrt blessed

KvanTTT:
It was merged to the `master` branch, it's strange.

KvanTTT:
I know only about changing default to dev. I don't like that branch should be changed for every MR because it cause error like this.

ericvergnaud:
Ok, I've switched the default branch to 'dev' such that future PRs should be against dev.
@parrt FYI

parrt:
thanks guys...

jimidle:
I think we can create protection rules for the master branch, but I have
not really looked into exactly what they can manage - not had time.

PR #4330, opened 2023-06-25
fix memory leak risk of cpp target

wangtao9:
See https://github.com/antlr/antlr4/pull/4237 for details:

image


wangtao9:
That's it, I've rebased the latest dev branch and all checks passed!

Before this fix, there was a risk of a memory leak when threads exited prematurely, which was also reported by the asan (AddressSanitizer) check. This PR fixes the above problem, you can use this example  to verify the effect: https://github.com/wangtao9/antlr4-perfopt-test/tree/sanitizer_check

Issue #4342, opened 2023-07-02
[Go target] API is missing EmitToken()


kaby76:
I am trying to port [python/python3](https://github.com/antlr/grammars-v4/tree/master/python/python3) to the Go target.

I was able to sort of bypass the issue noted [here](https://github.com/antlr/antlr4/issues/3446). I was also able to get [BaseLexer.NextToken()](https://github.com/antlr/antlr4/blob/ce7f5e73cd4787c5aebdc05c29d4d898d6f2d123/runtime/Go/antlr/v4/lexer.go#L181) to call the "superclass" Python3ParserBase.Emit() to be called by assignment to [Virt](https://github.com/antlr/antlr4/blob/ce7f5e73cd4787c5aebdc05c29d4d898d6f2d123/runtime/Go/antlr/v4/lexer.go#L39). But, now I cannot get past the fact that EmitToken() is not a virtual function as in all other targets. This is really frustrating.

[Emit(token) in C#](https://github.com/antlr/antlr4/blob/ce7f5e73cd4787c5aebdc05c29d4d898d6f2d123/runtime/CSharp/src/Lexer.cs#L314) and [Java](https://github.com/antlr/antlr4/blob/ce7f5e73cd4787c5aebdc05c29d4d898d6f2d123/runtime/Java/src/org/antlr/v4/runtime/Lexer.java#L233), which don't conflict with `Emit()` because the languages implement polymorphism unlike Go, are defined. For JavaScript, which doesn't have polymorphism, it's declared and exported as [emitToken(token)](https://github.com/antlr/antlr4/blob/ce7f5e73cd4787c5aebdc05c29d4d898d6f2d123/runtime/JavaScript/src/antlr4/Lexer.d.ts#L28C8-L28C8). The method is visible and virtualized.

[...]

PR #4343, opened 2023-07-02
Fix for #4342--EmitToken(token) must be exported and virtualized

kaby76:
Required for porting https://github.com/antlr/grammars-v4/tree/master/python/python3 to the Go target. This change just mirrors the APIs for the other targets, like Java, CSharp, and JavaScript/TypeScript.


parrt:
@jimidle ?

Issue #4318, opened 2023-06-19
[typescript] Runtime declares CommonToken(source, .....) wrong.


kaby76:
I need to create a CommonToken for [python3](https://github.com/antlr/grammars-v4/blob/5585fe7e9292a6a3855f2beef36cf75da2b41271/python/python3/JavaScript/Python3LexerBase.js#L59) but for TypeScript. Unfortunately, "source" is declared as a number [here](https://github.com/antlr/antlr4/blob/96f97704a107db99d90ea9503d637104c3bb7afe/runtime/JavaScript/src/antlr4/CommonToken.d.ts#L4). That's wrong. Completely blocked.

kaby76:
Tried and got a couple different things.
```
Python3LexerBase.ts:68:38 - error TS1005: ',' expected.

68         return new CommonToken(source: [ this, this.input_ ], type, 0, start, stop);
                                        ~
```

```
Python3LexerBase.ts:68:33 - error TS2345: Argument of type '(CharStream | this)[]' is not assignable to parameter of type 'number'.

68         return new CommonToken( [ this, this.input_ ], type, 0, start, stop);
                                   ~~~~~~~~~~~~~~~~~~~~~
```

kaby76:
grammars-v4 CI builds use official NPM published packages. 4.13.0 is the latest.

ericvergnaud:
ah.

I guess in the meantime you could write:
new CommonToken( [ this, this.input_ ] as any as number, ... ?

kaby76:
> I guess in the meantime you could write:
> new CommonToken( [ this, this.input_ ] as any as number, ... ?

Thanks, that works. What a weird type checking, coersion language.

The fix--***may be***--seems to be with [CommonToken.d.ts](https://github.com/antlr/antlr4/blob/96f97704a107db99d90ea9503d637104c3bb7afe/runtime/JavaScript/src/antlr4/CommonToken.d.ts).
```
import { Token } from "./Token";
import { CharStream } from "./CharStream";
import { Lexer } from "./Lexer";

export declare class CommonToken extends Token {
    constructor(source: (CharStream | Lexer)[], type: number, channel: number, start: number, stop: number);
    clone(): CommonToken;
    cloneWithType(type: number): CommonToken;
    toString(): string;
}
```
Good. I'll leave this off here for now as I need to back to cleaning up the horrifying chaos with the grammars-v4/python grammars.

Issue #2577, opened 2019-06-18
Repeating tokens for x times


IdkGoodName:
- [x] I am not submitting a question on how to use ANTLR
- [x] I have done a search of the existing issues to make sure I'm not sending in a duplicate

> **Note:** This is a suggestion for ANTLR4.

Currently, in ANTLR4, there is no other way to repeat token **x** times, other than typing `SomeToken SomeToken SomeToken ...`. This can be frustrating if you need to repeat token a lot of times and you need to either paste or type it out. Regex already has that thing, `[abc]{6}` - this matches either `a`, `b` or `c` 6 times. But Regex isn't parser generator, it's for matching things in text. I am suggesting having something similar to Regex's `{...}`.

> **Parse token few times**

Taking idea from Regex's `{...}`, there's a problem. ANTLR4 already uses `{...}`. Since ANTLR4 already uses it, we can take operator, instead of braces. We can use `@`(because `*` is already taken) and here's the syntax `Token @ x`. `'\\u' HexDigit @ 4` is equivalent to `'\\u' HexDigit HexDigit HexDigit HexDigit`.

> **Parse between x and y**

Regex also has ability to match item between **x** and **y**, by using `{x,y}`. Since we took operator, and not brackets, it might be less-readable if we use comma. Instead, we can use `-`, which isn't used anywhere in ANTLR4 as of now. Example of that is `Token @ 1 - 10`, which is equivalent to `Token | Token Token | Token Token Token | Token Token Token Token | Token Token Token Token Token...`. That's a big list right there. You'd ask, where's the usage of that? well, there are some places. For example, "extended unicode escape" can use it. `'\\u{' HexDigit @ 1 - 6  '}'`

----------------

Just like `*` and `+` can be used on anything, you could use `@` also on anything.

HuangLiPang:
@IdkGoodName I am wondering why do you close the issue? Did you figure out how to repeat x times?

KvanTTT:
> Token | Token Token | Token Token Token | Token Token Token Token | Token Token Token Token Token

It also can be written as

```g4
Token (Token (Token (Token Token?)?)?)?
```

Also, there are several similar issues on repository.

PR #4345, opened 2023-07-03
Proposal: Add action templates functionality

DavidGregory084:
This is an idea about how to resolve #4067 without developing an entirely new actions language.

This PR enables users to write cross-target grammars which make use of actions, by enabling users to provide action templates as StringTemplate `.stg` group files on the command line.

By providing different action templates for each target language, users can provide a different implementation of the action logic for each target.

`Java.stg`:
```
normalizerImports() ::= <<
import java.text.Normalizer;
import java.text.Normalizer.Form;
>>
normalize(s) ::= <<Normalizer.normalize(<s>, Form.NFKC)>>
getText() ::= <<getText()>>
setText(s) ::= <<setText(<s>);>>
```

`Javascript.stg`:
```
normalizerImports() ::= ""
normalize(s) ::= <<<s>.normalize("NFKC")>>
getText() ::= "this.text"
setText(s) ::= "this.text = <s>"
```

The example below is the motivating example for me - I have a grammar that I'd like to use in both a JVM-based compiler and a VS Code extension:

`Example.g4`:
```antlr
lexer grammar Example;

@lexer::header {
<normalizerImports()>
}

ID : (ID_START ID_CONTINUE* | '_' ID_CONTINUE+) { <setText(normalize(getText()))> } ;
ID_START : [\p{XID_Start}] ;
ID_CONTINUE: [\p{XID_Continue}] ;
WS : (' '|'\n') -> skip ;
```

Thanks to my employer @opencastsoftware for sponsoring this work!

### DIFF
--- a/tool-testsuite/test/org/antlr/v4/test/tool/TestActionTemplates.java
+++ b/tool-testsuite/test/org/antlr/v4/test/tool/TestActionTemplates.java
@@ -5,6 +5,7 @@ import org.antlr.v4.test.runtime.Stage;
 import org.antlr.v4.test.runtime.java.JavaRunner;
 import org.antlr.v4.test.runtime.states.ExecutedState;
 import org.antlr.v4.test.runtime.states.GeneratedState;
+import org.antlr.v4.test.runtime.states.JavaCompiledState;
 import org.antlr.v4.test.runtime.states.State;
 import org.junit.jupiter.api.Test;
 import org.junit.jupiter.api.io.TempDir;
@@ -275,8 +276,52 @@ public class TestActionTemplates {
 		assertEquals(expecting, ((ExecutedState) state).output);
 	}
 
+	@Test
+	void testActionTemplateSemanticPredicate(@TempDir Path tempDir) {
+		String actionTemplates = "pred() ::= <<true>>";
+
+		writeActionTemplatesFile(tempDir, actionTemplates);
+
+		String actionTemplatesFile = tempDir + FileSeparator + "Java.stg";
+
+		String grammar =
+			"grammar P;\n"+
+			"file : atom EOF ;\n"+
+			"atom : scientific | { <pred()> }? variable ;\n"+
+			"variable: VARIABLE ;\n"+
+			"scientific: SCIENTIFIC_NUMBER ;\n"+
+			"VARIABLE : VALID_ID_START VALID_ID_CHAR* ;\n" +
+			"SCIENTIFIC_NUMBER : NUMBER (E SIGN? UNSIGNED_INTEGER)? ;\n"+
+			"fragment VALID_ID_START : ('a' .. 'z') | ('A' .. 'Z') | '_' ;\n"+
+			"fragment VALID_ID_CHAR : VALID_ID_START | ('0' .. '9') ;\n"+
+			"fragment NUMBER : ('0' .. '9') + ('.' ('0' .. '9') +)? ;\n"+
+			"fragment UNSIGNED_INTEGER : ('0' .. '9')+ ;\n"+
+			"fragment E : 'E' | 'e' ;\n"+
+			"fragment SIGN : ('+' | '-') ;\n"+
+			"WS : (' '|'\\n') -> skip ;";
+
+		State state = execParser(grammar, "Bla", tempDir, "file", actionTemplatesFile);
+
+		assertInstanceOf(ExecutedState.class, state);
+
+		ExecutedState executedState = (ExecutedState) state;
+
+		// We can never match the input unless pred() expands to true
+		assertEquals("", executedState.output);
+		assertEquals("", executedState.errors);
+	}
+
 	void writeActionTemplatesFile(Path tempDir, String template) {
 		writeFile(tempDir.toString(), "Java.stg", template);
+	}
+
+	State execParser(String grammarStr, String input, Path tempDir, String startRule, String actionTemplates) {
+		RunOptions runOptions = createOptionsForJavaToolTests("P.g4", grammarStr, "PParser", "PLexer",
+			false, true, startRule, input,
+			false, false, Stage.Execute, actionTemplates);
+		try (JavaRunner runner = new JavaRunner(tempDir, false)) {
+			return runner.run(runOptions);
+		}
 	}
 
 	State execLexer(String grammarStr, String input, Path tempDir, String actionTemplates) {
--- a/tool-testsuite/test/org/antlr/v4/test/tool/TestActionTemplates.java
+++ b/tool-testsuite/test/org/antlr/v4/test/tool/TestActionTemplates.java
@@ -1,0 +1,131 @@
+package org.antlr.v4.test.tool;
+
+import org.antlr.v4.test.runtime.RunOptions;
+import org.antlr.v4.test.runtime.Stage;
+import org.antlr.v4.test.runtime.java.JavaRunner;
+import org.antlr.v4.test.runtime.states.ExecutedState;
+import org.antlr.v4.test.runtime.states.GeneratedState;
+import org.antlr.v4.test.runtime.states.State;
+import org.junit.jupiter.api.Test;
+import org.junit.jupiter.api.io.TempDir;
+
+import java.nio.file.Path;
+
+import static org.antlr.v4.test.runtime.FileUtils.writeFile;
+import static org.antlr.v4.test.runtime.RuntimeTestUtils.FileSeparator;
+import static org.antlr.v4.test.tool.ToolTestUtils.createOptionsForJavaToolTests;
+import static org.junit.jupiter.api.Assertions.*;
+
+public class TestActionTemplates {
+	@Test() void testIncorrectActionTemplateGroupExtension(@TempDir Path tempDir) {
+		String actionTemplates = tempDir + FileSeparator + "Java.st";
+
+		String grammar =
+			"lexer grammar L;"+
+			"WS : (' '|'\\n') -> skip ;";
+
+		State state = execLexer(grammar, "34 34", tempDir, actionTemplates);
+
+		assertInstanceOf(GeneratedState.class, state);
+		GeneratedState generated = (GeneratedState) state;
+		assertTrue(generated.containsErrors());
+		assertEquals(
+			"State: Generate; \n" +
+			"error(207):  error reading action templates file " + actionTemplates + ": " +
+			"Group file names must end in .stg: " + actionTemplates + "\n",
+			generated.getErrorMessage());
+	}
+
+	@Test() void testActionTemplateFileMissing(@TempDir Path tempDir) {
+		String actionTemplates = tempDir + FileSeparator + "Java.stg";
+
+		String grammar =
+			"lexer grammar L;"+
+			"WS : (' '|'\\n') -> skip ;";
+
+		State state = execLexer(grammar, "34 34", tempDir, actionTemplates);
+
+		assertInstanceOf(GeneratedState.class, state);
+		GeneratedState generated = (GeneratedState) state;
+		assertTrue(generated.containsErrors());
+		assertEquals(
+			"State: Generate; \n" +
+			"error(206):  cannot find action templates file " + actionTemplates + " given for L\n",
+			generated.getErrorMessage());
+	}
+
+	@Test void testActionTemplateLexerAction(@TempDir Path tempDir) {
+		writeActionTemplatesFile(tempDir, "writeln(s) ::= <<outStream.println(\"<s>\");>>");
+
+		String actionTemplates = tempDir + FileSeparator + "Java.stg";
+
+		String grammar =
+			"lexer grammar L;\n"+
+			"I : '0'..'9'+ {<writeln(\"I\")>} ;\n"+
+			"WS : (' '|'\\n') -> skip ;";
+
+		State state = execLexer(grammar, "34 34", tempDir, actionTemplates);
+
+		// Should have identical output to TestLexerActions.testActionExecutedInDFA
+		String expecting =
+			"I\n" +
+			"I\n" +
+			"[@0,0:1='34',<1>,1:0]\n" +
+			"[@1,3:4='34',<1>,1:3]\n" +
+			"[@2,5:4='<EOF>',<-1>,1:5]\n";
+
+		assertInstanceOf(ExecutedState.class, state);
+
+		assertEquals(expecting, ((ExecutedState) state).output);
+	}
+
+	@Test void testActionTemplateHeader(@TempDir Path tempDir) {
+		String sb =
+			"normalizerImports() ::= <<\n" +
+			"import java.text.Normalizer;\n" +
+			"import java.text.Normalizer.Form;\n" +
+			">>\n" +
+			"normalize(s) ::= <<Normalizer.normalize(<s>, Form.NFKC)>>\n" +
+			"getText() ::= <<getText()>>\n" +
+			"setText(s) ::= <<setText(<s>);>>";
+
+		writeActionTemplatesFile(tempDir, sb);
+
+		String actionTemplates = tempDir + FileSeparator + "Java.stg";
+
+		String grammar =
+			"lexer grammar L;\n"+
+			"@lexer::header {\n"+
+			"<normalizerImports()>\n"+
+			"}\n"+
+			"ID : (ID_START ID_CONTINUE* | '_' ID_CONTINUE+) { <setText(normalize(getText()))> } ;\n"+
+			"ID_START : [\\p{XID_Start}] ;\n"+
+			"ID_CONTINUE: [\\p{XID_Continue}] ;\n"+
+			"WS : (' '|'\\n') -> skip ;";
+
+		State state = execLexer(grammar, "This _is \ufb01ne", tempDir, actionTemplates);
+
+		String expecting =
+			"[@0,0:3='This',<1>,1:0]\n"+
+			"[@1,5:7='_is',<1>,1:5]\n"+
+			"[@2,9:11='fine',<1>,1:9]\n"+
+			"[@3,12:11='<EOF>',<-1>,1:12]\n";
+
+		assertInstanceOf(ExecutedState.class, state);
+
+		assertEquals(expecting, ((ExecutedState) state).output);
+	}
+
+	void writeActionTemplatesFile(Path tempDir, String template) {
+		writeFile(tempDir.toString(), "Java.stg", template);
+	}
+
+	State execLexer(String grammarStr, String input, Path tempDir, String actionTemplates) {
+		RunOptions runOptions = createOptionsForJavaToolTests("L.g4", grammarStr, null, "L",
+			false, true, null, input,
+			false, false, Stage.Execute, actionTemplates);
+		try (JavaRunner runner = new JavaRunner(tempDir, false)) {
+			return runner.run(runOptions);
+		}
+	}
+}
--- a/tool-testsuite/test/org/antlr/v4/test/tool/TestActionTemplates.java
+++ b/tool-testsuite/test/org/antlr/v4/test/tool/TestActionTemplates.java
@@ -133,7 +133,7 @@ public class TestActionTemplates {
 
 		assertEquals(
 			"State: Generate; \n" +
-			"error(210):  error rendering action template file " + actionTemplates + ": Java.stg 1:11: mismatched input ':' expecting '::='\n" +
+			"error(209):  error compiling action templates file " + actionTemplates + ": Java.stg 1:11: mismatched input ':' expecting '::='\n" +
 			"error(212): " + grammarFile + ":2:14: error rendering action template: 2:16: no such template: /writeln\n",
 			generated.getErrorMessage());
 	}
--- a/tool-testsuite/test/org/antlr/v4/test/tool/TestActionTemplates.java
+++ b/tool-testsuite/test/org/antlr/v4/test/tool/TestActionTemplates.java
@@ -5,7 +5,6 @@ import org.antlr.v4.test.runtime.Stage;
 import org.antlr.v4.test.runtime.java.JavaRunner;
 import org.antlr.v4.test.runtime.states.ExecutedState;
 import org.antlr.v4.test.runtime.states.GeneratedState;
-import org.antlr.v4.test.runtime.states.JavaCompiledState;
 import org.antlr.v4.test.runtime.states.State;
 import org.junit.jupiter.api.Test;
 import org.junit.jupiter.api.io.TempDir;
@@ -18,7 +17,9 @@ import static org.antlr.v4.test.tool.ToolTestUtils.createOptionsForJavaToolTests
 import static org.junit.jupiter.api.Assertions.*;
 
 public class TestActionTemplates {
-	@Test() void testIncorrectActionTemplateGroupExtension(@TempDir Path tempDir) {
+	@Test void testIncorrectActionTemplateGroupExtension(@TempDir Path tempDir) {
+		writeFile(tempDir.toString(), "Java.st", "");
+
 		String actionTemplates = tempDir + FileSeparator + "Java.st";
 
 		String grammar =
@@ -40,7 +41,7 @@ public class TestActionTemplates {
 			generated.getErrorMessage());
 	}
 
-	@Test() void testActionTemplateFileMissing(@TempDir Path tempDir) {
+	@Test void testActionTemplateFileMissing(@TempDir Path tempDir) {
 		String actionTemplates = tempDir + FileSeparator + "Java.stg";
 
 		String grammar =
@@ -276,8 +277,7 @@ public class TestActionTemplates {
 		assertEquals(expecting, ((ExecutedState) state).output);
 	}
 
-	@Test
-	void testActionTemplateSemanticPredicate(@TempDir Path tempDir) {
+	@Test void testActionTemplateSemanticPredicate(@TempDir Path tempDir) {
 		String actionTemplates = "pred() ::= <<true>>";
 
 		writeActionTemplatesFile(tempDir, actionTemplates);
--- a/tool-testsuite/test/org/antlr/v4/test/tool/TestActionTemplates.java
+++ b/tool-testsuite/test/org/antlr/v4/test/tool/TestActionTemplates.java
@@ -27,11 +27,14 @@ public class TestActionTemplates {
 		State state = execLexer(grammar, "34 34", tempDir, actionTemplates);
 
 		assertInstanceOf(GeneratedState.class, state);
+
 		GeneratedState generated = (GeneratedState) state;
+
 		assertTrue(generated.containsErrors());
+
 		assertEquals(
 			"State: Generate; \n" +
-			"error(207):  error reading action templates file " + actionTemplates + ": " +
+			"error(208):  error reading action templates file " + actionTemplates + ": " +
 			"Group file names must end in .stg: " + actionTemplates + "\n",
 			generated.getErrorMessage());
 	}
@@ -46,15 +49,144 @@ public class TestActionTemplates {
 		State state = execLexer(grammar, "34 34", tempDir, actionTemplates);
 
 		assertInstanceOf(GeneratedState.class, state);
+
 		GeneratedState generated = (GeneratedState) state;
+
 		assertTrue(generated.containsErrors());
+
 		assertEquals(
 			"State: Generate; \n" +
 			"error(206):  cannot find action templates file " + actionTemplates + " given for L\n",
 			generated.getErrorMessage());
 	}
 
-	@Test void testActionTemplateLexerAction(@TempDir Path tempDir) {
+	@Test void testUnlexableActionTemplate(@TempDir Path tempDir) {
+		writeActionTemplatesFile(tempDir, "writeln(s) ::= <<outStream.println(\"<s>\");>>");
+
+		String actionTemplates = tempDir + FileSeparator + "Java.stg";
+		String grammarFile = tempDir + FileSeparator + "L.g4";
+
+		String grammar =
+			"lexer grammar L;\n"+
+			"I : '0'..'9'+ {<¢>} ;\n"+
+			"WS : (' '|'\\n') -> skip ;";
+
+		State state = execLexer(grammar, "34 34", tempDir, actionTemplates);
+
+		assertInstanceOf(GeneratedState.class, state);
+
+		GeneratedState generated = (GeneratedState) state;
+
+		assertTrue(generated.containsErrors());
+
+		assertEquals(
+			"State: Generate; \n" +
+			"error(211): " + grammarFile + ":2:14: error compiling action template: 2:16: invalid character '¢'\n" +
+			"error(211): " + grammarFile + ":2:14: error compiling action template: 2:14: this doesn't look like a template: \" <¢> \"\n",
+			generated.getErrorMessage());
+	}
+
+	@Test void testUnlexableMultilineActionTemplate(@TempDir Path tempDir) {
+		writeActionTemplatesFile(tempDir, "writeln(s) ::= <<outStream.println(\"<s>\");>>");
+
+		String actionTemplates = tempDir + FileSeparator + "Java.stg";
+		String grammarFile = tempDir + FileSeparator + "L.g4";
+
+		String grammar =
+			"lexer grammar L;\n"+
+			"I : '0'..'9'+ {\n" +
+			"  <¢>\n" +
+			"};\n"+
+			"WS : (' '|'\\n') -> skip ;";
+
+		State state = execLexer(grammar, "34 34", tempDir, actionTemplates);
+
+		assertInstanceOf(GeneratedState.class, state);
+
+		GeneratedState generated = (GeneratedState) state;
+
+		assertTrue(generated.containsErrors());
+
+		assertEquals(
+			"State: Generate; \n" +
+			"error(211): " + grammarFile + ":2:14: error compiling action template: 3:3: invalid character '¢'\n" +
+			"error(211): " + grammarFile + ":2:14: error compiling action template: 3:0: mismatched input '  ' expecting EOF\n",
+			generated.getErrorMessage());
+	}
+
+	@Test void testInvalidActionTemplateGroup(@TempDir Path tempDir) {
+		writeActionTemplatesFile(tempDir, "writeln(s) := <<outStream.println(\"<s>\");>>");
+
+		String actionTemplates = tempDir + FileSeparator + "Java.stg";
+		String grammarFile = tempDir + FileSeparator + "L.g4";
+
+		String grammar =
+			"lexer grammar L;\n"+
+			"I : '0'..'9'+ {<writeln(\"I\")>} ;\n"+
+			"WS : (' '|'\\n') -> skip ;";
+
+		State state = execLexer(grammar, "34 34", tempDir, actionTemplates);
+
+		GeneratedState generated = (GeneratedState) state;
+
+		assertTrue(generated.containsErrors());
+
+		assertEquals(
+			"State: Generate; \n" +
+			"error(210):  error rendering action template file " + actionTemplates + ": Java.stg 1:11: mismatched input ':' expecting '::='\n" +
+			"error(212): " + grammarFile + ":2:14: error rendering action template: 2:16: no such template: /writeln\n",
+			generated.getErrorMessage());
+	}
+
+	@Test void testInvalidActionTemplate(@TempDir Path tempDir) {
+		writeActionTemplatesFile(tempDir, "writeln(s) ::= <<outStream.println(\"<s>\");>>");
+
+		String actionTemplates = tempDir + FileSeparator + "Java.stg";
+		String grammarFile = tempDir + FileSeparator + "L.g4";
+
+		String grammar =
+			"lexer grammar L;\n"+
+			"I : '0'..'9'+ {<writeln(\"I\")} ;\n"+
+			"WS : (' '|'\\n') -> skip ;";
+
+		State state = execLexer(grammar, "34 34", tempDir, actionTemplates);
+
+		GeneratedState generated = (GeneratedState) state;
+
+		assertTrue(generated.containsErrors());
+
+		assertEquals(
+			"State: Generate; \n" +
+			"error(211): " + grammarFile + ":2:14: error compiling action template: 2:29: premature EOF\n",
+			generated.getErrorMessage());
+	}
+
+	@Test void testInvalidMultilineActionTemplate(@TempDir Path tempDir) {
+		writeActionTemplatesFile(tempDir, "writeln(s) ::= <<outStream.println(\"<s>\");>>");
+
+		String actionTemplates = tempDir + FileSeparator + "Java.stg";
+		String grammarFile = tempDir + FileSeparator + "L.g4";
+
+		String grammar =
+			"lexer grammar L;\n"+
+			"I : '0'..'9'+ {\n"+
+			"  <writeln(\"I\")\n"+
+			"};\n"+
+			"WS : (' '|'\\n') -> skip ;";
+
+		State state = execLexer(grammar, "34 34", tempDir, actionTemplates);
+
+		GeneratedState generated = (GeneratedState) state;
+
+		assertTrue(generated.containsErrors());
+
+		assertEquals(
+			"State: Generate; \n" +
+			"error(211): " + grammarFile + ":2:14: error compiling action template: 4:1: premature EOF\n",
+			generated.getErrorMessage());
+	}
+
+	@Test void testValidActionTemplate(@TempDir Path tempDir) {
 		writeActionTemplatesFile(tempDir, "writeln(s) ::= <<outStream.println(\"<s>\");>>");
 
 		String actionTemplates = tempDir + FileSeparator + "Java.stg";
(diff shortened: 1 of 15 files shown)


tool/src/org/antlr/v4/tool/GrammarTransformPipeline.java
```diff
@@ -509,4 +517,43 @@ public GrammarRootAST extractImplicitLexer(Grammar combinedGrammar) {
...
+			});
+		} catch (IllegalArgumentException e) {
+			if (e.getMessage() != null && e.getMessage().startsWith("No such group file")) {
+				tool.errMgr.toolError(ErrorType.CANNOT_FIND_ACTIONS_TEMPLATE_FILE_GIVEN_ON_CMDLINE, e, fileName, root.g.name);
+			} else {
+				tool.errMgr.toolError(ErrorType.ERROR_READING_ACTION_TEMPLATES_FILE, e, fileName, e.getMessage());
+			}
+		} catch (STException e) {
+			tool.errMgr.toolError(ErrorType.ERROR_READING_ACTION_TEMPLATES_FILE, e, fileName, e.getMessage());
+		}
```
DavidGregory084, line 546: I'm flying blind here so I followed the example of `expandParameterizedLoops`

tool/src/org/antlr/v4/tool/GrammarTransformPipeline.java
```diff
@@ -509,4 +517,43 @@ public GrammarRootAST extractImplicitLexer(Grammar combinedGrammar) {
...
+	}
+
+	public GrammarAST expandActionTemplate(GrammarAST t, STGroupFile actionTemplates) {
+		String grammarFileInfo = t.g.fileName + " " + t.getLine() + 1 + ":" + t.getCharPositionInLine();
```
DavidGregory084, line 550: This is absolute rubbish - would love your ideas about how to connect up the error reporting from StringTemplate to the line/column offsets in the ANTLR grammar.

tool/src/org/antlr/v4/Tool.java
```diff
@@ -279,7 +279,8 @@ protected void handleOptionSetArg(String arg) {
 				return;
 			}
 			if ( Grammar.parserOptions.contains(option) ||
-				 Grammar.lexerOptions.contains(option) )
+				 Grammar.lexerOptions.contains(option) ||
+				option.startsWith("actionTemplates"))
```
DavidGregory084, line 283: I need to revert this as it's no longer needed

ericvergnaud:
Mmmm... this IS an "entirely new actions language". The fact that it leverages templates makes it look like not (and is elegant), but we'd still have to opine on the syntax and more importantly: the supported set of 'keywords', both characteristics of a language... 
The latter could be a blocker because we precisely don't want to find ourselves in a situation where we'd have to expand that set for every language, because the existing set does not support the specific use case a user needs to support...
Using templates for testing works because the set is limited to what the engine requires, and the task of supporting additional templates lies with the developer needing it, but with your proposal that task would lie with target authors who don't actually need it, so not sure it would work well.
As you know, there are other options being discussed.

DavidGregory084:
> Mmmm... this IS an "entirely new actions language". The fact that it leverages templates makes it look like not (and is elegant), but we'd still have to opine on the syntax and more importantly: the supported set of 'keywords', both characteristics of a language...

Hmm can you explain what you mean by that? The syntax is that of StringTemplate, and all of the template identifiers used are user-defined - they’re simply supplied as StringTemplate groups.

DavidGregory084:
Ah I think I understand - I’m not proposing that there is a centrally managed collection of StringTemplate groups shipped with ANTLR - this PR adds a command line option via which users can supply their own templates which they have written themselves.

Writing cross-target grammars is a rather advanced use case and I think it’s reasonable to expect users to learn StringTemplate syntax if they want to do that.

DavidGregory084:
Why [StringTemplate](https://www.stringtemplate.org/)?

* It's used in the implementation of ANTLR itself and in the ANTLR code generation process, so it's familiar to ANTLR contributors
* It's already on the classpath of the ANTLR tool, so there are no new dependencies needed
* It's a stable and well-tested template language that will allow users to provide whatever kind of action logic they want

kaby76:
I have a few observations and questions.

* I've been writing target-agnostic Antlr grammars probably more than most, porting grammars-v4 to different targets, pretty much every day of the week for several years, and just got through yet another, [python/python3](https://github.com/antlr/grammars-v4/pull/3540). The main difficulty is that the runtime APIs across targets are inconsistent. Three bugs were discovered (https://github.com/antlr/antlr4/issues/4318 https://github.com/antlr/antlr4/pull/4340 https://github.com/antlr/antlr4/pull/4343). While annoying, dealing with actions per target is manageable using [target agnostic format](https://github.com/antlr/antlr4/blob/ce7f5e73cd4787c5aebdc05c29d4d898d6f2d123/doc/target-agnostic-grammars.md).
* Importantly, why is this part of a parser generator? And, as with any parser generator, where is the boundary between parser generator and some other tool that uses it?
* Note, I've been writing, already, a string processor for Antlr grammars, [trgen](https://github.com/kaby76/Domemtech.Trash/tree/main/src/trgen), so I am familiar with it--and all it's problems--and the idea of separating this from the parser generator. (I've been thinking of a more general processor as well.)
* I am not at all happy with the way the model-view is separated in ST. For example, you cannot write a simple string equality test: `<if(target=="CSharp")>....<endif>`.

DavidGregory084:
> While annoying, dealing with actions per target is manageable using [target agnostic format](https://github.com/antlr/antlr4/blob/ce7f5e73cd4787c5aebdc05c29d4d898d6f2d123/doc/target-agnostic-grammars.md).

I think the Fortran90 grammar is actually a good example in itself of the need for some kind of solution to target-agnostic actions.

The target-agnostic format described there relies upon `superClass` in order to work, which means that it falls short for non-OO targets like Go and needs [adjustments](https://github.com/antlr/grammars-v4/blob/master/fortran/fortran90/Go/transformGrammar.py#L22) in order to produce valid code.

In many cases it would be really difficult for users to integrate such a script into their workflow (e.g. when the ANTLR generator is called by a build tool plugin or as part of some other complex workflow).

> Importantly, why is this part of a parser generator? And, as with any parser generator, where is the boundary between parser generator and some other tool that uses it?

ST is already a dependency of the ANTLR tool, so I guess to some extent the question of whether to embed it into ANTLR has already been decided. The expansion of ST templates in this PR is completely optional and is only active if the user provides the `-DactionTemplates` command line flag.

I suppose the boundary here is the point of template expansion, which happens early in the ANTLR generation process just after parsing the grammar file itself. It's also where the most troubling interactions occur, in terms of showing users error messages relating to the template expansion process.

> I am not at all happy with the way the model-view is separated in ST.

I think that is a moot point in this implementation as no model binding is performed - the user has to write all of the groups that they want to use in their `.stg` template, and they can use only those groups in their grammar. 😄

tool/src/org/antlr/v4/tool/GrammarTransformPipeline.java
DavidGregory084: I'm investigating using STErrorListener to connect StringTemplate's error reporting to ANTLR's here

kaby76:
Usually, I would say that ST should be supported outside the tool, but after thinking about this a bit, I think you are right. The main reason why it should be supported in the tool is because people would say a grammar that requires further processing shouldn't be called an Antlr grammar. In addition, the grammar could be tagged as requiring a certain version of Antlr. I will try out some examples shortly.

kaby76:
Semantic predicates are actions, too. It's not expanding template references there, but it is working elsewhere.

```
header() ::= <<
int count;
>>

action1() ::= <<
this.count = 0;
>>

pred() ::= <<
this.count <= 0
>>

action2() ::= <<
this.count++;
>>
```

```
grammar Expr;

options { actionTemplates='Temp.stg'; }

@members {
<header()>
}

file : x (SEMI x)* SEMI? EOF;
x : { <action1()> } expression;
expression : expression POW expression
 | expression (TIMES | DIV) expression
 | expression (PLUS | MINUS) expression
 | LPAREN expression RPAREN
 | (PLUS | MINUS)* atom ;
atom : scientific | { <pred1()> }? variable { <action2()> } ;
scientific : SCIENTIFIC_NUMBER ;
variable : VARIABLE ;

VARIABLE : VALID_ID_START VALID_ID_CHAR* ;
SCIENTIFIC_NUMBER : NUMBER (E SIGN? UNSIGNED_INTEGER)? ;
LPAREN : '(' ;
RPAREN : ')' ;
PLUS : '+' ;
MINUS : '-' ;
TIMES : '*' ;
DIV : '/' ;
GT : '>' ;
LT : '<' ;
EQ : '=' ;
POINT : '.' ;
POW : '^' ;
SEMI : ';' ;
WS : [ \r\n\t] + -> channel(HIDDEN) ;

fragment VALID_ID_START : ('a' .. 'z') | ('A' .. 'Z') | '_' ;
fragment VALID_ID_CHAR : VALID_ID_START | ('0' .. '9') ;
fragment NUMBER : ('0' .. '9') + ('.' ('0' .. '9') +)? ;
fragment UNSIGNED_INTEGER : ('0' .. '9')+ ;
fragment E : 'E' | 'e' ;
fragment SIGN : ('+' | '-') ;
```

DavidGregory084:
> Semantic predicates are actions, too. It's not expanding template references there, but it is working elsewhere.

Ah thanks for trying it out - I will take a look at semantic predicates tomorrow

kaby76:
Looks like you need to test for SEMPRED in addition to ACTION in `expandActionTemplates(GrammarRootAST root)`. Then, trim the extra '?' from the text of the predicate. Honestly, I don't understand why the text isn't just trimmed of '{', '}', and '?' when the parse tree is constructed. You might need to check the equivalent action blocks for a lexer grammar. I don't know if the parse tree is very different.

Otherwise, I think solution is better than the others we've thought of up to now. It fixes the "p" vs "l" problem, as well as the "this." vs "$this->" vs "self." vs "this->" problem. And it doesn't result in a proliferation of `if-then-else` code sprinkled through the grammar itself.

I don't think we have any more non-OO targets. The Go runtime is slammed into something that resembles an OO framework. It just wan't implemented well to begin with. But, ST actions could come in handy with a new target like C.

This doesn't take care of the target-specific code in local and return clauses. But, we don't use those features in grammars-v4.

You got my vote. @parrt Please consider this PR. It goes a long way to fixing the problems in writing grammars over different targets.

ericvergnaud:
A similar but more independent approach would be to write the grammar as a template and apply the transforms in order to produce a target specific g4.
Less convenient but does not require any change to antlr4 tool.

kaby76:
The problem with writing the grammar as .stg is that it is then not an Antlr grammar. The .stg could not be parsed using the grammars-v4 antlr4 grammar. Existing VSCode or VSIDE extensions would not work. People would not call it an Antlr grammar. The use of the template calls would not be restricted to just the actions. Not a good solution.

You could implement this as a separate command-line tool like trgen to parse as an Antlr4 grammar, then render the actions, but then you'd have a separate tool. It's would be labeled a third party tool that folks won't know how to install, screw up on install, say it's not an Antlr grammar. Antlr4-tools is official, directly referenced in antlr.org.

ericvergnaud:
@kaby76 re the template not being a valid grammar, that's a fair point.
That said I think actionTemplates should be a command line option only. Its purpose is to provide a different implementation per target. Making it a grammar option ( using `options { actionTemplates='Temp.stg'; }` ) pushes the responsibility of locating that template to antlr, which is wrong.
@DavidGregory084 there is a big downside to this which relates to automation. Many developers use the maven plugin to generate code at build time (rather than at dev time - which I prefer). The plugin would now need to know where to find the action templates, on a per target basis. Does your PR cater for this ?

tool/src/org/antlr/v4/tool/Grammar.java
```diff
@@ -80,6 +80,7 @@ public class Grammar implements AttributeResolver {
 		parserOptions.add("TokenLabelType");
 		parserOptions.add("tokenVocab");
 		parserOptions.add("language");
+		parserOptions.add("actionTemplates");
```
ericvergnaud, line 83: Not sure we want to support this as a grammar option

tool/src/org/antlr/v4/tool/Grammar.java
```diff
@@ -1178,6 +1179,10 @@ public String getLanguage() {
 		return getOptionString("language");
 	}
 
+	public String getActionTemplates() {
```
ericvergnaud, line 1182: same here

kaby76:
> Making it a grammar option ( using `options { actionTemplates='Temp.stg'; }` ) pushes the responsibility of locating that template to antlr, which is wrong. 

Well, I could solve the explicit dependency by applying a naming standard to the .g4's and .stg's, e.g., Python3Lexer.g4/Python3Lexer.stg and Python3Parser.g4/Python3Parser.stg. I already apply a coding standard for the location of target-specific files over in grammars-v4. It took years to straighten it all out, but all target-specific files are in sub-directories Cpp/, CSharp/, Dart/, Go/, Java/, JavaScript/, PHP/, Python3/, and TypeScript/. So, these .stg's would go into each of the sub-directories. That's fine because that's where the `superClass` files already exist. The base classes can be easily rewritten as .stg's. I'd know which grammar requires a command-line option for trgen, and could then generate a build script in whatever build system (dotnet, maven, npm, etc) to apply the command-line option.

But, why is an equivalent `actionTemplates` "wrong"? The Antlr tool already has a `superClass` option. Why even use `options { superClass=Python3LexerBase; }` when we can just do `antlr4 -DsuperClass=Python3LexerBase Python3Lexer.g4` and assume so because we have a naming scheme?

ericvergnaud:
@kaby76  Sorry to disagree, but that would have to be a new entry in the pom file. I am not supportive of an approach that requires the tool to resolve the stg file path in the first place, and differently for each target.

kaby76:
> @kaby76 Sorry to disagree, but that would have to be a new entry in the pom file. I am not supportive of an approach that requires the tool to resolve the stg file path in the first place, and differently for each target.

First, `actionTemplates` is not a directory, it's a top-level file name, including path. 

Second, nothing in the .g4 would need to be changed per target. The command line to generate the target-specific recognizers would still be the same: `antlr4 -actionTemplates=Python3Lexer.stg -Dlanguage=CSharp Python3Lexer.g4; antlr4 -actionTemplates=Python3Lexer.stg -Dlanguage=CSharp Python3Lexer.g4`, or antlr4 -actionTemplates=Python3Lexer.stg -Dlanguage=Java Python3Lexer.g4; antlr4 -actionTemplates=Python3Lexer.stg -Dlanguage=Java Python3Lexer.g4`. That's where I wholeheartedly prefer that the `-Dlanguage` option not affect the `-DactionTemplate` option. They should be completely independent.

Nothing in the pom.xml would need to change whatsoever other than now adding the option, which I wouldn't do, because I would always write the grammar in grammars-v4 with the option within the grammar, just like the `superClass` option.

DavidGregory084:
> That said I think actionTemplates should be a command line option only. Its purpose is to provide a different implementation per target. Making it a grammar option ( using `options { actionTemplates='Temp.stg'; }` ) pushes the responsibility of locating that template to antlr, which is wrong.

I agree that a target-specific option does not make a lot of sense in the grammar file - the only reason that I have done it this way is that the "language" option is already a grammar-level option and "actionTemplates" seems to belong wherever "language" is defined - happy to change it.

I expect most folks would provide "actionTemplates" as a command line option.

> @DavidGregory084 there is a big downside to this which relates to automation. Many developers use the maven plugin to generate code at build time (rather than at dev time - which I prefer). The plugin would now need to know where to find the action templates, on a per target basis. Does your PR cater for this ?

Not at the moment - I think that would require changes to the Maven plugin and I haven't look at that part of the project yet.

The Gradle plugin already requires manually manipulating the command line options for any advanced use-cases, e.g. here's the invocation from my current project:

```kotlin
tasks.generateGrammarSource {
    // See: https://github.com/antlr/antlr4/issues/2335
    val outputDir = file("build/generated-src/antlr/main/org/mina_lang/parser")
    // the directory must exist or ANTLR bails
    doFirst { outputDir.mkdirs() }
    arguments = arguments + listOf(
        "-visitor",
        "-no-listener",
        // the lexer tokens file can't be found by the parser without this
        "-lib", outputDir.absolutePath,
        "-package", "org.mina_lang.parser")
}
```

Having said that, I don't really know which use cases would require providing different arguments to the same build tool?

e.g. in my case, I will probably move my grammar to a top-level directory, and symlink it into my Gradle source directories and my npm source directories. Then each build tool will invoke the ANTLR tool with different command line arguments to generate the target that's relevant for that build tool - Java for my Gradle build, TypeScript for my npm build.

tool-testsuite/test/org/antlr/v4/test/tool/TestActionTemplates.java
```diff
@@ -0,0 +1,290 @@
...
+		assertInstanceOf(GeneratedState.class, state);
+
+		GeneratedState generated = (GeneratedState) state;
+
+		assertTrue(generated.containsErrors());
+
+		assertEquals(
+			"State: Generate; \n" +
+			"error(211): " + grammarFile + ":2:14: error compiling action template: 3:3: invalid character '¢'\n" +
+			"error(211): " + grammarFile + ":2:14: error compiling action template: 3:0: mismatched input '  ' expecting EOF\n",
```
DavidGregory084, line 113: This message is a little weird - I'm not sure why changing the action template to be multiline triggers this error message

DavidGregory084:
In https://github.com/antlr/antlr4/pull/4345/commits/f89c5205b38f054978b42263a71e3b0fa21afdcc I've tried to connect up the error reporting from StringTemplate with the positions in the ANTLR grammar file. You can see the results in tool-testsuite/test/org/antlr/v4/test/tool/TestActionTemplates.java.
I'm sure that there are cases that I haven't thought of which could result in janky position reporting - I think it will take a few folks kicking the tires to iron out the problems with it!

kaby76:
> > That said I think actionTemplates should be a command line option only. Its purpose is to provide a different implementation per target. Making it a grammar option ( using `options { actionTemplates='Temp.stg'; }` ) pushes the responsibility of locating that template to antlr, which is wrong.
> 
> I agree that a target-specific option does not make a lot of sense in the grammar file - the only reason that I have done it this way is that the "language" option is already a grammar-level option and "actionTemplates" seems to belong wherever "language" is defined - happy to change it.

> I expect most folks would provide "actionTemplates" as a command line option.

Yes, respectfully I disagree. The .stg's are just like imported .g4's. Collectively, it's the "grammar". Without either, it's not a whole grammar. The target-specific nature should in the `-lib/-Dlanguage` options, not the path in the actionTemplates option.

Following this logic, why even have `superClass` in the grammar? Why not always say `antlr4 -DsuperClass=Python3LexerBase`. It does not make sense; it is not consistent.

DavidGregory084:
> Yes, respectfully I disagree. The .stg's are just like imported .g4's. Collectively, it's the "grammar". Without either, it's not a whole grammar. The target-specific nature should in the `-lib/-Dlanguage` options, not the path in the actionTemplates option.

I can definitely see the viewpoint that it is like an imported file.

To try and explain the other viewpoint a bit more - the "actionTemplates" argument is the mechanism which allows the grammar to receive different target-specific template snippets dependent upon the given `-Dlanguage` option, so from that perspective, it makes sense that it would be a target-specific option.

By making the "actionTemplates" a part of the grammar file, users are forced to use relative paths in order to provide those target-specific templates differently for each target, which enforces a specific approach and folder structure onto users of the option.

I hope that explanation makes sense?

ericvergnaud:
@kaby76 a couple of thoughts:
 - superClass is _**NOT**_ target dependent. I can have the same superClass name regardless of the target. I personally do that.
 - the stg **_IS_** target dependent, and being target specific is its _**only**_ purpose - if I just need 1 target then I can write action code directly in the grammar. So where would the tool find that target-specific file ? 
  - If the full path is hardcoded as an option in the grammar, then it's not target specific. 
  - If it's a file name relative to a directory structure, then it forces every developer to replicate a specific directory structure (is that the one that we've picked for the antlr4-grammars), and places the responsibility of locating the actionTemplates file on the tool, 2 things which we absolutely want to avoid.

DavidGregory084:
> Semantic predicates are actions, too. It's not expanding template references there, but it is working elsewhere.

@kaby76 I added a test using a cut-down version of your example to ensure that semantic predicates are being expanded in https://github.com/antlr/antlr4/pull/4345/commits/b619c20e5381784d8461f9c2d0a2d4845b674fcb

ericvergnaud:
We'd also need to support this in the IntelliJ plugin

kaby76:
> if I just need 1 target then I can write action code directly in the grammar. So where would the tool find that target-specific file ?

I have my preferences, but you're going to have to pick your poison, i.e., choose a coding standard.  But, yes, you absolutely don't want to force a coding standard in the Antlr tool itself, i.e., "the .stg files must be named Java.stg, CSharp.stg, etc"--because there may be different top-level stg files for each grammar. Or, "the .stg files must be placed under a directory that is named after the target, and further named after the name of the grammar".

In the grammars-v4 repo, the target-specific files are placed in a sub-directory with a target name. That's been agreed upon after years of bantering about. I really don't want to revisit this because it's been hell just to get to this standard, common coding convention across all 300 grammars. Trgen will read the grammar and copy all files from the target-specific directory into the directory with the grammars. You can even fork a .g4, and place it in the target-specific directory. I needed something so I can just write trgen and do testing in a nice, consistent, template-driven manner.

And, yes, I would also never hardwire the target into the grammar, either in the `language` option, or via `actionTemplates` option, or the actions themselves.

But, I would hardwire the action template file name into the grammar, and **never use a relative or absolute path**. I would like that the feature honor `-lib` because I think that would be useful for other developers. But, whether we have a `-lib` option for searching for .stg's, it doesn't matter: trgen will just copy the files CSharp/* or Java/* or what have you to where the .g4's live. The advantage of hardwiring `options { actionTemplates=... }` into the grammar forces the tool to raise an error if the file is missing. Missing grammar files should both be flagged as an error at tool time. It is isn't, some one will type `antlr4 -Dlanguage=CSharp Python3Lexer.g4`, and expect a functioning lexer out the back. Then, I'll start seeing posts from folks wondering "The tool generated a parser, but it's not compiling. Why is this? Why isn't Antlr generating the parser for CSharp cause I used the -Dlanguage=CSharp option?". We get these questions at least a couple times a month.

BTW, if action templating is outside the Antlr tool, someone is going to take a grammar from grammars-v4, insist on not installing the third-party tool, and start manually editing the .g4, moving the template code into the .g4! We'll never stop hearing the complaints that "This is too much work! This is not an Antlr grammar!". At least right now folks can take a "target agnostic grammar" and manually adjusting the .g4, changing "this." to "this->" or what have you, because they refuse to run `python translateGrammar.py`.

ericvergnaud:
@kaby76 

A side-effect of this proposal, if accepted, will be to remove the need for a per language folder in grammars-v4, except when a base recogniser or parser is genuinely required (which I believe is rare, but you know better).

If for example I look at the folder for java, there would no longer be a need for Java9LexerBase, and I suspect the structure could become as simple as:
Java9Lexer.g4
Java9Parser.g4
java-action-templates.stg
csharp-action-templates.stg
...

So relying, within the tool, on a hard coded structure that is bound to disappear doesn't seem like the best move...

Plus, it's much less likely to be approved. The smaller the footprint, the higher the chances.

kaby76:
> @kaby76
> 
> A side-effect of this proposal, if accepted, will be to remove the need for a per language folder in grammars-v4, except when a base recogniser or parser is genuinely required (which I believe is rare, but you know better).
> 
> If for example I look at the folder for java, there would no longer be a need for Java9LexerBase, and I suspect the structure could become as simple as: Java9Lexer.g4 Java9Parser.g4 java-action-templates.stg csharp-action-templates.stg ...
> 
> So relying, within the tool, on a hard coded structure that is bound to disappear doesn't seem like the best move...
> 
> Plus, it's much less likely to be approved. The smaller the footprint, the higher the chances.

We could have the files either way. trgen, nor the Antlr tool, would care. I'll have to chit-chat it with the admins over in grammars-v4.

### Option 1
```
java/java9/Java9Parser.g4
java/java9/Java9Lexer.g4
java/java9/Java/action-templates.stg
java/java9/CSharp/action-templates.stg
...
```

### Option 2
```
java/java9/Java9Parser.g4
java/java9/Java9Lexer.g4
java/java9/Java-action-templates.stg
java/java9/CSharp-action-templates.stg
...
```

I agree, make this PR as small as possible to give @parrt less material to complain about. :)

trgen won't care if all the .stg's are in the same directory as the .g4's, or in the target-specific directory. All files are all copied into the directory containing the grammars from the target-specific directory.

The templates for the trgen driver code would be straightforward in either case. For example, with "Option 1" and the Java target, the [antlr4 call](https://github.com/antlr/grammars-v4/blob/c9e5c8caca0b562805711b22d04a8ce39d012461/_scripts/templates/Java/st.build.sh#L12) would require `-DactionTemplates='<x.GrammarName>.stg'`. For "Option 2", I'd need to add `-DactionTemplates='<target>-<x.GrammarName>.stg'`. I don't have an attribute `<target>`--I never had the need. But, easy enough to add.

I really hope this is merged. It cleans up so much junky hacks.

ericvergnaud:
Ok so I guess the next step is to get @parrt's in principle approval. Then we can look at side effects: maven plug-in, IntelliJ plug-in, antlr lab ... any other ? (Nobody's maintaining the eclipse plug-in I'm afraid...)
@parrt any thoughts ? It seems that this elegant proposal would address a very longly and highly desired cross-target capability with very little impact.

ericvergnaud:
@parrt the general idea follows what we implemented for testing i.e. string templates. The templates would be provided by the developer via a cmd line option, meaning we only provide the automation, not the templates themselves.

kaby76:
Trying this PR out on the [python/python3 grammar](https://github.com/kaby76/grammars-v4/tree/df4e27b98c1c9f3b5cfe8d2a1277a8bc25ddbfa3/python/python3). So far I have Cpp, CSharp, and Java working fine. It looks good. The "transformGrammar.py" hack is gone and there are fewer files for Cpp (no headers that I need to declare). I can probably combine the lexer and parser stg's together for a target. This would result in half the files that define the grammar.

[stg-python3.zip](https://github.com/antlr/antlr4/files/11963828/stg-python3.zip)

kaby76:
It is really unclear on how to call the Lexer.Emit() from the `(*Python3Lexer) Emit()` func that I define if there's no name to access the *unnamed field* `*antlr4.BaseLexer` in the damn Python3Lexer struct/type. `l.BaseLexer` does not work outside of the generated `NewPython3Lexer()` func. I don't think the Go target has ever been thought out.
```
type Python3Lexer struct {
	*antlr.BaseLexer
	channelNames []string
	modeNames    []string
	// TODO: EOF string
}
func NewPython3Lexer(input antlr.CharStream) *Python3Lexer {
	Python3LexerInit()
	l := new(Python3Lexer)
	l.BaseLexer = antlr.NewBaseLexer(input)
	staticData := &Python3LexerLexerStaticData
	l.Interpreter = antlr.NewLexerATNSimulator(l, staticData.atn, staticData.decisionToDFA, staticData.PredictionContextCache)
	l.channelNames = staticData.ChannelNames
	l.modeNames = staticData.ModeNames
	l.RuleNames = staticData.RuleNames
	l.LiteralNames = staticData.LiteralNames
	l.SymbolicNames = staticData.SymbolicNames
	l.GrammarFileName = "Python3Lexer.g4"
	// TODO: l.EOF = antlr.TokenEOF

	return l
}

func (l *Python3Lexer) Emit() antlr.Token {
	t := l.BaseLexer.Emit()                          <<<<<<<<<< SYNTAX ERROR.
	return t
}
```

ericvergnaud:
Haven't tried it but I believe you should be able to call l.Emit() directly

kaby76:
> Haven't tried it but I believe you should be able to call l.Emit() directly

Yes you are right. I just don't name the func Emit() but something like OverrideEmit().

DavidGregory084:
> Then we can look at side effects: maven plug-in, IntelliJ plug-in, antlr lab ... any other ? (Nobody's maintaining the eclipse plug-in I'm afraid...)

I had some time today to take a quick look at this aspect:

* Maven plugin - I think this will work out already due to the presence of the `Map<String, String> options` parameter for providing grammar options in the Antlr4Mojo. Users will be able to configure their action templates by adding the following to the ANTLR plugin configuration:

```xml
<plugin>
    <groupId>org.antlr</groupId>
    <artifactId>antlr4-maven-plugin</artifactId>
    <version>4.3</version>
    <executions>
        <execution>
            <goals>
                <goal>antlr4</goal>
            </goals>
        </execution>
        <configuration>
          <options>
             // This "options" element treats each tag as key and the tag content as value of a map entry
            <actionTemplates>path/to/action-templates.stg</actionTemplates>
          </options>
        </configuration>
    </executions>
</plugin>
```

However, the path aspect is a little unclear - what working directory will the Maven plugin use? As a result of that, I think makes sense to adopt @kaby76's suggestion that we respect the `libDirectory` configuration / `-lib` flag when looking for action templates. I think that the logic for this would be equivalent to [getImportedGrammarFile](https://github.com/antlr/antlr4/blob/ce7f5e73cd4787c5aebdc05c29d4d898d6f2d123/tool/src/org/antlr/v4/Tool.java#L787) from the ANTLR Tool code, except that we might try to load the path as an absolute path first before trying to use it relative to other directories.

* IntelliJ plugin - the IntelliJ plugin seems to treat all action code the same - it's lexed as a sequence of ACTION_CONTENT tokens that are mostly ignored other than checking that they have balanced braces and quotes so we know when the action code has ended. So, the StringTemplate action templates won't have syntax highlighting, but neither does existing action code. I guess it will need real testing to determine whether there might be other impacts.

kaby76:
Now I remember how this is supposed to work in Go. I have to type coerce. This is okay, should work. Took a while to remember how the runtime works.

DavidGregory084:
> I think makes sense to adopt @kaby76's suggestion that we respect the libDirectory configuration / -lib flag when looking for action templates.

The latest commit https://github.com/antlr/antlr4/pull/4345/commits/84866cc05d3eb65108f581305389ec22c8d5199b makes this change - the lookup priority mirrors that for looking up imported grammars, except that we try to read the action templates as an absolute path first:
* Absolute path
* Input directory
* Grammar file parent directory
* Lib directory

DavidGregory084:
I've updated the `options.md` to document the `-DactionTemplates` option. You can see the rendered file [here](https://github.com/antlr/antlr4/blob/c7dc4c80b3b3473501802a3eb34e3da21de66ea2/doc/options.md#actiontemplates).

DavidGregory084:
Would a maintainer be able to re-run the macos-11, typescript job? I'm pretty sure I can't have broken it by changing a Markdown file 😄

ericvergnaud:
@parrt can you opine on this proposal ?

kaby76:
I finally got the Python3 port of [python3](https://github.com/antlr/grammars-v4/tree/master/python/python3) working, but it required careful attention to the indentations of two attribute references in the grammar file. Works for CSharp, Cpp, Java, Go, Python3. Working on the rest (Dart, JavaScript, PHP, TypeScript). https://github.com/kaby76/stg-python3

kaby76:
While implementing the JavaScript target, I found out that the `@members` are placed in the constructor for the [parser](https://github.com/antlr/antlr4/blob/ce7f5e73cd4787c5aebdc05c29d4d898d6f2d123/tool/resources/org/antlr/v4/tool/templates/codegen/JavaScript/JavaScript.stg#L136) and [lexer](https://github.com/antlr/antlr4/blob/ce7f5e73cd4787c5aebdc05c29d4d898d6f2d123/tool/resources/org/antlr/v4/tool/templates/codegen/JavaScript/JavaScript.stg#L812). This differs significantly from the other targets, like the CSharp [parser](https://github.com/antlr/antlr4/blob/ce7f5e73cd4787c5aebdc05c29d4d898d6f2d123/tool/resources/org/antlr/v4/tool/templates/codegen/CSharp/CSharp.stg#L304) and [lexer](https://github.com/antlr/antlr4/blob/ce7f5e73cd4787c5aebdc05c29d4d898d6f2d123/tool/resources/org/antlr/v4/tool/templates/codegen/CSharp/CSharp.stg#L1014C2-L1014C24). That requires restructuring methods into delegates. There are similar subtleties for Python3 and indentation.

This information is pretty important, but it's not in the JavaScript doc. https://github.com/antlr/antlr4/blob/dev/doc/grammars.md#actions-at-the-grammar-level. The doc could be improved.

kaby76:
The [rewrite of the grammars-v4 python3 grammar](https://github.com/kaby76/stg-python3/tree/main) is complete. All ports except PHP and Swift work.

If this PR does not go in, I will create a tool in Trash that does the same. But, I recommend adoption.

ericvergnaud:
@parrt I guess your feedback would be much appreciated

parrt:
Going on vacation for the weekend tomorrow so hopefully I will have a chance to read through this. Sorry for the delay.

tool/src/org/antlr/v4/tool/GrammarTransformPipeline.java
```diff
@@ -509,4 +520,332 @@ public GrammarRootAST extractImplicitLexer(Grammar combinedGrammar) {
...
+			private void reportInternalError(STMessage stMessage) {
+				errorManager.toolError(ErrorType.INTERNAL_ERROR, stMessage.cause, stMessage.toString());
+			}
+		};
+	}
+
+	/**
+	 * Create an error listener for action templates embedded inside a grammar's actions and semantic predicates.
+	 */
+	private STErrorListener createGrammarErrorListener(GrammarAST ast) {
```
DavidGregory084, line 565: One thing to note about the code in this error listener, which mostly revolves around translating positions - it would probably be a good idea to tidy this up a bit by exposing the information that we need in the STMessage classes in https://github.com/antlr/stringtemplate4 instead, especially `getSTRuntimeMessageSourceLocation`, which is essentially a duplicate of `STRuntimeMessage#getSourceLocation` that doesn't call `.toString` before returning.

Perhaps it would make sense to add a `Coordinate getSourceCoordinate()` method to `STMessage` (which may return null).

parrt:
Very interesting and well thought out proposal. Away from my computer, but I am positively inclined to approve something in this nature. So it sounds like the grammar itself would mention the name of a template file, and then the -lib would indicate a directory with a template file. But what is the convention for changing the name of the template according to the language? we definitely don’t want to name the target language inside the grammar file as that defeats the purpose. Seems to me that you might simply want to have a command line option that indicates the specific template file to use, but then it means we don’t need to specify the template file name inside the grammar. But that has the negative that users will be mystified when the code generator does not compile as you point out

ericvergnaud:
Thanks for this. I’m all for making this a cmd line only option. We don’t went to embed any target directory logic. Re non compiling parser when the option is missing, it doesn’t sound difficult to detect a string template and throw an error if no template is provided.Envoyé de mon iPhoneLe 29 juil. 2023 à 19:04, Terence Parr ***@***.***> a écrit :﻿
Very interesting and well thought out proposal. Away from my computer, but I am positively inclined to approve something in this nature. So it sounds like the grammar itself would mention the name of a template file, and then the -lib would indicate a directory with a template file. But what is the convention for changing the name of the template according to the language? we definitely don’t want to name the target language inside the grammar file as that defeats the purpose. Seems to me that you might simply want to have a command line option that indicates the specific template file to use, but then it means we don’t need to specify the template file name inside the grammar. But that has the negative that users will be mystified when the code generator does not compile as you point out

kaby76:
>  I’m all for making this a cmd line only option. We don’t went to embed any target directory logic.

If you want to declare the .stg/.g4 pairing outside the grammar file--in a bash script, makefile, Maven file, etc, that's your choice. But, please don't force the coding style on me. I would prefer to declare the actionTemplates option in the grammar because it has to be specified somewhere.

In my example, there is no mention of target in either of the .g4's: `options { actionTemplate='Python3Parser.stg';}` is the same for all targets. There are no forked copies of the .g4's for each target (a maintenance nightmare). There is no target-specific logic in the .g4. The templated actions are the same for all targets in the .g4, e.g., `@header {<header()>}`. Really, there is zero target logic in the .g4 files in my example. The target-specific choice is in the generated build script. trgen overlays the grammar with target-specific files. Alternatively, it would be fine if the target-specific .stg's could be found using the `-lib` option.

If the pairing of top-level .stg with the .g4 is not specified in the grammar, I'll have to devise a way to add the `-DactionTemplates='...'` option to the generated build script. Some grammars in grammars-v4 have multiple .g4's, e.g., [csharp](https://github.com/antlr/grammars-v4/tree/15e59da1e75fac1ccc343bf8ddd2e2ee8a25dad5/csharp). If the .stg/.g4 pairing is not declared in the .g4, it would need to be declared in the desc.xml or computed by parsing the .g4's, .stg's. Driver code is not part of the grammar we shared in grammars-v4.

DavidGregory084:
> So it sounds like the grammar itself would mention the name of a template file, and then the -lib would indicate a directory with a template file.

Yes, that's right, `actionTemplates` is currently a grammar-level option, but note that grammar-level options can always be provided to the ANTLR tool via the command line, i.e. `antlr4 -DactionTemplates=Actions.stg`.

It seems that @ericvergnaud is strongly against keeping this as a grammar-level option and @kaby76 is strongly for it, so I hope you can be a tie breaker. 😅 

I will note though that I don't see any harm in leaving this at the grammar level if it suits use cases whereby the relative folder structure or `-lib` argument is used to determine where a fixed template file name mentioned in the grammar file is picked up from.

ericvergnaud:
Since -lib describes an input directory, I’m fine with the tool trying to locate the template file in that directory.But I will strongly oppose any logic where the template file name needs to be inferred from the target.Envoyé de mon iPhoneLe 30 juil. 2023 à 16:17, David Gregory ***@***.***> a écrit :﻿

So it sounds like the grammar itself would mention the name of a template file, and then the -lib would indicate a directory with a template file.

Yes, that's right, actionTemplates is currently a grammar-level option, but note that grammar-level options can always be provided to the ANTLR tool via the command line, i.e. antlr4 -DactionTemplates=Actions.stg.
It seems that @ericvergnaud is strongly against keeping this as a grammar-level option and @kaby76 is strongly for it, so I hope you can be a tie breaker. 😅
I will note though that I don't see any harm in leaving this at the grammar level if it suits use cases whereby the relative folder structure or -lib argument is used to determine where a fixed template file name mentioned in the grammar file is picked up from.

DavidGregory084:
@kaby76 that change should already have been implemented as of https://github.com/antlr/antlr4/pull/4345/commits/84866cc05d3eb65108f581305389ec22c8d5199b - are you using an earlier commit or is this something that I need to look into?

ericvergnaud:
@parrt this is not ready to merge, but I guess we have your green light to complete the work ?

kaby76:
> @parrt this is not ready to merge, but I guess we have your green light to complete the work ?

I am still waiting for the `-lib` option fix. That part does not work. I never received a reply from @DavidGregory084 addressing the bug that I found. If he doesn't fix the problem, I would still recommend the merge because it would still be good enough to redo the grammars in grammars-v4. I can fix the bug in a follow-up PR.

ericvergnaud:
> > @parrt this is not ready to merge, but I guess we have your green light to complete the work ?
> 
> I am still waiting for the `-lib` option fix. That part does not work. I never received a reply from @DavidGregory084 addressing the bug that I found. If he doesn't fix the problem, I would still recommend the merge because it would still be good enough to redo the grammars in grammars-v4. I can fix the bug in a follow-up PR.

mmm... I think we would need to support this in the maven and the IntelliJ plugins at minimal, and antlr-lab

DavidGregory084:
> I am still waiting for the `-lib` option fix. That part does not work. I never received a reply from @DavidGregory084 addressing the bug that I found. If he doesn't fix the problem, I would still recommend the merge because it would still be good enough to redo the grammars in grammars-v4. I can fix the bug in a follow-up PR.

@kaby76 I asked you a question about this in my last comment:

> @kaby76 that change should already have been implemented as of [84866cc](https://github.com/antlr/antlr4/commit/84866cc05d3eb65108f581305389ec22c8d5199b) - are you using an earlier commit or is this something that I need to look into?

Which revision are you using?

kaby76:
> Which revision are you using?

I was using the latest. I always use the latest. But of course, mistakes sometimes happen. But not this time.

I created a repo (https://github.com/kaby76/do-over-and-over.git) with a script (https://github.com/kaby76/do-over-and-over/blob/main/do-over-and-over.sh) that clones your repo, does a build and run for three different versions.
* For commit `84866cc`, which you assert fixes the `-lib` problem--it does not as shown [here](https://github.com/kaby76/do-over-and-over/blob/9cda4bb083770b8ce11a967bce7f13a2c92a1a30/output.txt#L559-L562), `"error(206):...."`.
* For tip of branch `action-templates`. The problem still occurs, as shown [here](https://github.com/kaby76/do-over-and-over/blob/9cda4bb083770b8ce11a967bce7f13a2c92a1a30/output.txt#L1112-L1115), `"error(206):...."`.
* For tip of branch `action-templates`, patching the code using [fix.txt](https://github.com/kaby76/do-over-and-over/blob/main/fix.txt) [at this point in the output](https://github.com/kaby76/do-over-and-over/blob/9cda4bb083770b8ce11a967bce7f13a2c92a1a30/output.txt#L1141). My patch works, as shown [here with no errors](https://github.com/kaby76/do-over-and-over/blob/9cda4bb083770b8ce11a967bce7f13a2c92a1a30/output.txt#L1670-L1671).

DavidGregory084:
Cool thanks for confirming - I'll take a look as soon as I can!

ericvergnaud:
@DavidGregory084 hi there. It's been a while. Any chance you will progress this ?

ericvergnaud:
@DavidGregory084 we'd like to include this feature in ANTLR5.
Would you be able to migrate this PR accordingly ?

DavidGregory084:
Apologies @ericvergnaud I have not had a great deal of time for open source over the past few months. What are the timelines like for ANTLR5? I will try to make some time for this next week!

ericvergnaud:
The overall timeline is flexible, but this feature will rapidly become critical due to the new architecture.

DavidGregory084:
@ericvergnaud apologies for the delay - @kaby76 was absolutely right about the cause of the issue, but I wanted to write a unit test to exercise the `-lib` directory code and never seemed to find the time.

I have incorporated @kaby76's fix now and added a test.

Next I will migrate the PR to the antlr5 repo?

Will you still accept this PR on antlr4 as well, or should we close it here?

ericvergnaud:
@DavidGregory084 I wouldn't close it. Once it is successfully implemented in antlr5, its usefulness might be proven well enough for Ter to accept it.

DavidGregory084:
Opened a PR for the port at https://github.com/antlr/antlr5/pull/51

kaby76:
Note, templates could be applied elsewhere in the grammar e.g., `UP_TO_THREE_DOTS: <repeat(3, '.')> ;`. Not a great example, but just to illustrate a point. https://github.com/antlr/antlr4/issues/2577

DavidGregory084:
Hey @parrt @ericvergnaud is there anything I can do to move things along here?
If not, is this something that would be considered for ANTLR5, or does anybody have any thoughts about what would?

parrt:
@DavidGregory084 Unfortunately I can't take on any biggish antlr stuff at the moment as this project is on the backburner for me.

DavidGregory084:
No probs at all! Thanks for getting back, I will mentally park this for now 💤

PR #4356, opened 2023-07-14
Drop unstable golang.org/x/exp dependency

liggitt:
This PR replaces the only two uses of the `golang.org/x/exp` dependency with an inlined slice equality check.

The golang.org/x/exp module is [explicitly unstable](https://pkg.go.dev/golang.org/x/exp#section-readme), so it is beneficial to avoid depending on it in ways that pass those dependencies on to downstream users.


jimidle:
I have already done this. Alongside sone other changes. But they won’t hit
as a PR fir a while as I have some other changes I need to test too.

Thanks for your contribution!

PR #4363, opened 2023-07-19
[SWIFT] Add Antlr4Dynamic product

GrzegorzSpryszynski-TomTom:
SPM cannot always properly recognise dependency type (dynamic or static) so it would be useful to be able to explicitly specify if static or dynamic is needed.
This is the case when eg. dependency is provided not as a source code but rather as a binary and dependency is already fixed into the file.
SPM is not able to detect binary dependencies. There is also no way to specify type of binary's dependencies.

There is already Antlr4Static in Package.swift so dynamic equivalent should be added for completeness.

This PR brings back the [change](https://github.com/antlr/antlr4/pull/3669) that as introduced a year ago but was later overwritten by [dev to master merge](https://github.com/antlr/antlr4/commit/49b69bb31aa34654676a864b229a369680122470#diff-a8ff7c526b830308074a30da78f5627b54ce5a6ed6b6e21b28013057e99a5a06).

### DIFF
--- a/Package.swift
+++ b/Package.swift
@@ -12,6 +12,10 @@ let package = Package(
             name: "Antlr4Static",
             type: .static,
             targets: ["Antlr4"]),
+        .library(
+            name: "Antlr4Dynamic",
+            type: .dynamic,
+            targets: ["Antlr4"]),
     ],
     targets: [
         .target(


parrt:
Hi. how certain are you that this change is correct for all of the swift folks?

GrzegorzSpryszynski-TomTom:
> Hi. how certain are you that this change is correct for all of the swift folks?

Hi, it will be invisible for most of them. it doesn't brake current setups. It is logical addition to the current target `Antlr4Static`. And it was already [discussed](https://github.com/antlr/antlr4/pull/3669).
It will be usefull for small number of developers who uses binary targets together with package source and maybe others.

PR #4373, opened 2023-08-03
Add "override" before EOF in TS

GP4cK:
Currently, when targeting TS, ANTLR will generate a parser file with this code:
image

This PR adds the override keyword to fix this issue.


mike-lischke:
Obsolete with my patch #4403. Can be closed.

ericvergnaud:
@parrt blessed

PR #4383, opened 2023-08-13
Add module-info

bowbahdoe:
This adds a module-info through moditect in a multi-release jar. This is required for downstream consumers to be able to use the runtime with jlink without configuration.

## After
demo


obfischer:
Any update on this pull request? I would like to see JPMS support in ANTLR.

Issue #4386, opened 2023-08-17
[CSharp] ReportAttemptingFullContext() call passes null for config sets, thwarts attempts to debug grammars.


kaby76:
I don't know why but [ReportAttemptingFullContext is passed `null` for the last parameter, the ATN config set](https://github.com/antlr/antlr4/blob/832b202476a199028f50eb5517349090ed627166/runtime/CSharp/src/Atn/ParserATNSimulator.cs#L2289).

For Java, [the code correctly passes the ATN config set](https://github.com/antlr/antlr4/blob/832b202476a199028f50eb5517349090ed627166/runtime/Java/src/org/antlr/v4/runtime/atn/ParserATNSimulator.java#L2132).

I use CSharp as my development environment. Besides the fact that the tools in C# are much, much better than Java, all my tools in Trash are C# based.

I need the ATN config sets to get the prediction context in order to debug grammars. I can only get the value when overriding the DiagnosticErrorListener class.

I repeat myself again and again: the runtimes are not one-to-one, and contain many sutle differences, sometimes critical. Without access to the ATN config set, I cannot debug grammars--especially when they contain actions.

IDEs and plugins do not replace essential functionality that I require from the runtime.

PR #4387, opened 2023-08-17
[CSharp] Fir for #4386

kaby76:
This is a fix for #4386.

In this PR, I changed the signatures for ReportAttemptingFullContext() and ReportContextSensitivity() for the CSharp target to be identical to that for all the other targets. It seems that the code before this change was written with the idea to implement exposing the prediction context in another way, but it never got completed and never worked. This change exposes the ATN config sets so one can use them to create meaningful error messages and perform a better analysis of why a grammar is performing poorly.

There are no tests for this API change. In fact, there was no test for any of the API to display diagnostic errors, and it's unfair to require I write all these tests from scratch at this point. It should have been done years ago, with this [a commit with just one word for the description](https://github.com/antlr/antlr4/commit/10a15804481fb8aad397bb9eba0c5e78f4b15220).

### DIFF
--- a/doc/go-target.md
+++ b/doc/go-target.md
@@ -78,7 +78,7 @@ If you are already using the repo and import `github.com/antlr4-go/antlr/v4` the
 using the standard.
 
 ```shell
-go get -u get github.com/antlr4-go/antlr
+go get -u github.com/antlr4-go/antlr
 ```
 
 If you have not yet upgraded existing projects to the `/v4` module path, consult the section *Upgrading to v4
--- a/runtime/CSharp/src/IParserErrorListener.cs
+++ b/runtime/CSharp/src/IParserErrorListener.cs
@@ -122,7 +122,7 @@ namespace Antlr4.Runtime
         /// the simulator state when the SLL conflict was
         /// detected
         /// </param>
-        void ReportAttemptingFullContext(Parser recognizer, DFA dfa, int startIndex, int stopIndex, BitSet conflictingAlts, SimulatorState conflictState);
+        void ReportAttemptingFullContext(Parser recognizer, DFA dfa, int startIndex, int stopIndex, BitSet conflictingAlts, ATNConfigSet configs);
 
         /// <summary>
         /// This method is called by the parser when a full-context prediction has a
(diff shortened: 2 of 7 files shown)


runtime/CSharp/src/IParserErrorListener.cs
```diff
@@ -122,7 +122,7 @@ public interface IParserErrorListener : IAntlrErrorListener<IToken>
         /// the simulator state when the SLL conflict was
```
ericvergnaud, line 122: need to update the comment ?

runtime/CSharp/src/IParserErrorListener.cs
```diff
@@ -173,6 +173,6 @@ public interface IParserErrorListener : IAntlrErrorListener<IToken>
         /// the simulator state when the unambiguous prediction
```
ericvergnaud, line 173: need to update the comment ?

doc/go-target.md
```diff
@@ -78,7 +78,7 @@ If you are already using the repo and import `github.com/antlr4-go/antlr/v4` the
 using the standard.
 
 ```shell
-go get -u get github.com/antlr4-go/antlr
+go get -u github.com/antlr4-go/antlr
```
KvanTTT, line 81: This change doesn't relate to the MR title.

runtime/Python2/src/antlr4/xpath/XPath.py
KvanTTT, line 1: Please move out changes not related to C# to separated MRs.

KvanTTT (changes requested):

doc/go-target.md
kaby76: (Thanks. The focus was in the wrong window in Emacs, meant to be in a bash shell, and somehow saved it by mistake.)

runtime/Python2/src/antlr4/xpath/XPath.py
```diff
@@ -66,12 +66,12 @@ class XPath(object):
     WILDCARD = "*" # word not operator/separator
```
ericvergnaud, line 66: How does this file relate to the PR ?

ericvergnaud (changes requested):

kaby76:
Apparently, I mistakenly created the branch for the PR on "master". That explains all the weird file changes reported by Github for the PR. The cleaned-up PR is https://github.com/antlr/antlr4/pull/4399

ericvergnaud:
FYI it's possible to change a PR target after its creation

PR #4399, opened 2023-08-29
[CSharp] Fix for #4386 -- change signatures for ReportAttemptingFullContext() and ReportContextSensitivity() to be identical to all other targets

kaby76:
This is a fix for #4386: fix signatures of [ReportAttemptingFullContext()](https://github.com/antlr/antlr4/blob/a27cf84839c4f794e43b83b9b922d9342367fe45/runtime/CSharp/src/DiagnosticErrorListener.cs#L88C30-L88C57) and [ReportContextSensitivity()](https://github.com/antlr/antlr4/blob/a27cf84839c4f794e43b83b9b922d9342367fe45/runtime/CSharp/src/DiagnosticErrorListener.cs#L97) for the CSharp.

In this PR, I changed the signatures for ReportAttemptingFullContext() and ReportContextSensitivity() for the CSharp runtime to be identical to that for all the other runtimes. The code before this change appears to be written with the idea of exposing the prediction context in some other way. But, it seems the idea never was thought completely out, and what was implemented was basically useless null pointers being passed around for no reason. This change exposes the ATN config sets so I can then actually create meaningful error messages and perform a better analysis of why a grammar is performing poorly. Without the change, we are left with code that doesn't work the same way as all the other targets, and doesn't really help us to figure out why a grammar is bad.

Incidentally, you may be wondering why [ReportAmbiguity()](https://github.com/antlr/antlr4/blob/a27cf84839c4f794e43b83b9b922d9342367fe45/runtime/CSharp/src/DiagnosticErrorListener.cs#L74C30-L74C45) does not have the same problem with the signature as the others. It appears that came from initial check-in of the runtime for CSharp. https://github.com/antlr/antlr4/commit/e8c4bc4b0942f755d98f1dd49b63c5b4d8ceccb8#diff-46a0f5d3e67ea8c3d1c4fe92bd6fbbe9bcdfcf01fd05b6cb8bcc4286ac532488R99 

There are no tests for this API change. In fact, there were no tests for any of the API to display diagnostic errors for CSharp. And, it's unfair to require I write all these tests from scratch at this point. It should have been done years ago, with this [a commit with just one word for the description](https://github.com/antlr/antlr4/commit/10a15804481fb8aad397bb9eba0c5e78f4b15220).

### DIFF
--- a/runtime/CSharp/src/Atn/ParserATNSimulator.cs
+++ b/runtime/CSharp/src/Atn/ParserATNSimulator.cs
@@ -2286,7 +2286,7 @@ namespace Antlr4.Runtime.Atn
 								   ", input=" + parser.TokenStream.GetText(interval));
 			}
 			if (parser != null)
-				parser.ErrorListenerDispatch.ReportAttemptingFullContext(parser, dfa, startIndex, stopIndex, conflictingAlts, null /*configs*/);
+				parser.ErrorListenerDispatch.ReportAttemptingFullContext(parser, dfa, startIndex, stopIndex, conflictingAlts, configs);
 		}
 
 		protected virtual void ReportContextSensitivity(DFA dfa, int prediction, ATNConfigSet configs, int startIndex, int stopIndex)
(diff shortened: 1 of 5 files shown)


ericvergnaud:
lgtm

PR #4403, opened 2023-08-31
Patch to fix a number of TypeScript issues

mike-lischke:
This is a large patch to fix a large number of TypeScript issue, making this runtime now usable by more than just simple parse tasks.

Note: while the changes work well, there's currently a problem with the Maven tests. They print the message " java.lang.Exception: TypeScript ANTLR runtime is not initialized", which I need help with, as I don't know how the runtime is initialized in the test framework. 

@ericvergnaud, @parrt


mike-lischke:
@ericvergnaud please review!

mike-lischke:
fyi: I had to switch from ts-node to [tsx](https://www.npmjs.com/package/tsx), because I could not get ts-node to work in the ANTLR4 test (even though I use it successfully in other projects). For test execution it makes no difference, however.

ericvergnaud:
Hey Mike, sorry was away from my computer for a couple of days.

mike-lischke:
Yes, that happens, no worries. I had to close this PR because of serious conflicts after the 4.13.1 release. Will open a new one in a minute.

PR #4407, opened 2023-09-04
Improved mypy type-checking for generated Python3 files.

chwmilcon:
Antlr4's generated code for Python3 target had some misuse of __slots__. All the base objects were not using slots so derived objects from them didn't correctly enforce slots. __slots__ were added to base objects. Additionally, a number of forward declarations were made as class_name = None, this causes issues with mypy when checking code. Imports and other forward declarations were added to correct this. Furthermore, return types and argument types were added to functions within runtime to contribute to mypy type checking.  Emitted code was verified by generating a parser for java8 using python 3.11

### DIFF
--- a/runtime/Python3/src/antlr4/BufferedTokenStream.py
+++ b/runtime/Python3/src/antlr4/BufferedTokenStream.py
@@ -17,21 +17,18 @@ from io import StringIO
 from antlr4.Token import Token
 from antlr4.error.Errors import IllegalStateException
 
-# need forward declaration
-Lexer = None
-
 # this is just to keep meaningful parameter types to Parser
 class TokenStream(object):
-
-    pass
+    def LT(self, k:int) -> Optional[Token]:
+        pass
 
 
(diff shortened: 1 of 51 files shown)


ericvergnaud:
Hi. Thanks for this.
I'm seeing a lot of duplicate code in this PR, I guess this was machine generated ? 
It's not currently mergeable, missing DCO, and the python 3 build now fails...

chwmilcon:
Are you referring to the import statements being duplicated? Yes, those
were generated by monkeytype. Are there others? That might be a merge issue
on my side.

What's DCO? And I'm currently fixing the CI/CD failure. It appears that
runtime files need to be able to stand more independently than I tested
them.

ericvergnaud:
Yes I am referring to the duplicate imports
See https://github.com/antlr/antlr4/blob/dev/CONTRIBUTING.md for contributing policy
DCO means you need to sign-off your commits

chwmilcon:
This branch appears incomplete. Separation from private branch not done correctly.

PR #4409, opened 2023-09-05
Clean up of the JS/TS target

mike-lischke:
@ericvergnaud OK, let's start with the non-functional clean up. This patch only contains things that improve formatting + copyrights, as well as a smaller production build. Should be easy to agree on these changes.

Would appreciate a quick review, so I can continue with the following patches.


runtime-testsuite/test/org/antlr/v4/test/runtime/typescript/TsNodeRunner.java
```diff
@@ -35,7 +35,7 @@ protected void initRuntime(RunOptions runOptions) throws Exception {
 	}
 
 	private void npmInstallTsNodeAndWebpack() throws Exception {
-		Processor.run(new String[] {NPM_EXEC, "--silent", "install", "-g", "typescript", "ts-node", "webpack", "webpack-cli"}, null);
+		Processor.run(new String[] {NPM_EXEC, "--silent", "--force", "install", "-g", "typescript", "npx", "tsx", "webpack", "webpack-cli"}, null);
```
ericvergnaud, line 38: We're not switching to tax. According to their web page, they don't use tsc, and don't perform type checking. 
Tsc compatibility is a requirement.

runtime/JavaScript/src/antlr4/CharStream.d.ts
```diff
@@ -1,5 +1,9 @@
-export declare class CharStream {
+/* Copyright (c) 2012-2022 The ANTLR Project. All rights reserved.
```
ericvergnaud, line 1: it's 2023 :-)

runtime/JavaScript/src/antlr4/CharStreams.d.ts
```diff
@@ -1,4 +1,9 @@
-import {CharStream} from "./CharStream";
+/* Copyright (c) 2012-2022 The ANTLR Project. All rights reserved.
+ * Use of this file is governed by the BSD 3-clause license that
+ * can be found in the LICENSE.txt file in the project root.
+ */
+
+import CharStream from "./CharStream.js";
```
ericvergnaud, line 6: We're not moving away from typescript style imports. Not sure what problem this is trying to solve but in my experience it's a recipe for problems.

runtime/JavaScript/src/antlr4/CommonToken.js
```diff
@@ -1,3 +1,9 @@
+/*
+ * Copyright (c) 2012-2017 The ANTLR Project. All rights reserved.
+ * Use of this file is governed by the BSD 3-clause license that
+ * can be found in the LICENSE.txt file in the project root.
+ */
+
 import Token from "./Token.js";
```
ericvergnaud, line 7: Please remove the extension

runtime/JavaScript/src/antlr4/InputStream.d.ts
```diff
@@ -1,4 +1,12 @@
+/*
+ * Copyright (c) 2012-2017 The ANTLR Project. All rights reserved.
+ * Use of this file is governed by the BSD 3-clause license that
+ * can be found in the LICENSE.txt file in the project root.
+ */
+
 export declare class InputStream {
-    constructor(data: string);
-    constructor(data: string, decodeToUnicodeCodePoints: boolean);
+    public constructor(data: string);
```
ericvergnaud, line 8: why does this constructor need to be public when the others don't ?

runtime/JavaScript/src/antlr4/TokenStream.d.ts
```diff
@@ -1,16 +1,26 @@
...
+ */
+
+import Token from "./Token.js";
+import TokenSource from "./TokenSource.js";
+import Interval from "./misc/Interval.js";
 
 export declare class TokenStream {
+    public tokenSource: TokenSource;
+    public index: number;
+    public get size(): number;
```
ericvergnaud, line 13: using a getter is left to the implementation, from a contract pov it's a field

runtime/JavaScript/src/antlr4/atn/ParserATNSimulator.d.ts
```diff
@@ -1,21 +1,26 @@
...
+export declare class ParserATNSimulator extends ATNSimulator {
+    public predictionMode: PredictionMode;
+    public decisionToDFA: DFA[];
+    public atn: ATN;
+    public debug?: boolean;
+    public trace_atn_sim?: boolean;
 
-    constructor(recog: Recognizer<Token>, atn: ATN, decisionToDFA: DFA[], sharedContextCache: PredictionContextCache);
-    adaptivePredict(input: TokenStream, decision: number, outerContext: ParserRuleContext) : number;
+    public constructor(recog: Parser, atn: ATN, decisionToDFA: DFA[], sharedContextCache: PredictionContextCache);
```
ericvergnaud, line 22: again, why public here ?

runtime/JavaScript/src/antlr4/atn/index.d.ts
```diff
@@ -1,9 +1,14 @@
...
-export * from './ATNDeserializer';
-export * from './LexerATNSimulator';
-export * from './ParserATNSimulator';
-export * from './PredictionMode';
-export * from './PredictionContextCache';
+/*
+ * Copyright (c) 2012-2017 The ANTLR Project. All rights reserved.
+ * Use of this file is governed by the BSD 3-clause license that
+ * can be found in the LICENSE.txt file in the project root.
+ */
```
ericvergnaud, line 6: ts imports should not have a suffix

runtime/JavaScript/src/antlr4/tree/ParseTreeVisitor.d.ts
```diff
@@ -1,16 +1,18 @@
...
+import ErrorNode from "./ErrorNode.js";
+import TerminalNode from "./TerminalNode.js";
+import ParseTree from "./ParseTree.js";
 
-    visitTerminal(node: TerminalNode): Result;
-
-    visitErrorNode(node: ErrorNode): Result;
+export declare class ParseTreeVisitor<Result> {
+    public visit(tree: ParseTree): Result;
+    public visitChildren(node: ParseTree): Result;
```
ericvergnaud, line 12: a node with children is necessarily a ruleNode, please revert

runtime/JavaScript/src/antlr4/tree/Trees.js
```diff
@@ -16,57 +17,57 @@ const Trees = {
...
         ruleNames = ruleNames || null;
         recog = recog || null;
-        if(recog!==null) {
+        if (recog !== null) {
             ruleNames = recog.ruleNames;
         }
-        if(ruleNames!==null) {
-            if (t instanceof RuleNode) {
+        if (ruleNames !== null) {
+            if (t instanceof RuleContext) {
```
ericvergnaud, line 52: this change is incorrect, it might a future bug in the java runtime i.e. it so happens that currently only RuleContext implements RuleNode, but that could change

ericvergnaud (changes requested):

runtime-testsuite/test/org/antlr/v4/test/runtime/typescript/TsNodeRunner.java
mike-lischke: Using tsx is just an attempt to get the build working. As you saw ts-node failed in GH actions and it fails for me locally. I use ts-node for other projects and know it a bit, but it still gives griefs, while tsx works out of the box.

runtime/JavaScript/src/antlr4/CharStream.d.ts
mike-lischke: My last information about copyrights in ANTLR4 is that we always keep the original ones and don't update any year, but maybe that only belongs to the first year, not the second one.

runtime/JavaScript/src/antlr4/CharStreams.d.ts
mike-lischke: What is a "typescript import"? This is a normal JS default import as being used in many other places in the JS runtime.

runtime/JavaScript/src/antlr4/CommonToken.js
mike-lischke: Two points here:

1) The .js extension is used already in the JS runtime in many places.
2) The .js extension is required for [full ESM support](https://nodejs.org/api/esm.html#mandatory-file-extensions).

runtime/JavaScript/src/antlr4/InputStream.d.ts
mike-lischke: If there's no access modifier then the default is public. But making it explicit which visibility a member has is considered good coding style. In JS it doesn't really matter, as there is no public or private way (the `#member' notation is an addendum to improve this situation), but for TS it makes sense as the transpiler enforces the correct access.

runtime/JavaScript/src/antlr4/TokenStream.d.ts
mike-lischke: It's not. A single getter makes this member readonly (and shows there's potentially more processing than just a single field). Why hiding such information, if the language allows that?

runtime/JavaScript/src/antlr4/CharStreams.d.ts
ericvergnaud: this file is a ts file, not a js file.
As such it needs to follow ts import conventions i.e. no suffix.
It makes no sense to import a js file in a d.ts file in the first place, since we're only looking to import types, not implementations. Think of the d.ts files as a set of files that need to live somewhat independently of the js ones, because the js ones get babelized, uglified and ultimately packaged - they might not exist in the distribution

runtime/JavaScript/src/antlr4/CommonToken.js
ericvergnaud: Yes sorry, this is a js file

runtime/JavaScript/src/antlr4/tree/ParseTreeVisitor.d.ts
mike-lischke: No, a node with children is a (Tree)[https://github.com/antlr/antlr4/blob/dev/runtime/Java/src/org/antlr/v4/runtime/tree/Tree.java], but that's an interface in Java and there's no real sense to separate the 3 interfaces `Tree`, `SyntaxTree` and `ParseTree`. The first two don't add anything useful to `ParseTree` hence they are combined into a single interface.

The `RuleNode` is another interface which only adds a single member. There's only one type which uses `RuleNode`, which is `RuleContext`, so it makes sense to remove `RuleNode` and directly work with `RuleContext`. Why keep a class/interface which is not being used?

runtime/JavaScript/src/antlr4/tree/Trees.js
mike-lischke: We are not programming for the future. What you think might never come. What we do is to solve the current problems.

runtime/JavaScript/src/antlr4/tree/ParseTreeVisitor.d.ts
ericvergnaud: "it makes sense" is an opinion. We just have different coding styles, and I don't change working code to satisfy coding style, otherwise I would change the code 10 times per day.

runtime/JavaScript/src/antlr4/tree/Trees.js
ericvergnaud: Well there isn't any problem with using RuleNode, so why are you proposing this change ?

runtime/JavaScript/src/antlr4/tree/ParseTreeVisitor.d.ts
mike-lischke: Then read "make sense" as "there's a good reason". It's not personal coding style, but something useful. You also have left out things in the current runtime that you have no use for. Btw. I did the same change in the C++ runtime in 2016. The 3 different interfaces are simply over-engineered and are useless in the context of the JS/TS runtime.

runtime/JavaScript/src/antlr4/CharStreams.d.ts
ericvergnaud: This is irrelevant here, but FYI for normal TS files, the extension is also not necessary and should be avoided, unless the imported file is actually a JS one. See for example https://github.com/hyperledger/aries-framework-javascript/blob/main/packages/core/src/agent/Agent.ts.

runtime/JavaScript/src/antlr4/InputStream.d.ts
ericvergnaud: "the transpiler enforces the correct access" -> you mean the type checker ? The transpiler drops this since it's not supported by JS.

runtime/JavaScript/src/antlr4/tree/ParseTreeVisitor.d.ts
ericvergnaud: I happen to find it more useful to show that the 3 types of nodes are covered by the if/else if... which is why I wrote it like that in the first place.

runtime-testsuite/test/org/antlr/v4/test/runtime/typescript/TsNodeRunner.java
mike-lischke: webpack doesn't do any type checking either. During development you would need an extra type checker anyway. VS Code has a project wide TS type checker option, which makes this super easy.

Anyway, I cannot change that back or I will not be able to run the ANTLR4 tests locally.

runtime/JavaScript/src/antlr4/CharStream.d.ts
mike-lischke: Once I have a definite answer from @parrt I'll update them.

runtime/JavaScript/src/antlr4/CharStreams.d.ts
mike-lischke: A link to a random framework doesn't help here. I thought I had provided the link, but cannot find it here. So here it is again. Full ESM support [requires the extension](https://nodejs.org/api/esm.html#mandatory-file-extensions) in Node.js (in the browser they are required anyway). It's the bundler that automatically handles that, but when running with tools like ts-node or tsx they are required.

Hence I'm not going to remove the extensions again.

runtime-testsuite/test/org/antlr/v4/test/runtime/typescript/TsNodeRunner.java
ericvergnaud: ts-code runs tsc that does type checking.

runtime/JavaScript/src/antlr4/tree/ParseTreeVisitor.d.ts
mike-lischke: Can you point me to a place in source code where you test, say, on `Tree`? How often do you check for `ParseTree` and for `SyntaxTree`? I cannot find any place where `Tree` or `SyntaxTree` is used in JS/TS code, except to define `ParseTree`.

And even in the other targets there's only very few use of `Tree`, namely in the `Trees` file (only Java, Dart and Swift) and in the Java UI tool, which is not relevant here.

I think these are enough points to warrant a simplification of this small corner of the ANTLR4 type tree. That's why I have no intention to revert that.

runtime/JavaScript/src/antlr4/CharStreams.d.ts
ericvergnaud: The link you provide is for node i.e. javascript, not for typescript.
This is [the official reference](https://www.typescriptlang.org/docs/handbook/modules.html#import) and you'll read that they don't use extensions except for side effects.

mike-lischke: But ultimately all TS files end up as JS files, so this applies anyway. If you are against including the .js extension, why did you include them in the existing JS/TS runtime? You even use the extension in the test cases (see Test.ts.stg).

The main point here is: when using tools like ts-node to run TS directly you need the extensions or they will not find the imports. I've been through that for a number of times already.

We are wasting our time here on something which is an absolute minor aspect and I have good reasons to keep the extensions included.

OskarDamkjaer: @mike-lischke The docs page you link is a little confusing, it makes it seem like the it uses a ".js" extension for side effects only but that is not the case. 

In fact the typescript documentation has a section on file extensions in imports [here](https://www.typescriptlang.org/docs/handbook/esm-node.html), where they explain why the `.js` extension is needed to support both commonjs and esm for node when using typescript.

PR #4411, opened 2023-09-06
[Javascript] Fix broken package entrypoints

Codex-:
Currently as of 4.13.1 the entrypoint for `browser` is invalid, and `main` specifies the `esm` build instead of the `commonjs` one.

This results in commonjs based sources attempting to load esm syntax when addressing the sources using `main`.

The error in `browser` causes this error when attempting to bundle the sources:
```
[commonjs--resolver] Failed to resolve entry for package "antlr4". The package may have incorrect main/module/exports specified in its package.json.
error during build:
Error: Failed to resolve entry for package "antlr4". The package may have incorrect main/module/exports specified in its package.json.
```

### Fixes
- `browser` field updated to fix the extension
  - `"dist/antlr4.web.js"` -> `"dist/antlr4.web.mjs"`
- `main` field updated to point to the commonjs bundle
  - "dist/antlr4.node.mjs"` -> `"dist/antlr4.node.cjs"`

### Added
- `module` field to point to the esm bundle
  - `"module": "dist/antlr4.node.mjs"`

More information on `main` vs `module` can be found here: https://nodejs.org/api/packages.html#dual-commonjses-module-packages

### DIFF
--- a/runtime/JavaScript/package.json
+++ b/runtime/JavaScript/package.json
@@ -3,8 +3,9 @@
   "version": "4.13.1",
   "type": "module",
   "description": "JavaScript runtime for ANTLR4",
-  "browser": "dist/antlr4.web.js",
-  "main": "dist/antlr4.node.mjs",
+  "browser": "dist/antlr4.web.mjs",
+  "main": "dist/antlr4.node.cjs",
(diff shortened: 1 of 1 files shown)


ericvergnaud:
Thanks for this
@parrt blessed

Issue #4310, opened 2023-06-10
[Java] Use Java 21 sealed class and pattern matching to take place the visitor pattern


XiLaiTL:
- [ ] I have reproduced my issue using the latest version of ANTLR
- [ ] I have done a search of the existing issues to make sure I'm not sending in a duplicate

 - target Java

Similar issue: #3282 
 https://github.com/antlr/antlr4/issues/3282#issue-1003867751

As Java 21 will be the next LTS version and it features the pattern matching for sealed class.

It will be possible that use this feature to take place the older visitor model in parser to build the ADT.
Or it will be a new option `-sealed` to generate Matcher.java like the Listener.java and Visitor.java for the new Java version. 

This way, the simple example will be like:

Calc.g4
```antlr
grammar Calc;
stat: ID '=' expr ';';
expr: NUM                   #singleExpr
    | ID                    #identifierExpr
    | expr ('+'|'-') expr   #addSubExpr
    | expr ('*'|'%') expr   #mulDivExpr
    | '(' expr ')'          #parenExpr
    ;
ID: [a-z]+;
NUM: [0-9]+;
```

CalcSealedParser.java
```java

	public sealed static class ExprContext extends ParserRuleContext permits SingleExprContext, AddSubExprContext, MulDivExprContext, ParenExprContext, IdentifierExprContext {

	}

	public final static class SingleExprContext extends ExprContext {

	}

	public final static class AddSubExprContext extends ExprContext {

	}

	public final static class MulDivExprContext extends ExprContext {

	}

	public final static class ParenExprContext extends ExprContext {

	}

	public final static class IdentifierExprContext extends ExprContext {

	}
```
Or write without permits but embed in class

```java
	public sealed static class ExprContext extends ParserRuleContext{

		public final static class SingleExprContext extends ExprContext {

		}
		
		public final static class AddSubExprContext extends ExprContext {

		}
		
		public final static class MulDivExprContext extends ExprContext {

		}
		
		public final static class ParenExprContext extends ExprContext {

		}
		
		public final static class IdentifierExprContext extends ExprContext {

		}

	}
```

And then, use pattern matching to replace the visit.

PR #4419, opened 2023-09-18
Added an option to generate the context classes final.

lkishalmi:
Well, I was trying to test some Java 21 features, when it hit me that the generated Context classes are not final, so I can't simply just announce a sealed super class for them.

This is a trivial fix for the issue. Other languages could benefit from this `final`-ization as well, however I'm not good enough in them, to take that responsibility.

Might be a step for #4310


lkishalmi:
@parrt Any chance to get this one reviewed?

parrt:
Hi @lkishalmi It sounds reasonable but my mind it has been elsewhere for a while and I just don't know the latest java stuff; I'm also sure how it should be broadcast across other targets. Sorry!

s7oev:
hey @parrt - do you think there is any chance of moving forward with this and #4310? based on the thumbs up of the issue, it sees like there is some interest in this

PR #4425, opened 2023-09-22
Fix for #4406 -- add in renaming for reserved word "accept"

### LINKED ISSUE #4406: [Dart, Go] "accept" in parser grammar conflicts with target.
The id "accept" should be added to the code for the Cpp, Dart, and Go target conflicts.

kaby76:
This PR fixes #4406.

The reserved word tables for Go and Dart were updated so that grammars with a parser rule name "accept" will not cause problems. This was found with a recent update to the PCRE grammar. https://github.com/antlr/grammars-v4/pull/3690


tool/src/org/antlr/v4/codegen/target/DartTarget.java
```diff
@@ -36,7 +36,9 @@ public class DartTarget extends Target {
 		"deferred", "hide", "return", "with",
 		"do", "if", "set", "yield",
 
-		"rule", "parserRule"
+                "rule", "parserRule",
+                "accept"
```
KvanTTT, line 40: Please fix indentation and use tabs instead of spaces.

KvanTTT (changes requested):

tool/src/org/antlr/v4/codegen/target/DartTarget.java
kaby76: I will make the change, but we really should not use tabs. Tab expansion display is dependent on the editor used. In GNU Emacs, it's 8 spaces. For Github, it's 4 spaces. For VSCode, 4. For vim, 8. It is well known: [tabs are evil](https://www.emacswiki.org/emacs/TabsAreEvil). There are many places where spaces are used for indentation already. Is there a coding style doc for Antlr?

KvanTTT: > I will make the change, but we really should not use tabs.

Sorry, but historically ANTLR code base was written with tabs and currently it's not rational to reformat all files. But much worse is to mix spaces and tabs withing the same file.

>  Is there a coding style doc for Antlr?

For java and stg files indentation style is specified: [indent_style = tab](https://github.com/antlr/antlr4/blob/dev/.editorconfig#L10).

>  Tab expansion display is dependent on the editor used. In GNU Emacs, it's 8 spaces. For Github, it's 4 spaces. For VSCode, 4. For vim, 8.

I suppose all tools should support [editorconfig](https://github.com/antlr/antlr4/blob/dev/.editorconfig#L4). Width of tab is specified there and it's 4 spaces.

PR #4428, opened 2023-09-26
fix: TypeScript: incorrect type in CommonTokenStream

johannesheesterman:


ericvergnaud:
Thanks for this!
@parrt blessed

PR #4444, opened 2023-10-17
Fix description of \uXXXX escape sequences

akosthekiss:
The number of hex digits to follow \u is exactly four, not up to 4.


akosthekiss:
Is there anything I should add to the PR?

akosthekiss:
It seems that the documentation is still slightly imprecise. Gentle ping for a fix?

akosthekiss:
Let me try and bring some attention to this tiny doc patch with this comment. I know that I may be the only one who is bugged by this minuscule wording issue. But I recall that it was misleading me back then -- so I just wanted to prevent others getting misled. Any spare blessings perchance?

Issue #4439, opened 2023-10-13
Generated Go parser fails "go vet"


SeanPMiller:
JDK version: OpenJDK 17
ANTLR version: v4.13.1
Target language: Go
Go version: v1.21.3

Reproduce with the following:
```$ cat Foo.g4
grammar Foo;
bar : 'frobozz' | 'xyzzy' ;
$ java org.antlr.v4.Tool -Dlanguage=Go Foo.g4 
$ go vet foo_parser.go foo_listener.go 
# command-line-arguments
./foo_parser.go:179:2: unreachable code
$ sed '179q;d' foo_parser.go 
	goto errorExit // Trick to prevent compiler error if the label is not used
$```

Please let me know if this is PEBCAK on my part.

Thanks for your time and attention.

SeanPMiller:
https://github.com/antlr/antlr4/blob/dev/tool/resources/org/antlr/v4/tool/templates/codegen/Go/Go.stg#L436

SeanPMiller:
@jimidle 
https://github.com/antlr/antlr4/commit/adbe946416f5f6d794ff4c3322739c3c5c20a90c

PR #4445, opened 2023-10-17
Band-Aid "go vet" error with Go codegen template

SeanPMiller:
Resolves https://github.com/antlr/antlr4/issues/4439

This may not work forever. If it ever stops working, then we can switch to using a `const` variable, as  that is how Rob Pike & Co. intend for debugging-logic to work. Refer to the following issue:

https://github.com/golang/go/issues/16370

### DIFF
--- a/tool/resources/org/antlr/v4/tool/templates/codegen/Go/Go.stg
+++ b/tool/resources/org/antlr/v4/tool/templates/codegen/Go/Go.stg
@@ -432,8 +432,10 @@ errorExit:
 	<finallyAction>
 	<endif>
 	p.ExitRule()
+	if false {
+		goto errorExit // Trick to prevent compiler error if the label is not used
+	}
 	return localctx
-	goto errorExit // Trick to prevent compiler error if the label is not used
 }
 >>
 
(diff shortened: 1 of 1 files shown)


simeonmiteff:
To any fellow travelers wondering how to get this (in lieu of this PR being merged), pipe your generated code through:
```bash
  tr '\n' '\f' | \
  sed 's|return localctx\f\tgoto errorExit // Trick to prevent compiler error if the label is not used|if false {\f          goto errorExit // Trick to prevent compiler error if the label is not used\f    }\f    return localctx|g' | \
  tr '\f' '\n'
```

SeanPMiller:
Given what a showstopper bug this is for Go developers, I'm surprised it has been languishing unmerged.

jimidle:
Why woudl you run go vet on generated code though?

SeanPMiller:
Really. Have you tried to `go vet` a Go module that depends on another module generated by ANTLR?

SeanPMiller:
I'll excuse your obnoxious comment.

jimidle:
Yeah. Sorry - I did not read this correctly. I will add it, but I am just overrun at the moment and have been for a whlie.

SeanPMiller:
Understood. I've been there. Are y'all short on volunteers?

vecknishwaran:
Hi, just checking in on the status of this PR. Are there any plans to merge it for now?

kentyman23:
Please merge this! It looks simple and safe and makes `go vet` finally work.

PR #4450, opened 2023-10-23
[PHP] fix: make BaseListener abstract by default

dkarlovi:
The `BaseListener` does nothing by default, it should be marked as abstract to signal it's not meant to be used directly (because it makes no sense to do so) but extended.


dkarlovi:
Failure unrelated.

PR #4454, opened 2023-10-28
docs: fix wrong way to import in go

KevinZonda:


jimidle:
@parrt This PR is fine Ter

PR #4480, opened 2023-11-30
Fix indirect left recursive rule operator precedence via delegated operator precedence

### LINKED ISSUE #4460: left recursive rule references and operator precedence
Hey there,

when using left recursive grammars with operator precedence, I am unable to split my left recursive rule into multiple ones while maintaining the operator precedence.
 
For example:
```g4
grammar TestGrammar;

start : expr EOF;
expr : notexpr | expr '&&' expr | ID;
notexpr : '!' expr;

ID: [a-zA-Z_][a-zA-Z_0-9]* ;
```

An input string of `!a && a` results in a parse tree is the equivalent to `!(a && a)`, ignoring the precedence. I would have expected a result such as `(!a) && (a)` instead.

The reason being, that operator precedence is only used with ordered alts and reset to 0 for references.
This is a snippet from the generated (Java) parser of this grammar:

```java
	private ExprContext expr(int _p) throws RecognitionException {
		enterRecursionRule(_localctx, 2, RULE_expr, _p);
		try {
			int _alt;
			enterOuterAlt(_localctx, 1);
			{
			setState(12);
			_errHandler.sync(this);
			switch (_input.LA(1)) {
			case EXCL:
				{
				setState(10);
				notexpr();        // no precedence
				}
				break;
			case ID:
```

And here, the same part for a grammar where notexpressions are inlined/not hidden behind another production.

```java
	private ExprContext expr(int _p) throws RecognitionException {
		enterRecursionRule(_localctx, 2, RULE_expr, _p);
		try {
			int _alt;
			enterOuterAlt(_localctx, 1);
			{
			setState(13);
			_errHandler.sync(this);
			switch (_input.LA(1)) {
			case EXCL:
				{
				setState(10);
				match(EXCL);
				setState(11);
				expr(3);     // precedence of 3
				}
				break;
			case ID:
```

The reason for hiding productions behind another is that we/I derive grammars from other grammars, resulting in ... larger grammars.
Due reusing productions (such as expressions) and the need to parse productions stand alone (e.g., a not expression), the generated java parsers are reaching the limits allowed in Java.

I spend some time to think about a possible solution in ANTLR: 
My suggesting is to introduce a "delegated/indirect operator precedence" parameter to some methods,
 which passes the operator precedence through (such as for the not expression in the example above).  

I've only implemented it for Java right now to receive some feedback:
https://github.com/antlr/antlr4/compare/dev...MisterErwin:antlr4:ind-operator-prec

(It contains three smaller tests, but I also ran some larger tests with the root cause of this issue :smile: )

In case the change seems okay, I would go ahead and prepare a PR for all languages (and not just a prototype in Java). 

Thanks
Alex

P.S.: @  kaby76, thanks for helping me with this over on stackoverflow - sorry for not being able to upvote the answers.

### LINKED ISSUE #4477: [Swift] only first rule parameter generated correctly
the following grammar generates invalid code for `rule2` and `rule3`:

```
grammar T;

rule1[Int arg1]
    :   .
    ;

rule2[Int arg1, Int arg2]
    :   .
    ;

rule3[Int arg1, Int arg2, Int arg3]
    :   .
    ;
```

=>

```
open func rule2(_ arg1: Int!, _arg2: Int!) throws -> Rule2Context {
open func rule3(_ arg1: Int!, _arg2: Int!, _arg3: Int!) throws -> Rule3Context {
```
it should generate something like:
```
open func rule2(_ arg1: Int!, _ arg2: Int!) throws -> Rule2Context {
open func rule3(_ arg1: Int!, _ arg2: Int!, _ arg3: Int!) throws -> Rule3Context {
```
(space between "_" and "arg2" / "arg3"

MisterErwin:
Resolves #4460 

Operator precedence is currently (incorrectly) reset to zero, when a left-recursive rule contains other (non-left)recursive elements.

```g4
grammar T
e : m | e '*' e * ID;
m: '-' e;
ID : 'a'..'z'+ ;
```

The ATN for rule e is shown below:
ATN of rule e

In case no indirect recursion was present (`e : '-' e; | e '*' e * ID;`), the ATN would look like the following:

ATN of rule `e: '-' e | e'*' ...  `

Notably, without indirectness, `<e[2]>` signals a precedence used.

For completeness, the ATN of rule m is shown below too:
ATN of rule m

The `<e[0]>` (or in Java: `e(0);`) in rule `m` resets the operator precedence to zero, which is not correct (see issue #4460 or the tests added by this PR).

## Proposed solution:

In case we detect a rule reference resulting in a binary or prefix recursion, we introduce an optional parameter `int _dp` (similar to the `p` parameter of left recursive rules) with a default of `0` to the targeted rule.
(This would replace the old `<e[0]>` by a `<e[_dp]>`.)

This PR does not touch the ATN behavior of ANTLR (well.. almost not. The possible precedence value `_dp` is now excluded during the ATN creation, but where this value is now set, any precedence was 0 before - yay). 

This MR also introduces tests for additional attributes for rules (but not for left recursive ones, because ANTLR outputs an error message for these: `rule e is left recursive but doesn't conform to a pattern ANTLR can handle`)

The precedence option for indirect left-recursive rules is now set to `"_dp"`, which is a new parameter introduced (similar to the existing `_p`).

## Breaking changes:
This PR would prevent optional parameters from being added to parsers for the Dart target via the grammar.
Normal parameters are not affected, thus I would argue that change to be not of importance.

The reason I found this bug at all was that with large grammars, 
 I wanted to use the same rules (such as formulas) in multiple other rules.

The expected token position of the TestSymbolIssues#testUndefinedLabel test had to be modified by 5 tokens (`(= dp 1)`).

Thanks to Ken (@kaby76) for the initial directions over on [SO](https://stackoverflow.com/questions/77173636/antlr-left-recursive-rule-precedence) 

(By accident this also fixes #4477 :) )

P.S.: I've run a bunch of tests on internal grammars generated using the changes proposed, which so far has not shown additional errors.

P.P.S.: I am not sure if a RuleWalker similar to the `LeftRecursiveRuleWalker` instead of my `LeftRecursiveRuleTransformer#isBinaryOrPrefix` method would have been the better option - (~~but my version is at least easier to step through while debugging~~)

P.P.P.S.: The DOTGenerator is not aware of the presence of the `_dp` value for rererences - if desired, I can add this functionality (but wanted to refrain from doing so before the other changes were discussed.)

### DIFF
--- a/tool/resources/org/antlr/v4/tool/templates/codegen/Dart/Dart.stg
+++ b/tool/resources/org/antlr/v4/tool/templates/codegen/Dart/Dart.stg
@@ -317,7 +317,7 @@ bool _<r.name>_sempred(dynamic _localctx, int predIndex) {
 
 RuleFunction(currentRule,args,code,locals,ruleCtx,altLabelCtxs,namedActions,finallyAction,postamble,exceptions) ::= <<
 
-<if(currentRule.modifiers)><currentRule.modifiers:{f | <f> }><else><endif><currentRule.ctxType> <currentRule.escapedName>(<[args, currentRule.hasDelegatedPrecedence:{_|{int dp=0\}}]; separator=", ">) {
+<if(currentRule.modifiers)><currentRule.modifiers:{f | <f> }><else><endif><currentRule.ctxType> <currentRule.escapedName>(<[args, currentRule.hasDelegatedPrecedence:{_|[int dp=0]}]; separator=", ">) {
   dynamic _localctx = <currentRule.ctxType>(context, state<currentRule.args:{a | , <a.escapedName>}>);
   enterRule(_localctx, <currentRule.startState>, RULE_<currentRule.name>);
   <namedActions.init>
@@ -541,7 +541,7 @@ cases(tokens) ::= <<
 
 InvokeRule(r, argExprsChunks) ::=<<
 state = <r.stateNumber>;
-<if(r.labels)><r.labels:{l | <labelref(l)> = }><endif><r.escapedName>(<[{<if(r.ast.options.pv)>dp<else><r.ast.options.p><endif>}, argExprsChunks, r.ast.options.dp:{v|dp:<v>}]; separator=",">);
+<if(r.labels)><r.labels:{l | <labelref(l)> = }><endif><r.escapedName>(<[{<if(r.ast.options.pv)>dp<else><r.ast.options.p><endif>}, argExprsChunks, r.ast.options.dp]; separator=",">);
 >>
 
 MatchToken(m) ::= <<


n-dragon:
any update on when this will be merged ?

MisterErwin:
> any update on when this will be merged ?
Some aspects of this PR offer up some discussion (mainly my decicion on the RuleWalker and DOTGenerator),
 other than I am just waiting for feedback (see also [this discussion](https://github.com/antlr/antlr4/discussions/4514#discussion-6089670)).
 - which is also why haven't resolved the merge conflict.

For my usages, I'm either working around this issue by preprocessing the input grammars or by patching these changes into antlr.

> side question, how did you generate the ATN ?
The -atn CLI option outputs the ATN in graphviz/dot format (which you can render using a graphviz or some websites)

n-dragon:
> Some aspects of this PR offer up some discussion (mainly my decicion on the RuleWalker and DOTGenerator),
> other than I am just waiting for feedback (see also [this discussion](https://github.com/antlr/antlr4/discussions/4514#discussion-6089670)).
> Which is also why haven't re

thanks for the answer,
may I have your opinion on this discussion?
https://github.com/antlr/antlr4/discussions/4686

PR #4171, opened 2023-03-10
Cpp: Fix the unused parameter warning in the sempred function.

0xFireWolf:
Consider the following simple grammar:

```antlr
Expression : Expression '*' Expression
           | Expression '+' Expression
```

ANTLR will generate the following member function:

```c++
bool SampleParser::expressionSempred(ExpressionContext *_localctx, size_t predicateIndex)
{
    switch (predicateIndex) 
    {
        case 0: return precpred(_ctx, 5);
        case 1: return precpred(_ctx, 4);

        default:
        break;
    }
    return true;
}
```

The first parameter is never used in the function body, resulting in an unused parameter warning. This PR fixes the warning by adding the attribute `[[maybe_unused]]` which is available as of C++17.

Thank you.

PR #4488, opened 2023-12-11
Fix warning: unused parameter 'context' [-Wunused-parameter].

rturrado:
I've just commented out the unused parameter as I've seen you do in other parts of the code, with /*variableName*/.


runtime/Cpp/runtime/src/tree/xpath/XPathLexer.cpp
```diff
@@ -155,7 +155,7 @@ void XPathLexer::action(RuleContext *context, size_t ruleIndex, size_t actionInd
   }
 }
 
-void XPathLexer::IDAction(antlr4::RuleContext *context, size_t actionIndex) {
+void XPathLexer::IDAction(antlr4::RuleContext */*context*/, size_t actionIndex) {
```
jmairboeck, line 158: This file is generated. I added a comment #4171 (which is about a similar issue) to fix the generator template instead.

rturrado: Great, many thanks!

Issue #4041, opened 2022-12-24
Python runtime test failure with 4.10 and later


mweinelt:
After upgrading the antlr4 python runtime past 4.9.1 (tested 4.10.1 and 4.11.1) in nixpkgs, we've been seeing its test suite fail with

```pytb
Traceback (most recent call last):
  File "/build/source/runtime/Python3/tests/ctest.py", line 10, in <module>
    from parser.cparser import CParser
  File "/build/source/runtime/Python3/tests/parser/cparser.py", line 631, in <module>
    class CParser ( Parser ):
  File "/build/source/runtime/Python3/tests/parser/cparser.py", line 635, in CParser
    atn = ATNDeserializer().deserialize(serializedATN())
  File "/nix/store/ch8i929c63av55h9nxkinifh61mazf1h-python3.10-antlr4-python3-runtime-4.11.1/lib/python3.10/site-packages/antlr4/atn/ATNDeserializer.py", line 28, in deserialize
    self.checkVersion()
  File "/nix/store/ch8i929c63av55h9nxkinifh61mazf1h-python3.10-antlr4-python3-runtime-4.11.1/lib/python3.10/site-packages/antlr4/atn/ATNDeserializer.py", line 50, in checkVersion
    raise Exception("Could not deserialize ATN with version " + str(version) + " (expected " + str(SERIALIZED_VERSION) + ").")
Exception: Could not deserialize ATN with version  (expected 4).
```

The version returned seems to be `\x03`, while the test suite expects it to be `4`.

While similar to #3997 and #3895 we are running the test suite, not our own code.

mweinelt:
Looking at the test directory we should probably execute `run.py` instead. But then we get a version mismatch, even though we pass in the correct antlr4 version (4.11.1). Is the generation code version the python runtime tests against outdated? Is that a problem?

```
............
ANTLR runtime and generated code versions disagree: 4.11.1!=4.9.3
ANTLR runtime and generated code versions disagree: 4.11.1!=4.9.3
ANTLR runtime and generated code versions disagree: 4.11.1!=4.9.3
ANTLR runtime and generated code versions disagree: 4.11.1!=4.9.3
ANTLR runtime and generated code versions disagree: 4.11.1!=4.9.3
ANTLR runtime and generated code versions disagree: 4.11.1!=4.9.3
ANTLR runtime and generated code versions disagree: 4.11.1!=4.9.3
ANTLR runtime and generated code versions disagree: 4.11.1!=4.9.3
[...]

PR #4522, opened 2024-01-31
Generate meaningful exception message on checkVersion() failure

raffaellod:
Currently, loading a parser from an incompatible version results in a message like:

> Could not deserialize ATN with version  (expected 4).

(see e.g. https://github.com/antlr/antlr4/issues/4041 )

Change the exception message generation to use `ord(version)`, so it shows e.g. the string 3 instead of the character with ordinal 3, which is non-printable and the cause for the above cryptic message.


ericvergnaud (approved):

ericvergnaud:
Thanks for this!

Issue #4110, opened 2023-02-12
TypeScript missing critical class definition exports 


kaby76:
I was trying to write a template for TypeScript for testing the new target on grammars-v4. There is (are) important classes that need to be exported. Otherwise, I am blocked, and there will be no testing of the target in grammars-v4 because the [trgen templates](https://github.com/kaby76/Domemtech.Trash/tree/main/src/trgen/templates) require a consistent runtime API in order to provide consistent functionality across targets.

* Recognizer missing in export list. https://github.com/antlr/antlr4/blob/c2bfa34a66e6449ff6fb2e388e17e2ff4b7e8e82/runtime/JavaScript/src/antlr4/index.d.ts#L10 
* List not sorted alphabetically for quick lookup for missing exports. https://github.com/antlr/antlr4/blob/c2bfa34a66e6449ff6fb2e388e17e2ff4b7e8e82/runtime/JavaScript/src/antlr4/index.d.ts#L1
* TokenSource missing export.
* CommonToken missing TypeScript wrapper declaration and export.

I don't know why certain classes do not have corresponding TypeScript `export class` definition files. It should be identical to the list of exported symbols in Java.

ericvergnaud:
As a principle, I only export what is strictly required, excluding internal classes.
That can be hard to define until you bump into issues.
Can you provide an example of failing generated trgen code ?

kaby76:
> As a principle, I only export what is strictly required, excluding internal classes. That can be hard to define until you bump into issues. Can you provide an example of failing generated trgen code ?

I'm still going through with the creation of the template for TypeScript. So far, it's just Recognizer, which is easy to fix. But, I'll let you know what else may be missing, probably not much because the JavaScript target was working fine.

kaby76:
Also missing is the CharStreams class. I use it in JavaScript ([here](https://github.com/kaby76/Domemtech.Trash/blob/0332403442462a69a322b7289f2f2810b6337559/src/trgen/templates/JavaScript/Test.js#L139), [here](https://github.com/kaby76/Domemtech.Trash/blob/0332403442462a69a322b7289f2f2810b6337559/src/trgen/templates/JavaScript/Test.js#L144), [here](https://github.com/kaby76/Domemtech.Trash/blob/0332403442462a69a322b7289f2f2810b6337559/src/trgen/templates/JavaScript/Test.js#L149)), but there's no TypeScript decl for the class.

ericvergnaud:
[...]

PR #4525, opened 2024-02-04
added some properties and methods

RobEin:
Lexer._modeStack
Lexer.mode()
Token.source
Recognizer.getSymbolicNames()
Recognizer.getErrorListenerDispatch()

[TypeScript missing critical class definition exports](https://github.com/antlr/antlr4/issues/4110#issuecomment-1925751601)

### DIFF
--- a/runtime/JavaScript/src/antlr4/Lexer.d.ts
+++ b/runtime/JavaScript/src/antlr4/Lexer.d.ts
@@ -16,16 +16,15 @@ export declare class Lexer extends Recognizer<number> {
     _tokenStartLine: number;
     _tokenStartColumn: number;
     _type: number;
-    _modeStack: number[];
-    _mode: number;
 
     constructor(input: CharStream);
     reset(): void;
     nextToken(): Token;
     skip(): void;
     more(): void;
-    more(m: number): void;
-    mode(m: number): void;
+    getModeStack(): number[];
+    getMode(): number;
+    setMode(m: number): void;    
     pushMode(m: number): void;
     popMode(): number;
     emitToken(token: Token): void;
--- a/runtime/JavaScript/src/antlr4/Lexer.d.ts
+++ b/runtime/JavaScript/src/antlr4/Lexer.d.ts
@@ -16,6 +16,8 @@ export declare class Lexer extends Recognizer<number> {
     _tokenStartLine: number;
     _tokenStartColumn: number;
     _type: number;
+    _modeStack: number[];
+    _mode: number;
 
     constructor(input: CharStream);
     reset(): void;
@@ -23,6 +25,7 @@ export declare class Lexer extends Recognizer<number> {
     skip(): void;
     more(): void;
     more(m: number): void;
+    mode(m: number): void;
     pushMode(m: number): void;
     popMode(): number;
     emitToken(token: Token): void;
--- a/runtime/JavaScript/src/antlr4/Recognizer.d.ts
+++ b/runtime/JavaScript/src/antlr4/Recognizer.d.ts
@@ -6,4 +6,6 @@ export declare class Recognizer<TSymbol> {
 
     removeErrorListeners(): void;
     addErrorListener(listener: ErrorListener<TSymbol>): void;
+    getSymbolicNames(): string[];
+    getErrorListenerDispatch(): ErrorListener<TSymbol>;
 }
--- a/runtime/JavaScript/src/antlr4/Recognizer.d.ts
+++ b/runtime/JavaScript/src/antlr4/Recognizer.d.ts
@@ -6,6 +6,7 @@ export declare class Recognizer<TSymbol> {
 
     removeErrorListeners(): void;
     addErrorListener(listener: ErrorListener<TSymbol>): void;
-    getSymbolicNames(): string[];
     getErrorListenerDispatch(): ErrorListener<TSymbol>;
+    getLiteralNames(): string[] | [];
+    getSymbolicNames(): string[] | [];
 }
--- a/runtime/JavaScript/src/antlr4/Token.d.ts
+++ b/runtime/JavaScript/src/antlr4/Token.d.ts
@@ -1,15 +1,13 @@
-import {CharStream} from "./CharStream";
-import {TokenSource} from "./TokenSource";
-import {InputStream} from "./InputStream";
+import { TokenSource } from "./TokenSource";
+import { CharStream } from "./CharStream";
 
 export declare class Token {
 
+    static INVALID_TYPE: number;
     static EOF: number;
-    
     static DEFAULT_CHANNEL: number;
     static HIDDEN_CHANNEL: number;
 
-    source: [ TokenSource, InputStream ];
     tokenIndex: number;
     line: number;
     column: number;
@@ -21,5 +19,6 @@ export declare class Token {
 
     clone(): Token;
     cloneWithType(type: number): Token;
+    getTokenSource(): TokenSource;
     getInputStream(): CharStream;
 }
--- a/runtime/JavaScript/src/antlr4/Token.d.ts
+++ b/runtime/JavaScript/src/antlr4/Token.d.ts
@@ -1,4 +1,6 @@
 import {CharStream} from "./CharStream";
+import {TokenSource} from "./TokenSource";
+import {InputStream} from "./InputStream";
 
 export declare class Token {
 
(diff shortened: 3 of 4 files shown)


RobEin:
Unfortunately, I cannot comment on this:
failing at "build (macos-11, php) " - Install dependecies ...

Thanks,
Robert

runtime/JavaScript/src/antlr4/Lexer.d.ts
```diff
@@ -16,13 +16,16 @@ export declare class Lexer extends Recognizer<number> {
...
+    _modeStack: number[];
+    _mode: number;
 
     constructor(input: CharStream);
     reset(): void;
     nextToken(): Token;
     skip(): void;
     more(): void;
     more(m: number): void;
+    mode(m: number): void;
```
ericvergnaud, line 28: I guess the intent with the 2nd `more` was to be `mode`. Can you merge these ?

runtime/JavaScript/src/antlr4/Token.d.ts
```diff
@@ -7,6 +9,7 @@ export declare class Token {
     static DEFAULT_CHANNEL: number;
     static HIDDEN_CHANNEL: number;
 
+    source: [ TokenSource, InputStream ];
```
ericvergnaud, line 12: I'd rather you add `getTokenSource` and `getInputStream`. `source` being a tuple is an internal detail

runtime/JavaScript/src/antlr4/Lexer.d.ts
```diff
@@ -16,13 +16,16 @@ export declare class Lexer extends Recognizer<number> {
     _tokenStartLine: number;
     _tokenStartColumn: number;
     _type: number;
+    _modeStack: number[];
+    _mode: number;
```
ericvergnaud, line 20: maybe rather add `getMode` an `getModeStack` ?

runtime/JavaScript/src/antlr4/Lexer.d.ts
RobEin: The 2 `more` are already in the current [Lexer.d.ts](https://github.com/antlr/antlr4/blob/3c06d0ac6f8750c252ec496eb3629b0a6e9e47fb/runtime/JavaScript/src/antlr4/Lexer.d.ts#L24C5-L25C27).

runtime/JavaScript/src/antlr4/Recognizer.d.ts
```diff
@@ -6,4 +6,6 @@ export declare class Recognizer<TSymbol> {
 
     removeErrorListeners(): void;
     addErrorListener(listener: ErrorListener<TSymbol>): void;
+    getSymbolicNames(): string[];
```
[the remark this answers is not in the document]
RobEin: I would add all JavaScript attributes, but you wrote: ["I only export what is strictly required"](https://github.com/antlr/antlr4/issues/4110#issuecomment-1427032474)

runtime/JavaScript/src/antlr4/Token.d.ts
RobEin: I have a method where I refer to the source property and a getTokenSource() would look strange in my opinion:.
```
private getCommonTokenByToken(oldToken: Token): CommonToken {
    const cToken = new CommonToken(oldToken.source, oldToken.type, oldToken.channel, oldToken.start, oldToken.stop);
    cToken.tokenIndex = oldToken.tokenIndex;
    cToken.line = oldToken.line;
    cToken.column = oldToken.column;
    cToken.text = oldToken.text;
    return cToken;
}
```

RobEin: Sorry, maybe what I wrote was a little misunderstood.
Perhaps it would have been more understandable if I had referred to a public field instead of a property.
However, VS Code [marks](https://putridparrot.com/blog/properties-and-fields-in-typescript/) public fields as properties.

runtime/JavaScript/src/antlr4/Lexer.d.ts
RobEin: The public fields would be better because I have also used them in my various portings:
[Java port](https://github.com/antlr/grammars-v4/blob/bf950dfcfb26477735f02402dadf98e1e507c051/python/python3_12_1/Java/PythonLexerBase.java#L293)
[JavaScript port](https://github.com/antlr/grammars-v4/blob/bf950dfcfb26477735f02402dadf98e1e507c051/python/python3_12_1/JavaScript/PythonLexerBase.js#L292)
[Pyton port](https://github.com/antlr/grammars-v4/blob/bf950dfcfb26477735f02402dadf98e1e507c051/python/python3_12_1/Python3/PythonLexerBase.py#L238)

runtime/JavaScript/src/antlr4/Lexer.d.ts
ericvergnaud: There should only be 1, it's a bug in the existing file.

runtime/JavaScript/src/antlr4/Recognizer.d.ts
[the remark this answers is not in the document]
ericvergnaud: I believe in encapsulation. I'm ok to provide data via methods, not so much direct read/write access to fields that may change anytime... Adding `getLiteralNames` is perfectly fine.

runtime/JavaScript/src/antlr4/Token.d.ts
ericvergnaud: I'm really not keen to expose such a poorly designed construct. The d.ts file already has `getInputStream`, so you only need to add `getTokenSource`.

runtime/JavaScript/src/antlr4/Lexer.d.ts
ericvergnaud: Well these fields are clearly private... (I guess there was a bit of laziness with the first implementation 30 years ago)

ericvergnaud:
As a heads-up, we've started working on antlr5 which will come with a cleaned-up API (and speed in JS and Python).
It's early days, but since you're dealing with rather low-level details, it could be useful to get involved at some point such that we understand your use cases and make sure they are supported ?

RobEin:
The new ANTLR5 looks very exciting.
If I can help in any way, I will gladly do so.

RobEin (commented):
I apologize for not showing up until now, but I've been very busy (including other ANTLR4 stuff).
I have implemented the changes you suggested.
I hope you thought so.

Instead of the getMode() and setMode() methods, wouldn't a mode property be better?

runtime/JavaScript/src/antlr4/Recognizer.d.ts
```diff
@@ -6,4 +6,7 @@ export declare class Recognizer<TSymbol> {
 
     removeErrorListeners(): void;
     addErrorListener(listener: ErrorListener<TSymbol>): void;
+    getErrorListenerDispatch(): ErrorListener<TSymbol>;
```
ericvergnaud, line 9: Can we rename this to say `getRootErrorListener()` (open to suggestions). Appreciate it's an instance of `ErrorListenerDispatch` but that's a terrible name...

ericvergnaud:
> Instead of the `getMode()` and `setMode()` methods, wouldn't a `mode` property be better?
It would, but you can't make it a property in the d.ts without doing the same in the .js, which would break backwards compatibility

ericvergnaud (changes requested):

runtime/JavaScript/src/antlr4/Recognizer.d.ts
RobEin: The name `mode` is really not a good property name, because there is already a `mode()` method in the JavaScript Lexer.
Maybe a property for JavaScript/TypeScript called `lexerMode` or `tokenizerMode`?
And the `mode()` method would be depreciated in the JavaScript Lexer.

I don't like the name getErrorListenerDispatch either.

ericvergnaud: Actually we can't change that name without breaking backwards compatibility in JS...

RobEin: Indeed, unfortunately the `getErrorListenerDispatch` method should not be renamed.
I think an independent TypeScript runtime (no `*.js`, no `*.d.ts`) would be ideal for this and other more modern designs.
Maybe in ANTLR5?

I remove my `setMode` method because there is already a `mode` method that does the same thing.
However, a `setMode`/`getMode` pair would have been more elegant than a `mode`/`getMode` pair.

PR #4529, opened 2024-02-16
Fix missing export of TokenStreamRewriter

weilence:


ericvergnaud:
Hi,
thanks so much for this!
However it seems to be missing the TS api ?

weilence:
Yes, the d.ts file exports this API, but in reality, it cannot be used. For example:
```javascript
import { TokenStreamRewriter } from 'antlr4';
```
Directly using `TokenStreamRewriter` results in it being undefined, although the d.ts file appears to be normal.
You must use it like this:
```javascript
import antlr4 from 'antlr4';
antlr4.TokenStreamRewriter
```

ericvergnaud:
So the purpose of this PR is to fix this issue ? Should it be named "Fix missing export of TokenStreamRewriter" ?

weilence:
Yes

ericvergnaud:
Thanks!

PR #4535, opened 2024-02-19
[WIP] Update deprecated usage for InputStream.

multiarc:
Fix CharStream in index files referencing deprecated InputStream instead. 
Only log exceptions (for example LexerNoViableAltException) if debug is enabled for _interp. 
There is no other way to control this console output right now, seems like missed if check.
[in progress]: unintentional exception is fired by parser along with passing it through error listener at the same time. This behavior changed


multiarc:
>

PR #4536, opened 2024-02-19
Default exception logging (no setting), 

multiarc:
Only log Lexer exceptions (for example LexerNoViableAltException) when debug is enabled for _interp.
There is no other way to control this console output right now, seems like missed if check.
Update index files to reference CharStream instead of deprecated InputStream as it's actual type.
Update usage of InputStream to use CharStream instead.
[in progress/question]: unintentional exception is fired by Parser (NoViableAltException) along with passing it through error listener at the same time. This behavior changed since previous release. 
Working on resolving this (if it was not intentional of course). Please comment if that was intentional or not.

### DIFF
--- a/runtime/JavaScript/spec/rewriter/TokenStreamRewriterSpec.js
+++ b/runtime/JavaScript/spec/rewriter/TokenStreamRewriterSpec.js
@@ -8,7 +8,7 @@ import calc from "./generatedCode/calc.js";
  * @param {string} input 
  */
 function getRewriter(lexerClass, input) {
-    const chars = new antlr4.InputStream(input);
+    const chars = new antlr4.CharStream(input);
     const lexer = new lexerClass(chars);
     const tokens = new antlr4.CommonTokenStream(lexer);
     tokens.fill();
@@ -385,7 +385,7 @@ describe("TokenStreamRewriter", () => {
     
     it("throws an error if second replace operation overlaps the first one on the left", () => {
         // Arrange
-        const chars = new antlr4.InputStream("abcccba");
+        const chars = new antlr4.CharStream("abcccba");
         const lexer = new abc(chars);
         const tokens = new antlr4.CommonTokenStream(lexer);
         tokens.fill();
--- a/runtime/JavaScript/src/antlr4/Lexer.js
+++ b/runtime/JavaScript/src/antlr4/Lexer.js
@@ -121,7 +121,9 @@ export default class Lexer extends Recognizer {
 							this.notifyListeners(e); // report error
 							this.recover(e);
 						} else {
-							console.log(e.stack);
+                            if (this._interp.debug) {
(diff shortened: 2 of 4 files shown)


runtime/JavaScript/spec/rewriter/TokenStreamRewriterSpec.js
```diff
@@ -8,7 +8,7 @@ import calc from "./generatedCode/calc.js";
  * @param {string} input 
  */
 function getRewriter(lexerClass, input) {
-    const chars = new antlr4.InputStream(input);
+    const chars = new antlr4.CharStream(input);
```
ericvergnaud, line 11: The correct code would be CharStreams.fromString

runtime/JavaScript/src/antlr4/Lexer.js
```diff
@@ -121,7 +121,9 @@ export default class Lexer extends Recognizer {
 							this.notifyListeners(e); // report error
 							this.recover(e);
 						} else {
-							console.log(e.stack);
+                            if (this._interp.debug) {
```
ericvergnaud, line 124: This code hasn't changed in 4 years, so the change in behavior relates to something else. Maybe some grammar specific code (predicate or production) that throws this exception, most probably unrelated to antlr itself ?

ericvergnaud (changes requested):

runtime/JavaScript/src/antlr4/Lexer.js
multiarc: I will investigate and look for independent reproduction.

runtime/JavaScript/spec/rewriter/TokenStreamRewriterSpec.js
```diff
@@ -8,7 +8,7 @@ import calc from "./generatedCode/calc.js";
  * @param {string} input 
  */
 function getRewriter(lexerClass, input) {
-    const chars = new antlr4.InputStream(input);
+    const chars = new antlr4.CharStreams.fromString(input);
```
ericvergnaud, line 11: This is a static function, please remove `new`

runtime/JavaScript/spec/rewriter/TokenStreamRewriterSpec.js
```diff
@@ -385,7 +385,7 @@ describe("TokenStreamRewriter", () => {
     
     it("throws an error if second replace operation overlaps the first one on the left", () => {
         // Arrange
-        const chars = new antlr4.InputStream("abcccba");
+        const chars = new antlr4.CharStreams.fromString("abcccba");
```
ericvergnaud, line 388: same here

ericvergnaud (changes requested):

ericvergnaud (approved):

PR #4540, opened 2024-02-23
fix(javascript): fix types not being recognized for NodeNext module resolution

### LINKED ISSUE #4218: Can't import from TypeScript runtime with `"moduleResolution": "nodenext"`
I'm honestly not sure what needs to change in the `package.json` because I thought `"type": "module"`, `"main"` and `"types"` would be enough, but apparently not.  The ESM situation is a mess.  When i turn on `"moduleResolution": "nodenext"`, I get errors like

```
test.ts:1:22 - error TS2305: Module '"antlr4"' has no exported member 'CommonTokenStream'.

1 import { CharStream, CommonTokenStream }  from 'antlr4';
```

It works with `"moduleResolution": "node"`, but according to TS docs that's really designed for CJS and doesn't support packages with `"export"` maps, so I wouldn't be able to use `antlr4` and packages with export maps in the same project.

KurtGokhan:
Fixes #4218

The problem was, since the `package.json` has `type: module`, Typescript is treating regular `d.ts` files as module. But the `d.ts` files are generated manually and they don't comply with how a module file should look like (e.g. missing file extensions on imports). So they are essentially more like CommonJS files.

Luckily, there is a way to override the module detection. If the file has `cts` or `cjs` extension, Typescript will treat that file as CommonJS. In this case, adding `cts` extension only to the `index` file was enough.

Ideally, the TS files should be built with `tsc` in the future to ensure the files comply with the corresponding module format.

Also added a `files` field to the package.json so that only relevant files are included in the NPM bundle. Previously it also included files like `spec`, `src`, and dev config files, which made it confusing to debug.

Tested my changes locally with various `moduleResolution` options like `Node, Node10, Node16, NodeNext, Bundler` and they work flawlessly. Only the `Classic` option doesn't work but it also doesn't work on upstream and it doesn't work for most other packages as well.

### Alternative solutions

All of the other solutions require changes in more files but I will leave them here as an option.

The first obvious solution is to add extensions to all the TS files.

I checked various other packages like `lodash-es` and checked why they don't have this problem even though they don't have extension in TS files. Turns out, they don't have `type: module` in their package.json, so files are treated as CommonJS by default, even though the packages are ES packages.

So one of the solutions is to remove `type: module` and convert all `js` files to `mjs`.


ericvergnaud:
Interesting, I'd never heard of its files before, thanks for this.
See my last comment in #4218, hopefully we can get close to a packaging implementation that we can trust.

KurtGokhan:
> Hopefully we can get close to a packaging implementation that we can trust.

Is the test suite for that absolutely necessary? Imho, we find a working configuration, test it manually in different scenarios, iterate if necessary. This isn't something that will change much. Most npm projects work with this approach.

Btw, sharing the results from AreTheTypesWrong:

Before:
image

After:
image

Here are the problems:
- [Internal Resolution Error](https://github.com/arethetypeswrong/arethetypeswrong.github.io/blob/main/docs/problems/InternalResolutionError.md) can cause issues if `skipLibCheck` is not false. But this seems to also exist on current setup.
- [Masquerading as CJS](https://github.com/arethetypeswrong/arethetypeswrong.github.io/blob/main/docs/problems/FalseCJS.md) isn't a big issue since this package doesn't have any default exports.

mike-lischke:
@KurtGokhan interesting tool you used for this information here. I didn't know that before.

As a test I let it test the ANTLR4 TypeScript target [antlr4ng](https://github.com/mike-lischke/antlr4ng) and it came out as:

Bildschirmfoto 2024-04-18 um 21 38 03

Maybe you want to try that out? I'm open to accept a PR to fix the last remaining issue.

KurtGokhan:
@mike-lischke  "Masquerading as ESM" shouldn't cause any problems as long as you are not using default exports. It may not need a fix.

runtime/JavaScript/package.json
```diff
@@ -51,14 +56,18 @@
   },
   "exports": {
     ".": {
+      "types": "./src/antlr4/index.d.cts",
+      "import": "./dist/antlr4.node.mjs",
+      "require": "./dist/antlr4.node.cjs",
+      "default": "./dist/antlr4.node.mjs",
       "node": {
-        "types": "./src/antlr4/index.d.ts",
+        "types": "./src/antlr4/index.d.cts",
```
ericvergnaud, line 64: I believe this should be removed ?

runtime/JavaScript/package.json
```diff
@@ -51,14 +56,18 @@
...
         "import": "./dist/antlr4.node.mjs",
         "require": "./dist/antlr4.node.cjs",
         "default": "./dist/antlr4.node.mjs"
       },
       "browser": {
-        "types": "./src/antlr4/index.d.ts",
+        "types": "./src/antlr4/index.d.cts",
```
ericvergnaud, line 70: I believe this should be removed ?

ericvergnaud:
@parrt let's see if we can get that one in

runtime/JavaScript/package.json
KurtGokhan: Yes, removed it and tested with different module resolutions like `NodeNext`, `Bundler` and `Node`. Seems working.

ericvergnaud:
@partt blessed

parrt:
Looks like there are conflicts.

ericvergnaud:
@parrt I see conflicts if you rebase, not if you merge ?

PR #4541, opened 2024-02-24
Remove dead link

### LINKED ISSUE #4537: Broken javascript download link
Fix the javascript target link to download please

ericvergnaud:
Fixes #4537 
@parrt you might want to also update the antlr.org web site


parrt:
I don't see a link on the download page so I think we're okay?

ericvergnaud:
There is a link here https://www.antlr.org/download.html in the JavaScript section

parrt:
Ah. thanks. updated: https://github.com/antlr/website-antlr4/blob/gh-pages/download.html

Issue #4491, opened 2023-12-14
[TypeScript] Generated files have inconsistent "import" statements, grammars-v4 testing of TypeScript now fails


kaby76:
This is a possible problem with the TypeScript target. It is best demonstrated by the [plsql grammar](https://github.com/antlr/grammars-v4/tree/5792cf2ddf4a70933a6f0fec051a34fa3d9655f4/sql/plsql). This grammar is different than most because it contains a "superClass" option for lexer and parser.

If you generate the target TypeScript code for the grammar (`antlr4 -v 4.13.1 -encoding utf-8 -Dlanguage=TypeScript   PlSqlLexer.g4; antlr4 -v 4.13.1 -encoding utf-8 -Dlanguage=TypeScript   PlSqlParser.g4`), you get code with two types of "import" statements, many with curly braces around an imported class:
```
PlSqlParserListener.ts:import { Sql_scriptContext } from "./PlSqlParser";
PlSqlParserListener.ts:import { Unit_statementContext } from "./PlSqlParser";
PlSqlParserListener.ts:import { Alter_diskgroupContext } from "./PlSqlParser";
PlSqlParserListener.ts:import { Add_disk_clauseContext } from "./PlSqlParser";
PlSqlParserListener.ts:import { Drop_disk_clauseContext } from "./PlSqlParser";
PlSqlParserListener.ts:import { Resize_disk_clauseContext } from "./PlSqlParser";
PlSqlParserListener.ts:import { Replace_disk_clauseContext } from "./PlSqlParser";
PlSqlParserListener.ts:import { Wait_nowaitContext } from "./PlSqlParser";
PlSqlParserListener.ts:import { Rename_disk_clauseContext } from "./PlSqlParser";
PlSqlParserListener.ts:import { Disk_online_clauseContext } from "./PlSqlParser";
```
and others without curly braces:
```
PlSqlLexer.ts:import PlSqlLexerBase from './PlSqlLexerBase';
PlSqlParser.ts:import PlSqlParserListener from "./PlSqlParserListener.js";
PlSqlParser.ts:import PlSqlParserBase from './PlSqlParserBase';
```
My tsconfig.json file is this:
```
{
  "compilerOptions": {
    "module": "ES2020",
    "moduleResolution": "node",
    "target": "ES6",
    "noImplicitAny": true,
  },
  "ts-node": {
    "esm": true,
    "experimentalSpecifierResolution": "node"
  }
}
```

[...]

PR #4549, opened 2024-03-08
Use arrayview for bit set

ericvergnaud:
Implements perf improvement by using array view for bit set

### DIFF
--- a/runtime/JavaScript/src/antlr4/misc/BitSet.js
+++ b/runtime/JavaScript/src/antlr4/misc/BitSet.js
@@ -28,7 +28,7 @@ export default class BitSet {
     }
 
     values() {
-        return Object.keys(this.data);
+        return Object.keys(this.data); // .map(s => parseInt(s))
     }
 
     minValue() {
--- a/runtime/JavaScript/src/antlr4/misc/BitSet.js
+++ b/runtime/JavaScript/src/antlr4/misc/BitSet.js
@@ -11,13 +11,13 @@ export default class BitSet {
         this.data = new Uint32Array(1);
     }
 
-    add(index) {
+    set(index) {
         this._checkIndex(index)
         this._resize(index);
         this.data[index >>> 5] |= 1 << index % 32;
     }
 
-    has(index) {
+    get(index) {
         this._checkIndex(index)
         const slot = index >>> 5;
         if (slot >= this.data.length) {
@@ -26,7 +26,7 @@ export default class BitSet {
         return (this.data[slot] & 1 << index % 32) !== 0;
     }
 
-    remove(index) {
+    clear(index) {
         this._checkIndex(index)
         const slot = index >>> 5;
         if (slot < this.data.length) {
--- a/runtime/JavaScript/src/antlr4/misc/BitSet.js
+++ b/runtime/JavaScript/src/antlr4/misc/BitSet.js
@@ -12,13 +12,13 @@ export default class BitSet {
     }
 
     set(index) {
-        this._checkIndex(index)
+        BitSet._checkIndex(index)
         this._resize(index);
         this.data[index >>> 5] |= 1 << index % 32;
     }
 
     get(index) {
-        this._checkIndex(index)
+        BitSet._checkIndex(index)
         const slot = index >>> 5;
         if (slot >= this.data.length) {
             return false;
@@ -27,7 +27,7 @@ export default class BitSet {
     }
 
     clear(index) {
-        this._checkIndex(index)
+        BitSet._checkIndex(index)
         const slot = index >>> 5;
         if (slot < this.data.length) {
             this.data[index >>> 5] &= ~(1 << index);
@@ -56,7 +56,7 @@ export default class BitSet {
             let l = this.data[k];
             while (l !== 0) {
                 const t = l & -l;
-                result[pos++] = (k << 5) + this._bitCount(t - 1);
+                result[pos++] = (k << 5) + BitSet._bitCount(t - 1);
                 l ^= t;
             }
         }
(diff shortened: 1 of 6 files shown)


ericvergnaud:
ignore the unrelated php failure

kaby76:
I will be doing the review in three comments: "performance", "correctness", and "code issues".

### Performance

This graph shows for java/java the runtimes for CSharp, Antlr4ng, TS 4.13.1 and TS with PR. There is about a 2.8176/2.7061 = 1.04 speed up (4%).

java

ericvergnaud:
4% is great for such a small and low-risk change :-)

kaby76:
All (271) grammars in grammars-v4 (except three, which have the ["module not found" import issue uncorrected in base class code](https://github.com/antlr/antlr4/issues/4491)) pass using the runtime of this PR.

runtime/JavaScript/src/antlr4/misc/BitSet.js
```diff
@@ -8,31 +8,74 @@ import equalArrays from "../utils/equalArrays.js";
...
+                result[pos++] = (k << 5) + BitSet._bitCount(t - 1);
+                l ^= t;
+            }
+        }
+        return result;
     }
 
     minValue() {
-        return Math.min.apply(null, this.values());
+         for (let k = 0; k < this.data.length; ++k) {
```
kaby76, line 67: Indent off.

runtime/JavaScript/src/antlr4/misc/BitSet.js
```diff
@@ -48,6 +91,33 @@ export default class BitSet {
...
-        return this.values().length;
+        return this.data.map(l => BitSet._bitCount(l)).reduce((s,v) => s + v, 0);
+    }
+
+    _resize(index) {
+        const count = index + 32 >>> 5;
+        if (count <= this.data.length) {
+            return;
+        }
+        const data = new Uint32Array(count);
```
kaby76, line 102: Why are you constructing a new array when I think Uint32Array.resize() is available?

ericvergnaud: I'm afraid your thinking is incorrect. You can resize an ArrayBuffer but not a TypedArray. I suspect the reason for this is that the underlying buffer is shared (a poor design decision if you ask me) so resizing a view may actually silently resize other views. Resizing the buffer would be a hack, not something I'm comfortable with

runtime/JavaScript/src/antlr4/misc/BitSet.js
ericvergnaud: Fixed

kaby76:
This code is definitely not correct:
https://github.com/antlr/antlr4/blob/e4a77df168ee0c74c0fd760b6410730018dbbe3b/runtime/JavaScript/src/antlr4/misc/BitSet.js#L33

You should not shift bit by index but by index mod 32.

See C# code.

https://github.com/antlr/antlr4/blob/1b3150b5de7400757fd418de4362bd28e44af106/runtime/CSharp/src/Sharpen/BitSet.cs#L127

ericvergnaud:
> This code is definitely not correct:
> 
> https://github.com/antlr/antlr4/blob/e4a77df168ee0c74c0fd760b6410730018dbbe3b/runtime/JavaScript/src/antlr4/misc/BitSet.js#L33
> 
> You should not shift bit by index but by index mod 32.
> 
> See C# code.
> 
> https://github.com/antlr/antlr4/blob/1b3150b5de7400757fd418de4362bd28e44af106/runtime/CSharp/src/Sharpen/BitSet.cs#L127

mmm... `it("clears 1 value", () => {` in BitSetSpec says it's ok... digging...

kaby76:
https://github.com/antlr/antlr4/blob/e4a77df168ee0c74c0fd760b6410730018dbbe3b/runtime/JavaScript/src/antlr4/misc/BitSet.js#L31-L33

Let's consider some examples.

index = 0 => slot = index/32 = 0. 1<<(index%32) = 0x00000001
index = 1 => slot = index/32 = 0. 1<<(index%32) = 0x00000002
index = 31 => slot = 31/32 = 0. 1<<(index%32) = 0x80000000.
index = 32 => slot = 32/32 = 1. 1<<(index%32) = 0. data[0] = 0x00000000, data[1] = 0x00000001.

For index = 32, if you compute "1<<index" instead of "1<<(index%32)", the result is 0 (assuming 32-bit integer value of the shift). That's not correct.

Also "index >>> 5" is computed twice in the code.

kaby76:
Actually, "1<<index" is probably ok because JS has a very very weird left-shift operator unlike every programming language in existence. 1<<67 = 8. But, that is the same as 1<<(index % 32) because JS operator "<<" is a rotate, not a shift. Very weird language.

ericvergnaud:
> Actually, "1<<index" is probably ok because JS has a very very weird "<<" operator unlike every programming language in existence. 1<<67 = 8 in JS. But, that is the same as 1<<(67% 32). Very weird language. Still I would need to think about this.

I just checked manually, it's ok.

ericvergnaud:
> Also "index >>> 5" is computed twice in the code.

Fixed

kaby76:
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Left_shift

`The right operand will be converted to an unsigned 32-bit integer and then taken modulo 32, so the actual shift offset will always be a positive integer between 0 and 31, inclusive. For example, 100 << 32 is the same as 100 << 0 (and produces 100) because 32 modulo 32 is 0.`

ericvergnaud:
If all ok with you can you approve ?

kaby76 (approved):

PR #4546, opened 2024-03-04
cache ATNConfig hashCode in Java, C#, Python and JavaScript

ericvergnaud:
This proposal follows findings in https://github.com/mike-lischke/antlr4ng/commit/77d0d4760f40555aae7241564a72b5d6d3f1b53a where caching ATNConfig hashCode seems to improve performance significantly during warmup

PR #4551, opened 2024-03-08
implement faster HashMap

ericvergnaud:
new faster HashMap

### DIFF
--- a/runtime/JavaScript/spec/HashMapSpec.js
+++ b/runtime/JavaScript/spec/HashMapSpec.js
@@ -1,0 +1,60 @@
+import HashMap from "../src/antlr4/misc/HashMap.js";
+import HashCode from "../src/antlr4/misc/HashCode.js";
+
+class Thing {
+
+    value1 = Math.random();
+    value2 = Math.random();
+
+    hashCode() {
+        return HashCode.hashStuff(this.value1);
+    }
+
+    equals(other) {
+        return other instanceof Thing
+            && other.value1 === this.value1
+            && other.value2 === this.value2;
+    }
+}
+
+describe('test HashMap', () => {
+
+    it("sets a thing", () => {
+        const t1 = new Thing();
+        const map = new HashMap();
+        map.set("abc", t1);
+        expect(map.containsKey("abc")).toBeTrue();
+        expect(map.containsKey("def")).toBeFalse();
+        expect(map.length).toEqual(1);
+    })
+
+    it("gets a thing", () => {
+        const t1 = new Thing();
+        const map = new HashMap();
+        map.set("abc", t1);
+        const t2 = map.get("abc");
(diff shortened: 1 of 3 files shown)


ericvergnaud:
ignore unrelated php failure

kaby76:
For this PR, there is a 40% speed up. (2.8537 +- SD 0.0124 vs 2.0492 +- SD 0.033).

java

kaby76:
(There may be other optimizations that Antlr4ng does, but I haven't gotten that far.)

ericvergnaud:
So we’re talking 60-65% cumulated improvements ? Whoosh!!!Envoyé de mon iPhoneLe 10 mars 2024 à 23:32, Ken Domino ***@***.***> a écrit :﻿
(There may be other optimizations that Antlr4ng does, but I haven't gotten that far.)

kaby76:
Yes, something like ~60%. There are probably other optimizations to look at.

At some point, can we get a release of 4.13.2?? No hurries though, after other fixes are considered. I did not like this 4.13.1-patch1 release because that breaks my testing code, and I don't think GitHub Dependabot even considers this a release.

kaby76:
This PR passes all grammars-v4 tests.

kaby76 (approved):

ericvergnaud:
> At some point, can we get a release of 4.13.2?? No hurries though, after other fixes are considered. I did not like this 4.13.1-patch1 release because that breaks my testing code, and I don't think GitHub Dependabot even considers this a release.

That's a question for @parrt.

ericvergnaud:
@kaby76 it's be interesting to run your performance benchmark on the merged dev version

kaby76:
The runtime for the tip of dev is better.

java

But, it is still missing something as Antlr4ng is faster. I am still surprised the hash value caching didn't help. I must have done something wrong, or it should have been applied somewhere else. Looking....

ericvergnaud:
antlr4ng also implements specialized constructors which avoid testing instance types. It might be worth doing this for the objects that are instantiated millions of times (such as ATNConfig). But we might be reaching a point where profiling becomes necessary to better understand the remaining bottlenecks.
Sam mentioned that the reason for not caching ATNConfig hashCode is memory footprint.

ericvergnaud:
> But, it is still missing something as Antlr4ng is faster. I am still surprised the hash value caching didn't help. I must have done something wrong, or it should have been applied somewhere else. Looking....

I can revise and rebase that branch if you're willing to double-check that...

kaby76:
I'm profiling the CSharp target because [your PR changed one file in that target](https://github.com/antlr/antlr4/pull/4546/files#diff-20dbca93785728353c109eb4fee6a1c1a308ddb178a58af9e1ddb17d915e383b). (IMO, C# has far-superior profiling tools to anything I've seen in all other targets.)

PR #4558, opened 2024-03-16
Add parse listener methods to Parser typescript

Phlosioneer:
This PR adds the four parse listener methods. The `_parseListeners` field could also be set to `ParseTreeListener[]`, but it would _technically_ be a breaking change.

I chose to use `ParseTreeListener` instead of `any` in these methods because parse listeners cannot actually be `any`; their functions are called without checking if they're defined. The current ParseTreeListener interface is the minimum requirement to avoid exceptions.


ericvergnaud (approved):
Thanks!

PR #4559, opened 2024-03-16
Add override specifier to typescript copyfrom codegen

Phlosioneer:
Required for `"noImplicitOverride": true` typescript setting.

I couldn't figure out how to add a test for this situation in the typescript test file. It only happens if a rule context class has a subclass.


ericvergnaud:
I think it happens when you have labeled alternatives, such as:

```
generic_rule:
 specific_rule_1 # SepecificRule1
 | specific_rule_2 # SepecificRule2
 ;

```

We already have tests for these (for example runtime-testsuite/resources/org/antlr/v4/test/runtime/descriptors/LeftRecursion/AmbigLR_1.txt), which leads me to suspect that the need for this comes from tsconfig ? Or a specific version of tsc ?

Phlosioneer:
It is a tsconfig setting:

> Required for `"noImplicitOverride": true` typescript setting.

I don't see where the expected typescript translation of that test file lives...? Is the generated parser source compared against a real file, or is it just a "does it generate and run" test?

ericvergnaud:
The test sample is used to generate a unit test for each target. You'd need to follow the build instructions to actually see the corresponding code since it's dynamically generated.

PR #4561, opened 2024-03-18
remove setMode method

RobEin:
Added:
- Lexer.`mode(...)`
- Lexer.`getMode()` (instead of `_mode` field)
- Lexer.`getModeStack()`
- Recognizer.`getErrorListenerDispatch()`
- Recognizer.`getLiteralNames()`
- Recognizer.`getSymbolicNames()`
- Token.`INVALID_TYPE`
- Token.`getTokenSource()`

Removed (fixed):
- `more(m: number)` method

### DIFF
--- a/runtime/JavaScript/src/antlr4/Parser.d.ts
+++ b/runtime/JavaScript/src/antlr4/Parser.d.ts
@@ -6,6 +6,7 @@ import {ParserATNSimulator} from "./atn";
 import {Token} from "./Token";
 import {ParserRuleContext} from "./context";
 import {Printer} from "./utils";
+import {ParseTreeListener} from "./tree";
 
 export declare class Parser extends Recognizer<Token> {
 
@@ -24,6 +25,10 @@ export declare class Parser extends Recognizer<Token> {
     constructor(input: TokenStream);
     match(ttype: number): Token;
     matchWildcard(): Token;
+    getParseListeners(): ParseTreeListener[];
+    addParseListener(listener: ParseTreeListener): void;
+    removeParseListener(listener: ParseTreeListener): void;
+    removeParseListeners(): void;
     consume(): Token;
     enterRule(localctx: ParserRuleContext, state: number, ruleIndex: number): void;
     exitRule() : void;


runtime/JavaScript/src/antlr4/Lexer.d.ts
```diff
@@ -22,7 +22,9 @@ export declare class Lexer extends Recognizer<number> {
     nextToken(): Token;
     skip(): void;
     more(): void;
-    more(m: number): void;
+    mode(m: number): void;
```
ericvergnaud, line 25: can you add a setMode both here and in js and mark this one as deprecated ?

runtime/JavaScript/src/antlr4/Recognizer.d.ts
```diff
@@ -6,4 +6,7 @@ export declare class Recognizer<TSymbol> {
 
     removeErrorListeners(): void;
     addErrorListener(listener: ErrorListener<TSymbol>): void;
+    getErrorListenerDispatch(): ErrorListener<TSymbol>;
```
ericvergnaud, line 9: can you rename to getErrorListener() and add to js ?

runtime/JavaScript/spec/HashMapSpec.js
```diff
@@ -0,0 +1,60 @@
+import HashMap from "../src/antlr4/misc/HashMap.js";
```
ericvergnaud, line 1: Looks like you need to rebase first

ericvergnaud (changes requested):

runtime/JavaScript/src/antlr4/Recognizer.d.ts
```diff
@@ -6,4 +6,7 @@ export declare class Recognizer<TSymbol> {
...
+    getLiteralNames(): string[] | [];
```
ericvergnaud, line 10: I don't think the type is correct. The array cannot be empty and it contains `string | null `elements

ericvergnaud: Same here

ericvergnaud (changes requested):

runtime/JavaScript/src/antlr4/Recognizer.d.ts
RobEin: I started from [this](https://github.com/antlr/antlr4/blob/b4c12ece2c6e44f4faf1f30c258e52c8142c2dcc/runtime/JavaScript/src/antlr4/Recognizer.js#L32).
I will correct it to:
```
    getLiteralNames(): string[];
    getSymbolicNames(): string[];
```

runtime/JavaScript/src/antlr4/Recognizer.d.ts
RobEin: ... and I think the `getErrorListenerDispatch` method should be marked as depricated in js.

runtime/JavaScript/src/antlr4/Lexer.d.ts
RobEin: What will be the next ANTLR version, i.e. from which version should it be marked as deprecated?
since 4.13.2?

ericvergnaud:
YesEnvoyé de mon iPhoneLe 18 mars 2024 à 18:14, Robert Einhorn ***@***.***> a écrit :﻿
@RobEin commented on this pull request.



In runtime/JavaScript/src/antlr4/Lexer.d.ts:
> @@ -22,7 +22,9 @@ export declare class Lexer extends Recognizer<number> {
     nextToken(): Token;
     skip(): void;
     more(): void;
-    more(m: number): void;
+    mode(m: number): void;

What will be the next ANTLR version, i.e. from which version should it be marked as deprecated?
since 4.13.2?

PR #4566, opened 2024-03-26
Fixing crash bug in TokenStreamRewriter

hs-apotell:
Fixing crash bug in TokenStreamRewriter

Order of operations issue - use before deleting not vice-versa.


parrt:
Thanks, @hs-apotell !

PR #4570, opened 2024-03-31
[JavaScript] Fix incorrect this.channel access

Phlosioneer:
`BufferedTokenStream::nextTokenOnChannel` was accessing the field `this.channel` instead of the parameter `channel`. The buffered token stream class doesn't have a field named `channel`, so this exhausted the token buffer instead.

The bug had not been caught because it is only called from `CommonTokenStream` with parameter `this.channel`.


ericvergnaud:
Pfff!! How did this one survive 10 years ? Thanks!

PR #4571, opened 2024-04-01
golang: Add to quick checks to PredictionContext Equals checks (they …

ehmm:
…are in use

in other runtimes):

1. If pointers are equal it's the same object
2. If we are comparing singletons both types must be singletons

### DIFF
--- a/runtime/Go/antlr/v4/prediction_context.go
+++ b/runtime/Go/antlr/v4/prediction_context.go
@@ -115,6 +115,9 @@ func (p *PredictionContext) Hash() int {
 }
 
 func (p *PredictionContext) Equals(other Collectable[*PredictionContext]) bool {
+	if p == other {
+		return true
+	}
 	switch p.pcType {
 	case PredictionContextEmpty:
 		otherP := other.(*PredictionContext)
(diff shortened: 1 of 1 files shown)


jimidle:
I would need to do some serious re-checking to accept this PR. I seem to think that I already went down that path and it fails in cases that I do not recall right now.

ehmm:
Hi @jimidle - I saw some of your previous PRs and the fixes you made, and I understand it is a delicate area - this patch only has 2 additions.

1. Add a check that if it's the same object itself return true - I don't think there can be a problem with this one as it must be true
2. If we are comparing a SinglePredictionContext to any other type then return false, I looked at the code (and the Java/C# implementations they both have this check) I can't see where this will provide a false positive, but if you want I can try and take another look or run tests.

We are using antlr for an internal project with quite a complex grammar and I have checked the results after the change and the parse trees are the same, but if you have other tests that you can think of I can run them as well

parrt:
Hi all. I'm definitely supportive of improving the runtime, but we definitely have to be careful not to mess up the performance improvements and restructuring @jimidle did.  seems that all of the unit tests are passing but maybe @jimidle  can try it on some of his super secret internal grammars?

jimidle:
The change does not look incorrect to me and the reasoning is sound. But I definitely want to double check as it is a very change sensitive area that was completely incorrect before the Go changes. I will try to verify this this week.

Also, maybe I could take a look at the grammar and see if we could avoid ambiguities that cause a drop into LL mode.

parrt:
Maybe @mike-lischke's fancy new benchmarking system could be applied here to test Go target?

jimidle:
OK - I took a longer look at this and it should not cause any issues. So let's bring this in and I will migrate it to the read-only module for a release shortly.

PR #4574, opened 2024-04-02
[TS] Add property versions of raw fields to TS

Phlosioneer:
JS properties exist for the `_type` and `_input` fields, but they aren't present in the typescript definition file. This commit adds those properties.

The commit adds doc comments pointing developers toward the corresponding properties. I don't know if there's a deprication process for removing the old fields. `inputStream`'s setter function does a reset of the lexer, but otherwise the properties are normal.

### DIFF
--- a/runtime/JavaScript/src/antlr4/Lexer.d.ts
+++ b/runtime/JavaScript/src/antlr4/Lexer.d.ts
@@ -7,7 +7,11 @@ export declare class Lexer extends Recognizer<number> {
 
     static DEFAULT_MODE: number;
 
+    /**
+     * Use inputStream property instead
+     */
     _input: CharStream;
+    inputStream: CharStream;
     _interp: LexerATNSimulator;
     text: string;
     line: number;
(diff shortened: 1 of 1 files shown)


runtime/JavaScript/src/antlr4/Lexer.d.ts
```diff
@@ -7,15 +7,23 @@ export declare class Lexer extends Recognizer<number> {
 
     static DEFAULT_MODE: number;
 
+    /**
```
ericvergnaud, line 10: Please use 
/**
 * @deprecated .../...
 */

ericvergnaud (changes requested):

runtime/JavaScript/src/antlr4/Lexer.d.ts
ericvergnaud: @Phlosioneer Hi there, it's been a while... any progress ?

PR #4575, opened 2024-04-02
[JS] Fix Parser.getSourceName returning undefined

Phlosioneer:
`Parser.getSourceName` was expecting a field `sourceName` on `TokenStream`, but that doesn't exist on any `TokenStream` implementation. This commit fixes the bug by calling `getSourceName`, which *is* defined on `BufferedTokenStream`.

This commit also removes some nearby commented-out code that didn't make any sense.


ericvergnaud:
Thanks!

PR #4579, opened 2024-04-07
[TS] add missing channel constants

### LINKED ISSUE #4572: Tool does not generate any constants for channels for the TypeScript target
I don't know if this is by design: channels declared in a lexer grammar file are not generated into constants in the generated lexer. For an example grammar, see [PhpLexer.g4](https://github.com/antlr/grammars-v4/blob/c7883a594bf622200c561db181848d277fcd49a7/php/PhpLexer.g4#L34), then run the tool with `-Dlanguage=TypeScript` and look at the generated .ts files.

It seems that this should be possible because it generates constants for token types, e.g., `public static readonly PHPStart = 4;`.

ericvergnaud:
The title says it all

Fixes #4572

### DIFF
--- a/tool/resources/org/antlr/v4/tool/templates/codegen/TypeScript/TypeScript.stg
+++ b/tool/resources/org/antlr/v4/tool/templates/codegen/TypeScript/TypeScript.stg
@@ -870,6 +870,9 @@ import <superClass> from './<superClass>.js';
 export default class <lexer.name> extends <superClass; null="Lexer"> {
 	<lexer.tokens:{k | public static readonly <k> = <lexer.tokens.(k)>;}; separator="\n">
 	public static readonly EOF = Token.EOF;
+	<if(lexer.escapedChannels)>
+	<lexer.escapedChannels:{c| public static readonly <c> = <lexer.escapedChannels.(c)>;}; separator="\n">
+	<endif>
 	<rest(lexer.modes):{m | public static readonly <m> = <i>;}; separator="\n">
 
 	public static readonly channelNames: string[] = [ "DEFAULT_TOKEN_CHANNEL", "HIDDEN"<if (lexer.channelNames)>, <lexer.channelNames:{c| "<c>"}; separator=", ", wrap, anchor><endif> ];


kaby76:
Thank you. This fixes the problem.

Note, there are still missing methods for [lexer commends](https://github.com/antlr/antlr4/blob/master/doc/lexer-rules.md#lexer-commands), like [setChannel()](https://github.com/antlr/antlr4/blob/1855d9aa4c5ebb15284b95a07048b6260304629c/runtime/Java/src/org/antlr/v4/runtime/Lexer.java#L315-L317). Some of the set/get methods, which are in Java, are inconsistent in TypeScirpt. For example, [_type in TypeScript is a public field](https://github.com/antlr/antlr4/blob/1855d9aa4c5ebb15284b95a07048b6260304629c/runtime/JavaScript/src/antlr4/Lexer.d.ts#L18), whereas in Java there are [get](https://github.com/antlr/antlr4/blob/1855d9aa4c5ebb15284b95a07048b6260304629c/runtime/Java/src/org/antlr/v4/runtime/Lexer.java#L311) and [set](https://github.com/antlr/antlr4/blob/1855d9aa4c5ebb15284b95a07048b6260304629c/runtime/Java/src/org/antlr/v4/runtime/Lexer.java#L307) methods.

But, I wonder about [text in TypeScript](https://github.com/antlr/antlr4/blob/1855d9aa4c5ebb15284b95a07048b6260304629c/runtime/JavaScript/src/antlr4/Lexer.d.ts#L12) vs [_text in JavaScript](https://github.com/antlr/antlr4/blob/1855d9aa4c5ebb15284b95a07048b6260304629c/runtime/JavaScript/src/antlr4/Lexer.js#L81), whereas in [setText() in Java sets _text](https://github.com/antlr/antlr4/blob/1855d9aa4c5ebb15284b95a07048b6260304629c/runtime/Java/src/org/antlr/v4/runtime/Lexer.java#L296-L298). There is no `this.text` in the JS code.

The fields in the lexer object are used when calling [emit()](https://github.com/antlr/antlr4/blob/1855d9aa4c5ebb15284b95a07048b6260304629c/runtime/JavaScript/src/antlr4/Lexer.js#L213-L216).


I can work around the lexer interface by directly manipulating the token itself by overriding emit().

ericvergnaud:
@kaby76 can you file separate issues/PRs for these other missing items ?

PR #4584, opened 2024-04-10
Remove combineCommonParents function which does nothing

andreasbuhr:
The function combineCommonParents in PredictionContext.cpp had no effect. But it consumed a lot of CPU cycles. This patch removes it.

@jcking it would be nice if you could comment was your intention was here.


andreasbuhr:
No, this is the wrong fix, I'll come up with a new fix.

PR #4593, opened 2024-04-14
[Python3] Replace deprecated `assertEquals()` to fix py3.12 compat

mgorny:
Replace the `assertEquals()` method calls in the Python3 test suite with the modern `assertEqual()`.  The former was removed in Python 3.12, and therefore is causing the tests to fail.  The latter is available in all versions of Python 3.


mgorny:
Thanks!

PR #4597, opened 2024-04-17
Add projects to test typescript support with different compiler options

JesusTheHun:
Following #4218, this PR adds three test projects that use ESM and CJS with npm and one that uses ESM with pnpm.

The projects are compiled with different typescript compiler options to ensure compatibility.

This PR focuses on testing that antlr4 can be imported in TS projects. This is why it does not include non-TS test projects.


runtime/JavaScript/package.json
```diff
@@ -42,7 +42,8 @@
   },
   "scripts": {
     "build": "webpack",
-    "test": "jasmine",
+    "test": "jasmine && npm run test:projects",
+    "test:projects": "bash test-projects.sh",
```
ericvergnaud, line 46: can we rename `projects` to `builds` ?

runtime/JavaScript/spec/imports/setups/node-cjs-ts/test.sh
```diff
@@ -0,0 +1,19 @@
...
+
+failure=0
+
+for tsconfig in "${tsconfigFiles[@]}"; do
+  echo -n "$tsconfig "
+
+  ./node_modules/.bin/tsc -p $tsconfig || { failure=1 ; echo "FAIL tsc: $tsconfig"; }
+  result=$(node ./tsOutput/index.js)
+  [[ $result == "OK" ]] && echo "OK"
+  [[ $result != "OK" ]] && { failure=1 ; echo "FAIL runtime: $tsconfig"; }
```
ericvergnaud, line 15: 'FAIL loading runtime with config' might be easier to understand in logs ?

runtime/JavaScript/spec/imports/setups/node-esm-ts/test.sh
```diff
@@ -0,0 +1,20 @@
...
+
+failure=0
+
+for tsconfig in "${tsconfigFiles[@]}"; do
+  echo -n "$tsconfig "
+
+  ./node_modules/.bin/tsc -p $tsconfig || { failure=1 ; echo "FAIL tsc: $tsconfig"; }
+  result=$(node ./tsOutput/index.js)
+  [[ $result == "OK" ]] && echo "OK"
+  [[ $result != "OK" ]] && { failure=1 ; echo "FAIL runtime: $tsconfig"; }
```
ericvergnaud, line 16: same here

runtime/JavaScript/package.json
```diff
@@ -42,7 +42,8 @@
   },
   "scripts": {
     "build": "webpack",
-    "test": "jasmine",
+    "test": "jasmine && npm run test:projects",
```
ericvergnaud, line 45: can we rename `projects` to `builds` ?

ericvergnaud (changes requested):

runtime/JavaScript/package.json
JesusTheHun: do you also want me to rename the directory "setups" to "builds" ?

ericvergnaud: Yes that would make sense

JesusTheHun:
> Looks like the tests fail ? See for example https://github.com/antlr/antlr4/actions/runs/8723991967/job/23933679496?pr=4597

I fail to run the tests locally using maven. I haven't touch to java in a while so I'm a bit rusty at identifying compilation chain issues 😅
In addition to the requested changes I've committed a fix, hopefully the one we need.

runtime/JavaScript/spec/imports/builds/node-cjs-ts/package.json
```diff
@@ -0,0 +1,18 @@
...
+  "description": "",
+  "main": "index.js",
+  "type": "commonjs",
+  "scripts": {
+    "test": "bash test.sh"
+  },
+  "author": "",
+  "license": "ISC",
+  "dependencies": {
+    "antlr4": "file:../../../.."
```
ericvergnaud, line 13: shouldn't we be using a local build rather than the source folder ?

runtime/JavaScript/spec/imports/builds/node-esm-ts/package.json
```diff
@@ -0,0 +1,18 @@
...
+  "description": "",
+  "main": "index.js",
+  "type": "module",
+  "scripts": {
+    "test": "bash test.sh"
+  },
+  "author": "",
+  "license": "ISC",
+  "dependencies": {
+    "antlr4": "file:../../../.."
```
ericvergnaud, line 13: shouldn't we be using a local build rather than the source folder ?

ericvergnaud (changes requested):
Fixes the failing tests, but not sure it satisfies the requirement to test against the built package (vs the source project)

JesusTheHun:
> Fixes the failing tests, but not sure it satisfies the requirement to test against the built package (vs the source project)

I was hesitant to test against the build because that would require to compile every time you make a change, which is annoying during development. But I can make the change if you want. It is safer indeed.

ericvergnaud:
> @ericvergnaud just pinging in case you didn't see my edit above

mmm... not sure what you mean by _your build process bundle the javascript, but do not include the type definitions_ ?
if you call `npm build`, it will create the `dist` folder that gets published to npm

JesusTheHun:
> Ah I see what you mean. But how can we be sure that your test samples are using the `dist` folder ?

The test uses the package.json. The package.json references the `dist` folder for the javascript, and the `src` for the type definitions.

JesusTheHun:
@ericvergnaud after migrating the project using this lib to pnpm, I noticed the types stopped to work altogether. After looking around I noticed that the imports inside the definition files did not include the file extension, which is required for modern node usage.

ericvergnaud:
Hi, sorry for the delay.
I suspect that the reason you need the .js extensions is because the tests are running against the source code rather than the built package.
Node does not support TypeScript, and should be using the web packed module.
I suggest you revert commit 40bb02cb53137ea6e4aee3a43488729d90d84f05, which makes this PR way too broad.

JesusTheHun:
@ericvergnaud The need for the extension comes from the ˋnode16ˋ TypeScript target.
TypeScript uses the ˋd.tsˋ files, so it's unrelated to the packaged source files.

ericvergnaud:
@JesusTheHun sorry but I don't follow:
 - I've been using these .ts files "as is" for years now without any problem
 - Typescript uses the .ts files, and doesn't need the .js extension
 - node16 is unaware of these .ts files, and should be using the webpacked module 
...
What am I missing ?
It seems the issue appears when introducing pnpm ? That's a separate topic altogether. 
Can we restrict this PR back to its primary intent ?

JesusTheHun:
@ericvergnaud When I mention `node16` I'm talking about the TS compiler option "target". Not the node runtime version.

`pnpm` is the most strict package manager. It enforces standards and prevent bad usage instead of working around them. This may be the reason why only `pnpm` issues errors.

The intent of the PR is to make sure projects can import Antlr without issue, right ? 
If so, the commit is required, otherwise `pnpm` users won't be able to import Antlr.

Also, I'm on holidays without my machine. So I wont be able to revert the commit myself before August 5th.

That being said, I'm curious about the reason for rejecting the commit. Ok it makes a lot of file changes, but they are trivial changes, it's not like they each require a review.

ericvergnaud:
The reason is that I'm in favor of small steps. 
We currently don't support pnpm and I don't need that additional responsibility (similarly we only support webpack).
So let's keep this PR as small as possible and ensure it covers the currently supported environments:
 - node >= 16
 - typescript >= 4.8
 - webpacked code from the `dist` folder

JesusTheHun:
I understand the need for small steps.

`pnpm` falls into the support for node 16, since it's officially distributed by Node.js via `corepack`.

I understand the commit touches a lot of files but the changes are tiny and identical in each file.
I don't see any drawback.

ericvergnaud:
@parrt not sure what is your ETA for the release ?

ericvergnaud:
> @ericvergnaud I'll be back in a few days, can you hold the release until then ? I assume you also want me to delete the test project for `pnpm` ?

Yes please rollback pnpm related changes

parrt:
I’m gonna try for this weekend maybe tomorrow

Dictation in use. Please excuse homophones, malapropisms, and nonsense.

ericvergnaud:
@JesusTheHun hi there, it's been a while... Any progress ?

ericvergnaud:
@JesusTheHun Thanks, this is looking good. The TS tests fail though. Could you take a look ?

JesusTheHun:
> @JesusTheHun Thanks, this is looking good. The TS tests fail though. Could you take a look ?

This looks unrelated to this PR, as a previous action failed the same way. See https://github.com/antlr/antlr4/actions/runs/10863756846/job/30148277515

ericvergnaud:
@JesusTheHun Can you rebase ?

JesusTheHun:
@ericvergnaud all clear ☝️

ericvergnaud:
@parrt blessed
@JesusTheHun Many thanks for this !!!

PR #4617, opened 2024-05-15
[Cpp] Fix cpp memory leaks

njnobles:
This PR fixes memory leaks caused by the StaticData objects associated with the generated Parser and Lexer classes. 

Found using the CRT library in VS2022. With the demo project, it produces a leak report similar to this:
```
Detected memory leaks!
Dumping objects ->
{21126} normal block at 0x000001AAB1D11180, 32 bytes long.
 Data: <                > 00 95 CF B1 AA 01 00 00 A0 9A CF B1 AA 01 00 00 
{20999} normal block at 0x000001AAB1D141E0, 32 bytes long.
 Data: <`       `       > 60 1F D1 B1 AA 01 00 00 60 1F D1 B1 AA 01 00 00 
{20982} normal block at 0x000001AAB1D134C0, 32 bytes long.
 Data: <         1      > C0 A8 CD B1 AA 01 00 00 C0 31 D1 B1 AA 01 00 00 
{20981} normal block at 0x000001AAB1D20DF0, 128 bytes long.
 Data: <                > 00 19 D1 B1 AA 01 00 00 00 19 D1 B1 AA 01 00 00 
...<~7000 lines long>
```
Valgrind on Ubuntu produces a similarly large leak detection.

Switching the StaticData objects to be unique_ptr instead of raw pointers removes all but 1 leak.
The remaining leak is the ATNDeserializationOptions which is also fixed by a unique_ptr.

### DIFF
--- a/runtime/Cpp/runtime/src/atn/ATNDeserializationOptions.cpp
+++ b/runtime/Cpp/runtime/src/atn/ATNDeserializationOptions.cpp
@@ -14,7 +14,7 @@ ATNDeserializationOptions::ATNDeserializationOptions(ATNDeserializationOptions *
       _generateRuleBypassTransitions(options->_generateRuleBypassTransitions) {}
 
 const ATNDeserializationOptions& ATNDeserializationOptions::getDefaultOptions() {
-  static const ATNDeserializationOptions* const defaultOptions = new ATNDeserializationOptions();
+  static const std::unique_ptr<const ATNDeserializationOptions> defaultOptions = std::make_unique<const ATNDeserializationOptions>();
   return *defaultOptions;
 }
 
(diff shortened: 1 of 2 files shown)


njnobles:
@ericvergnaud 
Hi Eric, sorry to ping you directly, but I was wondering if you'd be able to review this PR? Or do you know of someone else who would be more appropriate as a reviewer?
Thank you!

njnobles:
Hi @parrt, would it be possible to get someone to review this PR? Thank you!

parrt:
Unfortunately I'm not qualified to evaluate this :(  Maybe @mike-lischke ? @jcking ?

mike-lischke (approved):

parrt:
Thanks, @mike-lischke !

njnobles:
Thank you, @parrt and @mike-lischke!

PR #4618, opened 2024-05-17
Add a Force Compile Property

georgecao:
When I upgrade antlr to a new version, but keep the grammar files untouched, the grammar files will not be compiled but there is a warning message complaining the runtime version changed.

So I add a property forceCompile to skip the checksum and last modification date check and just re-compile the grammars.

### DIFF
--- a/antlr4-maven-plugin/src/main/java/org/antlr/mojo/antlr4/Antlr4Mojo.java
+++ b/antlr4-maven-plugin/src/main/java/org/antlr/mojo/antlr4/Antlr4Mojo.java
@@ -106,6 +106,12 @@ public class Antlr4Mojo extends AbstractMojo {
 	protected boolean forceATN;
 
 	/**
+	 * Skip grammar file change detect and just compile them.
+	 */
+	@Parameter(property = "antlr4.forceCompile", defaultValue = "false")
+	protected boolean forceCompile;
+
+	/**
 	 * A list of grammar options to explicitly specify to the tool. These
 	 * options are passed to the tool using the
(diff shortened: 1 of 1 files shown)


jimidle:
I think you just need mvn clean. I am about to make a few tweaks to the
plugin though. So I’ll take a look. But I doubt we can accept this.

mvn clean compile

georgecao:
I add the plugin as follow: 
`              <plugin>
                    <groupId>org.antlr</groupId>
                    <artifactId>antlr4-maven-plugin</artifactId>
                    <version>4.13.1</version>
                </plugin>`
and I tested `mvn clean compile`,  but the grammar files were not re-compiled. 
I read the code, beside the modification date of the grammar file, the plugin also check `*.tokens` files.  I can force the re-compile by deleting the *.tokens files then run `mvn clean compile`.

jimidle:
Your configuration is wrong I think.

```xml
...
<executions>
 <execution>
  <goals>
   <goal>antlr4</goal>
  </goals>
 </execution>
</executions>
<configuration>
  <visitor>true</visitor>
  <sourceDirectory>src/main/antlr4</sourceDirectory>
</configuration>  
```
If your configuration is correct then check permissions and file times. There is no need for this flag.

khmarbaise:
I'm using a usual configuration https://github.com/khmarbaise/rpn-calculator/blob/master/pom.xml and it work without any issue... no supplemental source directory configuration needed... just follow the convention... https://github.com/khmarbaise/rpn-calculator/blob/master/rpn/pom.xml

PR #4620, opened 2024-05-21
feat: Use Go min/max builtins

tklauser:
The `min` and `max` builtins are available since Go 1.21[^1]. The github.com/antlr4-go/antlr/v4 is specifying Go 1.22 in its go.mod file, so the builtins can be used.

[^1]: https://go.dev/doc/go1.21#language

### DIFF
--- a/runtime/Go/antlr/v4/tokenstream_rewriter.go
+++ b/runtime/Go/antlr/v4/tokenstream_rewriter.go
@@ -641,22 +641,3 @@ func reduceToSingleOperationPerIndex(rewrites []RewriteOperation) map[int]Rewrit
 	}
 	return m
 }
-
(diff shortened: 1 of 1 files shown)


tklauser:
The `go.mod` in this repo still specifies Go 1.20:

https://github.com/antlr/antlr4/blob/6197d6cdadb206576a302b3d14d035646d738a02/runtime/Go/antlr/v4/go.mod#L3

Which obviously this PR is not compatible with. I got confused because https://github.com/antlr4-go/antlr is using Go 1.22:

https://github.com/antlr4-go/antlr/blob/4d7e18847d8881a176db3d7cca909e3f3942a05f/go.mod#L3

Closing for now.

PR #4621, opened 2024-05-22
Rewrite SubContext* rules and update tests

jcjaskula-aws:
I added an additional argument to `SubContextLocal` and `SubContextMember` to avoid hardcoding indices in the test, i.e. 
```
SubContextLocal({<Production("e")>(0)}, {<Result("v")>})>
```

To give more context, I have been able to create a runtime/generator for Julia. This is the only modification I need to push to the main Antlr4 repos to keep my implementation "standalone". The change will help me release my Julia runtime/generator eventually and I believe that it is transparent for all other target languages.

FYI, in Julia, I need to write `SubContextLocal` as:
```
SubContextLocal(ctx, subctx, local, arg) ::= "<subctx>(<ctx>, <arg>).<local>"
```


parrt:
Unfortunately I can't accept any more new targets to main repo; it's hard enough managing what I've got. especially since I just don't have time for this project at the moment. So, as it is on the way to accepting a new target, probably not worth me digging into this. Sorry!!

jimidle:
I took a look and it seems fairly innocuous, but before I look further, I would like to know a bit more about why you have to do this for Julia and cannot find any other way? Also, hard coding for tests isn't the end of the world. But can you add a comment here that explains exactly why you have to do this? It may be a quirk of Julia, but requiring changes in other target templates usually does not pass the sniff test. But right now, there is no context for me to see why this is the only solution in Julia.

jimidle:
I will need to look at why you need test changes for other targets though. It does not seem natural. I will try to give it some time and give Terence my recommendation.

PR #4622, opened 2024-05-22
fix(TypeScript): update and add missing definition in IntervalSEt.d.ts

yrue:
In the TypeScript target, I noticed that the IntervalSet.d.ts file didn't match the code in IntervalSet.js—they really should be the same. Because of this mismatch, I ran into the error TS2339: "Property 'addRange' does not exist on type 'IntervalSet'" when I tried to use IntervalSet utilities like addOne. So, in this PR, I've made changes to ensure they are consistent.

### DIFF
--- a/runtime/JavaScript/src/antlr4/misc/IntervalSet.d.ts
+++ b/runtime/JavaScript/src/antlr4/misc/IntervalSet.d.ts
@@ -1,13 +1,41 @@
 import {Interval} from "./Interval";
+import { Token } from "../Token";
 
-export declare class IntervalSet {
+export default class IntervalSet {
+    private intervals: Interval[] | null;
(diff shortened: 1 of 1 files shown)


mike-lischke:
Instead of fixing each problem in the standard TS runtime step by step you could use a really complete TS runtime, which is also faster ([antlr4ng](https://github.com/mike-lischke/antlr4ng)). Saves you some headache.

yrue:
> Instead of fixing each problem in the standard TS runtime step by step you could use a really complete TS runtime, which is also faster ([antlr4ng](https://github.com/mike-lischke/antlr4ng)). Saves you some headache.

Thanks for the recommendation, I'll check it out :)

PR #4624, opened 2024-05-22
Mark all generated C# classes with [GeneratedCode] attribute

martinskuta:
Hi,

we use ANTLR to generate a C# parser for our grammar and we run for each build static code analysis and for obvious reasons we exclude generated code, including the generated ANTLR classes. For ANTLR generated code we have to manually exclude the generated namespace from the inspection, because not all of the generated classes are marked with [[GeneratedCode]](https://learn.microsoft.com/en-us/dotnet/api/system.codedom.compiler.generatedcodeattribute) attribute, none of the public nested classes are, only the top level parent ones for Parser/Visitor are marked as generated.

We are using NDepend and JetBrains InspectCode to analyse the code. With the changes to the CSharp target template I propose in this PR no further work has to be done to exclude ANTLR generated code from the inspection as both tools understand that attribute, as do probably any other C# static analysis tools.

### DIFF
--- a/tool/resources/org/antlr/v4/tool/templates/codegen/CSharp/CSharp.stg
+++ b/tool/resources/org/antlr/v4/tool/templates/codegen/CSharp/CSharp.stg
@@ -860,6 +860,8 @@ CaptureNextTokenType(d) ::= "<d.varName> = TokenStream.LA(1);"
 
 StructDecl(struct,ctorAttrs,attrs,getters,dispatchMethods,interfaces,extensionMembers,signatures,
            superClass={ParserRuleContext}) ::= <<
+[System.CodeDom.Compiler.GeneratedCode("ANTLR", "<file.ANTLRVersion>")]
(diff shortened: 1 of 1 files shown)


sharwell:
📝 This should not be necessary since the entire file is marked as generated code:
https://github.com/antlr/antlr4/blob/5a68af0109d911e3bcd2e2c0eaaaecf25dbee66c/tool/resources/org/antlr/v4/tool/templates/codegen/CSharp/CSharp.stg#L244-L252

If a tool is failing to account for this designator at the top of the file, it would be considered a bug in that tool that should be corrected.

martinskuta:
You are right. I didn't think about it this way, but makes sense that a symbol that is inside of a containing symbol that is generated should be also considered generated. My thinking was that if you create a partial class then you could create a nested class that would not be considered auto generated, but in that case the symbol would be the partial class without the attribute I assume.

Thx for the reply, I will report it to NDepend.

PR #4626, opened 2024-05-22
Add some methods for Go runtime

RobEin:
Added methods for CommonToken class:
SetChannel
SetStop
SetText    // also exists in Token (Token.SetText may be unnecessary)
SetTokenType

Added constructor for CommonToken class:
CommonToken.NewCommonTokenFromToken

Removed method in CommonToken class:
Clone

Added methods for Lexer class:
GetModeStack
GetMode

The equivalent of the NewCommonTokenFromToken constructor can also be found in the [Java CommonToken class](https://www.antlr.org/api/Java/org/antlr/v4/runtime/CommonToken.html#%3Cinit%3E(org.antlr.v4.runtime.Token)).
The NewCommonTokenFromToken constructor is similar to the clone method, in that a cloned CommonToken instance is created in both.
NewCommonTokenFromToken accepts both Token and CommonToken parameters for cloning.
This made the clone method redundant.

### DIFF
--- a/doc/cpp-target.md
+++ b/doc/cpp-target.md
@@ -1,6 +1,6 @@
 # C++
 
-The C++ target supports all platforms that can either run MS Visual Studio 2013 (or newer), XCode 7 (or newer) or CMake (C++11 required). All build tools can either create static or dynamic libraries, both as 64bit or 32bit arch. Additionally, XCode can create an iOS library. Also see [Antlr4 for C++ with CMake: A practical example](http://blorente.me/beyond-the-loop/Antlr-cpp-cmake/).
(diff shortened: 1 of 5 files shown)


jimidle (commented):
I will take a look at these at the weekend and if all is good I will put it in with the generic tree walker

RobEin:
Can I expect a review on this?
I have developed a Go target (about 400 lines) for a Python parser where these new features would be needed.
I also spent a lot of time modifying the Go runtime for these new features.
So I had a lot of work in it and it may be useful for others.

RobEin:
[Here](https://github.com/RobEin/ANTLR4-parser-for-Python-3.12/tree/main/port_Go) is a Python parser with a Go port that already uses the above modifications.

PR #4644, opened 2024-06-21
[JavaScript-runtime] Support TypeScript bundler module resolution for types

guidsdo:
When using the published `npm` package in my TypeScript project, TypeScript gave the error that it couldn't find the types.

This fix makes sure that the type field in the `package.json` is not in a specific exports target, which fixes the issue. The same types are exposed for both the web and node target, so they were combined in the same field.

### DIFF
--- a/runtime/JavaScript/package.json
+++ b/runtime/JavaScript/package.json
@@ -52,17 +52,16 @@
   "exports": {
     ".": {
       "node": {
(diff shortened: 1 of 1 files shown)


ericvergnaud:
@parrt blessed
@guidsdo Thanks for this!

PR #4646, opened 2024-06-23
Add new features and modify final fields

RobEin:
**Added method:**
getMode()

**Added read-only property:**
modeStack

**Modified fields:**
DEFAULT_MODE
MORE
SKIP
DEFAULT_TOKEN_CHANNEL
HIDDEN
MIN_CHAR_VALUE
MAX_CHAR_VALUE

Final fields cannot be used in the case clause of the switch statement.
error message: 
> "The expression of a constant pattern must be a valid constant."

e.g:
The folowing example does not work with final field but it works with constant field:
```
switch (curMode) {
  case Lexer.DEFAULT_MODE:
    ...
}

```
It would be worthwhile to review the other classes in this regard.


vHanda:
ping. I'd love to have this merged soon.

RobEin:
Can we expect a review on this?
I developed a Dart target (about 300 lines) for a Python parser where these new features would be needed.
I also spent a lot of time modifying the Dart runtime for these new features.
So I had a lot of work in it and it may be useful for others.

RobEin:
[Here](https://github.com/RobEin/ANTLR4-parser-for-Python-3.12/tree/main/port_Dart) is a Python parser with a Dart port that already uses the above modifications.

PR #4661, opened 2024-07-22
[Cpp] Use target_include_dirs for cmake targets

njnobles:
Use target_include_dirs to attach include directories to targets. This allows setting include directories for the `INSTALL_INTERFACE` which will get exported into `antlr4-targets.cmake`.

`antlr4-targets.cmake` will now have an `INTERFACE_INCLUDE_DIRECTORIES` property set like so:
```
set_target_properties(antlr4_shared PROPERTIES
  INTERFACE_COMPILE_DEFINITIONS "ANTLR4CPP_EXPORTS"
  INTERFACE_INCLUDE_DIRECTORIES "${_IMPORT_PREFIX}/include/antlr4-runtime"
)
```
When downstream users pull in antlr4-runtime using `find_package(antlr4-runtime)`, the imported targets `antlr4_static` and `antlr4_shared` will automatically pick up the appropriate include directories instead of needing to manually use `ANTLR4_INCLUDE_DIR`.

### DIFF
--- a/runtime/Cpp/runtime/CMakeLists.txt
+++ b/runtime/Cpp/runtime/CMakeLists.txt
@@ -7,7 +7,9 @@ if (NOT ANTLR_BUILD_SHARED AND NOT ANTLR_BUILD_STATIC)
   message(FATAL_ERROR "Options ANTLR_BUILD_SHARED and ANTLR_BUILD_STATIC can't both be OFF")
 endif()
 
-include_directories(
+set(libantlrcpp_INCLUDE_INSTALL_DIR "include/antlr4-runtime")
+
+set(libantlrcpp_INCLUDE_DIRS
   ${PROJECT_SOURCE_DIR}/runtime/src
   ${PROJECT_SOURCE_DIR}/runtime/src/atn
   ${PROJECT_SOURCE_DIR}/runtime/src/dfa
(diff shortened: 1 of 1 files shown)


njnobles:
Hi @mike-lischke, 
Would you be able to review this when you have some time?
Thanks!

mike-lischke:
Sorry, I don't know enough about cmake to validate the changes.

njnobles:
Thanks for responding @mike-lischke 
I put together this small test repo that should help validate the changes: https://github.com/njnobles/antlr-cmake-example
It sets up 2 simple projects, one to build the runtime and one to consume the runtime. It has options to build without my fix applied which should demonstrate the issue I'm trying to fix and another option to build with my fix. 
Should be able to build on any platform.

Would this be enough to help you validate? Or is there someone else that I could request a review from that might have more cmake knowledge?
Thank you!

mike-lischke (approved):

mike-lischke:
Thanks for providing that test repo, it helps to understand. 

@parrt I think this PR can be merged.

hzeller:
This does seem to create some wierd artifacts where the location of the includes is actually in the wrong place

Here is one issue report in nix: https://github.com/NixOS/nixpkgs/issues/379757 it seems like the dev component (with the includes) ends up being installed where the library is installed, so the generated `antlr4-targets.cmake` has a `set(_IMPORT_PREFIX  ...)` that points to a binary install prefix, not the dev prefix.

I don't know enough cmake to make sense of it, but maybe @njnobles does ?

PR #4675, opened 2024-08-12
Update to eclipse-temurin:21 and remove appuser

james-m-henderson:
It seems this appuser wasn't being used at all, and the USER line wasn't being set to enable it. In addition, the adduser utility didn't exist in the old or updated version of temurin. This seems simpler and works.


ericvergnaud:
Thanks for this
@parrt blessed

PR #4682, opened 2024-08-25
Update python-target.md

medoror-mo:
Updating examples migrating start_() -> root()


ericvergnaud:
Hi, thanks for this, but your proposal is incorrect. The example grammar does not have a `root` rule, hence why the example calls the `start_` rule (it could actually call any rule)

PR #4691, opened 2024-09-05
Remove unused function createThin()

alexeyerm:
Remove unused function createThin() from common_token_factory.go, CommonTokenFactory.py, CommonTokenFactory.js.

The unexported function 'antlr.createThin' is never used but has a potential bug. Using nil constant at common_token_factory.go:52, it is passed as 1st parameter in call to function 'antlr.NewCommonToken' at common_token_factory.go:52, where it is dereferenced at token.go:140.


alexeyerm:
not enough signed

PR #4709, opened 2024-10-02
Explicitly ignore unused var in TypeScript and @typescript-eslint

guidsdo:
When using some strict TypeScript options, you can't compile the code anymore because of this specific line. Unfortunately I don't know enough about the necessity of the line to help in another way, so that's why I added this simple fix.

There is another error for unused imports in the Lexer file, but I wasn't able to test that one out since that import seemed to have been changes recently.


ericvergnaud (approved):
Thanks for this!

ericvergnaud:
@parrt blessed

PR #4711, opened 2024-10-03
Add LexerNoViableAltException typescript class

happytomatoe:
Currently I am using  latest antlr package from npm `"antlr4": "^4.13.2"` and I don't know how to check if an exception is the proposed class without importing js sources.

### DIFF
--- a/runtime/JavaScript/src/antlr4/error/LexerNoViableAltException.d.ts
+++ b/runtime/JavaScript/src/antlr4/error/LexerNoViableAltException.d.ts
@@ -1,0 +1,13 @@
+import {ATNConfigSet} from "../atn";
+import {Recognizer} from "../Recognizer";
+import { Token } from "../Token";
+import {RecognitionException} from "./RecognitionException";
+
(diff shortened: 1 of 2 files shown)


ericvergnaud:
Thanks for this. You also need to add it to `error/index.d.ts` !

happytomatoe:
> Thanks for this. You also need to add it to `error/index.d.ts` !

Done

ericvergnaud:
You also need to sign the DCO. See https://github.com/antlr/antlr4/pull/4711/checks?check_run_id=31039893768 on how to do this.

jimidle:
I would also comment that levers should never throw this error. Always have a last rule:BADCHAR: . ;And let the parser handle it. On Oct 3, 2024, at 12:33, RomanLukash340 ***@***.***> wrote:﻿

You also need to sign the DCO. See https://github.com/antlr/antlr4/pull/4711/checks?check_run_id=31039893768 on how to do this.

Done

ericvergnaud:
@parrt blessed

PR #4718, opened 2024-10-19
Adding Typescript definition files to export Trees module.

### LINKED ISSUE #4715: antlr4 Trees module missing in JS runtime
I am trying import Trees module from JS runtime, but get "Cannot find module 'antlr4/tree/Trees' or its corresponding type declarations" error. See below screenshot.

Screenshot 2024-10-16 at 8 52 55 PM

`
import Trees from "antlr4/tree/Trees";
`

Based on the source code [Trees.js](https://github.com/antlr/antlr4/blob/dev/runtime/JavaScript/src/antlr4/tree/Trees.js), I should be able to import use above syntax.

Can someone look into why this module was not packaged into the runtime?

Target runtime: JavaScript / Typescript
Runtime version: 4.13.2

ZengLawrence:
Adding Typescript definition files to export Trees module.

Fixes #4715

### DIFF
--- a/runtime/JavaScript/src/antlr4/tree/Trees.d.ts
+++ b/runtime/JavaScript/src/antlr4/tree/Trees.d.ts
@@ -1,0 +1,15 @@
+import { Parser } from "../Parser";
+import { ParseTree } from "./ParseTree";
+import { Tree } from "./Tree";
+
+export default Trees;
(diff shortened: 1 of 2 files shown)


ericvergnaud:
Thanks for this, LGTM.
@parrt blessed

PR #4726, opened 2024-11-06
Improve Python version handling

cjwatson:
Checking just the second element of a version tuple is never good style, and it makes it less clear what actual versions of Python are in question.

### DIFF
--- a/runtime/Python3/src/antlr4/Lexer.py
+++ b/runtime/Python3/src/antlr4/Lexer.py
@@ -11,7 +11,7 @@
 from io import StringIO
 
 import sys
-if sys.version_info[1] > 5:
+if sys.version_info >= (3, 6):
     from typing import TextIO
 else:
     from typing.io import TextIO
--- a/runtime/Python3/src/antlr4/Parser.py
+++ b/runtime/Python3/src/antlr4/Parser.py
@@ -3,7 +3,7 @@
 # Use of this file is governed by the BSD 3-clause license that
 # can be found in the LICENSE.txt file in the project root.
 import sys
-if sys.version_info[1] > 5:
+if sys.version_info >= (3, 6):
     from typing import TextIO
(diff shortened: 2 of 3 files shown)


runtime/Python3/src/antlr4/xpath/XPathLexer.py
```diff
@@ -2,7 +2,7 @@
 from antlr4 import *
 from io import StringIO
 import sys
-if sys.version_info[1] > 5:
+if sys.version_info >= (3, 6):
```
ericvergnaud, line 5: This file is generated so it'll be updated automatically on the next gen

runtime/Python3/tests/expr/ExprLexer.py
```diff
@@ -2,7 +2,7 @@
 from antlr4 import *
 from io import StringIO
 import sys
-if sys.version_info[1] > 5:
+if sys.version_info >= (3, 6):
```
ericvergnaud, line 5: This file is generated so it'll be updated automatically on the next gen

runtime/Python3/tests/expr/ExprParser.py
```diff
@@ -3,7 +3,7 @@
 from antlr4 import *
 from io import StringIO
 import sys
-if sys.version_info[1] > 5:
+if sys.version_info >= (3, 6):
```
ericvergnaud, line 6: This file is generated so it'll be updated automatically on the next gen

ericvergnaud (changes requested):

ericvergnaud:
@cjwatson thanks for this. Just rollback the changes on the generated files and we'll be good to go.

cjwatson:
OK, done.

ericvergnaud:
@cjwatson Thanks!
@parrt blessed

PR #4731, opened 2024-11-15
fix(JS: ErrorListener types)

elorzafe:
Fixes: https://github.com/antlr/antlr4/issues/4730

Updates TypeScript declaration file for `ErrorListener`. Additionally updates `DiagnosticErrorListener` that implements `ErrorListener` and adds a declaration file for `BitSet` class (`BitSet.d.ts`).

PR #4746, opened 2024-12-12
fix(JS: BitSet): Export the BitSet class

hutcheb:
As the BitSet class is now used in the ErrorListener class it should be exported as users are required to implement these new functions which use the BitSet class as one of their parameters.

It expands on #4731


ericvergnaud (approved):

ericvergnaud:
@parrt blessed

PR #4688, opened 2024-09-01
Upgrade to go 1.21, replace golang.org/x/exp/slices with slices

Tiscs:
Upgrade to go 1.21 to use `slices` package instead of `golang.org/x/exp/slices` package.

In the [antlr4-go/antlr](https://github.com/antlr4-go/antlr) project, the go version declared in [go.mod](https://github.com/antlr4-go/antlr/blob/main/go.mod) is already 1.22.

Issue #4663, opened 2024-07-26
Go runtime should not depend on Go 1.22 yet


mpsijm:
@jimidle bumped the Go version to 1.22 in https://github.com/antlr4-go/antlr/commit/4d7e18847d8881a176db3d7cca909e3f3942a05f (released as v4.13.1), but the Go version in this repo is still set to 1.20: https://github.com/antlr/antlr4/blob/master/runtime/Go/antlr/v4/go.mod#L3

My request is to downgrade the Go version in the https://github.com/antlr4-go/antlr repo, because it promises to be a clone of the Go runtime in this repository.

Another reason is that the version bump would force projects using the ANTLR Go runtime as dependency are forced to use at least Go 1.22, while Go 1.21 is still officially supported.

jimidle:
The source in that other tree is not yet updated. The “fix” is the other
way around.

mpsijm:
Thanks for the clarification! In that case, I would like to repeat my request to not yet depend on Go 1.22, since Go 1.21 is still supported :slightly_smiling_face:

jimidle:
as a apolicy, I keep the runtime at the latest version of Go whenever we make a new release. I won't go backwards though. Given Go's backward compatibility guarantee, it should not be an issue for anyone to upgrade, though I know corporate devops can be slow to do such things. Think of it this way: We know that Go 1.22.x has security fixes over Go 1.21 - so we should upgrade. 

What is your particular problem with upgrading?

mpsijm:
The problem is that, with this policy, this library _forces_ other modules that depend on it to upgrade their Go version, even when this is not strictly required. Our module structure is quite complex: while we like to keep our dependencies up-to-date as much as possible using `go get -u`, we only update the Go language version in a fixed cadence to keep the impact of this predictable. For example, we have a bunch of Docker images and other infrastructure that all need to be upgraded when one of our modules suddenly bumps their Go language version. When a dependency suddenly uses a higher Go language version, we have to manually revert the upgrade of this dependency, which has now been happening with https://github.com/antlr4-go/antlr.

Note that, since Go 1.21, the `go` command forcibly upgrades the Go version in a module to be the maximum version of its dependencies ([source](https://go.dev/ref/mod#go-mod-file-go)), which is probably why this problem only started popping up recently.

[...]

PR #4754, opened 2025-01-10
runtime(go): minimum requirement go1.18

xiaost:
golang.org/x/exp,
which requires newer go version,
is too heavy for previous implementation.

may duplicated with #4688 #4356 #4294, related #4663

### DIFF
--- a/.github/workflows/hosted.yml
+++ b/.github/workflows/hosted.yml
@@ -21,7 +21,7 @@ jobs:
       fail-fast: false
       matrix:
         os: [
-          macos-12,
+          macos-13,
           ubuntu-20.04,
           windows-2022
         ]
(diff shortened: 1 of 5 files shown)


xiaost:
We're proudly using antlr4, but we need to support users who're using go1.18. 
The go runtime of antlr4 doesn't use any new features, and go1.18 is good enough. 

I found that https://github.com/antlr4-go/antlr/ is using go1.22 while this repos is go1.20
Possibly it's caused by golang.org/x/exp

xiaost:
I'm checking the CI. 
seems like if we have no dependency, `go mod tidy` won't generate `go.sum` and it causes the test runner `java.io.FileNotFoundException` 

working on it.

xiaost:
JFYI. 
I also changed CI settings `macos-12` -> `macos-13`
see: https://github.com/actions/runner-images/issues/10721

jimidle:
I’m sorry. I thought I was fairly clear that we won’t be going back to
unsupported go releases. You can always make your own fork of course

PR #4759, opened 2025-01-21
Fix issue with index access on set in ParserInterpreter.py

PatrickStahl:
In the parserinterpreter.py file, there was an issue where an attempt was made to access a set using an index. However, sets do not support indexing, which led to an error.

This change fixes the issue by modifying the logic to access the elements of the set in a valid manner.

### DIFF
--- a/runtime/Python3/src/antlr4/ParserInterpreter.py
+++ b/runtime/Python3/src/antlr4/ParserInterpreter.py
@@ -108,7 +108,7 @@ class ParserInterpreter(Parser):
         tt = transition.serializationType
         if tt==Transition.EPSILON:
 
-            if self.pushRecursionContextStates[p.stateNumber] and not isinstance(transition.target, LoopEndState):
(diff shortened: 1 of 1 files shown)


kaby76:
Do you have a grammar that demos the error? This seems like a fairly important bug. I'd like to see if it occurs in grammars-v4. Thanks.

kaby76:
Interesting. I use ParserInterpreter to find all ambiguous parse trees, which is vital in understanding how to find grammar issues. I too needed to override the implementation because the one in the CSharp runtime is not synched up with the code in Java, e.g., the beginning of VisitState() is wrong in the CSharp code, and Reset() does not reset all the state variables in the CSharp code.

PatrickStahl:
So it appears that more than the Python implementation of the ParserInterpreter are flawed. That's interesting to know, and for future references, it would probably be good to review those implementations again.My problem was solved by extracting the information I needed from the parser instead, but the fix I proposed to the ParserInterpreter should definitely benefit its usage.Thank you for sharing your insights with me. I hope I could provide something beneficial to the ANTLR project.

PR #4760, opened 2025-01-21
Fix issue with index access on set in parserinterpreter.py

PatrickStahl:
In the parserinterpreter.py file, there was an issue where an attempt was made to access a set using an index. However, sets do not support indexing, which led to an error.

This change fixes the issue by modifying the logic to access the elements of the set in a valid manner.

### DIFF
--- a/runtime/Python3/src/antlr4/ParserInterpreter.py
+++ b/runtime/Python3/src/antlr4/ParserInterpreter.py
@@ -58,7 +58,7 @@ class ParserInterpreter(Parser):
         self._interp = ParserATNSimulator(self, atn, self.decisionToDFA, self.sharedContextCache)
 
     # Begin parsing at startRuleIndex#
-    def parse(self, startRuleIndex:int):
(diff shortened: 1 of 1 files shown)


runtime/Python3/src/antlr4/ParserInterpreter.py
```diff
@@ -58,7 +58,7 @@ def __init__(self, grammarFileName:str, tokenNames:list, ruleNames:list, atn:ATN
         self._interp = ParserATNSimulator(self, atn, self.decisionToDFA, self.sharedContextCache)
 
     # Begin parsing at startRuleIndex#
-    def parse(self, startRuleIndex:int):
+        def parse(self, startRuleIndex:int):
```
ericvergnaud, line 61: incorrect indent ?

ericvergnaud (changes requested):
The fact that we have an incorrect indent in this PR cruelly highlights the lack of testing for this ParserInterpreter. Is that something you could create ?

PR #4762, opened 2025-01-24
[C# target] Use Collections.EmptyList to prevent zero length allocations

Corniel:
I noticed when enabling some static code analysis rules (by setting the property `<EnableNETAnalyzers>true</EnableNETAnalyzers>` in the csproj file), that [CA1825: Avoid zero-length array allocations](https://learn.microsoft.com/dotnet/fundamentals/code-analysis/quality-rules/ca1825) was raised.

Because the C# solution still targets net4.5 (is that still needed? I would argue that net4.8 or netstandard2.0 should be sufficient, it is 2025 after all) the use of [`Array.Empty<T>`](https://learn.microsoft.com/dotnet/api/system.array.empty) can not be used. However, fortunately, `Collections.EmptyList<T>`  already provided a solution. I just applied them where applicable.

### DIFF
--- a/runtime/CSharp/src/Atn/ATN.cs
+++ b/runtime/CSharp/src/Atn/ATN.cs
@@ -77,10 +77,10 @@ namespace Antlr4.Runtime.Atn
         private readonly PredictionContextCache contextCache = new PredictionContextCache();
 
         [NotNull]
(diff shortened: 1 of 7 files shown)


Corniel:
Anyone?

ericvergnaud:
@parrt blessed

PR #4764, opened 2025-02-05
Add option for splitting generated Java parser into 3 files

ericvergnaud:
The current generator for Java embeds, within the `Parser` class, the serialized `ATN` and `ParserRuleContext` subclasses (as static child classes).

With large or complex grammars, this can lead to very large parser files that exceed 65k lines.
This is a problem because the JVM does not accept line numbers > 65535, thus making it impossible to debug these parsers.

This PR fixes the problem by adding a `-split-parser` option, that instructs the Tool to generate the serialized `ATN` and the `ParserRuleContext` subclasses in dedicated files, thus dramatically reducing the number of lines of the generated `Parser`. 

For the maven plugin, this option can be set using the `<arguments>-split-parser</arguments>` configuration entry.

There is no change if this option is not used.

### DIFF
--- a/tool/resources/org/antlr/v4/tool/templates/codegen/Java/Java.stg
+++ b/tool/resources/org/antlr/v4/tool/templates/codegen/Java/Java.stg
@@ -56,9 +56,11 @@ import org.antlr.v4.runtime.*;
 import org.antlr.v4.runtime.*;
 import org.antlr.v4.runtime.tree.*;
 import java.util.List;
+import java.util.ArrayList;
 <elseif(file.genDFA)>
 import org.antlr.v4.runtime.atn.*;
 import org.antlr.v4.runtime.dfa.DFA;
+import org.antlr.v4.runtime.misc.Utils;
 <else>
 import org.antlr.v4.runtime.atn.*;
 import org.antlr.v4.runtime.dfa.DFA;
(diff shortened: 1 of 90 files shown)


tool/resources/org/antlr/v4/tool/templates/codegen/Java/Java.stg
```diff
@@ -380,7 +416,7 @@ case <index>:
  * overriding implementation impossible to maintain.
  */
 RuleSempredFunction(r, actions) ::= <<
-private boolean <r.name>_sempred(<r.ctxType> _localctx, int predIndex) {
+private boolean <r.name>_sempred(<if(parser.file.genLean)><parser.name>Contexts.<endif><r.ctxType> _localctx, int predIndex) {
```
ericvergnaud, line 419: @parrt `RuleSempredFunction` is used in both `Lexer` and `Parser` contexts. What would be a clean way to avoid a ST warning here ?

parrt: Not sure what the warning is, but I'm sure you thought carefully about it

parrt:
Nothing seems crazy, but of course it's a widespread change. I'm just going to have to trust you. :)  should we update the documentation to include the new options for Java target?

ericvergnaud:
Thanks

PR #4738, opened 2024-11-28
Include `<chrono>` for high_resolution_clock

LilyWangLL:
I am a member of Microsoft vcpkg, due to there are new changes merged by microsoft/STL#5105, which revealed a conformance issue in `antlr4`. It must add include `<chrono>` to fix this error.

Compiler error with this STL change:
```
D:\b\antlr4\src\4.13.2-0987b822d0.clean\runtime\Cpp\runtime\src\atn\ProfilingATNSimulator.cpp(38): error C2653: 'high_resolution_clock': is not a class or namespace name
```

PR #4779, opened 2025-02-21
Fix "chrono" C++ library not being included on Windows

0xENDER:
On Linux, the build is successful because this library is usually included in the headers by default. But on Windows, when using Visual Studio (at least VS 2022 Community - the version I tested this on), the build fails because the "chrono" library is not included by default.

To avoid troubling new Windows adopters of this C++ runtime, it would be good to include this in the code!

(Adding this line does not break builds on Linux, and does not cause any warnings on Windows or Linux! - I tested it on both Windows and Linux)

### DIFF
--- a/runtime/Cpp/runtime/src/atn/ProfilingATNSimulator.cpp
+++ b/runtime/Cpp/runtime/src/atn/ProfilingATNSimulator.cpp
@@ -10,6 +10,7 @@
 #include "support/CPPUtils.h"
 
 #include "atn/ProfilingATNSimulator.h"
+#include <chrono>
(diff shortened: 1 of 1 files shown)


deanmsands3:
I was about to do this exact same PR. Thanks @0xENDER.

Hammie:
Duplicate of #4738 
Also ran into the same issue

0xENDER:
This change has been made through a different pull request

SimoneMalvezzi:
Please provide a new antlr stable release 4.13.2+ for this.

PR #4790, opened 2025-03-08
Use 'override' in virtual methods of derived classes.

hzeller:
... and clean up duplicates where a method is marked `virtual` _and_ `override` (`override` implies `virtual).

Use of `override` informs the developer reading the code and reduces mistakes (e.g. accidentally messing up a signature while overridding).


hzeller:
FYI @parrt

parrt:
hiya. gosh, that's a lot of changes, adding virtual everywhere etc!   Is it worth the risk of such a big change? I will defer to your experience.

hzeller:
there is no risk. Also, if you noticced, most of the change is just removing the superfluous `virtual` in front where there is already `override` at the end.

hzeller:
or to put it this way: the code is identical to the previous one for all intents and purposes, just more readable.
The changes are made fully automatic to avoid errors.

parrt:
Thanks :)

PR #4791, opened 2025-03-09
Fix a couple missing standard headers

hzeller:

Done half-automatically by running clang-tidy, then using a little script that inserts the header

```
~/bin/insert-header '<limits>' $(grep "no header providing.*misc-include-cleaner" Cpp_clang-tidy.out | awk -F: '/^runtime\/.*"std::numeric_limits"/ { print $1}' | sort -u)
```

Overall, this reduced the amount of `misc-include-cleaner` warnings in the runtime from 1923 downto 1192. More work to do, but already better :)


hzeller:
FYI @parrt

PR #4792, opened 2025-03-09
Add include `antlr4-common.h` to files using symbols from it.

hzeller:
All files using

  * `ANTLR4CPP_PUBLIC`
  * `ANTLR4CPP_USING_ABSEIL`
  * `ANTLR4CPP_HAVE_BUILTIN`
  * `INVALID_INDEX`
  * `Ref`
  * `ssize_t`

... and not yet directly including `antlr4-common.h`


hzeller:
FYI @parrt

PR #4793, opened 2025-03-09
Add top missing antlr4 headers inside impl.

hzeller:

Done semi-automatically by taking the output of `clang-tidy`'s `misc-include-cleaner` reports and inserting them where needed.

```
~/src/henner-tools/dev-tools/insert-header.cc 'tree/ParseTreeType.h' $(grep "no header providing.*misc-include-cleaner" Cpp_clang-tidy.out | awk -F: '/^runtime\/.*"antlr4::tree::ParseTreeType"/ { print $1}' | sort -u)
```

This is not done for all missing headers yet, some of the remaining reports point to cyclic dependencies that now work because everyone sees the forward declarations `support/Declarations.h`. My goal is minimize the reliance on these forward declarations and reach a good `IWYU` situation for a clear view of the internal dependencies.

Anyway, here a first round.

### DIFF
--- a/runtime/Cpp/runtime/src/ANTLRErrorListener.h
+++ b/runtime/Cpp/runtime/src/ANTLRErrorListener.h
@@ -8,6 +8,7 @@
 #include <string>
 #include <cstddef>
 #include "antlr4-common.h"
+#include "Token.h"
 #include "RecognitionException.h"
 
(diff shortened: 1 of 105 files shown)


hzeller:
FYI @parrt

PR #4800, opened 2025-03-19
Use the correct install prefix for cmake includes.

hzeller:
Not entirely sure if this is the correct fix for https://github.com/antlr/antlr4/pull/4661#issuecomment-2735315108 ... passing on to the original contributor of #4661


hzeller:
@njnobles does this look like the correct solution ?

jmairboeck:
I already did this in #4109 (including the docs).

hzeller:
Thanks, closing this in favor of yours. Also pinged maintainer there to take a look.

PR #4805, opened 2025-03-24
[JavaScript-runtime] Export ATN transitions with types

### LINKED ISSUE #4810: JavaScript-Runtime makes it hard to work with the ATN
I'm trying to build a suggestion system for my ANTLR4 grammer in a TypeScript project, based on [this guide by Daniel Bazaco](https://www.abstractsyntaxseed.com/blog/antlr4-autocomplete-library/writing-the-core). It makes use of the ATN generated by ANTLR and distignuishes between the ATN-transitions using `instanceof` checks. However since the npm-bundle does neither export the ATN-transition, nor its types, it's virtualy impossible to distinguish between the different transition-types like `AtomTransition`, `RuleTransition`, etc.

Exporting the ATN transitions and adding its types wouldn't impact the functionality of the remaining runtime at all, there for I'd really like to see them added soon. It would make working with the ATN a lot easier.

I have already setup a PR #4805 which addresses this issue,

MatthiasHarzer:
The current JS runtime does not export ATN transitions, which makes it hard to walk the ATN from outside.

This PR adds types for the ATN transitions (`src/antlr4/transition`) as well as exporting them so they can be imported from outside.

closes #4810


ericvergnaud:
@MatthiasHarzer thanks for this, but this is internal stuff. Exporting it would create an obligation to maintain it. What problem are you trying to solve ?

MatthiasHarzer:
Hi there @ericvergnaud,
thanks for your response. I'm sorry to hear that, but let me further explain the issue:

As I mentioned in my issue #4810, I'm trying to build a suggestion system to recommend tokens, that could follow a partial input. 

I really like the idea of leveraging the existing ANTLR ATN for this, as ANTLR is already a direct dependency, and I want to keep external dependencies to a minimum. Using the ATN also ensures that any changes made to the grammar (`.g4` file) are automatically reflected in the suggestion system, avoiding the need for additional manual adjustments.
While the ATN is already exported and partially typed in the current version, it is not very usable in its current state because its transitions cannot be distinguished without access to the transition classes.

Projects like [antlr4-c3](https://github.com/mike-lischke/antlr4-c3) use the ATN to predict following tokens and rules from an ANTLR-grammar and a partial input. However, I hesitate to use such tool as it depends on an alternative ANTLR-generator [antlr4ng](https://github.com/mike-lischke/antlr4ng). My goal is to build on the official ANTLR target and runtime as they appear to be supported the best. Additionaly, JavaScript isn't the only target in my custom language and I strongly prefer using the official ANTLR generator for all targets.


Exporting and typing these classes would not impact the rest of the runtime, but it would enable features like the one I’m trying to build. In my opinion, the benefits of making them available outweigh the maintenance overhead.

I hope this clarifies my reasoning, and I welcome any further questions or feedback.

kaby76:
Another application of the ATN data structures is to enumerate all the ambiguous parse trees at decision states. The [code I wrote to do this](https://github.com/antlr/grammars-v4/blob/master/_scripts/templates/CSharp/MyParserInterpreter.cs), in C# and part of a generated driver app for testing grammars in Trash, is used instead of the [erroneous one currently in the CSharp runtime](https://github.com/antlr/antlr4/blob/dev/runtime/CSharp/src/ParserInterpreter.cs). I could not do this without access to the ATN data structures. To do this tree enumeration, you also need the DecisionInfo data structure, aka "profiling", to know which decision states are ambiguous. Unfortunately, that's not implemented in the JS/TS runtime. If profiling and ambiguous parse tree enumeration are not part of the runtime (and they should not), then give us the tools so we can implement them. Once enumerated, Trash has about a half-dozen different ways to display the parse trees, which are much faster and more intuitive than generating .dot files and trying to find the decision state by eyeball. Ambiguity in grammars-v4 is [pervasive](https://github.com/antlr/grammars-v4/pull/4276): out of 363 grammars, 215 are ambiguous.

ericvergnaud:
@MatthiasHarzer Some time ago I started looking at implementing a code-completion core. The code is here https://github.com/ericvergnaud/antlr4/tree/suggestion-helper. The intent is precisely to provide a built-in tool for suggestions without exhibiting all internals of ANTLR (which create stickiness). It's still early days but maybe it's worth glancing at it ?

MatthiasHarzer:
@ericvergnaud Thank you very much for pointing me to this prototype. It already looks great and works very well based on what I’ve tested - it would fit my use case perfectly.
However, since there hasn’t been much activity on the associated PR (#4557), I went ahead and tried implementing it myself on top of the current JS-runtime, inspired by your work. It works like a charm - thank you again! 
Although I'm not particularly familiar with the inner working of ANTRL and the ATN, I still managed to understand the approach you persued.
I hope it’s okay for me to use parts of your code from your PR for my own purposes. If not, please let me know.

PR #4820, opened 2025-04-21
[Go] Use strings.Builder instead of string concat

daLingZhong:
This commit optimizes the performance of the GetTextFromInterval method when a single SQL text is too large


ericvergnaud:
@daLingZhong can you touch this PR in order to re-run the CI ?

daLingZhong:
> @daLingZhong can you touch this PR in order to re-run the CI ?

I don't know how to trigger it manually, and I don't think I have the permission. Can you do it?

PR #4832, opened 2025-05-09
fix: Cpp $rule.text fail if $rule is nullptr

Gashmob:
With a grammar like this:

```antlr
rule: (VAL | VAR {
        is_constant = false;
    }) name=IDENTIFIER (COLON type {
        type_name = $type.text; <-- we look this line
    })? (EQ value=expression {
        value = $value.tree;
    })?
```

When generating for Cpp target we get this specific result:

```cpp
type_name = (antlrcpp::downCast<Variable_declarationContext *>(_localctx)->typeContext != nullptr ? _input->getText(antlrcpp::downCast<Variable_declarationContext *>(_localctx)->typeContext->start, antlrcpp::downCast<Variable_declarationContext *>(_localctx)->typeContext->stop) : nullptr);
```

If input doesn't match the `type` branch then we fallback on nullptr. As of cpp23 standard this line doesn't compile as it waits for `std::string`.

AS `Token::getText` returns `std::string`, then fallback value should be empty string


ericvergnaud:
@Gashmob can you touch this PR in order to re-run the CI ?

Gashmob:
Up, this issue blocks one of my project, did you know when it is planned for review and merge ?

PR #4842, opened 2025-06-03
perf(runtime): Optimize thread safety and performance for DFA state management 

dingxin-tech:
Hi, team

As mentioned in issue #2454, I also found a performance bottleneck in ANTLR caused by the lock on the `DFA.states` object. Due to lock contention, the performance degrades significantly under high concurrency.

In my test case — parsing a SQL file with 400,000 characters — a single parse took about 200ms. However, when I increased the concurrency to 32 threads, the per-request latency jumped to 5–10 seconds. While not all SQL statements are this long, and it's expected that longer queries take more time, the current implementation introduces a serious issue: a single long SQL parsing task can block many short ones due to the global lock, causing the latency of short SQL statements to increase dramatically beyond expectations.

This problem can be resolved with a very simple change: replacing the `synchronized` HashMap with a `ConcurrentHashMap`.

### DIFF
--- a/runtime/Java/src/org/antlr/v4/runtime/atn/LexerATNSimulator.java
+++ b/runtime/Java/src/org/antlr/v4/runtime/atn/LexerATNSimulator.java
@@ -687,18 +687,17 @@ public class LexerATNSimulator extends ATNSimulator {
 		}
 
 		DFA dfa = decisionToDFA[mode];
-		synchronized (dfa.states) {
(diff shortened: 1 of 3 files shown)


jimidle:
It is very likely that your grammar is the root of the problem. What grammar are you using?

ericvergnaud:
@dingxin-tech can you touch this PR in order to re-run the CI ?

PR #4856, opened 2025-06-25
[JavaScript-runtime] Add getInvokingContext to Parser.d.ts

dudemaga:
When using non-local attribute references in the grammar (ex: `{$ruleName::MyAttribute}?`), the target [generates code](https://github.com/dudemaga/antlr4/blob/694898420da04b23de77032ba10befd737e5e21b/tool/resources/org/antlr/v4/tool/templates/codegen/JavaScript/JavaScript.stg#L577) that uses the `getInvokingContext` function in the parser. However, currently the function is not in the Parser type definition. This causes issues when using the runtime in a Typescript project.

This PR just adds the function to the Parser type definition as it is used in the code generation.


ericvergnaud:
@dudemaga can you touch this PR in order to re-run the CI ?

ericvergnaud:
@parrt blessed

Issue #4877, opened 2025-09-03
Critical vulnerability in dependency com.ibm.icu:icu4j:72.1


sanjay-kumar-sk:
The dependency(com.ibm.icu:icu4j) has a critical vulnerability.

Ref: https://www.sourceclear.com/vulnerability-database/security/integer-overflow/java/sid-7705

https://github.com/antlr/antlr4/blob/bd8f9930d053ec6a83e7d751e8c6f69138957665/tool/pom.xml#L47C1-L48C1

seb314:
(Edit: the content of this comment has been moved to a separate issue: https://github.com/antlr/antlr4/issues/4900)

PR #4878, opened 2025-09-03
Update icu4j version from 72.1 to 77.1

sanjay-kumar-sk:
For: https://github.com/antlr/antlr4/issues/4877


eoraita:
Any Idea when this will be committed and a new build of Antlr4 made available that resolves the Vulnerability

Issue #4843, opened 2025-06-03
[BUG] org.antlr.v4.runtime.atn.EpsilonTransition cannot be cast to org.antlr.v4.runtime.atn.RuleTransition


Dr-Bluemond:
I'm implementing a 'predicting the next token' function. I got this exception when I call `parser.getExpectedTokens()`

The grammar file is listed below, which I think is LL(1)
```g4
grammar Expr;

expr   : term ((PLUS | MINUS) term)* ;
term   : factor ((MUL | DIV) factor)* ;
factor : INT | LPAREN expr RPAREN ;

PLUS   : '+' ;
MINUS  : '-' ;
MUL    : '*' ;
DIV    : '/' ;
LPAREN : '(' ;
RPAREN : ')' ;
INT    : [0-9]+ ;

WS     : [ \t\r\n]+ -> skip ;
```

And I use this program to predict the next token. I create a special 'DirectTokenStream' class which provides existing token to the parser and calls 'getExpectedTokens' when all existing token are consumed.

```java
import expr.ExprParser;
import org.antlr.v4.runtime.*;
import org.antlr.v4.runtime.Vocabulary;
import org.antlr.v4.runtime.misc.Interval;

import java.util.Collections;
import java.util.List;
import java.util.ArrayList;
import java.util.Scanner;

public class Main {
    static ExprParser parser = new ExprParser(null);

    public static void main(String[] args) {
        Vocabulary vocabulary = parser.getVocabulary();
        List<Token> prefix = new ArrayList<>();
        Scanner scanner = new Scanner(System.in);
        while (true) {
            List<Integer> expectedTokens = getExpectedTokens(prefix);
            System.out.println("Expected tokens:");
            expectedTokens.forEach(token -> {
                System.out.println(token + ": " + vocabulary.getDisplayName(token));
            });

            System.out.print("Choose next token type: ");
            int chosenTokenType = scanner.nextInt();

            if (chosenTokenType == Token.EOF) {
                System.out.println("Exiting...");
                break;
            }

            Token nextToken = new CommonToken(chosenTokenType);
            prefix.add(nextToken);
        }

        scanner.close();
    }

    private static List<Integer> getExpectedTokens(List<Token> prefix) {
        DirectTokenStream tokenStream = new DirectTokenStream(prefix);
        parser.setTokenStream(tokenStream);
        tokenStream.setParser(parser);
        try {
            parser.expr();
[...]

PR #4888, opened 2025-09-17
fix(java): ATN.getExpectedTokens expects RuleTransition but gets EpsilonTransition (issue #4843)

Dr-Bluemond:
Resolved the issue raised in #4843. According to the convention, invokingState.transition(0) should be a RuleTransition, but I found that sometimes it can also be an EpsilonTransition. So I tried following this EpsilonTransition to find the actual RuleTransition.

The method to reproduce this issue can be found in the issue #4843


kaby76:
This problem exists in at least one other target (see the [CSharp code](https://github.com/antlr/antlr4/blob/bd8f9930d053ec6a83e7d751e8c6f69138957665/runtime/CSharp/src/Atn/ATN.cs#L249)). Shouldn't the fix be applied across all targets?

PR #4903, opened 2025-11-19
Interval: make public fields `final`

stevenschlansker:
The class docs already state that Interval is immutable, but this is not enforced.
Followup to #4901

I tried to fix up existing sites where this supposedly immutable class is mutated, to instead replace with a new value.

### DIFF
--- a/runtime/Java/src/org/antlr/v4/runtime/misc/IntervalSet.java
+++ b/runtime/Java/src/org/antlr/v4/runtime/misc/IntervalSet.java
@@ -664,18 +664,21 @@ public class IntervalSet implements IntSet {
             }
             // if on left edge x..b, adjust left
             if ( el==a ) {
-                I.a++;
+                I = Interval.of(a+1, b);
+                intervals.set(i, I);
                 break;
             }
             // if on right edge a..x, adjust right
             if ( el==b ) {
-                I.b--;
+                I = Interval.of(a, b-1);
+                intervals.set(i, I);
                 break;
             }
             // if in middle a..x..b, split interval
             if ( el>a && el<b ) { // found in this interval
                 int oldb = I.b;
-                I.b = el-1;      // [a..x-1]
+                I = Interval.of(a, el-1); // [a..x-1]
+                intervals.set(i, I);
                 add(el+1, oldb); // add [x+1..b]
             }
         }
--- a/runtime/Java/src/org/antlr/v4/runtime/misc/Interval.java
+++ b/runtime/Java/src/org/antlr/v4/runtime/misc/Interval.java
@@ -9,8 +9,8 @@ package org.antlr.v4.runtime.misc;
 public class Interval {
 	public static final Interval INVALID = new Interval(-1,-2);
 
-	public int a;
-	public int b;
+	public final int a;
+	public final int b;
 
 	public Interval(int a, int b) { this.a=a; this.b=b; }
 
--- a/tool/src/org/antlr/v4/parse/ANTLRLexer.g
+++ b/tool/src/org/antlr/v4/parse/ANTLRLexer.g
@@ -699,7 +699,7 @@ UNICODE_ESC
 				Interval badRange = Interval.of(getCharIndex()-2-hCount, getCharIndex());
 				String lastChar = input.substring(badRange.b, badRange.b);
 				if ( lastChar.codePointAt(0)=='\'' ) {
-					badRange.b--;
+					badRange = Interval.of(badRange.a, badRange.b-1);
 				}
 				String bad = input.substring(badRange.a, badRange.b);
 				Token t = new CommonToken(input, state.type, state.channel, badRange.a, badRange.b);


runtime/Java/src/org/antlr/v4/runtime/misc/IntervalSet.java
```diff
@@ -664,18 +664,21 @@ public void remove(int el) {
...
             // if on right edge a..x, adjust right
             if ( el==b ) {
-                I.b--;
+                I = Interval.of(a, b-1);
+                intervals.set(i, I);
                 break;
             }
             // if in middle a..x..b, split interval
             if ( el>a && el<b ) { // found in this interval
                 int oldb = I.b;
```
stevenschlansker, line 679: I think this variable `oldb` is redundant since we already saved `I.b` above at line 656. Let me know if you'd like me to remove this. (I didn't, in the spirit of keeping the change minimal)

ericvergnaud: Only test coverage would tell us that. Let's not touch this.

ericvergnaud (approved):

ericvergnaud:
@parrt blessed.
I'm very supportive of this change. The class comment is as follows:
`/** An immutable inclusive interval a..b */`
Note the _immutable_ in there, which expresses the original intent. 
The fact that we violated it internally might have been justified by performance, memory, whatever... I suspect mutations of intervals in the cache is the root cause of issue #4901, more than the cache itself. There doesn't seem to be any code checking whether the interval being mutated is in the interval cache, and should therefore absolutely not be mutated. This change ensures that the behavior complies with the intent.  
It probably also helps HotSpot do a better job since it can cache `a` and `b`.

parrt:
I agree there is an inconsistency there, but in the back of my mind, I remember there being a reason. Perhaps @sharwell who has a young agile mind still can remember.  In a nutshell, that is an insanely complicated algorithm and finally tuned implementation. We should be very careful about changing it unless there's a really good reason. :D

ericvergnaud:
The good reason is that changing fields of intervals deemed immutable has proven to break in a multithreaded scenario.
This PR affects in no way the algorithm. Let's see what @sharwell thinks.
In the meantime maybe @kaby76 can run performance tests to check whether it affects performance?

KvanTTT:
> In the meantime maybe @kaby76 can run performance tests to check whether it affects performance?

In my opinion correctness and consistency matter more than performance.

wendigo:
What's the decision here? Is there going to be a release once this change lands?

ericvergnaud:
> What's the decision here? Is there going to be a release once this change lands?

Hey we're all very busy with our daily jobs, you might need to give this a bit more time :-)

ericvergnaud:
@kaby76 Hey Ken, as a mean to consolidate our confidence that this change is both useful and harmless, would you be able to test it against a significant set of grammars?

kaby76:
Ok, will do. Let you know tomorrow or sooner if there's something unexpected. K

parrt:
Thanks folks. If Ken can show that it works across all of his tests then let’s go for it

sharwell:
I'd want to see allocation benchmarks on big scenarios before moving forward here. Unfortunately, I'm not set up to run these anymore. Do we have any concrete examples of bugs in the wild caused by this?

martint:
@sharwell, yes, this was the issue that led to it: https://github.com/jdbi/jdbi/issues/2898

ericvergnaud:
> I'd want to see allocation benchmarks on big scenarios before moving forward here. Unfortunately, I'm not set up to run these anymore. Do we have any concrete examples of bugs in the wild caused by this?

@sharwell If that helps, your C# implementation uses readonly fields.

kaby76:
Sorry for taking a bit of time, but I wanted to be sure. The PR causes no regressions in any of the 370 grammars in grammars-v4. For most grammars, there was no statistical change in speed with the PR. However, there was for a few grammars, all slightly faster. For plsql, there was a speed up of 2% (sample size=20, two-sided t-test, t-statistic 8.97, df=38, p=6.3e-11).  I made sure to make sure no bias in order of testing the PR versus control.

parrt:
@sharwell that sounds like pretty good evidence. What do you think? I'm always nervous about these kind of changes, but it seems to fix a bug and not affect performance.

parrt:
Thanks to @kaby76 for runnin all those tests!

sharwell:
I reviewed again and noticed that the only mutations of these fields occurred in `IntervalSet.remove`. This method does not appear to be used on hot paths.

> @sharwell If that helps, your C# implementation uses readonly fields.

It doesn't hurt, but I have to keep in mind that the ANTLR 4 C# optimized target was constructed a bit differently from my old work on the ANTLR 3 C# target. I missed a few details during the initial conversion (e.g. Interval and some tokens could have been structs) and my dedication to avoiding breaking changes caused some of the items to stay indefinitely. I was more focused on algorithm improvements for ANTLR 4 and somewhat less on micro-optimizations.

parrt:
Ok, given @sharwell's comment:

> I reviewed again and noticed that the only mutations of these fields occurred in IntervalSet.remove

I think it is safe to proceed!  Closing eyes, crossing fingers, and merging!

wendigo:
@parrt thank you!

stevenschlansker:
Thanks everyone for reviewing and merging this. Is there any chance of a new release soon? Jdbi would like to ship a fixed ANTLr for our users.

PR #4907, opened 2025-12-02
Change printf to use System.out instead of System.err

rulle-io:
It seems to be an overkiil to log such problems as errors.


jimidle:
Why? That is fatal error.

PR #4913, opened 2025-12-31
Add IVocabulary.getMaxTokenType

mlptownsend:
Adding getMaxTokenType to the interface for IVocabulary in the C# source for the runtime.

It's available in the Java but not in C#.
https://www.antlr.org/api/Java/org/antlr/v4/runtime/Vocabulary.html


ericvergnaud:
@mlptownsend Thanks for this!
@parrt blessed

Happy new year !!!

Issue #4924, opened 2026-03-07
C++ runtime installs headers in the wrong location


pstieber:
Currently, the C++ runtime installs header files in the work location.  For example if the install prefix is `/usr/local` the headers will be installed under `/usr/local/antlr4-runtime` instead of `/usr/local/includeantlr4-runtime`.

PR #4925, opened 2026-03-07
build: fix location of installed c++ headers

pstieber:
Prior to this fix, include files would be installed in the wrong location.  For example, if the install prefix was /usr/local, the headers would be installed in /usr/local/antlr4-runtime instead of /usr/local/include/antlr4-runtime.

This resolves issue #4924

### DIFF
--- a/runtime/Cpp/runtime/CMakeLists.txt
+++ b/runtime/Cpp/runtime/CMakeLists.txt
@@ -7,7 +7,7 @@ if (NOT ANTLR_BUILD_SHARED AND NOT ANTLR_BUILD_STATIC)
   message(FATAL_ERROR "Options ANTLR_BUILD_SHARED and ANTLR_BUILD_STATIC can't both be OFF")
 endif()
 
-set(libantlrcpp_INCLUDE_INSTALL_DIR "${CMAKE_INSTALL_INCLUDEDIR}/antlr4-runtime")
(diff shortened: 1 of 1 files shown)


runtime/Cpp/runtime/CMakeLists.txt
```diff
@@ -7,7 +7,7 @@ if (NOT ANTLR_BUILD_SHARED AND NOT ANTLR_BUILD_STATIC)
   message(FATAL_ERROR "Options ANTLR_BUILD_SHARED and ANTLR_BUILD_STATIC can't both be OFF")
 endif()
 
-set(libantlrcpp_INCLUDE_INSTALL_DIR "${CMAKE_INSTALL_INCLUDEDIR}/antlr4-runtime")
+set(libantlrcpp_INCLUDE_INSTALL_DIR "include/antlr4-runtime")
```
jmairboeck, line 10: This fix is wrong. There are systems which don't use "include" as the include path, e.g. Haiku, which uses "develop/headers". I think the correct fix would be to move the `include(GNUInstallDirs)` in the parent CMakeLists.txt further up before this subdirectory is added. I missed this apparently in #4109. Sorry for that.

jmairboeck: See #4971 for the proper fix.

PR #4933, opened 2026-04-10
refactor(JS: explicitly using TerminalNode): move the implementation to TerminalNode/ErrorNode and actually use them.

jerry4718:
- add an abstract `accept()` method to the `ParseTree` base class.
- remove `ErrorNodeImpl` and `TerminalNodeImpl`, use direct class implementations.
- update `Parser`, `ParserRuleContext`, and `ParseTreeWalker` to use instanceof checks.


jerry4718:
The type export and declaration are based on `TerminalNode`, yet the actual implementation used is `TerminalNodeImpl`; furthermore, the type definition for `TerminalNodeImpl` is not exported, making it impossible to perform `instanceof` checks during usage.

PR #4943, opened 2026-06-09
carry trailing high surrogate across CodePointBuffer appends

rootvector2:
`appendArrayInt` emitted a chunk-trailing high surrogate immediately instead of holding it in `prevHighSurrogate`, so a surrogate pair split across two `append()` calls (e.g. `CharStreams.fromReader` filling its read buffer mid-emoji) decoded to a spurious lone surrogate before the real code point; found tracing the chunked decode path, the fix keeps it pending for the next `append()` and flushes a still-dangling one in `build()`.

### DIFF
--- a/runtime-testsuite/test/org/antlr/v4/runtime/TestCodePointCharStream.java
+++ b/runtime-testsuite/test/org/antlr/v4/runtime/TestCodePointCharStream.java
@@ -9,6 +9,8 @@ package org.antlr.v4.runtime;
 import org.antlr.v4.runtime.misc.Interval;
 import org.junit.jupiter.api.Test;
 
+import java.nio.CharBuffer;
+
 import static org.junit.jupiter.api.Assertions.*;
 
 public class TestCodePointCharStream {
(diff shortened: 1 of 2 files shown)


rootvector2:
the failing `typescript` check is unrelated to this change. the whole runtime suite fails to compile with `TS2591: Cannot find name 'process'` (missing `@types/node`); the `java` build and the new test are green. happy to rebase if that job gets sorted on `dev`.

GeorgEchterling:
We've encountered this bug in production through GraphQL Java. It returns an `Invalid syntax with ANTLR error 'token recognition error at: '"...?` error to the client if a chunk of the query ends between a high surrogate and its matching low surrogate, since the lone surrogate doesn't match GraphQL's `StringCharacter` grammar.

rootvector2:
same path: graphql-java builds its stream with `CharStreams.fromReader`, which passes each `read()` result to `Builder.append()` as its own chunk, so a chunk ending on a high surrogate gets emitted by `appendArrayInt` as a lone code point before the low surrogate shows up in the next chunk. `splitSurrogatePairAcrossAppendsProducesOneCodePoint` in this PR is that case in isolation; the branch is current with `dev`.

PR #4953, opened 2026-07-18
Cpp: make DFAState edge lookups lock-free to speed up lexing

analog-cbarber:
> **CI note:** this branch is stacked on #4963 (a 3-file test-harness fix for the repo-wide `typescript` job failures, see #4962) so the full matrix can run green here. The C++ change is the single `Cpp:` commit; the diff collapses to it once #4963 merges.

## What

Replaces `DFAState::edges` — a `FlatHashMap<size_t, DFAState*>` guarded by the
ATN-wide `_edgeMutex` — with a lazily-allocated array of
`std::atomic<DFAState*>`, giving a lock-free `getEdge` on the per-character hot
path. Writers still take the existing mutex; only the read path changes.

## Why

The lexer consults the DFA edge table once per input character, so the hottest
loop in the C++ runtime pays a shared-lock acquire plus a hash-map probe per
codepoint. The Java reference runtime has neither cost: `DFAState.edges` there
is a plain array indexed by symbol, read without locking. This change brings
the C++ runtime to parity with the Java design.

Edges are indexed by `t + 1` (as in the Java runtime) so EOF (`t == -1`) maps
to slot 0; the table is sized `maxTokenType + 2`.

## Performance

Measured on a 4.7 MB JSON parse (Apple M5 Max, best of 7): ~12% off the
single-thread wall time (551 ms → 486 ms). JSON is lexer-heavy, which flatters
this change; parser-heavy grammars see less.

## Safety / verification

- Memory-safety: an earlier revision of this patch dropped the `t + 1` offset
  and produced an out-of-bounds write when a parser decision cached an edge on
  EOF lookahead. That is fixed in this submission and the scenario is covered:
  the patch has been running under **AddressSanitizer in CI** (single-token /
  empty / truncated / malformed inputs across two grammars) in a downstream
  project, clean.
- Semantics: atomic release-store on write, acquire-load on read; the map →
  array change does not alter which edges are cached.
- The full downstream test suite (event-stream comparisons against
  pure-Python-runtime oracles) runs against this patch continuously.

## Context

This patch (with #4954) is shipped today in [antlrope](https://github.com/zuzukin/antlrope), a
C++-accelerated Python runtime that drives
`ParserInterpreter`/`LexerInterpreter` from serialized ATNs. antlrope
currently vendors a patched copy of this runtime; once this change is merged
and released, it can drop the patch and vendor the unmodified upstream runtime
again. Benchmarks and methodology:
https://zuzukin.github.io/antlrope/concepts/ (the "vendored runtime and its
patches" section) and https://github.com/zuzukin/srdl-bench.

---

**Authorship note:** this patch was developed with AI assistance (Claude Opus
4.7, reflected in the commit trailer); I reviewed the design and code, and the
verification above (ASan CI, downstream test suite, benchmarks) is what I'd
stake it on.

### DIFF
--- a/runtime-testsuite/resources/org/antlr/v4/test/runtime/helpers/tsconfig.json
+++ b/runtime-testsuite/resources/org/antlr/v4/test/runtime/helpers/tsconfig.json
@@ -4,8 +4,5 @@
     "moduleResolution": "node",
     "target": "ES6",
     "noImplicitAny": true,
-  },
-  "ts-node": {
-    "esm": true
   }
 }
--- a/runtime-testsuite/test/org/antlr/v4/test/runtime/typescript/TsxRunner.java
+++ b/runtime-testsuite/test/org/antlr/v4/test/runtime/typescript/TsxRunner.java
@@ -15,7 +15,7 @@ import java.nio.file.Files;
 import static org.antlr.v4.test.runtime.FileUtils.writeFile;
 import static org.antlr.v4.test.runtime.RuntimeTestUtils.isWindows;
 
-public class TsNodeRunner extends RuntimeRunner {
+public class TsxRunner extends RuntimeRunner {
 
 	/* TypeScript runtime is the same as JavaScript runtime */
 	private final static String NORMALIZED_JAVASCRIPT_RUNTIME_PATH = getRuntimePath("JavaScript").replace('\\', '/');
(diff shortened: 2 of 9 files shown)


analog-cbarber:
Note on the failing `typescript` checks: they are pre-existing on every PR that runs the full matrix (diagnosed in #4962 — unpinned ts-node vs typescript@latest) and unrelated to this C++-only change. A fix is proposed in #4963. The `cpp` jobs, which exercise this change, are green on all three platforms.

PR #4954, opened 2026-07-18
Cpp: move DFA write locks from the ATN onto the DFA

analog-cbarber:
> **Note:** stacked on #4953 (and, for green CI, on #4963 — the test-harness fix for the repo-wide `typescript` failures, see #4962). This change is the single `Cpp: move DFA write locks` commit; the diff collapses to it as the others merge.

## What

Moves the DFA state/edge **write** locks off the ATN
(`ATN::_stateMutex` / `ATN::_edgeMutex`, removed) and onto the DFA itself
(`dfa::DFA::stateMutex()` / `edgeMutex()`, heap-allocated so `DFA` stays
movable). The lexer/parser simulators lock the owning DFA instead of the ATN.
`ATN::_mutex` (the lazy `nextTokens` cache) is unchanged.

## Why

Each `ParserInterpreter` / `LexerInterpreter` owns its own `decisionToDFA`,
but the write locks lived on the **shared** ATN — so concurrent interpreters
over one (read-only) ATN serialized on a single lock for DFA writes they do
not even share. Measured effect: parsing a shared spec across 4 threads ran at
**~0.6× of serial** — concurrency was a net loss.

With per-DFA locks, independent DFAs use independent locks: the same 4-thread
workload scales to **~2.1×** (≈4× better wall-clock than before the patch).
A DFA genuinely shared across threads (a generated recognizer's static
`decisionToDFA`) still serializes its own writers — same locking discipline,
per-decision rather than per-ATN granularity.

## Safety / verification

- **ThreadSanitizer**: a TSan build of the runtime with an 8-thread concurrent
  parse harness over a shared ATN reports no data races.
- AddressSanitizer runs continuously in a downstream CI against this patch.
- Read paths are unchanged by this commit (edge reads are lock-free via #4953).

## Context

Shipped today (with #4953) in
[antlrope](https://github.com/zuzukin/antlrope); this is what makes its
shared-spec parallel parsing scale. Measurements and methodology:
https://zuzukin.github.io/antlrope/concepts/ ("How much do the patches
contribute?").

---

**Authorship note:** this patch was developed with AI assistance (Claude Opus
4.8, reflected in the commit trailer); I reviewed the design and code, and the
verification above (TSan harness, ASan CI, benchmarks) is what I'd stake it on.

### DIFF
--- a/runtime-testsuite/resources/org/antlr/v4/test/runtime/helpers/tsconfig.json
+++ b/runtime-testsuite/resources/org/antlr/v4/test/runtime/helpers/tsconfig.json
@@ -4,8 +4,5 @@
     "moduleResolution": "node",
     "target": "ES6",
     "noImplicitAny": true,
-  },
-  "ts-node": {
-    "esm": true
   }
 }
--- a/runtime-testsuite/test/org/antlr/v4/test/runtime/typescript/TsxRunner.java
+++ b/runtime-testsuite/test/org/antlr/v4/test/runtime/typescript/TsxRunner.java
@@ -15,7 +15,7 @@ import java.nio.file.Files;
 import static org.antlr.v4.test.runtime.FileUtils.writeFile;
 import static org.antlr.v4.test.runtime.RuntimeTestUtils.isWindows;
 
-public class TsNodeRunner extends RuntimeRunner {
+public class TsxRunner extends RuntimeRunner {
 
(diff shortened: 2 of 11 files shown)


analog-cbarber:
Note on the failing `typescript` checks: pre-existing CI breakage diagnosed in #4962, fix proposed in #4963 — unrelated to this C++-only change. The `cpp` jobs are green on all three platforms.

PR #4957, opened 2026-07-20
reject tokenVocab names containing a path separator

abdul7056:
The `tokenVocab` option value is used verbatim to build the `.tokens` path in `TokenVocabParser`:
- `getImportedVocabFile` resolves it against `libDirectory`, `outputDirectory` and the grammar directory, none of which constrain the result
- option values may be quoted string literals, so `options { tokenVocab='../../secret/creds'; }` resolves outside all three roots
- lines that fail the token-definition regex are reported via `TOKENS_FILE_SYNTAX_ERROR`, putting the contents of the opened file into tool output

Rejecting names that carry a path separator before anything is opened. `antlr4-maven-plugin` already reduces this same value with `stripPath`, so a bare grammar name is the expected shape.


abdul7056:
gentle ping, any thoughts?

PR #4961, opened 2026-07-24
Fix typo: MY_CXX_WARNING_FLGAS -> MY_CXX_WARNING_FLAGS in Cpp CMakeLists.txt

### LINKED ISSUE #4864: misspelled flag name MY_CXX_WARNING_FLGAS
https://github.com/antlr/antlr4/blob/dev/runtime/Cpp/CMakeLists.txt

I found the flag MY_CXX_WARNING_FLGAS, I think it is MY_CXX_WARNING_FLAGS

arimu1:
Fixes #4864.

## What

`runtime/Cpp/CMakeLists.txt` has two occurrences of a misspelled variable
reference, `MY_CXX_WARNING_FLGAS`, in the `CMAKE_CXX_FLAGS_RELEASE` lines
(one in the MSVC branch, one in the non-MSVC branch). The correctly spelled
variable is `MY_CXX_WARNING_FLAGS`.

This PR corrects both occurrences to `MY_CXX_WARNING_FLAGS`.

## Why this is purely cosmetic

Since `MY_CXX_WARNING_FLGAS` was never defined, it silently expanded to
nothing in the `set(CMAKE_CXX_FLAGS_RELEASE ...)` lines — it did not cause a
build error. It also does **not** mean Release builds compile without
warning flags: line 89 (`set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS}
${MY_CXX_WARNING_FLAGS}")`) already appends the warning flags to the base
`CMAKE_CXX_FLAGS`, and CMake concatenates `CMAKE_CXX_FLAGS` with the
per-config `CMAKE_CXX_FLAGS_<CONFIG>` variable when invoking the compiler.
So warning flags were already applied in Release builds through that path;
the typo only dropped a redundant second copy of the same flags being
appended a second time in the Release-specific line.

## Notes

- Fixing the typo consistent with the correctly-spelled usage elsewhere in
  the same file (Debug, MinSizeRel, and RelWithDebInfo lines already use
  `MY_CXX_WARNING_FLAGS` correctly).
- No behavioral test added — this is a build-file wording/typo fix with no
  observable behavior change.
- A prior PR for the same issue, #4865, was opened and self-closed by the
  reporter without merging (it appears to have been opened from a branch
  literally named `dev`, against this repo's contribution guidelines). The
  typo is still present on current `dev`, so this PR supersedes it.


arimu1:
C++ typo-only change (`MY_CXX_WARNING_FLGAS` → `MY_CXX_WARNING_FLAGS`). Rebased onto current `dev`.

Note: the red TypeScript build jobs look unrelated to this one-line CMake fix (cpp/java/js jobs green). Happy to dig if maintainers want, or they can ignore for this PR.

Issue #4962, opened 2026-07-25
TypeScript runtime tests fail in CI since ~April: ts-node incompatible with typescript@latest (TS 7)


analog-cbarber:
Every `build (*, typescript)` CI job currently fails, on every PR that triggers them. Each test dies at the Execute stage, before any ANTLR code runs, with:

```
TypeError: Cannot read properties of undefined (reading 'fileExists')
    at readConfig (.../ts-node/dist/configuration.js:91:33)
```

## Cause

The test harness installs its toolchain unpinned ([TsNodeRunner.java](https://github.com/antlr/antlr4/blob/dev/runtime-testsuite/test/org/antlr/v4/test/runtime/typescript/TsNodeRunner.java#L38)):

```java
Processor.run(new String[] {NPM_EXEC, "--silent", "install", "-g", "typescript", "ts-node", "webpack", "webpack-cli"}, null);
```

npm's `typescript@latest` is now the TypeScript 7 (Go-based) rewrite, which no longer exposes the `ts.sys` internal API that ts-node 10.9.x (unmaintained since 2022) reaches into — ts-node crashes on startup before compiling anything.

## Evidence it's the toolchain, not the tree

- `dev` tip `7d5770395` passed all three typescript jobs on 2026-02-16 (run [22078029856](https://github.com/antlr/antlr4/actions/runs/22078029856)); the identical tree fails them today.
- The failure appears on every full-matrix PR regardless of content: #4943, #4947, #4949, #4953, #4954, #4961. PRs that look green (e.g. #4957–#4960) simply didn't trigger the typescript jobs, so the breakage is easy to miss.
- Reproduced locally with the harness setup (ESM package.json + helper tsconfig + a test file importing `antlr4`): ts-node under `typescript@7.0.2` fails with exactly the CI error; the same file runs fine under an executor that doesn't use the TypeScript compiler API.

## Fix options

1. **Swap ts-node for [tsx](https://tsx.is)** (esbuild-based, no dependency on TypeScript's compiler API, actively maintained) — structurally immune to TypeScript's release cadence. PR to follow.
2. Pin the install (`typescript@5 ts-node@10.9.2`) — also goes green today, but freezes the tests to an aging TypeScript under an unmaintained runner.

---

*Diagnosis developed with AI assistance (Claude Fable 5); I reviewed and reproduced it.*

PR #4967, opened 2026-08-07
Fix TokenStreamRewriter bugs in the Python3 runtime

### LINKED ISSUE #4818: [python] TokenStreamRewriter _reduceToSingleOperationPerIndex error when reducing mixture of Replace and Insert operations
## Description
The [Python `TokenStreamRewriter` class](https://github.com/antlr/antlr4/blob/cb850120378d000e97da38afb852aa3c554b81f5/runtime/Python3/src/antlr4/TokenStreamRewriter.py) has a bug when there's a mixture of both Replace and Insert operations queued up, and some of the Insert operations overlap. The ["Walk inserts" section of `_reduceToSingleOperationPerIndex`](https://github.com/antlr/antlr4/blob/cb850120378d000e97da38afb852aa3c554b81f5/runtime/Python3/src/antlr4/TokenStreamRewriter.py#L169-L189) uses the wrong index when combining Inserts, which risks setting the wrong operation to `None`.

## To Reproduce
1. Define a list of `rewrites` which has a Replace, then two Inserts which overlap. Eg `[ReplaceOp(tokens, 2, 4, "foo"), InsertBeforeOp(tokens, 6, "bar"), InsertBeforeOp(tokens, 6, "baz")]`
2. Try call `_reduceToSingleOperationPerIndex` (either directly in a unit test, or by calling `getText()`
    - Expected behaviour: should combine the Insert operations and leave the ReplaceOp untouched, leaving: `[ReplaceOp(tokens, 2, 4, "foo"), None, InsertBeforeOp(tokens, 6, "bazbar")]`
    - Actual behaviour: Incorrectly sets the `ReplaceOp` to `None` because the Index reference is wrong, leaving: `[None, InsertBeforeOp(tokens, 6, "bar"), InsertBeforeOp(tokens, 6, "baz")]`
3. It then [fails with a `ValueError`](https://github.com/antlr/antlr4/blob/cb850120378d000e97da38afb852aa3c554b81f5/runtime/Python3/src/antlr4/TokenStreamRewriter.py#L194) because two different operations are still defined for the same index, 6.

## Suggested Fix
The "# Walk Replaces" section of the same method is already solving this issue by referencing `.instructionIndex`. I suspect that Inserts don't do the same is an oversight. So I'd suggest implementing the same approach for Inserts checks as well:
1. Change both [line 177](https://github.com/antlr/antlr4/blob/cb850120378d000e97da38afb852aa3c554b81f5/runtime/Python3/src/antlr4/TokenStreamRewriter.py#L177) and [line 180](https://github.com/antlr/antlr4/blob/cb850120378d000e97da38afb852aa3c554b81f5/runtime/Python3/src/antlr4/TokenStreamRewriter.py#L180):
    - from: `rewrites[prev_index] = None`
    - to: `rewrites[prevIop.instructionIndex] = None`
2. After that change, `prev_index` is no longer needed anywhere, so we can remove the `enumerate` call and change [line 174](https://github.com/antlr/antlr4/blob/cb850120378d000e97da38afb852aa3c554b81f5/runtime/Python3/src/antlr4/TokenStreamRewriter.py#L174):
    - from: `for prev_index, prevIop in enumerate(prevInserts):`
    - to: `for prevIop in prevInserts:` (this is same code as the Walk Replace section)

SEPURI-SAI-KRISHNA:
Fixes #4818.

The Python3 `TokenStreamRewriter` diverges from every other runtime in
`_reduceToSingleOperationPerIndex` and `ReplaceOp`. I checked all seven
runtimes that ship a rewriter (Java, C++, C#, JavaScript, Go, Swift,
Python3 — Dart has none): the other six are correct on each point below,
Python3 is the only outlier. Expected values in this PR were taken by
running the Java runtime, not assumed.

### 1. Wrong index when combining inserts (#4818)

`rewrites[prev_index] = None` used the index into the filtered
`prevInserts` list instead of `prevIop.instructionIndex`, so combining two
inserts nulls out an unrelated operation.

```python
rewriter.replaceIndex(2, 'R')
rewriter.insertBeforeIndex(0, 'x')
rewriter.insertBeforeIndex(0, 'y')
rewriter.getDefaultText()
```

Java returns `'yxabR'`. Python3 raised `ValueError: should be only one op per index`,
because the `ReplaceOp` was destroyed and both inserts survived at index 0.

### 2. Overlapping deletes

Two problems in the same branch:

- the merged right boundary used `min()` where Java, C++, C#, JavaScript,
  Go and Swift all use `max()`;
- the branch was unreachable regardless, because `delete()` stores `""` as
  the text while the guard tested `text is None`.

So overlapping deletes took the error path instead of merging:

```python
rewriter.delete('default', 0, 1)
rewriter.delete('default', 1, 3)
```

Java returns `''`. Python3 raised.

### 3. `ReplaceOp.__str__` returns `None` for deletes

The method has no `return` when `text` is empty, so `str(op)` raised
`TypeError: __str__ returned non-string (type NoneType)`. Every other
runtime emits `<DeleteOp@...>`.

This compounded #2: the `ValueError` message for an overlapping replace
could not be constructed, so the real diagnostic was replaced by a
confusing `TypeError` from deep inside string formatting.

I also dropped a stray `print('New rop ...')` left on the merge path — it
wrote to stdout from library code. (The equivalent prints in the Java and
C# runtimes are left for a separate PR, to keep this one to one runtime.)

### Why this went unnoticed

`runtime/Python3/tests/` is not run by CI, and the rewriter tests could not
run at all: `mocks/TestLexer.py` was generated by ANTLR 4.5.3 (serialized
ATN v3, runtime requires v4), and the import was commented out in `run.py`.

I regenerated the mocks with the current tool, checked in the `.g4` sources
so they can be regenerated in future, and re-enabled the suite.

### Testing

- `python3 runtime/Python3/tests/run.py` → **59 tests, all pass**
- `mvn -Dtest='python3.**' test` in `runtime-testsuite` → **362 tests, 0 failures**

The 39 pre-existing rewriter tests pass both before and after this change —
they never exercised these paths, which is how the bugs survived. The 4 new
regression tests fail 4/4 against the unfixed runtime.

### DIFF
--- a/runtime/Python3/src/antlr4/TokenStreamRewriter.py
+++ b/runtime/Python3/src/antlr4/TokenStreamRewriter.py
@@ -158,11 +158,12 @@ class TokenStreamRewriter(object):
                     rewrites[prevRop.instructionIndex] = None
                     continue
                 isDisjoint = any((prevRop.last_index<rop.index, prevRop.index>rop.last_index))
-                if all((prevRop.text is None, rop.text is None, not isDisjoint)):
+                # Delete special case of replace (text is empty):
+                # D.i-j.u D.x-y.v | boundaries overlap    combine to max(min)..max(right)
+                if all((not prevRop.text, not rop.text, not isDisjoint)):
                     rewrites[prevRop.instructionIndex] = None
                     rop.index = min(prevRop.index, rop.index)
-                    rop.last_index = min(prevRop.last_index, rop.last_index)
-                    print('New rop {}'.format(rop))
+                    rop.last_index = max(prevRop.last_index, rop.last_index)
                 elif (not(isDisjoint)):
                     raise ValueError("replace op boundaries of {} overlap with previous {}".format(rop, prevRop))
 
@@ -171,13 +172,13 @@ class TokenStreamRewriter(object):
             if any((iop is None, not isinstance(iop, TokenStreamRewriter.InsertBeforeOp))):
                 continue
             prevInserts = [op for op in rewrites[:i] if isinstance(op, TokenStreamRewriter.InsertBeforeOp)]
-            for prev_index, prevIop in enumerate(prevInserts):
+            for prevIop in prevInserts:
                 if prevIop.index == iop.index and type(prevIop) is TokenStreamRewriter.InsertBeforeOp:
                     iop.text += prevIop.text
-                    rewrites[prev_index] = None
+                    rewrites[prevIop.instructionIndex] = None
                 elif prevIop.index == iop.index and type(prevIop) is TokenStreamRewriter.InsertAfterOp:
                     iop.text = prevIop.text + iop.text
-                    rewrites[prev_index] = None
+                    rewrites[prevIop.instructionIndex] = None
             # look for replaces where iop.index is in range; error
             prevReplaces = [op for op in rewrites[:i] if isinstance(op, TokenStreamRewriter.ReplaceOp)]
             for rop in prevReplaces:
@@ -253,3 +254,4 @@ class TokenStreamRewriter(object):
             if self.text:
                 return '<ReplaceOp@{}..{}:"{}">'.format(self.tokens.get(self.index), self.tokens.get(self.last_index),
                                                         self.text)
+            return '<DeleteOp@{}..{}>'.format(self.tokens.get(self.index), self.tokens.get(self.last_index))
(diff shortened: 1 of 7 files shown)


SEPURI-SAI-KRISHNA:
The two failing typescript jobs are #4962 (ts-node vs. typescript@latest), unrelated to this PR — it only touches runtime/Python3/. All three python3 jobs pass.

PR #4976, opened 2026-09-12
[C++] Add -header-dir option to write generated header files to a separate directory

### LINKED ISSUE #2439: C++ generated code doesn't seperate source and header files
The generated code for C++ puts source and header files in the same directory. It would be desirable to have seperate src and include folder. 
This is required to share only the public interface (in the headers) with other targets that include the compiled generated source as a library. Those should only be able to link against symbols defined in the .h files but not in the .cpp files. 
To achieve this currently the only way is to move the headers into a separate directory manually as the compiler is given a directory of headers instead of single files. 
I am using Release Tag 4.7.1.

### LINKED ISSUE #4975: [C++] Option to write generated header files to a separate directory
Hi all,

When generating C++ code, ANTLR writes the headers and the `.cpp` files into the same directory, whatever `-o` points at. I'd like a way to send the headers somewhere else.

This came up for me while packaging a grammar as a C++ library. The usual layout is a public `include/` directory for consumers and a private `src/` for the implementation. Today the only way to get there is to put the whole generated directory on the include path, `.cpp` files and all, or to add a build step that moves the headers after ANTLR runs. Neither is terrible, but both feel like working around the tool.

The CMake helpers in this repo run into the same thing. `ANTLR_TARGET` in `FindANTLR.cmake` takes a single `OUTPUT_DIRECTORY`, and its README suggests `include_directories` on that same folder. `antlr4_generate` in `antlr4-generator.cmake.in` also uses its generated source directory as the include directory.

#2439 asked for this back in 2018 but never got a proposal or any discussion, so I thought it was worth picking up again with a more specific suggestion.

### Reproducing

Tried with 4.13.2 and with current `dev`:

```antlr
grammar T;
s : ID ;
ID : [a-z]+ ;
WS : [ \t\r\n]+ -> skip ;
```

```bash
antlr4 -Dlanguage=Cpp -o gen/src -visitor T.g4
ls gen/src
```

Everything ends up in `gen/src`: `TLexer.h`, `TLexer.cpp`, `TParser.h`, `TParser.cpp`, the listener and visitor pairs, plus the `.tokens` and `.interp` files. What I'm after is being able to put the `.h` files in, say, `gen/include` instead.

### What I have in mind

A new option, `-header-dir <dir>`:

```bash
antlr4 -Dlanguage=Cpp -o gen/src -header-dir gen/include -visitor T.g4
```

Only the headers would move. The `.cpp`, `.tokens`, `.interp` and `.dot` files would stay under `-o`. The new directory would be resolved the same way `-o` is, so grammar subdirectories and `-Xexact-output-dir` behave as people already expect. If you don't pass it, nothing changes, and targets that don't generate headers would simply ignore it.

The generated `#include` lines would stay as they are. You'd add the header directory to your include path, as with any other C++ library.

To make it usable from build tools, `-depend` would report the new header paths, `ANTLR_TARGET` in `FindANTLR.cmake` would get an optional `HEADER_OUTPUT_DIRECTORY` argument, `antlr4_generate` an optional header directory argument, and the Maven plugin an optional `headerOutputDirectory` parameter. Leaving those out gives exactly the command line you get today.

### Why not a `-source-dir` too, or changing `-o`?

I went back and forth on this. A matching `-source-dir` sounds tidy, but `-o` already controls every file for every target, and the Maven plugin, both CMake helpers and plenty of existing build scripts rely on it. A `-source-dir` would do the same job as `-o` for every target except C++, and we'd need rules for what wins when both are given. Keeping `-o` as the base directory and adding one optional option for headers avoids all of that, and there's nothing to deprecate. It also doesn't rule out adding more options later if someone needs them.

On the name, I stuck with a single dash like every other ANTLR option, and kept it short enough to line up with the rest of the `-help` output.

### Where it would go in the code

`CodeGenerator.writeRecognizer` and its siblings already know whether they're writing a header, via `SourceType`. That information gets dropped in `Target.genFile` before `Tool.getOutputFileWriter` picks the directory. Passing it through, and only using a different directory when the option is set and the file is a header, keeps every other code path as it is.

Related: #2001, about inaccurate `-depend` output for C++, and #4102, the proposed CMake overhaul. I've kept the CMake change small so it shouldn't get in the way of that work.

I'm happy to put together a few PRs (tool, Maven plugin and a C++ runtime) if this sounds like a reasonable direction. Thanks for taking a look!

hamidelmaazouz:
Fixes #4975
Closes #2439

Hi! This adds a `-header-dir` option so generated C++ headers can live in their own directory.

Today the C++ target puts every `.h` and `.cpp` file into the `-o` directory. That works, but most C++ projects keep public headers apart from their sources, so people end up either exposing a folder full of generated `.cpp` files on the include path or moving headers around after generation. #2439 asked for this back in 2018, and #4975 describes the problem in more detail.

With this change you can write:

```bash
antlr4 -Dlanguage=Cpp -o gen/src -header-dir gen/include MyGrammar.g4
```

and the headers land in `gen/include`, while the `.cpp`, `.tokens` and `.interp` files stay in `gen/src`. If you don't pass the option, nothing changes, for C++ or for any other target.

## How it works

The code generator already knew whether it was writing a header, through `SourceType`. That information just got dropped in `Target.genFile` before the tool worked out where to write the file. This PR passes it along, so headers can be routed to the new directory while everything else takes the same path as before.

The header directory follows exactly the same rules as `-o`. A grammar in a subdirectory gets the same subdirectory under the header directory, and `-Xexact-output-dir` flattens both. To keep the two in step, the first commit pulls the existing `-o` logic into helpers that take the base directory as a parameter. That commit doesn't change any behaviour on its own.

Generated `#include` lines stay as they are, so you add the header directory to your include path like any other C++ library.

Here's where files end up in each case:

| `-o` | `-header-dir` | `-Xexact-output-dir` | Grammar | Sources | Headers |
|---|---|---|---|---|---|
| `gen` | not set | either | any | as today | same as sources |
| `gen` | `inc` | off | `T.g4` | `gen/` | `inc/` |
| `gen` | `inc` | off | `sub/T.g4` | `gen/sub/` | `inc/sub/` |
| `gen` | `inc` | on | `sub/T.g4` | `gen/` | `inc/` |
| `gen` | `inc` | either | `/abs/T.g4` | `gen/` | `inc/` |
| not set | `inc` | off | `sub/T.g4` | `sub/` | `inc/sub/` |
| not set | `inc` | on | `sub/T.g4` | `sub/` | `inc/` |
| `gen` | an existing file | either | any | nothing, `OUTPUT_DIR_IS_FILE` | nothing |

A trailing `/` or `\` works too, just like with `-o`.

## Build tool support

It seemed only fair to wire this through the places people actually call ANTLR from:

- `-depend` reports headers in the header directory. While in there, it now also lists `TLexer.h` for combined grammars, which it used to leave out because of an `isDefined("headerFile")` check that never matched any template.
- The Maven plugin has a new optional `headerOutputDirectory` parameter.
- `ANTLR_TARGET` in `FindANTLR.cmake` accepts `HEADER_OUTPUT_DIRECTORY` and sets `ANTLR_<name>_HEADER_OUTPUT_DIR`. If you leave the argument out, the command it runs is exactly the same as before, so older jars still work.
- `antlr4_generate` from the `antlr4-generator` CMake package takes the header directory as an optional ninth argument, and `ANTLR4_INCLUDE_DIR_<target>` then points at it. Without the argument its commands are unchanged too.
- `doc/tool-options.md`, `doc/cpp-target.md`, the CMake README and `Antlr4Package.md` explain how to use it.

## Compatibility

The Java API only grows. The new pieces are `Tool.headerOutputDirectory`, `Tool.getHeaderOutputDirectory`, a three-argument `Tool.getOutputFileWriter`, a four-argument `Target.genFile`, a three-argument `CodeGenerator.write` and `BuildDependencyGenerator.getHeaderOutputFile`.

One thing worth calling out for anyone who subclasses the tool. Without `-header-dir`, every existing override is still called for every file, so nothing breaks. With `-header-dir`, header files skip overrides of the two-argument `Tool.getOutputFileWriter`, of `Tool.getOutputDirectory` and `Tool.new_getOutputDirectory`, and of the three-argument `Target.genFile`. Inside this repository that only touches `Antlr4Mojo.CustomTool`, which now has a matching override, and `GoTarget`, which doesn't generate headers anyway.

## Testing

- `TestHeaderOutputDirectory` in `tool-testsuite` walks through the table above, plus listeners and visitors, the exact `-depend` file list, the Java target ignoring the option, and existing overrides still being called when the option isn't used.
- `CppHeaderDirTest` in `runtime-testsuite` generates a C++ parser with headers in a separate directory, then builds and runs it. `CppRunner` and `Test.vcxproj.stg` take an optional header directory for this, and render just as before without one.
- `Antlr4MojoTest` has a new C++ project that sets `headerOutputDirectory`.

Everything passes locally on macOS:

| Suite | Result |
|---|---|
| `tool-testsuite` | 718 run, 0 failures, 5 skipped |
| `runtime-testsuite` `cpp.**` | 363 passed |
| `runtime-testsuite` `java.**` | 438 passed |
| `runtime-testsuite` `go.**` | 362 passed |
| `antlr4-maven-plugin` | 5 passed |

A heads-up about the Maven plugin tests. They fail on `dev` as it stands with the default `mavenVersion` of 3.8.5, because takari-plugin-testing 3.0.0 hits a null plugin in `DefaultMojoExecutionConfigurator`. Running `mvn verify -DmavenVersion=3.6.3` gets them going, and that's how they were run here. CI skips these tests, so it isn't affected.

For CMake, small scratch projects were built with and without `HEADER_OUTPUT_DIRECTORY`, and with `antlr4_generate` given no header directory, a relative one and an absolute one. All of them built and ran, a second build had nothing left to do, and without the new arguments the generated commands matched `dev` exactly.

## Noticed along the way

On `dev`, `-depend` for the Java target lists `TParserContexts.java` and `TParserDFA.java` even when the split parser option is off, so those files are never written. That's unrelated to headers, so it's left out of this PR.

@KvanTTT @kaby76 @jimidle @ftomassetti @ericvergnaud @parrt, would you have a moment to take a look? Thanks!


hamidelmaazouz:
A quick note on the red TypeScript jobs here. They're not caused by this change. The TypeScript target doesn't generate headers, so `-header-dir` never touches its output.

Every test fails before any ANTLR code runs, because `ts-node` crashes on startup:

```
TypeError: Cannot read properties of undefined (reading 'fileExists')
    at readConfig (.../ts-node/dist/configuration.js:91:33)
```

The test runner installs `typescript` without a version, npm now serves TypeScript 7, and `ts-node` 10 doesn't work with it. The same failure shows up on other open PRs. It's tracked in #4962, and #4963 fixes it by running the tests with `tsx`. Once that's merged, a re-run here should turn these jobs green.

